PR #60, opened 2011-07-15
package.json

ghost:
To be able to install this from npm I've added a package.json. I've hard-coded most of the module versions since there's no guarantee that a new backwards-incompatible version of a dependency won't get published.

I didn't see any obvious candidate for "main", but somebody more familiar with the codebase could probably figure out a nice way to make that work.


japj:
This should also allow one to perform "npm install -d" in the extracted directory to install all dependencies without having to manually install those

PR #110, opened 2011-07-31
Fixes the test of JAVA environment variable instead of UGLIFY

christiankakesa:
Fixes the test of JAVA environment variable instead of UGLIFY which is already tested.


lloyd:
thanks @fenicks!

PR #187, opened 2011-08-21
Ensure req.body exists before trying to use it.

brianloveswords:
I spent the latter half of yesterday learning a painful lesson: express.bodyParser() will not parse a request body unless the content-type is 'application/x-www-form-urlencoded'. It won't even set req.body to an empty dict, it leaves it undefined.

I suppose it's my own fault for neglecting to set this header when making my requests, but this was causing browserid.org/verify to consistently give me 500s.

This, at the very least, will make it so the verifier doesn't 500 on malformed requests.

### DIFF
--- a/verifier/app.js
+++ b/verifier/app.js
@@ -39,13 +39,14 @@ const   path = require('path'),
    httputils = require('./lib/httputils.js'),
  idassertion = require('./lib/idassertion.js'),
          jwt = require('./lib/jwt.js'),
-     express = require('express');
+     express = require('express'),
      metrics = require('../libs/metrics.js'),
      logger = require('../libs/logging.js').logger;
 
 logger.info("verifier server starting up");
 
 function doVerify(req, resp, next) {
+  req.body = req.body || {}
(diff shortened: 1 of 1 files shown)


lloyd:
oh, and more to the point, I wonder if we really want to require that clients of the API add this header?  It seems a needless restriction.

brianloveswords:
I can adjust the method so if req.body is undefined it populates it by using querystring.parse on req.rawBody. This is part of what bodyParser does under the hood, once it determines querystring.parse is the correct parser by examining the content-type.

lloyd:
Hmm.  Actually, since the body is actually form encoded at this point, perhaps maybe just a very clear and concise error message would be a better approach?  Like a bad request code with a verbose string explaining the requirement.

/cc @benadida An aside, I would typically prefer these post requests to have a content type of application/json, given that we're already using JSON throughout (and this causes us to double encode data), the apis are invoked programatically, and we don't have any need to directly invoke the apis as the result of a form completion.  form-encoded key values == tired, json post bodies == wired.  

now to that error message...

PR #193, opened 2011-08-22
Change the csrf response type to 'html'.

shane-tomlinson:
This makes sure that the non-XML response is parsed correctly and jQuery does not throw an error.

### DIFF
--- a/browserid/static/dialog/resources/browserid-network.js
+++ b/browserid/static/dialog/resources/browserid-network.js
@@ -45,7 +45,7 @@ var BrowserIDNetwork = (function() {
       $.get('/wsapi/csrf', {}, function(result) {
         csrf_token = result;
         cb();
-      });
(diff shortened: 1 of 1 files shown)


lloyd:
@shane-tomlinson - Related:  Shall we change the csrf response to application/json and send down a json encoded token?  By having the server content-type explicitly set correctly (right now it's not set to anything) client libs will interpret properly by default.  By using JSON we improve the symmetry across our wsapi.

```
$ curl -i https://dev.diresworb.org/wsapi/csrf | grep -i content
<nothing>
```

shane-tomlinson:
@lloyd - 100% in agreement.  Things like am_authed should probably be as well.

shane-tomlinson:
@lloyd - I opened a new issue #217 for having all calls return JSON responses.

PR #228, opened 2011-08-30
initial push toward the new design

clarkbw:
Sorry for the large pull request.  Here's the list of major things that will need fixing and come with these changes.
- Account Manager: removing an email address doesn't currently refresh the list and/or reload the page to normal
- Sign In and Sign Up need to be hooked up to actually do something
- Forgot password is another system that needs to be wired up to something real


shane-tomlinson:
@clarkbw - thanks for this.  Can you do me a favor and merge in the latest dev changes into your code to make the reverse merge a bit easier?

shane-tomlinson:
@clarkbw - I am having a look at localhost:10002 - looking hot!

lloyd:
Hey, this is hot!

I took the liberty of rebasing these changes on a fresh branch, I think I did the right thing on the merge conflicts.  

https://github.com/mozilla/browserid/tree/design

I don't mind integrating your new changes onto that branch (you can fork it and work off of it), if this is helpful.

@lloyd

shane-tomlinson:
@clarkbw, @lloyd - I have started the initial hooking up of behaviors - "happy case" behaviors for the authentication, new user, the initial page load to sync users, as well as the delete email.  

We need some error messages, also some cleanup is needed like putting all of the js into js files - right now I have form handlers included in the .ejs files.  @clarkbw, would you have an objection of me changing the id of signUpForm in the sign in page to signInForm and using a class to share styles with signUpForm?  This would enable me to put the form handler JS into JS files and remove them from the .ejs templates.

@clarbw, for the identity list, I took out the in-js text string template and put it into index.ejs in its own template where it is easier to see what is going on at a glance and work with.  

"I Forgot My Password" I have not yet touched.

shane-tomlinson:
@clarkbw - You will also want to fork off of the "design" branch that @lloyd set up - that is the one I am working against now.

shane-tomlinson:
Closing this as we have the design branch, no longer needing a pull request until the end.

PR #229, opened 2011-08-30
Relay frame

stomlinson:
Quite a major request to have IE support BrowserID.


ghost:
Verified as fixed with IE8 in Windows XP.

PR #256, opened 2011-09-15
no global dependencies and include jwcrypt using submodules

lloyd:
this change uses features in npm to obviate the need for an explicit "dependency setup" step (it's run by npm install), and also uses bins provided by npm libraries rather than requiring global installation.

after this is landed, the steps to run the browserid development server are:
1. git clone browserid
2. cd browserid && npm install
3. ./run.js


shane-tomlinson:
I have looked at the code, checked out the branch, and ran `npm install` as if a fresh install.  All works as desired.

PR #297, opened 2011-09-24
Fix broken URL for the reference validator

fmarier:
Looks like the master branch got renamed to dev at some point.


benadida:
with the new UX, this page is going away. Thanks for the patch though!

PR #327, opened 2011-09-29
Updating link to verfier source, fixes Issue#326

ozten:
Epic patch!


ozten:
llyod already got this while I as one the phone and finally hit submit :)

Issue #329, opened 2011-09-29
need to sync time with server before signing assertion


benadida:
otherwise problem logging in.

lloyd:
@benadida suggests we get server time.  I suggest we combine this with the /wsapi/csrf call, meaning that call moves to return an object which returns server time and csrf token, then at the time assertion signing is performed we calculate offset from the time csrf response was received, and that is the assertion generation time.

we might want to rename csrf call to `session_context` or some such.

benadida:
agreed.

PR #344, opened 2011-10-03
Don't break when a client machine has a system clocks that is off by more than 2 minutes

lloyd:
@shane-tomlinson and @benadida - I'd like careful review of these commits.  Basically I create a new `session_context` API which serves three pieces of information:
1. authentication status of the connection
2. current server time
3. csrf token

Client code is reworked so session status is only hit when needed to save on traffic, and usually in a normal usage it will only need to be hit once per dialog interaction.

Also, now the assertion generation code uses server time instead of local time to sign assertions.

If this code looks ok we'll merge into the rolling train and get QA on it.

### DIFF
--- a/browserid/lib/wsapi.js
+++ b/browserid/lib/wsapi.js
@@ -115,7 +115,7 @@ function checkAuthed(req, resp, next) {
 }
 
 function setup(app) {
-  // return the CSRF token and current server time (for assertion signing)
+  // return the CSRF token, authentication status, and current server time (for assertion signing)
   // IMPORTANT: this is safe because it's only readable by same-origin code
   app.get('/wsapi/session_context', function(req, res) {
     if (typeof req.session == 'undefined') {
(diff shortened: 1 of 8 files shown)


lloyd:
oh, this is for issue #329, which is 5 stars.

benadida:
this commit looks good to me.

lloyd:
with respect to the semantics of 'withCSRF', it meant "execute this callback with a valid CSRF token, fetching one from the server if required"  Now a CSRF token is one component of the session context.  `withContext` means "execute this callback with valid session context, fetching context from the server if required".

the naming changed because it's confusing for a caller wanting to know if the user is authenticated to call a function which seems like it provides a CSRF token, less so for that caller to say "In order to answer this question, I need the server supplied session context"...

lloyd:
@benadida & @shane-tomlinson - once you guys are happy with this, say so explicitly in the pull request and I'll merge down and push to beta servers.  I promised @jbonacci that if we do include this feature in the upcoming train it'll be ready to test by end of day tues.

shane-tomlinson:
@lloyd - please see https://github.com/mozilla/browserid/commit/246e88b8feac9a13c29d03e24c70eda1480240ad#commitcomment-628943.  Once that is done, I am all good with this pull request.

PR #431, opened 2011-10-14
Initial repoze.who browser id plugin

catlee:
repoze.who is an identification and authentication framework for arbitrary WSGI applications.

This module implements a repoze.who plugin that uses browserid to authenticate a user.


lloyd:
this is awesome!  but I think it belongs in its own repository.  Would you mind building a browserid-repoze repository for it, and then we can add it to [the wiki of 3rd party BrowserID libraries](https://github.com/mozilla/browserid/wiki/BrowserID-Libraries)?

catlee:
I have it here: https://github.com/catlee/repoze.who.plugins.browserid

PR #467, opened 2011-10-20
include full origin (inc. scheme) in generated assertions.  

### LINKED ISSUE #82: make sure audience captures full origin
scheme and port, in particular.

lloyd:
include full origin (inc. scheme) in generated assertions.  To preserve compatibility, allow the rp to omit scheme in post data, and subsequent to verification, return whatever the rp sent us.  issue #82.

once merged, closes #82.


ghost:
Verified that on the BrowserID dialog, the site name is filtered (no protocol, no port). Nothing else specific for QA to test.

PR #474, opened 2011-10-25
Ops-related features & fixes

fetep:
- script & spec file for building an RPM
- if writing out a secret cookie, make the file mode 0600
- add a /**heartbeat** end-point (so load balancers can see if the node.js stack is healthy)
- schema enhancements
- wrap schema creation around a database setting
- support NODE_EXTRA_CONFIG
- better conditional around URL substitution
- use "cluster" module to manage node processes


lloyd:
all merged except for cluster support, which we want to think about.

Issue #273, opened 2011-09-21
A message should be displayed for unsupported browsers


shane-tomlinson:
If we check for window.localStorage and window.postMessage, we should be able to find a large portion of browsers that are not supported.

shane-tomlinson:
@jbonacci: our minimum technical requirements are window.localStorage and window.postMessage.  Even with these requirements met, there are performance issues to worry about - for instance iOS takes ~20 seconds per key generated, IE8 about the same.  The other hitch that seems to be getting us right now is whether we can communicate between windows, this is especially problematic on mobile browsers.

shane-tomlinson:
a8230ccb0c26bef94f98701ec457b3d828686f33 displays an alert and will not open the dialog.  We could use a cleaner solution, perhaps a nice page?

lloyd:
labeled 'needs UX input'

clarkbw:
OK, instead of a dialog I want to use the center space for our error messages.
http://cl.ly/2h453W2r0E2d460M2D21

The important thing here is to decide what we want our users to be able to do.  We can say "IE doesn't work" but we should explain why and what they _could_ do to actually sign in, like use Firefox.

It would be great to have a breakdown of the various browser support situations on a wiki page that we can link to.  I think as a minimal approach, starting from the basics of what I have, I want to say "This isn't supported but modern browsers like Firefox are.  Check out our full support page here for other options."

Something more like this: http://cl.ly/3Y311l0g421k0A0A2T1G  however possibly less Firefox branding.

shane-tomlinson:
@clarkbw, we need a solution that addresses other browsers that are incompatible as well, right now we check for Internet Explorer <= 8 and display an IE specific message.  We also check for browsers that do not support localStorage, postMessage, and if the browser has problems with JSON.decode(null) (Android < 3.0) and display a generic message.

This brings up a question of whether we should have an API method that an RP can call to see if the user's browser supports BrowserID.  If the browser does not support BrowserID, they could avoid displaying the button?

shane-tomlinson:
@clarkbw - your response from our email -

"Agreed, I'm wondering if we want to do a bit of both.  For mobile browsers we have a bit of different layout and I think we'll want to continue that in this kind of error page.  I'd suggest we file a separate bug for the query API. "

I'll tackle this as you suggest with http://cl.ly/3Y311l0g421k0A0A2T1G.  We'll get an initial rev and then iterate.

shane-tomlinson:
[...]

PR #484, opened 2011-10-26
Unsupported dialog and LOADS of related changes, for issue #273

shane-tomlinson:
This is the unsupported dialog and all of its related changes.  There are so many commits because the original browser check was placed into the main dialog code.  This did not work for browsers that do not have postMessage as the channel is never established.  Afterwards I moved a lot of this logic into include.js and created a new dialog page which is served up to unsupported browsers. 

The most important part to review is the additional backend unit test named page-requests-test.js.  This now makes use of the wsapi client so it is standard.  https://github.com/mozilla/browserid/blob/245e4f6879656b607a1274ad8bbede67a406d7c4/browserid/tests/page-requests-test.js

I retained the BrowserSupport.js file and related unit tests even though the BrowserSupport code is pulled into include.js because in the near future I want to split include.js into several files for testing purposes and do not want to rewrite this code.

dialog.ejs has been split into dialog.ejs and dialog_layout.ejs.  This is so we can have one generic layout template to use across all current and future dialogs, including the new unsupported_dialog.ejs


lloyd:
I don't find the individual commits here very useful, I would prefer to squash merge this...

PR #496, opened 2011-10-27
Make vagrant boot instructions render nicely and copy+pastable

leto:
Just tidying up the readme a bit.


lloyd:
thank you kindly, @leto!

PR #718, opened 2011-12-06
Easier setup for configuring email in development

jrburke:
This patch:
- adds the ability to configure the SMTP port, which allows optional TLS use.
- removes email_to_console flag set in dev/local config in favor of just checking if smtp_params have been configured.
- given the new check for smtp_params && smtp_params.host in email.js, it moves the console.log() of the verification link to the "else" so that some potentially confusing ! tests can be avoided.

These changes worked for me when I wanted to configure SMTP in dev, and I also tested the case where I did not configure the SMTP env variables, and I got the console.log call.

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -113,7 +113,6 @@ g_configs.production = {
 // local development configuration
 g_configs.local =  {
   URL: 'http://127.0.0.1:10002',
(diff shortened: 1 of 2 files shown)


lloyd:
the world has changed since this pull request, I'm sorry we didn't get to it earlier.  There's a completely new configuration system.  have a look!

PR #726, opened 2011-12-08
Train 2011.12.01 hotfix 601

shane-tomlinson:
This is a hotfix to support the "Native API" for Addons (BrowserID plugin, OWA) and native implementations of BrowserID (Soup, Pancake).  It adds a new namespace, BrowserID.internal.  BrowserID.internal contains two functions:

setPersistent(origin, callback) - sets the persistent flag for an origin
get(origin, options, callback) - works like navigator.id.get. options.silent defaults to true.  To get an assertion for a required email address (OWA/Soup receipts case) specify options.requiredEmail.


shane-tomlinson:
Closing, opened this against the wrong branch.

PR #789, opened 2011-12-21
use crypto.getRandom instead of urandom

benadida:
added tests, added async ability for generation, added weakGenerate

### DIFF
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -171,35 +171,35 @@ function addEmailToAccount(existing_email, email, cb) {
 }
 
 exports.stageUser = function(email, cb) {
-  var secret = secrets.generate(48);
-
-  // overwrite previously staged users
-  sync();
-  db.staged[secret] = {
-    type: "add_account",
-    email: email,
-    when: (new Date()).getTime()
-  };
-  db.stagedEmails[email] = secret;
-  flush();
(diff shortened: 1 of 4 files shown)


lib/db/json.js
```diff
@@ -171,35 +171,35 @@ function addEmailToAccount(existing_email, email, cb) {
...
+    db.staged[secret] = {
+      type: "add_email",
+      existing_email: existing_email,
+      email: new_email,
+      when: (new Date()).getTime()
+    };
+    db.stagedEmails[new_email] = secret;
+    flush();
+    
+    setTimeout(function() { cb(secret); }, 0);
```
lloyd, line 201: given the whole thing is in a callback, this setTimeout is now not necc.

lib/db/json.js
```diff
@@ -171,35 +171,35 @@ function addEmailToAccount(existing_email, email, cb) {
...
+    // overwrite previously staged users
+    sync();
+    db.staged[secret] = {
+      type: "add_account",
+      email: email,
+      when: (new Date()).getTime()
+    };
+    db.stagedEmails[email] = secret;
+    flush();
+    setTimeout(function() { cb(secret); }, 0);
```
lloyd, line 184: given the whole thing is in a callback, this setTimeout is now not necc.

PR #807, opened 2011-12-23
Feature randomseed from server

benadida:


lloyd:
fully merged individually these commits.  the first have (server apis) went into train-2011.12.28.  the second half (client code) will go into 2012.01.04

benadida:
@lloyd I don't know the exact implementation of node's getRandom, we should test it.

PR #808, opened 2011-12-25
Removed a trailing comma that prevents Google Closure Compiler from mini...

JoshyPHP:
...fying the file.


lloyd:
thanks for the contribution!

PR #859, opened 2012-01-10
Upgrade license to MPL 2.

gerv:
Here's a first cut at upgrading the licence of the BrowseriD codebase to MPL 2, including adding licence blocks to files which lack them. Hopefully I haven't relicensed anything I shouldn't :-)


benadida:
I love that you just removed 3200 lines. I'm sold on MPL2 based only on that fact ;)

PR #862, opened 2012-01-10
Bug706572 localization

ozten:
This is a first stab to get some feedback.

Good place to start is https://github.com/ozten/browserid/blob/Bug706572_localization/docs/I18N.md

then 
https://github.com/ozten/browserid/blob/Bug706572_localization/bin/browserid#L65
and
https://github.com/ozten/browserid/blob/Bug706572_localization/lib/i18n.js

TODOs left in for discussion.

This PR doesn't address emails, which I'll look into next.


benadida:
closing this one in favor of v2.

PR #897, opened 2012-01-12
Bug706572 localization v2

ozten:
This is a rebased version of the earlier pull request.


lib/browserid/views.js
```diff
@@ -94,6 +94,18 @@ exports.setup = function(app) {
...
+  /**
+   * 
+   * XXX benadida or lloyd, I tried to use straight up regexp to do this, but.
+   * is there a better way to do this?
+   */
+  function QUnit(req, res) {
+    res.render('test.ejs', {title: 'BrowserID QUnit Test', layout: false});
+  }
+
+  app.get("/test", QUnit);
```
benadida, line 113: I would rather one of these redirect to the other, so we stop bookmarking both.

benadida:
overall, this patch looks good, except for one fairly important thing I would strongly suggest tweaking.

The i18n overhead on a request appears to be non-negligible, just from eyeballing the amount of code that's run. But in fact we only need to do all of this stuff on _two_ API calls that relate to sending email. Fortunately, every wsapi call is in its own file with its own metadata now. So we can easily add

exports.i18n = true;

in those wsapi calls that need it, and your middleware can bail early if i18n is not needed for a given call.

ozten:
I'm waiting for r+ before landing this on dev.

I'll merge again to keep the PR merge button green.

lloyd:
@ozten and I consulted gerv, and now understand that the final outstanding issue is removing the GPL licensed implementation of gettext(), as it would mean we couldn't release browserid under MPL without violating his license.

Specifically, this library is the issue: https://github.com/mozilla/browserid/blob/a725d667e4ce537404006037d1d72a7c5fd6a1d7/resources/static/shared/gettext.js

Choose your own adventure: rewrite, choose another, or contact the author?

ozten:
Lame. Sorry about that and thanks @lloyd for spotting this legal bug.

There are several client side gettext to choose from. I'll remove and stub out, then replace with a functioning gettext.

ozten:
Following up on other new code:

LGPL libraries:
node-gettext
lib-iconv

Issue #911, opened 2012-01-18
include.orig.js missing


anantn:
https://browserid.org/include.js says that the uncompressed source for itself can be found at https://browserid.org/include.orig.js, but the latter returns a 404.

shane-tomlinson:
This is fixed in https://github.com/mozilla/browserid/pull/922

lloyd:
fixed on dev: https://dev.diresworb.org/include.orig.js

jrgm:
This is now available on stage for train-2012.02. https://diresworb.org/include.orig.js

Issue #921, opened 2012-01-20
Add routes for include.js and include.orig.js so that symlinks are not overwritten


shane-tomlinson:
Currently whenever we run the compression script, the symlink to include.js is overwritten, frequently leading to an incorrect checkin.  Instead of updating the symlink, we could add routes that point to to these files.

include.js should be served from:

development: include_js/include.js
production: production/include.js

include.orig.js should always be served from include_js/include.js

Also, remove the symlink from the repo, remove the symlink creation step from the compression script.

lloyd:
merged to dev.

PR #922, opened 2012-01-20
Setting up views for include.js and include.orig.js

shane-tomlinson:
@lloyd, @ozten - could this be reviewed?  I have set up routes for include.js and include.orig.js as follows:
- include.js points to include_js/include.js or production/include.js depending on environment.
- Added tests checking for correct responses to 'include.js' and 'include.orig.js'.
- Added missing tests for correct response for '/test' and '/test/'
- Remove the symlink to include.js
- Remove symlink update in compress.sh

issue #921

Areas of concern - what was our previous caching policy on these files?  What should our caching policy be?

Any suggestions or updates are welcome!

### DIFF
--- a/lib/browserid/views.js
+++ b/lib/browserid/views.js
@@ -17,8 +17,6 @@ path = require('path');
 
 
 const VIEW_PATH = path.join(__dirname, "..", "..", "resources", "views");
-const DEV_INCLUDE = path.join(__dirname, "..", "..", "resources", "static", "include_js", "include.js");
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
Two other areas of concern - how do you run unit tests as if the environment were "production" - and how do we handle the case where the environment is production but the production resource is not found?

shane-tomlinson:
When merged, close #911

shane-tomlinson:
@lloyd, I was unaware that you could change the URL in such a fashion.  Updated the pull request with changes.

lloyd:
this looks fantastic.

PR #926, opened 2012-01-20
Move the declaration of window.gettext and window.format into gettext.

shane-tomlinson:
- This makes it available before any gettext calls are done in wait-messages and error-messages.
- Create's only a single instance that is used universally.


jrgm:
verifed in train-2012.02.02 code

PR #943, opened 2012-01-24
Fix IE8 showing a black background on the main site.

shane-tomlinson:
Everywhere we use an opacity, add an IE8 filter.

issue #929


lloyd:
cherry picked into train-2012.01.18 - testing commences.

Issue #947, opened 2012-01-24
scroll bar appears in required email flow when verifying with primary


shane-tomlinson:
Steps to reproduce:

1) Open dev.myfavoriteshow.org
2) Enter in a "primary" supported email address - something at eyedee.me
3) Check out the scroll bar.

Check out screen shot:
https://skitch.com/set117/g5882/browser-id

shane-tomlinson:
Labeling 5 star though it is a new feature just because it is such an ugly ugly user facing blemish on an otherwise awesome feature.

lloyd:
hotfixed in 2012.12.18.02

shane-tomlinson:
This is back with longish email addresses.

https://skitch.com/set117/g7say/verify-required-primary

jrgm:
verified on stage for multiple browser/os including android 2.1 stock.

PR #950, opened 2012-01-24
Fix scrollbar showing in the Required Email screen if email is a primary...

shane-tomlinson:
....

Set the max-height to 130px instead of 110px.

issue #947


lloyd:
cherry picked into train-2012.01.18 - testing commences.

PR #953, opened 2012-01-24
Putting locale into etag recipie

ozten:
Untested and I won't be around for much of the next two days, but... here you go :)


ozten:
local should be [a-z][a-z](_[A-Z][A-Z])?

Examples:
en
en_US

As long as '_' is okay, we should be good.

Issue #955, opened 2012-01-24
IE8: Provisioning with sample IdP throws exceptions


shane-tomlinson:
Steps to reproduce on local dev:

1) Go to sample RP
2) Enter in `<username>`@example.domain into required email.
3) Click "Get an assertion"
4) Watch IE8 blow up.

Screen clip: https://skitch.com/set117/g58xr/win7-pro-ie8-running

Extra Info from Debugger:
Webpage error details

User Agent: Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1; WOW64; Trident/4.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; Media Center PC 6.0)
Timestamp: Tue, 24 Jan 2012 12:25:56 UTC

Message: Exception thrown and not caught
Line: 244
Char: 11
Code: 0
URI: http://192.168.1.88:10002/provisioning_api.js

Message: Object doesn't support this property or method
Line: 17
Char: 3
Code: 0
URI: http://192.168.1.88:10005/provision.html

shane-tomlinson:
This is caused by the MPL license being above the DOCTYPE, putting the IdP into quirks mode.

shane-tomlinson:
@jbonacci - there was a missing bit in the initial steps, I forgot to surround `<username>` with the quotes.  Make sure you add a username before the @example.domain

Closing this issue, waiting for verification.

shane-tomlinson:
@jbonacci - I'm not seeing any errors anymore on my dev machine.  It looks like navigator.id is not being found in your case which normally means there is a problem getting include.js

if you go to http://10.251.25.155:10002/include.js, do you get a response?

PR #961, opened 2012-01-24
Fixing IE8 going into quirks mode in the provisioning page.

shane-tomlinson:
issue #955


lloyd:
This bug was introduced during MPL2 license conversion, with the license header put before doctype.

PR #998, opened 2012-01-28
Fix typo

fmarier:


shane-tomlinson:
@fmarier - thanks for the submission - merging.

Issue #1026, opened 2012-02-01
completing registration in a browser other than the initiating browser results in non-functional dialog


shane-tomlinson:
Steps to reproduce

1) Open up two browsers
2) In browser 1, go to beta.myfavoritebeer.org
3) Sign up for a new account
4) Keep dialog open
5) In browser 2, open email provider
6) Open verification link in browser 2
7) Enter passwords and complete flow.
8) Original dialog goes back to authentication screen, but buttons are unresponsive.

Desired result.

User needs to authenticate.

jrgm:
verified flow above, with osx safari 5.1.2, opera 11.6, FF 10, Chrome 16. Also went through adding an email to the 
first account. Moving on to winxp/{ie8,FF10,Chrome16,safari,opera}.

jrgm:
verified flow above, with winxp safari 5.1.2, opera 11.61, FF 10, Chrome 16, IE8

jrgm:
verified flow above on stage, with win7/32bit safari 5.1.2, opera 11.61, FF 10/11/12, Chrome 16, IE9

jrgm:
verified flow above on stage, with ubuntu 11.0 for opera 11.61, FF 10/11/12, Chrome 16
Trying android if the network here works.

PR #1029, opened 2012-02-01
Fixing the complete registration flow when the browser the user completes registration is different from the one they started at.                               

shane-tomlinson:
- in check_registration.js, the "auth" message has been changed to "authenticate    "
- in user.js, when the registration poll completes, set the "known" field of the     address in the addressInfo cache to "true"

issue #1026


shane-tomlinson:
closed with checkin 0cd88231443e30d3e4d579e3b6f42307acb49187.

PR #1030, opened 2012-02-01
BrowserID instead of Browser ID

mathjazz:
We use BrowserID (without space) everywhere else.


ozten:
Great work!
I had already changed in 
https://github.com/mozilla/browserid/commit/4b3c2c73f3290bbdb21f910c64329ea1aae9fa96

Issue #1058, opened 2012-02-06
Two new issues with the dialogs for 'forgot password' (train-2012.02.02)


jrgm:
This is, I think, related to the fixes for GH-984 (and possibly GH-1001). While GH-984 is fixed for what it described, there are two new bugs in the flow.

1) go to http://beta.myfavoritebeer.org with a clean browser profile (or simply ensure you are signed out of diresworb.org).

2) On beta.myfavoritebeer.org, click Sign in. The /sign_in popup window comes up. Enter an email address that is known to diresworb. Click next.

3) Click on 'forgot your password?' when it is shown.

Bug #1: the button is labelled just 'password', not 'reset password'.

4) Click on 'password' when that button is shown.

Bug #2: On the 'Check your email!' screen, it says "We sent a confirmation email to <b>undefined</b>"

shane-tomlinson:
@jrgm - Thanks for the report - when there are two issues, can you file them separately so we can better track them?

shane-tomlinson:
@ozten - in resources/static/i18n/en_US/client.json, I am seeing the following for Reset Password and Resetting Password:

```
 67       "Reset Password" : [                                                                                        
 68          null,
 69          "Password"
 70       ],
 71       "Resetting Password" : [
 72          null,
 73          "Password"
 74       ],
```

If I change the 

```
 67       "Reset Password" : [                                                                                        
 68          null,
 69          "Password"
 70       ],
```

to 

```
 67       "Reset Password" : [                                                                                        
 68          null,
 69          "Reset Password"
 70       ],
```

the buttons hows up correctly in the dialog.

ozten:
The en_US po files have some msgstr with values.

#: resources/static/shared/error-messages.js:108  
#, fuzzy  
msgid "Resetting Password"  
msgstr "Password"  

This happens when we change a string, it gets marked fuzzy so we know it changed.

I'll go through en_US client.po and messages.po and remove all msgstr values, so they are empty strings.

jrgm:
@shane-tomlinson yes, sorry; should have filed as two issues.

Anyways, all works as expected for the button label (in en-US) and for the email address not being undefined for IE8/WinXP, IE9/Win7, {opera,chrome,firefox} on {osx,ubuntu,winxp/win7}, safari on {osx,winxp,win7}

PR #1063, opened 2012-02-07
Hotfix 1058 forgot password

shane-tomlinson:
Fix the missing email address in the "check your email" screen for the forgot password flow.

issue #1058


jrgm:
Anyways, all works as expected for the button label (in en-US) and for the email address not being undefined for IE8/WinXP, IE9/Win7, {opera,chrome,firefox} on {osx,ubuntu,winxp/win7}, safari on {osx,winxp,win7}

PR #1064, opened 2012-02-07
Removing double space between sentences.

mathjazz:


shane-tomlinson:
@mathjazz - does this need merged into dev only, or does this require a merge into beta/prod as well?

mathjazz:
@shane-tomlinson, dev will be enough.

shane-tomlinson:
thanks @mathjazz!

Issue #1083, opened 2012-02-09
rate limiting responses should use HTTP 429


lloyd:
requests which refuse to service a request because it would cause too many emails being sent to a single address are currently using HTTP 403, HTTP 429 is a better fit.

jrgm:
verified in unit tests. Small note: node returns 'HTTP/1.1 429 unknown' in the response header.

PR #1084, opened 2012-02-09
use HTTP 429 for throttling responses.  issue #1083

lloyd:


stomlinson:
@lloyd - overall looks good - the front end unit tests will fail.  In resources/static/test/mocks/xhr.js, it refers to HTTP 403, can you update this to 429?

Issue #1087, opened 2012-02-10
IE9 in IE7 mode: Unsupported dialog shows no content


shane-tomlinson:
STR:

1) Open up IE9
2) Open the dev tools (F12), put IE into IE7 mode.
3) go to beta.myfavoritebeer.org
4) Click on "Sign In"

Dialog is empty.

Expected: Dialog says that you need a new browser.

shane-tomlinson:
Fixed with checkin c8f02d643fecb7fb29eea7ececd4e87e507fec18.

lloyd:
@jbonacci - you tested beta, but shanes fix hasn't made it to beta yet.  you can get a preview of the fix on dev.diresworb.org in about 3 minutes. as soon as https://dev.diresworb.org/ver.txt says something other than ce1b28c.  If it looks good, I can send a request to pete to merge the hotfix.

lloyd:
I cherrypicked c8f02d6 into dev and it doesn't appear to fix the problem.  I still see a grey background with no text displayed.

shane-tomlinson:
@lloyd - I just cherry picked the commit into train-2012.02.02 and the fix worked as expected.  

Somebody mentioned in the irc conversation linked to above, this _appears_ to be an IE9 in IE7 mode only bug, but it is a bug for any browser that supports the "opacity" CSS tag but does not meet the other requirements (JSON, localStorage, postMessage).  IE6 through IE8 lack support for opacity so they display the contents as expected.  I am not sure what all browsers that includes at this time.

This is labeled a 5 star because it is a regression.

shane-tomlinson:
@lloyd, @jbonacci - The discussion above and in IRC did lead me to add to the original hotfix to fix IE8 in IE7 mode - IE8 does not support the opacity CSS tag but does support their own "-ms-filter" which takes care of opacity. 

IE8 is missing an -ms-filter on the error screen which would prevent any error screens from having its contents displayed.

shane-tomlinson:
@jbonacci - Thanks for all the verifications!  w.r.t. IE7, there is already an issue for that - and I closed it yesterday with a fix. I changed the layout slightly, but that is because IE6 was giving me a pain in the butt.  You can see the fixed versions in the issue: #892

shane-tomlinson:
@jbonacci - as requested, I removed the contractions in checkin d19639961ecd4606c3d1531d2948c1714970edf8

PR #1088, opened 2012-02-10
Hotfix 1087 empty unsupported dialog

shane-tomlinson:
1 line CSS fix to ensure that the unsupported browser dialog message is always shown.

issue #1087


lloyd:
scratch that!  I cherry picked this commit into dev, didn't see improvement, and did not merge it into `train-2012.02.02`.

See conversation on irc: http://irclog.gr/#show/irc.mozilla.org/identity/54571

Issue #835, opened 2012-01-04
Generic error message when authenticating with cookies disabled.


chrismore:
See more here:

https://bugzilla.mozilla.org/show_bug.cgi?id=715017

chrismore:
Bug contents:

Created attachment 585597 [details]
BrowserID error with cookies disabled on iOS

If you have cookies disabled when you try to authenticate on BrowserID, you get an ambiguous error message.

To reproduce:

1) Disable cookies in your browser
2) Log in to Mozillians or Affiliates with BrowserID

Expected:

Error message about needing cookies to log in.

Actual:

Generic error message. (https://bugzilla.mozilla.org/attachment.cgi?id=585597)

shane-tomlinson:
@lloyd, @benadida - Since all of our cookies are HttpOnly, is there a way we can know if cookies have been enabled on the user's device?

shane-tomlinson:
Also related to #374

benadida:
@shane-tomlinson the way would be to set a flag cookie on first resource load and check that flag cookie on session_context, which could return a hasCookies: false if the flag cookie is not there.

benadida:
@shane-tomlinson another solution is to set a flag cookie that is not http-only, since it's just a flag, and check that in JS.

shane-tomlinson:
Firefox with cookies disabled does not even get the popup - it throws an exception in include.js when trying to see if window.localStorage exists.

https://skitch.com/set117/8yqhs/browserid-cookies-disabled-exception

shane-tomlinson:
Chrome with cookies disabled opens the popup, but blows up when trying to check authentication - QUOTA_EXCEEDED_ERR when trying to sync to localStorage.

See https://skitch.com/set117/8yq92/chrome-cookies-disabled

shane-tomlinson:
Fennec with cookies disabled acts like Firefox desktop, the dialog never appears.

Safari Desktop, Opera Desktop, and Safari iOS work as expected.

shane-tomlinson:
@lloyd - While testing cookies disabled support IE, I had the following exception thrown by the server:

```
browserid (22327): info: GET /wsapi/address_info?email=testuser%40testuser.com 200 1ms
browserid (22327): info: forwarding request: /wsapi/stage_user -> http://172.16.84.106:10004/wsapi/stage_user
dbwriter (22322): info: POST /wsapi/stage_user 200 3ms
[...]

PR #1103, opened 2012-02-13
Adding cookies/localStorage disabled support.

shane-tomlinson:
- Removing the explicit cookie_check at the start of the dialog site.
- Remove cookie_check.js from tests, build script, layout.
- Removing cookieCheck code from network.js and tests.
- In include.js, check for disabled localStorage - takes care of Chrome and Firefox disabling access to localStorage when cookies are disabled.
- Add a new "localstorage_disabled" view.
- Add backend tests for the existence, cachability of "localstorage_disabled"
- Add a case in error.ejs for cookies being disabled, which is indicated by a 403 status.

issue #835


shane-tomlinson:
I'm not happy with this pull request, closing for now.

PR #1106, opened 2012-02-14
Beefing up the cookies disabled check so that it handles almost all cases without user interaction.

shane-tomlinson:
- Add a new view, cookies_disabled - this is displayed to browsers that do not allow access to localStorage when cookies are disabled (Firefox, Chrome, Fennec)
- Add localStorage check in include.js to check for browsers that do not allow write access to localStorage.
- Add a Javascript cookie check in network.js->cookiesEnabled using a test cookie with duration of 1 second so it is never sent to the server.
- Remove the old cookies_enabled check that was theorized to come from session_context - remove this from the XHR mock as well.

issue #835

### DIFF
--- a/resources/static/dialog/views/error.ejs
+++ b/resources/static/dialog/views/error.ejs
@@ -11,7 +11,7 @@
     <h2 id="error_403">
       <%= gettext("BrowserID required cookies") %>
     </h2>
-    <%= gettext("Please close this window, <a target='_blank' href='http://support.mozilla.org/en-US/kb/Websites%20say%20cookies%20are%20blocked'>enable cookies</a> and try again") %>
+    <%= format(gettext("Please close this window, <a target='_blank' href='%s'>enable cookies</a> and try again"), ['http://support.mozilla.org/en-US/kb/Websites%20say%20cookies%20are%20blocked']) %>
   <% } else { %>
     <h2 id="defaultError">
       <%= gettext("We are very sorry, there has been an error!") %>
--- a/resources/static/dialog/views/error.ejs
+++ b/resources/static/dialog/views/error.ejs
@@ -9,27 +9,31 @@
     </h2>
   <% } else if (typeof network !== "undefined" && network.status == 403) { %>
     <h2 id="error_403">
-      <%= gettext("We are sorry, BrowserID required cookies") %>
+      <%= gettext("BrowserID required cookies") %>
     </h2>
-    <%= gettext("BrowserID requires your browser's cookies to be enabled to operate. Please enable your browser's cookies.") %>
+    <%= gettext("Please close this window, <a target='_blank' href='http://support.mozilla.org/en-US/kb/Websites%20say%20cookies%20are%20blocked'>enable cookies</a> and try again") %>
   <% } else { %>
     <h2 id="defaultError">
       <%= gettext("We are very sorry, there has been an error!") %>
     </h2>
   <% } %>
 
-  <p>
-  <% if (typeof dialog !== "undefined" && dialog !== false) { %>
-    <%= gettext("To retry, you will have to reload the page and try again.") %>
-  <% } else { %>
-    <%= gettext("To retry, you will have to close this window and try again.") %>
+  <% if (!(typeof network !== "undefined" && network.status == 403)) { %>
+    <p>
+      <% if (typeof dialog !== "undefined" && dialog !== false) { %>
+        <%= gettext("To retry, you will have to reload the page and try again.") %>
+      <% } else { %>
+        <%= gettext("To retry, you will have to close this window and try again.") %>
+      <% } %>
+    </p>
   <% } %>
-  </p>
 
   <% if(typeof action !== "undefined" || typeof network !== "undefined") { %>
-    <a href="#" id="openMoreInfo">
-      <%= gettext("See more info") %>
-    </a>
+    <p>
+      <a href="#" id="openMoreInfo">
+        <%= gettext("See more info") %>
+      </a>
+    </p>
 
     <ul id="moreInfo">
       <% if (typeof action !== "undefined") { %>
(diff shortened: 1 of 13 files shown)


resources/static/dialog/views/error.ejs
```diff
@@ -7,24 +7,33 @@
...
       <%= gettext("We are very sorry, there has been an error!") %>
     </h2>
   <% } %>
 
-  <p>
-  <% if (typeof dialog !== "undefined" && dialog !== false) { %>
-    <%= gettext("To retry, you will have to reload the page and try again.") %>
-  <% } else { %>
-    <%= gettext("To retry, you will have to close this window and try again.") %>
+  <% if (!(typeof network !== "undefined" && network.status == 403)) { %>
```
shane-tomlinson, line 21: Since we have "please close this window, ..." above, I am supressing the duplicate "action" message.

resources/static/dialog/views/error.ejs
```diff
@@ -7,24 +7,33 @@
...
+      <% } %>
+    </p>
   <% } %>
-  </p>
 
   <% if(typeof action !== "undefined" || typeof network !== "undefined") { %>
-    <a href="#" id="openMoreInfo">
-      <%= gettext("See more info") %>
-    </a>
+    <p>
```
shane-tomlinson, line 32: Surrounded this with a <p> in case the "To retry, you will have to..." text is supressed - this shows up next to the "cookies disabled" action text otherwise.

resources/static/include_js/include.js
```diff
@@ -840,30 +840,55 @@
...
+          // cookies are disabled, some browsers (Chrome) will except here.
+          win['localStorage'].setItem("test", "true");
+          win['localStorage'].removeItem("test");
+        }
+        else {
+          // Browser does not have local storage.
+          return "LOCALSTORAGE_MISSING";
+        }
+      } catch(e) {
+          return "LOCALSTORAGE_DISABLED";
```
shane-tomlinson, line 869: Made the error codes more explicit for easier grokking.

resources/static/shared/network.js
```diff
@@ -570,8 +568,19 @@ BrowserID.Network = (function() {
...
+      // Make sure we get context first or else we will needlessly send
+      // a cookie to the server.
       withContext(function() {
-        complete(onComplete, cookies_enabled);
+        try {
+          // set a test cookie with a duration of 1 second.
+          // NOTE - The Android 3.3 default browser will still pass this.
+          // http://stackoverflow.com/questions/8509387/android-browser-not-respecting-cookies-disabled/9264996#9264996
+          document.cookie = "test=true; max-age=1";
+          var enabled = document.cookie.indexOf("test") > -1;
```
shane-tomlinson, line 579: @ozten - Unfortunately, adding the additional test (for max-age) returned true for Android 3.3's default browser.

ozten, line 579: So much for un-tested advise ;)

shane-tomlinson, line 579: @ozten - Strangely enough, I tried reading the cookie value in Android 3.3 browser with cookies disabled.  The value comes back.  Whack.

ozten:
Looks good. r+

resources/views/cookies_disabled.ejs
```diff
@@ -0,0 +1,17 @@
...
+
+  <section id="error" class="cookies_disabled">
+      <div class="table">
+        <div class="vertical contents">
+          <h2 id="reason">
+            <%= gettext("BrowserID requires cookies") %>
+          </h2>
+
+          <p>
+            <%- gettext("Please close this window, <a target='_blank' href='http://support.mozilla.org/en-US/kb/Websites%20say%20cookies%20are%20blocked'>enable cookies</a> and try again") %>
```
ozten: You want to keep as much gunk out of a string as possible. I'd suggest:

<p class='no-cookies'>
<%- format(gettext("Please close this window, <a target='_blank' href='%s'>enable cookies</a> and try again",
                   'http://support.mozilla.org/en-US/kb/Websites%20say%20cookies%20are%20blocked') %>
</p>


Is target critical? Can we add this attribute via JavaScript? `$('#no-cookies a').attr('target', '_blank')`. We can leave it in to KISS, of course, but URLs should never be in strings, as they might change.

ozten:
r-
New string shouldn't have a link in it. Otherwise, great work!

resources/views/cookies_disabled.ejs
stomlinson: @ozten - great point, I hadn't thought of that.  Will update.

stomlinson: @ozten - I removed the URL from the strings and now use "format" to insert them.  I kept the target in the string because the cookies_disabled dialog as a whole does not have any javascript included and including JS to add a target attribute seems like overkill.  But, I am thinking I can take this a step further.  Would it be alright to take out both the target and href attributes and replace them with a single "%s"?

For instance:

```
<a target='_blank' href='http://support.mozilla.org/en-US/kb/Websites%20say%20cookies%20are%20blocked'>enable cookies</a>
```

would change to

```
<a %s>enable cookies</a>
```

with the value for the %s replacement being `target='_blank' href='....`

That way the translators have very little to worry about, but the target="_blank" can still be inserted.

resources/static/dialog/views/error.ejs
```diff
@@ -7,24 +7,33 @@
     <h2 id="error_503">
       <%= gettext("We are very sorry, the server is under extreme load!") %>
     </h2>
+  <% } else if (typeof network !== "undefined" && network.status == 403) { %>
+    <h2 id="error_403">
+      <%= gettext("BrowserID required cookies") %>
+    </h2>
+    <%= format(gettext("Please close this window, <a target='_blank' href='%s'>enable cookies</a> and try again"), ['http://support.mozilla.org/en-US/kb/Websites%20say%20cookies%20are%20blocked']) %>
```
ozten, line 14: This looks good. @mathjazz is always good to get a code review from early on for tricky strings.

mathjazz, line 14: My name is @mathjazz, and I approve this message. :)

PR #1107, opened 2012-02-14
Cherry picking the cookies enabled check and test from the feature_835_cookies_disabled branch.

shane-tomlinson:
This is to inform iOS users whose cookies have been disabled during a sync that they need to turn their cookies on.

close #1056


shane-tomlinson:
@ozten - thanks for the tip on checking for "max-age" - unfortunately Android still passes it.  I am going to try to read the cookie value in the other cookie related-pull request that I have.

@ozten, @lloyd - can this be added to this week's train before it goes to beta?

PR #1114, opened 2012-02-14
Fix for Issue 620

ozten:
This PR adds connect-cachify as a dependency.

It simplifies our layout ejs to use a cachify_js and cachify_css helper. The list of assets are managed in lib/static_resources.js.

In production, a single asset file will be served up with /v/<md5 hash>/original/path and expires headers of 1 year.

In development, multiple asset files will be served up with their paths. No special expires headers are sent (just static middleware headers).

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
         "convict": "0.0.6",
         "cjson": "0.0.6",
         "client-sessions": "0.0.3",
+        "connect-cachify": "0.0.4",
         "connect-cookie-session": "0.0.2",
         "connect-logger-statsd": "0.0.1",
         "ejs": "0.4.3",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -27,6 +28,7 @@
         "temp": "0.2.0",
         "uglify-js": "1.0.6",
         "uglifycss": "0.0.4",
+        "underscore": "1.3.1",
```
benadida, line 31: +1 to using underscore more thoroughly everywhere! Much nicer code.

benadida:
this patch looks great to me, especially the way it cleans up the templates.

shane-tomlinson:
+r - backend tests pass, from my understanding of the backend looks good.  

Prod mode requests have md5 hash.
Debug mode requests do not have md5 hash.
External file should enable easy inclusion into a compressor.

ozten:
Thanks! landed on dev.

PR #1120, opened 2012-02-15
Adding a 20px left and right margin to the delay, wait, and error screens

### LINKED ISSUE #1117: [l10n] margin needed for confirmation text in /sign_in, e.g., German
Minor nit, but in some languages (e.g., German) the layout of the confirmation message in /sign_in dialog could use a bit of margin.

img

shane-tomlinson:
This prevents long text from butting directly up to the edges in these screens. 

close #1117


benadida:
r+, merging.

PR #1136, opened 2012-02-16
Removing the "text-transform: lowercase" on all buttons.

### LINKED ISSUE #1113: Remove altering copy case via CSS
As mentioned in https://github.com/mozilla/browserid/issues/1105#issuecomment-3960781

It's a best practice not to mess with changing the case of strings via CSS. Sounds like some buttons we force the case to all lowercase.

We should change to lower case in English by editing the po file. We should remove any CSS that alters copy in this way.

shane-tomlinson:
@ozten - can you review/merge this one?

close #1113

### DIFF
--- a/resources/static/css/common.css
+++ b/resources/static/css/common.css
@@ -146,7 +146,6 @@ button,
     font-family: 'Droid Serif', Georgia, serif;
     color: #fff;
     text-shadow: -1px -1px 0 #37A6FF;
-    text-transform: lowercase;
     cursor: pointer;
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@ozten, @mathjazz - To go along with this change, I edited the .po files in en_US to change button copy to lowercase.  The four changes made are:

messages.po

Finish->finish

client.po

Reset Password->reset password
Cancel->cancel
Verify->verify

I have not yet run compile-mo.sh or compile-json.sh, I would like to make sure I did this correctly first.

PR #1137, opened 2012-02-16
iOS - when clicking one mail address, select radio button.

### LINKED ISSUE #1133: iOS - must click on radio button to select email address.
Expected - should be able to click the email address to select the radio button.

shane-tomlinson:
- Add a handler to the label surrounding the radio button.  If clicking on label, select address.
- Add a test to check for requirement.

@lloyd - I know you have an iPhone and iPad, can you check this?

close #1133

### DIFF
--- a/resources/static/dialog/controllers/pick_email.js
+++ b/resources/static/dialog/controllers/pick_email.js
@@ -65,6 +65,13 @@ BrowserID.Modules.PickEmail = (function() {
     return identities;
   }
 
+  function selectEmail(event) {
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@ozten - that is indeed the case.  When you click on a label in the iPhone, it does not select it's corresponding radio button.

lloyd:
I'll deploy and test shortly, assigned to me.

shane-tomlinson:
@lloyd - thanks for the review/merge.

PR #1138, opened 2012-02-16
Changing the case of the buttons from upper to lowercase.

shane-tomlinson:
@ozten - mind checking this one out?

issue #1113


shane-tomlinson:
@ozten - thanks for the review and merge!

PR #1143, opened 2012-02-17
Upgrading to connect-cachify 0.0.5 which has a bug fix for Issue#1141

ozten:
See https://github.com/mozilla/connect-cachify/commit/b85c91d5ce7f6fc4e7682e1091c24d6fb19afe29 for actual diff.

Tested on Fx, Chrome, iOS 5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
         "convict": "0.0.6",
         "cjson": "0.0.6",
         "client-sessions": "0.0.3",
-        "connect-cachify": "0.0.4",
+        "connect-cachify": "0.0.5",
         "connect-cookie-session": "0.0.2",
         "connect-logger-statsd": "0.0.1",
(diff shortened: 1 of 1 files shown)


lloyd:
r+ - my only concern is noted in the commit.  I'm curious about what the override semantics should be of overriding cache-control headers set by middleware.  the way it's implemented now, it seems like if middleware is installed after cachify that sets cache-control, it will overwrite the headers specified by cachify.

it might be nice if cachify always set the cache-control header upon res.emit('header'), which would cause it always to win, then optionally you can add configuration to change that behavior later (only if not present, or don't touch my headers)

ozten:
Agreed, mucking with headers after the fact is kind of "magic" that can bite us. I'll update the docs to note behavior and noodle on how to best let a client programmer override this module.

PR #1147, opened 2012-02-17
refactor httputils to remove needless logic and better leverage express....

lloyd:
...  easier to read and maintain.  also remove unused and unnec functions

### DIFF
--- a/lib/httputils.js
+++ b/lib/httputils.js
@@ -5,63 +5,33 @@
 // various little utilities to make crafting boilerplate responses
 // simple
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Nice cleanup r+.

ozten:
npm test pass

QUnit test failure, but I don't see how this branch could be causing it.
radio button is correctly selected
Expected:  
true
Result: 
false
Diff:  
true false 
Source: 
    at Object.<anonymous> (http://10.0.1.16:10002/test/cases/controllers/pick_email.js:167:5)

PR #1148, opened 2012-02-17
disable /test at the app level in stage or production.

lloyd:


lloyd:
good point, @ozten.  embedding diresworb.org and browerid.org in various places in our code is bad.  

I did want, however, tests to be possible to run in dev and aws environments.  We could take a moment and define what environments we have and where they are used.  maybe we want "local", "dev", and "production"?  then:

```
app.configuration("local", "dev", ...); 
```

would give us the ability to make this configuration specific to local and aws, without having it run in production or stage.  the thing I don't like about this is that "production" is special and according to expressjs.com:

```
"This is very important, as many caching mechanisms are only enabled when in production."
```

I felt we'd want said caching mechanisms in our aws envs (so `NODE_ENV=production`), but then there's application level stuff we'd want in aws but not in stage (like tests).

maybe we should open a new issue to figure this out.

lloyd:
no rush on this, so I'll just let it sit here.  open questions:

@shane-tomlinson: does the added commit address everything?

@ozten: do you think we should move how we centrally represent "deployment type" into a new issue?

shane-tomlinson:
@lloyd - this addresses my concerns.  r+ for my concerns.  If @ozten's concerns are met, good to go!

ozten:
Don't worry about a new bug, I just wanted to understand. Thanks!

Feel free to merge. I'll always put a "r-" if I don't think something is ready to merge.

Issue #423, opened 2011-10-14
Sign-in popup has no scrollbars → confusing when zoomed in


aleksejrs:
If the zoom in ratio for the browserid.org is big, some content of the pop-up window is not visible. That’s very confusing, because there are no scrollbars.

lloyd:
firefox seems to not add scrollbars, chrome does.

lloyd:
for the record, what browser and OS were you on?

aleksejrs:
Firefox (probaby 8.0 Beta) on GNU/Linux.

lloyd:
I too can repro this on Linux, arch Linux Firefox.  This really sounds like a platform specific browser bug.  Let's look for this upstream in bugzilla...

shane-tomlinson:
@lloyd - in include.js, we have scrollbars=0 - suppressing scrollbars.

In MDN's window.open article under Notes on Scrollbars - https://developer.mozilla.org/en/DOM/window.open#Note_on_scrollbars, it notes 

```
user sets a minimum font size for pages which is bigger than the font-size the 
web author requested. People over 40 years old or with particular viewing habit 
or reading preference often set a minimal font size in Mozilla-based browsers.
```

Under the scrollbars option description, it notes 

```
Tip: For accessibility reasons, it is strongly encouraged to set this feature always to yes.
```

I experimented with setting `scrollbars=1` in OSX, Windows and Linux and all current browsers act sane w.r.t. when it displays the slider within the scrollbar.  All browsers show a scrollbar placeholder if the scrollbar is not needed, but show the slider when necessary.

I think we should enable the scrollbars for users who have their browser zoomed in.

lloyd:
If we don't have scrollbars in the default (non-zoomed) view in any of the browsers we support, then I'm +1

shane-tomlinson:
Screen shots:

https://skitch.com/set117/8n1pu/win7-firefox-unzoomed
https://skitch.com/set117/8n1qm/win7-firefox-zoomed

https://skitch.com/set117/8n1qh/safari-osx-unzoomed
https://skitch.com/set117/8n1xi/safari-osx-zoomed

https://skitch.com/set117/8n1tb/ubuntu-chrome-unzoomed
https://skitch.com/set117/8n1tx/ubuntu-chrome-zoomed

shane-tomlinson:
Update on this:

Testing various combinations of
resizable=0/1
scrollbars=0/1

In OSX

In Windows:
[...]

PR #1149, opened 2012-02-17
Fix zoom/scrolling issues across platforms.

shane-tomlinson:
- Set both resizable and scrollbars to 1 for users who have their browsers zoomed.
- Set the body's overflow-y to auto so no scrollbars are shown unless they are needed.

issue #423


lloyd:
r+ - given shanes extensive testing and discussion on IRC, I think we get this into dev and verify it.

Issue #871, opened 2012-01-11
The "keep me signed in" option disappears from the Sign In dialog when refreshed after removing an email address


ghost:
Environments:
1: HTC Desire A8181 + Android 2.2.2 + Native browser
2: Windows 7 + Firefox 10 (beta3)

STR:
1. Launch http://beta.myfavoritebooze.org in the browser.
2. Click on the Sign In button.
3. Log in to a BrowserID account with multiple email addresses.
4. Leave the Sign in dialog with the email addresses list open.
5. Open https://diresworb.org and remove one email address from the logged in account.
6. Go back to the Sign In dialog, select the removed email address and click on the Sign In button.

An error message is displayed and the Sign In dialog is refreshed. After the refresh, the "keep me signed in" option is missing from the dialog.

PR #1151, opened 2012-02-18
Fix the issue #871

ghost:
Rather than invoking the doPickEmail state directly, publish pick_email
to preserve the original dialog options.


shane-tomlinson:
@fawek - nice pull request, thanks for the included test - perfect solution.  r+

PR #1152, opened 2012-02-19
Sync the e-mail list before attempting to remove an e-mail

### LINKED ISSUE #1123: When the user has 2 email addresses, he can cancel his BID account by removing the same address twice
OS: Windows 7
Browser: Firefox 11.0

STR:
1. Open https://diresworb.org/ in the browser.
2. Sign in with an account that has 2 email addresses associated to it.
3. Leave the page open on the Account Manager page.
4.  Open https://diresworb.org/ in another tab of the browser (the user may have multiple tabs and groups of tabs open so he might forget he already opened the page).
5. Remove the 2nd email address from this account and close the tab you removed it from.
6. Go to the diresworb.org tab that remained open. Don't reload it.
7. Try to remove the same email.

The following error message is displayed:
"Removing the last address will cancel your BrowserID account.
Are you sure you want to continue?"

If the user doesn't pay enough attention to the message and clicks "OK", his BID account is canceled although he never selected to remove the first email address associated to the account.

This issue only reproduces when the user has exactly 2 email addresses associated to BID.

ghost:
Fixes #1123.

This is probably a suboptimal fix. Ideally, there would be an API call that removes an e-mail and cancels the account (or just make it the default behavior) iff there are no other addresses but it would be hard to combine this with the UI confirmation step. Instead of a confirmation dialog, there could be a small footnote in the Account manager that removing the last e-mail is synonymous with deleting the account. In that case, you could just do the "remove the e-mail and the account" atomically on the server without having to check again with the user.

### DIFF
--- a/resources/static/pages/manage_account.js
+++ b/resources/static/pages/manage_account.js
@@ -39,33 +39,37 @@ BrowserID.manageAccount = (function() {
   }
 
   function removeEmail(email, oncomplete) {
-    var emails = user.getStoredEmailKeypairs();
-
     function complete() {
       oncomplete && oncomplete();
     }
 
-    if (_.size(emails) > 1) {
-      if (confirmAction("Remove " + email + " from your BrowserID?")) {
-        user.removeEmail(email, function() {
-          displayStoredEmails(oncomplete);
-        }, pageHelpers.getFailure(errors.removeEmail, oncomplete));
+    user.syncEmails(function() {
+      var emails = user.getStoredEmailKeypairs();
+      if (!emails[email]) {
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@fawek - Thanks for this request - I will review this further tomorrow (22/02/2012)

benadida:
@shane-tomlinson should we merge this? I'll let you review, just want to make sure we're merging pull requests :)

shane-tomlinson:
@fawek - Thanks for this!  Looks good, I tested it, and thanks for the additional tests.  r+

PR #1153, opened 2012-02-19
Can't uncheck the persistence check box

ghost:
Adjust the CSS selector in pick_email.js to only match e-mail radio buttons

### DIFF
--- a/resources/static/dialog/controllers/pick_email.js
+++ b/resources/static/dialog/controllers/pick_email.js
@@ -64,11 +64,14 @@ BrowserID.Modules.PickEmail = (function() {
     return identities;
   }
 
-  function selectEmail(event) {
+  function selectInput(event) {
     var target = dom.getAttr(event.currentTarget, "for");
-    if(target) {
-      dom.setAttr("#" + target, "checked", "checked");
+    if (target == "remember" && dom.hasAttr("#" + target, "checked")) {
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
Link to #1155 for tracking.

shane-tomlinson:
@fawek - Awesome, thanks for the contribution here - esp the tests.

The original reason for using Javascript to select the radio buttons is because iOS does not select the radio buttons when clicking on the corresponding labels.  It looks like check boxes suffer from this as well.

Would it be possible for you to expand your pull request so that when clicks on the "Always sign in using this email" label are received, it toggles the checkbox on iOS?

Thanks again!

ghost:
Thanks for the review! I can't test this on iOS but this should work.

shane-tomlinson:
@fawek - Thanks for the update!  I did some cross browser/device testing and came up with some strangeness - namely, you can no longer click on the radio or checkbox itself with any effect.  Labels are all good.

I have added a bunch of unit tests and am going to have a deeper look tomorrow.

shane-tomlinson:
@fawek - Thanks for this pull request.  I used a modified version of this request along with some additional tests for pull request 1163 - https://github.com/mozilla/browserid/pull/1163.

Thanks again!

ghost:
@shane-tomlinson - Awesome, yours looks much cleaner!

Issue #1160, opened 2012-02-20
remove target="<x>" from template strings.


shane-tomlinson:
This makes more work for the translators, replace this with a %s.

jrgm:
verified fixed in code and localization strings on stage train-2012-03-03. Noted one other place to cleanup in GH-1269

PR #1161, opened 2012-02-20
Removing the href/target attributes from translated anchor tags.

shane-tomlinson:
@ozten, @mathjazz - can you guys review this?

Instead, translators will see a %s and the href and target attributes will be added dynamically.

issue #1160

### DIFF
--- a/resources/static/dialog/views/error.ejs
+++ b/resources/static/dialog/views/error.ejs
@@ -11,7 +11,7 @@
     <h2 id="error_403">
       <%= gettext("BrowserID required cookies") %>
     </h2>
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
@ozten - Thanks for the tip, do you think it will be valuable to go back and change the `'`s to `"`s?

jrgm:
verified in code. Whether to convert ' to " can be a separate issue. Two other lines that could use conversion are
#: resources/static/dialog/views/confirm_email.ejs:18  
msgid ""
"If this is a mistake, just ignore the sent email and <a href=\"#\" id="
"\"cancel\">cancel</a>."
msgstr ""

#: resources/static/dialog/views/confirm_email.ejs:20
msgid ""
"If this is a mistake, just ignore the sent email and <a href=\"#\" id=\"back"
"\">use another email address</a>."
msgstr ""

But I'll file that separately.

PR #1184, opened 2012-02-24
Strings

mathjazz:
Two changes:
- Fixing typo (comunicating, only one m)
- Using ' instead of &#39;

(No idea why 6 commits are showing up instead of 2).


mathjazz:
Oh, we already pulled 4 of these commits. Only Typo and Unescape apostrophe will make change.

jrgm:
verified in train-2012.03.03

PR #1187, opened 2012-02-24
Hotfix 1170 graduating banner

shane-tomlinson:
@lloyd - can you review this?

I have taken the original train-2012.02.16 from browserid_private and made CSS changes to ensure that the banner is both the correct size and does not get pushed under the header.

It shows up as 3 commits because this includes the original hotfix commits to the other repo.

The relevant changes are in style.css and m.css, 3 sections were added/modified.  The relevant changes can be seen in: b40442b5fd2de8f58c343c271514ef4ca1fd8381

Things to test:
1) Banner looks right in both mobile and desktop.
2) Banner never goes under the header.
3) When window is narrow, banner expands to 2 or 3 lines as needed.
4) Sign in/sign up/manage/verify email/user are not affected.

When complete, can you cherry-pick b40442b into dev?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -38,6 +38,7 @@ train-2012.02.02:
   * (hotfix 2012.02.07) Modify build process to pick up locales from a .json file
   * (hotfix 2012.02.07) fix production-locales.sh script to defer to the environment for configuration
   * (hotfix 2012.02.13) fix for IE users not seeing error screens sometimes: #1087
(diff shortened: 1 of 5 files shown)


lloyd:
this looks fine to me.  I'll merge into dev, review layout there, and if that looks good merge this into 2012.02.16

lloyd:
looks good in dev.  r+.  merging.

Issue #1188, opened 2012-02-24
Almost identical strings


mathjazz:
As pointed out by localizers, these two strings are almost identical:

resources/static/dialog/views/offline.ejs:9
We are sorry, but we cannot communicate with BrowserID while you are offline.

resources/static/shared/error-messages.js:83
Unfortunately, BrowserID cannot communicate while offline!

I think we should pick (the first) one and use it everywhere.

shane-tomlinson:
@mathjazz - @lloyd actually disconnected the "offline" check quite a while back, this code should probably be removed all together.  Let me make sure this is the case, and if so, I'll do a pull request.  If the check is still active, I will combine the strings.

lloyd:
looks fixed!  please test

jrgm:
looks gone to me

PR #1190, opened 2012-02-24
Removing the offline check code that was already disabled.

shane-tomlinson:
@lloyd - can you review this one?  You initially removed the check with checkin d31a50ddee5f14a6d258af0e6faf80946188c3a6, I am removing the rest of the remnants here.
- Remove checks, tests and strings.

issue #1188


lloyd:
r+ - looks pretty straightforward.  The only thing further I'd do is grep around for offline post-merge.

Issue #1154, opened 2012-02-19
To retry, try again


mathjazz:
These two strings should be changed probably.
# : resources/static/dialog/views/error.ejs:18

msgid "To retry, you will have to reload the page and try again."
# : resources/static/dialog/views/error.ejs:20

msgid "To retry, you will have to close this window and try again."

/cc @ozten, @skinny97214

ozten:
Changed how? Make them the same string?

@shane-tomlinson do these serve two different purposes?

ozten:
Ah from dev-l10n-web Julen Ruiz Aizpuru says:

Hi,

I have some doubts regarding two strings:

"To retry, you will have to close this window and try again."
"To retry, you will have to reload the page and try again."

From my understanding, "retry" and "try again" means the same thing,
so unless the context is shown (which is not) I'm not sure what does
this exactly mean.

Any hints?

mathjazz:
No no, it sounds kind of funny, doesn't it?

Retry and try again is the same thing.

shane-tomlinson:
@ozten, @mathjazz - These strings DO serve different purposes - "close the window" is used for errors within the dialog, "reload the page" is for errors within the main site.

Better copy needed yes, but I think we should keep two distinct strings.

mathjazz:
Absolutely, these are two different things.

I was talking about using "to retry" and "try again" in the same sentence.

shane-tomlinson:
See bug #1167, closing as duplicate.

shane-tomlinson:
"To retry, ... try again" is shown in all the error messages _except_ the cookies disabled message.  Because of this, reopening this issue.

lloyd:
merged.

PR #1191, opened 2012-02-24
Fix the redundant and circular "To retry, ... try again"

shane-tomlinson:
Getting rid of the "To retry, ... and try again" and instead using "Please ... and try again."

issue #1154


lloyd:
r+

PR #1193, opened 2012-02-24
Fix Java Fennec not opening the BrowserID dialog.

shane-tomlinson:
@lloyd - you are the best to check this, it involves changes to WinChan.
- The UA string between XUL and Java versions of Fennec have changed, must now sniff for both strings.

As per :mbrubeck's suggestion in https://bugzilla.mozilla.org/show_bug.cgi?id=730015, I am sniffing for both "Fennec/" by itself and "Firefox/" && "Android".

issue #1176

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -656,7 +656,10 @@
     // checking Mobile Firefox (Fennec)
     function isFennec() {
       try {
-        return (navigator.userAgent.indexOf('Fennec/') != -1);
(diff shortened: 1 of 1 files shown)


benadida:
let's not merge this for now, I believe the user-agent change is a bug in fennec.

benadida:
I was wrong, this user-agent change is permanent, will review this now.

benadida:
r+

Issue #841, opened 2012-01-04
Present site's TOS and privacy policy to user


vthunder:
Some sites require the user to agree to their terms of service and/or privacy policy. There is interest in having the user be aware / agree to them during the sign-in flow.

Note that some sites require explicit agreement (checking a box), so we should allow for that.

I don't think we need to present the actual TOS/PP, links to them (that open in new pages) should be sufficient.

vthunder:
Crystal - we'll need some UX help here to make this not suck :-)

In particular, we want to make it as clear as possible that the TOS/PP are for the site the user is signing into, _not_ for BrowserID itself.

benadida:
OK, we have a path forward here:
- design API for RPs to send privacy and ToS URLs to the dialog
- dialog shows "by clicking sign-in, you agree to the ToS and privacy policy" with two links (or just one if only one is provided by RP)
- when we prompt the user for a password, we also note agreement to BrowserID ToS and privacy policy (not the RP).

That is all for now, should do this in time for the next train.

skinny97214:
Popup type email http://cl.ly/1M3x3q1d1i432L1p1T3m
Popup select email http://cl.ly/3k1S3v431N3o0Q2L1C3Y

Page http://cl.ly/0X1a3l381m3B0o0v1K2Z

fligtar:
Looks great to me, as long as legal is okay with it. My understanding was that EU requires a checkbox assent. Thanks!

benadida:
@lloyd this is closed right?

PR #1216, opened 2012-02-29
allow RP to specify tos and PP for display to user inside the browserid ...

lloyd:
...dialog.  issue #841


benadida:
Some thoughts:
- no TOS and PP in the assertion. I was expecting them to be there. But this is probably the right call for now. I think we should consider adding this later.
- doesn't work if only a TOS or PP is provided. I think we should make that work, but that can also be deferred, and this patch is okay as is for now.

so, r+.

PR #1221, opened 2012-02-29
Adding script to check if all locales have committed to SVN

mathjazz:
Most of the localiers work with our web-based tool Verbatim:
https://localize.mozilla.org/projects/browserid/

Some of them keep forgetting to commit their translations to SVN when they're done.

I need this script to quickly check if all of the locales that are ready for shipping also committed their work to the repo from which we're taking translations.


ozten:
Great work on this tool.

PR #1222, opened 2012-02-29
Handle moz-prefixed navigator.id property for Gecko's prototype native implementation

khuey:
We will want to vendor prefix the prototype implementation in Gecko.  The shim needs to pick up the moz-prefixed property.

### DIFF
--- a/resources/static/authentication_api.js
+++ b/resources/static/authentication_api.js
@@ -7,6 +7,10 @@
 (function() {
   "use strict";
 
+  if (navigator.mozid) {
+    navigator.id = navigator.mozid;
+  }
+
   if (!navigator.id) {
     navigator.id = {};
   }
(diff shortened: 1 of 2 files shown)


benadida:
@khuey thanks for this. I'm going to hold off on merging it for now because I want to make sure we don't break Nightly on production BrowserID. I think we should find some manual way to make this happen so we can test stuff out. But I'll leave this pull request open so we can merge it when the time comes.

lloyd:
I'll deploy this branch in a test env so we can test

lloyd:
apparently I dropped the ball on this.  Is this patch still relevant?  if so I'll test and merge.  Else close.

benadida:
@lloyd I suggest we don't do this yet. We need a strategy for testing whether the .mozId implementation is compatible with what we want before dropping down to it. In a test deployment, sure, but in the pipeline to production, I think it's too early.

lloyd:
@benadida shall we keep the PR open?

PR #1236, opened 2012-03-01
Enhacing best language match to also match a language regardless of loca...

ozten:
...le. Fixes Issue#1128

https://github.com/mozilla/browserid/issues/1128


lloyd:
r+ - there's a lot of discussion on irc about how to make this even better, but this change is def. an improvement.

there's also feelings that this is Good Enough.

http://irclog.gr/#show/irc.mozilla.org/identity/64052

PR #1242, opened 2012-03-02
Integrate Front-End unit tests running under PhantomJS into CI

lloyd:
These changes make it possible to test the following three things individually via `npm test` and some combination of env vars:
1. back end unit tests using mysql
2. back end unit tests using the json db
3. front end unit tests running under phantomjs

Here is verification that tests pass:

http://travis-ci.org/#!/mozilla/browserid/builds/780483

When you merge this, you should close issue 635.


lloyd:
ok, I'll merge now, then go immediately fix GH-1245, updating docs as I go.

Issue #403, opened 2011-10-11
The "This is not me" and "Use a different email" links should be rearranged on the Signing in dialog


ghost:
OS: All
Browser: All

STR:
1. Open http://beta.myfavoritebeer.org.
2. Click on the Sign in button.
3. Enter and email address and a password on sign in.
4. Observe the dialog with the list of emails.

The "This is not me" and "Use a different email" links:
- could be larger
- should be displayed closer to the line separating the footer from the rest of the page
- should the aligned with the email addresses list

benadida:
assigning this to @skinny97214

skinny97214:
http://cl.ly/1R0i0I3r1o3D0c2O1Q0b
- rename "use a different email" to "add another email"
- move "this is not me"

ozten:
I've done this work and submitted for review:
https://github.com/mozilla/browserid/pull/1252

PR #1252, opened 2012-03-05
Changing link copy and position. Fixes Issue#403.

ozten:
Fairly straightforward tweaks to fix
https://github.com/mozilla/browserid/issues/403


lloyd:
r+

PR #1253, opened 2012-03-05
fix deprecation warnings by updating temp and nodes-statsd - issue #1234

### LINKED ISSUE #1234: The "sys" module is now called "util". It should have a similar interface.
any time you see this error on the console, it's because we're using libraries that use deprecated APIs.  We should hunt down all these cases and upgrade the underlying libraries, if we can do so without moving to versions we're uncomfortable with.

lloyd:
please include 'closes #1234' in the merge commit


ozten:
Awesome! Thanks for grabbing upgrades.

jrgm:
verified in stage and my dev environment.

PR #1260, opened 2012-03-06
by default, run all tests during npm test, and emit clear errors when de...

### LINKED ISSUE #1245: change default tests run
once 1242 is merged, `npm test` will only run backend unit tests with the json database driver.  instead we should run:
1. backend tests with json driver
2. backend tests with mysql driver (or a nice message if not available)
3. frontend tests with phantomjs (or a nice message if not installed)

lloyd:
...pendencies (mysql or phantomjs) aren't installed

when merged, closes #1245

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,15 @@ You can stop the servers with a Cntl-C in the terminal.
 
 ## Testing
 
-Unit tests can be run by invoking `npm test` at the top level, and you
-should run them often.  Like before committing code.  To fully test
-the code you should install mysql and have a well permissions `test`
-user (can create and drop databases).  If you don't have mysql installed,
-code testing is still possible (it just uses a little JSON database).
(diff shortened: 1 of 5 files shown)


lloyd:
tests are not passing due to github outages: http://travis-ci.org/#!/mozilla/browserid/builds/807050

this is upstream and out of our control.

jrgm:
verified with train-2012.03.14 in two of my dev environments. By default all tests  (back_mysql, back (json), front (phantomjs)) are run and there are error messages printed when either mysql is not installed/configured, or when phantomjs is not installed. 

This is a comment from the peanut gallery, but for someone new to node, npm and browserid, the error messages could be clearer about what to do about the error messages. Someone should be able to 'git clone; npm install; npm test' and get a quick OK with two messages about mysql and phantomjs that they have two tests that need fixing. (Not npm test ending with a message that gives a sense that things are massively wrong).

Issue #1271, opened 2012-03-08
Support Primary Authority Delegation


ozten:
An organization should be able to delegate Primary authority to another domain.

Example: Mozilla.com could create the following file at `https://mozilla.com/.well-known/browserid`

```
{
    "authority": "browserid.mozilla.org"
}
```

BrowserID should retrieve `https://browserid.mozilla.org/.well-known/browserid` to determine if mozilla.com is a primary.

browserid.mozilla.org should be able to serve as the IdP for several domain names, such as mozilla.com, mozilla.org, mozilla.co.jp, etc.

ozten:
Pull Request:
https://github.com/mozilla/browserid/pull/1281

ozten:
Merged to dev.

lloyd:
this feature should make it to production 2012/03/28

ozten:
We don't currently have an environment.

We could setup a new website, say clortho.org that delegates to vinz.clortho.org. Can we file a bug, if we want that? This will happen naturally with the Mozilla IdP project.

MAX_AUTHORITY_DELEGATIONS has a unit test, but same answer. You'd setup a series of webservers on different domains that delegate to each other. Might be worth testing once outside of the unit tests, but I don't think we want to maintain these servers.

PR #1279, opened 2012-03-08
Issue1205

ozten:
This removes caching. This code will change quite a bit in [Issue 1271](https://github.com/mozilla/browserid/issues/1271), which I've already started based off this branch.

This was a bunch of commits that I rebased into one. Sigh... I don't know why there is a l10n commit from a month ago in here... I've spent a bunch of time trying to remove it, but it is harmless.


ozten:
Great Odin's beard, I managed to make a clean PR in
https://github.com/mozilla/browserid/pull/1280

PR #1280, opened 2012-03-08
Removing caching, fixes Issue#1205

ozten:
Depending on Squid for caching primary's /.well-known/browserid

This removes caching. This code will change quite a bit in Issue 1271, which I've already started based off this branch.

(I deleted PR 1279 which had extra commits)

### DIFF
--- a/lib/primary.js
+++ b/lib/primary.js
@@ -19,10 +19,9 @@ config = require("./configuration.js");
 
 const WELL_KNOWN_URL = "/.well-known/browserid";
 
-// cache .well-known/browserid for six hours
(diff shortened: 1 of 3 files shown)


lloyd:
@ozten: I don't know if this is any better or worse, it was just my first thought at how to go about it.  I'm going to finish reviewing this request and merge if everything seems cool, we can change it later if you feel that putting it in the proxy would be better.

ozten:
As noted in Issue 1205, closing.

https://github.com/mozilla/browserid/pull/1281 is ready for review. I've added the fixes from this PR there too.

PR #1285, opened 2012-03-09
don't log more information than is needed in browserid-metrics.json - st...

### LINKED ISSUE #1283: Referrer info logged to browserid.json should be stripped to origin only.
https://github.com/mozilla/browserid/blob/dev/lib/metrics.js#L77-87

Not stripping causes distinct counts on the "rp" field to not match up with the validator.json "rp" field which is based on the Audience attribute.

lloyd:
...rip everything from referrer except origin - issue #1283

please include `closes #1283` when merging this commit, or manually close the bug.

### DIFF
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -21,7 +21,8 @@ const
 winston = require("winston"),
 configuration = require("./configuration"),
 path = require('path'),
-fs = require('fs');
+fs = require('fs'),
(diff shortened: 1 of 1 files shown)


lloyd:
tests running here: http://travis-ci.org/#!/mozilla/browserid/builds/830464

benadida:
r+

Issue #1254, opened 2012-03-05
server logs should record millisecond resolution timestamps (i.e., the time of day)


jrgm:
Helps when trying to corelate requests spread across multiple servers, and in determining order of failing sequences of requests.

ozten:
Okay, I've added @milliseconds after the UTC time, not wanting to break any parsers that currently scrape [<UTC time>]

dev_bid
browserid (26068): info: [Sat, 10 Mar 2012 00:15:03 GMT]@3786 GET /shared/modules/page_module.js 304 - ms

default_bid
browserid (26000): info: 10.0.1.41 - - [Sat, 10 Mar 2012 00:14:13 GMT]@3517 "GET /shared/modules/page_module.js HTTP/1.1" 304 - "http://10.0.1.32:10002/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:12.0a2) Gecko/20120308 Firefox/12.0a2"

Pull request
https://github.com/mozilla/browserid/pull/1291

ozten:
Done on dev. Let's file a new bug for any logging tweaks. Thanks.

jrgm:
Verified. The timestamp field in the output to /var/browserid/log/*.log is now a javascript toISOString() on stage servers with train-2012.03.14

Issue #1255, opened 2012-03-05
http log format should record connect's :response-time millis


jrgm:

ozten:
I don't know how this works, but would something like pid be useful?

In previous lives we'd use server ip/thread id to help with tracing logs.

jrgm:
The log line formatting lives in https://github.com/senchalabs/connect/blob/master/lib/middleware/logger.js

We use `default` in stage and production, `dev` in dev. I'm just thinking we'd define our own custom formats `default_bid` and `dev_bid` to include the built-in :response-time token in the format string. The time it took to process the request is useful in logs. 

pid doesn't mean much with node (except to post-facto locate a restart in the logfiles, or if you have workers like verifier where the pid is recorded in the logs already).

ozten:
Thanks for all the detail!

Added :response-time to the format:

dev_bid
browserid (26068): info: [Sat, 10 Mar 2012 00:15:03 GMT]@3786 GET /shared/modules/page_module.js 304 - ms

default_bid
browserid (26000): info: 10.0.1.41 - - [Sat, 10 Mar 2012 00:14:13 GMT]@3517 "GET /shared/modules/page_module.js HTTP/1.1" 304 - "http://10.0.1.32:10002/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:12.0a2) Gecko/20120308 Firefox/12.0a2"

Pull request:
https://github.com/mozilla/browserid/pull/1291

ozten:
Done on dev. Let's file a new bug for any logging tweaks. Thanks.

jrgm:
verified done on dev. I'lll file a followup to add :response-time to format 'default_bid'

PR #1291, opened 2012-03-10
Adding dev_bid and default_bid which have millisecond resolution.

ozten:
Fixes Issue #1254 and Issue #1255

### DIFF
--- a/bin/browserid
+++ b/bin/browserid
@@ -135,7 +135,7 @@ app.use(function(req, resp, next) {
 
 var static_root = path.join(__dirname, "..", "resources", "static");
 
-app.use(cachify.setup(assets(config.get('supported_languages')), 
+app.use(cachify.setup(assets(config.get('supported_languages')),
         {
           prefix: 'v',
           production: config.get('use_minified_resources'),
--- a/config/local.json
+++ b/config/local.json
@@ -8,7 +8,7 @@
   "database": {
     "driver": "json"
   },
-  "express_log_format": "dev",
+  "express_log_format": "dev_bid",
   "email_to_console": true
 }
 
--- a/config/production.json
+++ b/config/production.json
@@ -33,7 +33,7 @@
   "debug_lang": "it-CH",
   // locale directory should be overridden
   "locale_directory": "/home/app/code/locale",
-  "express_log_format": "default",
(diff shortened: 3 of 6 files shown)


ozten:
Also @jrgm if you could review, that would be great!

jrgm:
So a couple of things:
- default_bid is missing the addition of :response-time
- I know the comments in connect show ':response-time ms' but the 'ms' seems to me unneccessary
- Do we actually have parsers that are digesting that timestamp? Would they not be fine with .toISOString() which includes the milliseconds as part of it?
- I think you meant .slice(-3), not .slice(-4)

ozten:
I assumed metrics was using these logs (so it might break them?). ISO String lacks a time zone, which is probably fine. I'll make all these changes, thanks for the review!

lloyd:
So looking at production formatted logs, there are two types of messages that make it in.  First we've got webserver logging:

```
{"level":"info","message":"127.0.0.1 - - [Mon, 12 Mar 2012 17:07:18 GMT] \"GET /ver.txt HTTP/1.0\" 200 86 \"-\" \"-\"","timestamp":"12 Mar 17:07:18"}
```

next, we've got programmatic logging:

```
{"level":"debug","message":"connecting to database: browserid","timestamp":"10 Mar 01:59:21"}
```

Rather than adding the millisecond resolution timestamp to the contents of the log message (which only is present in the former, how about we remove the date from the logging string the webserver emits, and change the timestamp present in all messages to include ms?

lloyd:
How about this: 

```
--- a/lib/logging.js
+++ b/lib/logging.js
@@ -37,7 +37,7 @@ var filename = path.join(log_path, configuration.get('process_type') + ".log");

 exports.logger = new (winston.Logger)({
   transports: [new (winston.transports.File)({
-    timestamp: true,
+    timestamp: function() { return (new Date()).toISOString(); },
     filename: filename,
     colorize: true,
```

ozten:
Putting all the feedback together
- Switch to now for timestamp
- Switch timestamp gobally

ozten:
- Removed date-millis, we're good with the global update.

jrgm:
verified on stage train-2012-03-14

Issue #1233, opened 2012-03-01
Lock down locales and verify them 24 hours before prod release


ozten:
We had a hiccup in our prod push, because invalid config was manually entered.

To fix the root cause we shoud:
- Capture final locales in a new config `config/l10n-prod.json`
- Create a prod config verification script
- Freeze locales 24 hours before the planned push

@mathjazz once config/l10n-prod.json exists, you could edit it,  instead of filing bugs and doing etherpad for the final set of locales.

@mathjazz At least 24 hours before push to production, once you've edit that file for the last time, you would ping the l10n community and ask them to verify their locale is working on stage.

How does that sound?

mathjazz:
This sounds cool, updating the config file is actually easier for me than filing bugs and editing Etherpad.

Why do we need this file 24 hours before push to production?

ozten:
It has to be 24 hours, because @fetep builds RPMs that far in advance and QA does a final pass.

Okay, @mathjazz, @fetep please review 
https://github.com/mozilla/browserid/pull/1296

New config - https://github.com/mozilla/browserid/commit/5a6b9dd658c9342a26afb7ab1a9d80ea107eb020#diff-1

Starting with https://github.com/mozilla/browserid/commit/5a6b9dd658c9342a26afb7ab1a9d80ea107eb020#diff-2
will explain the two new scripts.

ozten:
Landed on dev.

PR #1296, opened 2012-03-13
Adding a new config l10n-prod.json to capture shipping locales.

ozten:
Added two new scripts:
- check_l10n_config.js
- every_locale.js

Fixes Issue #1233

### DIFF
--- a/docs/I18N.md
+++ b/docs/I18N.md
@@ -29,6 +29,17 @@ The request object and the response's template context have the following variab
  * gettext, ngettext - Gettext functions
  * format - for string interpolation
 
+## Tools
+
+``scripts/every_locale.js`` - will output a list of every locale under the locale directory. This can
+be copied into ``config/l10n-all.json``.
+
+``scripts/check_l10n_config.js`` - Run with the ``CONFIG_FILES`` environment variable set to the one json
+config file which you'll like to validate. Examples:
+
+    CONFIG_FILES=/home/ozten/Projects/browserid/config/l10n-prod.json node ./scripts/check_prod_l10n.js
+    ONFIG_FILES=/home/ozten/Projects/browserid/config/l10-nall.json node ./scripts/check_prod_l10n.js
+
 ## Debugging
 
 If code is evaluated in node.js (server-side) then node-gettext is providing the string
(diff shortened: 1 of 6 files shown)


docs/I18N.md
```diff
@@ -29,6 +29,17 @@ The request object and the response's template context have the following variab
...
+## Tools
+
+``scripts/every_locale.js`` - will output a list of every locale under the locale directory. This can
+be copied into ``config/l10n-all.json``.
+
+``scripts/check_l10n_config.js`` - Run with the ``CONFIG_FILES`` environment variable set to the one json
+config file which you'll like to validate. Examples:
+
+    CONFIG_FILES=/home/ozten/Projects/browserid/config/l10n-prod.json node ./scripts/check_prod_l10n.js
+    ONFIG_FILES=/home/ozten/Projects/browserid/config/l10-nall.json node ./scripts/check_prod_l10n.js
```
benadida: is that a missing "C" in CONFIG_FILES?

lib/i18n.js
```diff
@@ -216,6 +216,27 @@ exports.localeFrom = localeFrom = function (language) {
 };
 
 /**
+ * Given a locale code, return a language code
+ */
+exports.languageFrom = function (locale) {
+  if (!locale || !locale.split) {
+    return "";
```
benadida, line 223: do you want an empty string here, or a null?

scripts/check_l10n_config.js
```diff
@@ -0,0 +1,44 @@
...
+  process.exit(1);
+}
+
+var config = require('../lib/configuration.js'),
+    error = 0,
+    logged = false;
+
+console.log("Checking ", config.get('supported_languages').length, "languages from ", process.env['CONFIG_FILES']);
+
+config.get('supported_languages').forEach(function (lang, i) {
```
benadida, line 17: any chance config.get() returns non-array value?

benadida:
check comments above, otherwise looks good for merge.

docs/I18N.md
ozten: Sigh. Ya, will update.

lib/i18n.js
ozten, line 223: Not sure how we do low level error handling. My gut is to avoid nulls. This is currently only used by this validation script and the error will be caught at a higher level.

scripts/check_l10n_config.js
ozten, line 17: Good call. I found that convict config layer would give us a default locale if a config was empty, so I think we're good here.

ozten:
Landed on dev. Thanks for the review!

PR #1299, opened 2012-03-13
Issue 1133 email with plus

shane-tomlinson:
@lloyd or @ozten - can you check this?

The original problem stems from DOM id's not being allowed to contain certain characters such as a "+" and jQuery not being able to search for an element with an ID that contains a ".".  Instead of basing the radio button's id on the email address, using a generic "email_" + <number>.  <number> is the index of the address in the list, so each id is unique.  @tonychung's update in issue #1133 is incomplete, an email address containing a `+` was not able to be selected by clicking on the text in any browser/OS - this is not an iOS specific problem.

Things to check:
- email addresses that contain `+` and `.` in the local portion work.
- email addresses can be selected by clicking on the text on iPhones/iPods

Post Merge:
Please cherry-pick or merge into dev and re-close issue #1133.

issue #1133

### DIFF
--- a/resources/static/dialog/views/pick_email.ejs
+++ b/resources/static/dialog/views/pick_email.ejs
@@ -7,10 +7,10 @@
 
   <div id="selectEmail" class="form_section">
       <ul class="inputs">
-          <% _.each(identities, function(item) { var emailAddress = item.address; var cleanedEmail = emailAddress.replace("@","_").replace(".", "_"); %>
+          <% _.each(identities, function(item, index) { var emailAddress = item.address, id = "email_" + index; %>
               <li>
-                  <label for="<%= cleanedEmail %>" class="serif<% if (emailAddress === siteemail) { %> preselected<% } %> selectable" title="<%= emailAddress %>">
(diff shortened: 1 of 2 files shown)


ozten:
r+ Merged to dev, see feedback on commit.

shane-tomlinson:
Thanks for the review @ozten!

The ids are used to associate the labels to the radio buttons.  This connection is needed for several things.  First, _most_ browsers will automatically act as if the user clicked on the radio button itself if they clicked on it's associated label.  For iOS 5 devices, this is not the case so we had to do our trickery of proxying clicks from the label to its associated radio button.  To proxy, I used IDs to associate the two together.  Finally, screen readers use this association to read out which button the user is currently "hovered" over or has selected.

Fortunately, the ids are only used for associating the labels and radio buttons and do not need to be kept in sync with other parts of the system, if that were a requirement, I'd definitely say use a hash.  Do you think it is worthwhile to do that now, or would it be premature?

Thanks again @ozten

ozten:
Perfect. No, I wouldn't switch to a hash, I was spacing out on them being used in the for attribute of the label tag.

Issue #1339, opened 2012-03-27
network.cookiesEnabled can call its onComplete callback twice


shane-tomlinson:
If the browser's cookies are enabled but the onComplete callback to network.cookiesEnabled throws an exception, onComplete will be called a second time with a false status, leading developers (or users) to falsely believe that cookies are disabled.

Steps to reproduce:
See #1323

PR #1340, opened 2012-03-27
Fix cookiesEnabled being possible to call it's onComplete callback twice if onComplete throws an exception the first time.

shane-tomlinson:
If the onComplete callback excepted, the catch block would re-call onComplete with a false status, indicating the cookies were disabled.

issue #1339


lloyd:
r+

PR #1343, opened 2012-03-27
Locales list for train-2012.03.14

mathjazz:


ozten:
Per IRC, we're losing fi, sr, uk.
After merge, I'll add en-US to this list.

PR #1344, opened 2012-03-28
Add the groundwork for browser specific "Enable Cookies" URLs.

shane-tomlinson:
@ozten or @lloyd - Could you guys do a review/give some feedback here?

I have started with browser specific enable cookie URLs, at the moment only differentiating between iOS and the rest of the world.  I would like to make a list of specialized URLs for popular desktop and mobile browsers.

issue #1167
issue #1302

### DIFF
--- a/lib/static_resources.js
+++ b/lib/static_resources.js
@@ -38,6 +38,7 @@ var common_js = [
   '/shared/helpers.js',
   '/shared/screens.js',
   '/shared/browser-support.js',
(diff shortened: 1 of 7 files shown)


ozten:
Great work! This will really help our users. We should ask the SUMO team to work towards supporting OS and Browser detection for flavors of an article.

r+

automated tests run, didn't manually test.

PR #1360, opened 2012-03-29
Only print failing frontend tests

shane-tomlinson:
issue 1326


lloyd:
code looks good, ran locally, output looks good, tests pass.  r+

PR #1365, opened 2012-03-30
Extracting header removal logic to run before etagify middleware

ozten:
Fixes Issue#1353

### DIFF
--- a/lib/browserid/views.js
+++ b/lib/browserid/views.js
@@ -45,9 +45,9 @@ exports.setup = function(app) {
   // Issue#1353 This is kind of dirty, but this is our last chance
   // to fixup headers for an ETag cache hit
   // x-frame-options - Allow these to be run within a frame
-  app.use(function (req, resp, next) {
+  app.use(function (req, res, next) {
     if (req.path === '/communication_iframe') {
-      resp.removeHeader('x-frame-options');
+      res.removeHeader('x-frame-options');
     } else if (req.path === '/relay') {
       res.removeHeader('x-frame-options');
     }
--- a/lib/browserid/views.js
+++ b/lib/browserid/views.js
@@ -41,6 +41,20 @@ function renderCachableView(req, res, template, options) {
 }
 
 exports.setup = function(app) {
+
+  // Issue#1353 This is kind of dirty, but this is our last chance
+  // to fixup headers for an ETag cache hit
+  // x-frame-options - Allow these to be run within a frame
+  app.use(function (req, resp, next) {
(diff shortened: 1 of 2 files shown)


lib/browserid/views.js
```diff
@@ -41,6 +41,20 @@ function renderCachableView(req, res, template, options) {
 }
 
 exports.setup = function(app) {
+
+  // Issue#1353 This is kind of dirty, but this is our last chance
+  // to fixup headers for an ETag cache hit
+  // x-frame-options - Allow these to be run within a frame
+  app.use(function (req, res, next) {
+    if (req.path === '/communication_iframe') {
```
shane-tomlinson, line 49: @ozten - Is it dirty because it is a middleware to override the x-frame-options or just the code?  Maybe to keep things a bit more DRY, you could use an array of path names and .indexOf to see if the current path is a match to remove the header.

```
var paths = ['/relay', '/communication_iframe'];
if(paths.indexOf(req.path) > -1) {
  res.removeHeader('x-frame-options');
}
```

tests/cache-header-tests.js
```diff
@@ -87,7 +87,11 @@ function hasProperCacheHeaders(path) {
     },
     "followed by a request with an if-modified-since cache header, and bogus etag": {
       topic: function(err, r) {
-        var etag = r.headers['etag'].replace(/"$/, "bogus\"");
+        var etag = r.headers['etag'] = '"bogus"';
+        // No ETag present in iframes, make one
+        if (['/communication_iframe', '/relay'].indexOf(path) === -1) {
+          etag = r.headers['etag'].replace(/"$/, "bogus\"");
```
shane-tomlinson, line 104: Does "bogus" get re-set to a real etag?

lib/browserid/views.js
ozten, line 49: I marked it dirty because the code doesn't live in-line with the route, so you have to remeber to look up here when thinking about communication_iframe or relay.

We could do the indexOf style, I did if/else in case in the future there are other headers we need to tweak on a route by route basis.

tests/cache-header-tests.js
ozten, line 104: I think this is the last step in a unit test per path, where we test the behavior of an unknown etag. It won't get reset to a real etag, but we've already run through the logic with real etags several times.

lloyd:
how about a unit test that verifies when iframe content is served with a 304 (if-none-match present) we don't have X-Frame-Options.  Then a test that the converse is true for other content?

ozten:
Will do.

Issue #1364, opened 2012-03-29
etag headers incorred in train-2012.03.28 - breaks l10n


lloyd:
etagify doesn't respect Vary headers, so the result is we serve the same ETag for all content under the same URL despite the fact that it varies by `Accept-Language`.

Demonstration:

```
$ curl -q -H 'Accept-Language:it-CH;' -i https://dev.diresworb.org/sign_in 2> /dev/null | egrep '^ETag' 
ETag: "49285e8b798f3421ac912e901fa6fb9f"
$ curl -H 'Accept-Language:it-CH;' https://dev.diresworb.org/sign_in 2> /dev/null | md5
1775d2eb10a6223ea3caeab45f5166b6
$ curl -i -H 'Accept-Language:en-US;' https://dev.diresworb.org/sign_in 2> /dev/null | egrep '^ETag'
ETag: "49285e8b798f3421ac912e901fa6fb9f"
$ curl -H 'Accept-Language:en-US;' https://dev.diresworb.org/sign_in 2> /dev/null | md5
49285e8b798f3421ac912e901fa6fb9f
```

Notice here that the it-CH content is served with the same ETag header as the en-US content.  Assuming that clients properly respect vary headers, this could cause false 304s in the case where content changes for one locale, but not another.

Practically speaking, given the way our software is constructed, this does not seem like an issue.  However this is ugly and if it ever were to bite us it might cost us a lot of time.

Solution:  fix etagify upstream to respect headers by which content varies and upgrade to that version.

lloyd:
The previous comment is wrong in assessing the severity of the issue.  it's a little more ugly that described above.  This is an issue tested via google chrome on osx.  STR:
1. set it-CH as your language
2. load https://dev.diresworb.org/sign_in
3. notice "‮Ｍǝ ɐɹǝ ʌǝɹʎ soɹɹʎ´ ʇɥǝɹǝ ɥɐs qǝǝu ɐu ǝɹɹoɹ¡
4. That error is expected when directly loading the dialog.  now switch your language to en-US
5. hit "enter" in the url bar to soft-reload the resource
6. notice you'r still looking at the debug language
7. hard reload with command-r
8. notice we switched to english.

A user would see a failure to switch dialog language after changing their browser preferences.

jrgm:
Veritied on stage (diresworb.org) with curl at the command line, Chrome 18 OSX, FF 11 OSX, IE8 WinXP, FF 11 WinXP, 
[...]

PR #1369, opened 2012-03-30
bump to etagify 0.0.2 which respects Vary headers.  issue #1364

lloyd:
This is the full changeset from etagify 0.0.2:
https://github.com/lloyd/connect-etagify/commit/4e2c6ef592daa6e05ab00832c844d0ca4683559a

once merged, we should close issue 1364.


ozten:
r+ etagify, merge at will!

Issue #1374, opened 2012-04-03
ie8 does not have Array.prototype.forEach


shane-tomlinson:
Array.prototype.forEach is used in dialog/controllers/authenticate.js->showHint

shane-tomlinson:
4c16576 was removed as a suboptimal solution, replacing with e892635a50fe24897aaac7b8d34122fbc482941b.

PR #1375, opened 2012-04-03
Replacing the use of Array.prototype.forEach with _.each, which we are guaranteed to have.

shane-tomlinson:
issue #1374


lloyd:
r+

PR #1387, opened 2012-04-04
implement navigator.id.experimental.watch().  closes #1354.  closes #134...

### LINKED ISSUE #1354: decide on and implement final API for deeper BrowserId involvment in session duration
we've implemented an event api, we're talking about an alternative.  Before we branch on april 11th we need a decision, implementation, and documentation of the new API we're going to propose to the world.

### LINKED ISSUE #1357: New API does not work with IE8 - IE8 does not have document.createEvent('Event');
in include.js, IE8 is trying to create an event using `document.createEvent("Event")`  IE8 does not have this.

We have to do some feature-query event-fu:

http://stackoverflow.com/questions/827716/emulate-clicking-a-link-with-javascript-that-works-with-ie

We could:

```
var evt;
if(document.createEvent) {
      evt = document.createEvent('Event');                                                                                  
      evt.initEvent(type, true, true);
}
else if(document.createEventObject) {
      evt = document.createEventObject();
}

if(evt) {
     if (params) {
     ....
```

If the user is using IE8 to sign in to beta.myfavoritebeer.org, an error is shown in the developer tools, but the user can still sign in.

lloyd:
...8.  closes #1357

### DIFF
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -161,12 +161,12 @@ $(document).ready(function() {
 
     $(".specify button.assertion").attr('disabled', 'true');
 
-    navigator.id.request({
+    navigator.id.experimental.request({
       privacyURL: $('#privacy').attr('checked') ? "/privacy.html" : undefined,
       tosURL: $('#tos').attr('checked') ? "/TOS.html" : undefined,
       requiredEmail: requiredEmail,
-      oncomplete: function() {
-        loggit("oncomplete");
+      onclose: function() {
+        loggit("onclose");
         $(".specify button.assertion").removeAttr('disabled');
       }
     });
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -90,6 +90,11 @@ pre {
   <pre> ... </pre>
 </div>
 
+<div class="readiness">
+  <h2>readiness</h2>
+  <pre> ... </pre>
+</div>
+
 </body>
 
 <script src="jquery-min.js"></script>
@@ -102,6 +107,8 @@ function loggit() {
   } catch(e) {}
 }
 
+var serial = 1;
+
 // a function to check an assertion against the server
 function checkAssertion(assertion) {
   $.ajax({
@@ -113,7 +120,8 @@ function checkAssertion(assertion) {
       audience: window.location.protocol + "//" + window.location.host
     },
     success: function(data, textStatus, jqXHR) {
-      $(".loginEvents > pre").text(JSON.stringify(data, null, 4));
+      var old = $(".loginEvents > pre").text() + "\n";
+      $(".loginEvents > pre").text(old + JSON.stringify(data, null, 4));
     },
     error: function(jqXHR, textStatus, errorThrown) {
       var resp = jqXHR.responseText ? JSON.parse(jqXHR.responseText) : errorThrown;
@@ -126,14 +134,20 @@ navigator.id.experimental.watch({
   email: (localStorage.loggedInUser === 'null') ? null : localStorage.loggedInUser,
   onready: function () {
     loggit("onready");
+    var txt = serial++ + ' navigator.id ready at ' + (new Date).toString();
+    $(".readiness > pre").text(txt);
+
   },
   onlogin: function (assertion) {
     loggit("onlogin");
+    var txt = serial++ + ' got assertion at ' + (new Date).toString();
+    $(".loginEvents > pre").text(txt);
+
     checkAssertion(assertion);
   },
   onlogout: function () {
     loggit("onlogout");
-    var txt = 'logout callback invoked at ' + (new Date).toString();
+    var txt = serial++ + ' logout callback invoked at ' + (new Date).toString();
     $(".logoutEvents > pre").text(txt);
   }
 });
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -17,7 +17,7 @@ a:link, a:visited { font-style: italic; text-decoration: none; color: #008; }
 a:hover { border-bottom: 2px solid black ; }
 .title { font-size: 2em; font-weight: bold; text-align: center; margin: 1.5em auto 1.5em auto; }
 .intro { font-size: 1.2em;  }
-.specify { font-size: 1.1em; padding-top: 2em; }
+.specify, .session { font-size: 1.1em; padding-top: 2em; }
 body div { width: 600px; margin: auto; }
 
 pre {
@@ -57,7 +57,7 @@ pre {
 </div>
 
 <div class="specify">
-  <p>What flavor of assertion would you like?</p>
+  <p><b>What flavor of assertion would you like?</b></p>
   <ul>
     <li>
       <input type="checkbox" id="privacy">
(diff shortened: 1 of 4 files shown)


example/rp/index.html
```diff
@@ -146,14 +145,26 @@
...
+        $(".specify button.assertion").removeAttr('disabled');
+      }
     });
   });
 
   $(".specify button.logout").click(navigator.id.logout);
+
+  $(".session button.update_session").click(function() {
+    localStorage.loggedInUser = $.trim($('#loggedInUser').val());
+    $(".session input").fadeOut(100).fadeIn(350);
```
shane-tomlinson, line 183: @lloyd - I am confused by the behavior here, once I said who the user was, I expected an attempt at getting an assertion.

resources/static/include_js/include.js
```diff
@@ -770,12 +770,13 @@
             try {
               var d = JSON.parse(e.data);
               if (d.a === 'ready') messageTarget.postMessage(req, origin);
-              else if (d.a === 'error') cb(d.d);
-              else if (d.a === 'response') {
+              else if (d.a === 'error') {
+                if (cb) { cb(d.d); cb = null; }
+              } else if (d.a === 'response') {
                 removeListener(window, 'message', onMessage);
```
shane-tomlinson, line 776: @lloyd - why remove the listeners/perform cleanup only in the 'response' case? Is it possible for there to be an error and then a valid response?

resources/static/include_js/include.js
```diff
@@ -1001,53 +980,39 @@
...
-      var i;
-      for (i = 0; i < listeners[type].length; i++) {
-        if (listeners[type][i] === listener) break;
-      }
-      if (i < listeners[type][i].length) {
-        listeners[type].splice(i, 1);
+      if (options.email !== undefined) {
+        commChan.notify({
+          method: 'loggedInUser',
+          params: options.email
```
shane-tomlinson, line 1017: Before passing this to the dialog, should we do some sanitation/verification to make sure the specified email address is really an email address?

resources/static/include_js/include.js
```diff
@@ -1001,53 +980,39 @@
...
-
-    function internalRemoveEventListener(type, listener ) {
-      // remove event from listeners table
-      var i;
-      for (i = 0; i < listeners[type].length; i++) {
-        if (listeners[type][i] === listener) break;
-      }
-      if (i < listeners[type][i].length) {
-        listeners[type].splice(i, 1);
+      if (options.email !== undefined) {
```
shane-tomlinson: I would change this to 

```
if (typeof options.email !== "undefined") {
```

for the reasons outlined in comment 2 of:
http://stackoverflow.com/questions/27509/detecting-an-undefined-object-property-in-javascript

If another script somewhere else in the RP redefines "undefined", this check as currently written will break silently.

resources/static/include_js/include.js
```diff
@@ -1001,53 +980,39 @@
...
       }
     }
 
-    navigator.id.removeEventListener = function(type, listener/*, useCapture */) {
-      checkCompat(false);
-      internalRemoveEventListener(type, listener);
+    navigator.id.experimental = {
+      watch: function(options) {
+        checkCompat(false);
+        internalWatch(options);
```
shane-tomlinson: The types for  onlogin, onlogout and onready should probably be checked to make sure we are not trying to call non-functions later on.

resources/static/include_js/include.js
```diff
@@ -936,11 +937,11 @@
...
-    // table of registered event listeners
-    var listeners = {
-      login: [ ],
-      logout: [ ],
-      loginCanceled: [ ]
+    // table of registered observers
+    var observers = {
+      login: null,
+      logout: null,
+      ready: null
```
shane-tomlinson, line 944: Where is observers.ready used?  I can see observers.login and observers.logout being called, but not observers.ready.

resources/static/include_js/include.js
```diff
@@ -1001,53 +980,39 @@
...
-          emitEvent('login', { assertion: params });
+          if (observers.login) observers.login(params);
         });
       }
     }
 
-    function internalAddEventListener(type, listener) {
-      // add event to listeners table if it's not there already
-      if (!listeners[type]) throw "unsupported event type: '" + type + "'";
+    function internalWatch(options) {
```
shane-tomlinson, line 998: Can the internal\* functions be grouped together to increase mental cohesion of the blocks?

resources/static/include_js/include.js
```diff
@@ -1055,28 +1020,27 @@
...
-
-      // 2. send a "loggedInUser" message to iframe
-      commChan.notify({ method: 'loggedInUser', params: email });
-    };
-
     // backwards compatibility function
     navigator.id.get = function(callback, options) {
       checkCompat(true);
-
+      internalWatch({
```
shane-tomlinson: Why the calls to both internalWatch and internalRequest - which can call internalWatch with an empty options block?  can internalWatch return before internalRequest?

shane-tomlinson:
Nice work on this @lloyd.  Most of my comments are inline, one more here.

I think now would be an opportune time to do a bit of cleanup, it is becoming difficult to find the various API calls inside the file without doing a search for navigator.id.  I would try to combine all the various navigator.id.\* under a single object so that readers of the code can find all the API calls easily.

```
extend(navigator.id, {
  get: ...,
  getVerifiedEmail: ...,
  logout: ...,
  experimental: {
     watch: ...
  }
});
```

example/rp/index.html
```diff
@@ -146,14 +145,26 @@
...
+        loggit("oncomplete");
+        $(".specify button.assertion").removeAttr('disabled');
+      }
     });
   });
 
   $(".specify button.logout").click(navigator.id.logout);
+
+  $(".session button.update_session").click(function() {
+    localStorage.loggedInUser = $.trim($('#loggedInUser').val());
```
shane-tomlinson, line 182: Directly calling localStorage like this will break on IE6 or IE7 when testing to see if the unsupported_dialog works as expected.  We could use a shim or an abstraction layer.

resources/static/include_js/include.js
lloyd, line 776: great question @shane-tomlinson - I'm noticing when I cancel a dialog I get an error from winchan inside the dialog (user closed dialog) then I get a null response - presumably both browserid code and winchan are trying to handle the cancellation case.  This defense ensures we don't multiply invoke client code, but we should probably go and remove the handling of client closing window in dialog?

resources/static/include_js/include.js
lloyd, line 1017: maybe we should do that sanitation _inside_ our code running in the dialog, because we have to do that anyway?  (all of include.js runs in an untrusted environment, and we have to harden the postMessage API from the dialog side.)

resources/static/include_js/include.js
lloyd: like it.  fix coming

resources/static/include_js/include.js
lloyd, line 944: fixed.  see additional commits.

resources/static/include_js/include.js
lloyd: fixed.

lloyd:
@shane-tomlinson I restructured the public API so that it's all together at the end of the file.  see 1ff57d1

example/rp/index.html
lloyd, line 183: @shane-tomlinson for you: 3cd8065

resources/static/include_js/include.js
lloyd, line 998: done, see 1ff57d1

lloyd:
k dudes, ready for another round of review.  I've gone through all comments and tried to hit all the important issues.

tests pass: http://travis-ci.org/#!/mozilla/browserid/builds/1027601

lloyd:
Ok, possibly controversial move.  I'm going to merge this PR, given that it closes 3 open 5 stars in dev, and what we have in dev isn't shippable anyway.  

Can all reviewers open bugs against me with further work required after reviewing the 7 commits I made in response to the first round of review?

Ultimately, dev will definitely not ship without this and might ship with it.  If I can get it merged now, I have a shot at doing the requisite testing in time to get this out in `train-2012.04.11`.

I hope this is ok, please complain loudly if I'm being overly aggressive.

Issue #1390, opened 2012-04-05
IE6 and IE7 errors in communication frame


shane-tomlinson:
STR

1) Open IE6
2) Go to local dev RP
3) Double click error icon in bottom left of browser chrome.

Many errors:

```
Line: 245
Char: 34
Error: Exception thrown and not caught
Code: 0
URL: http://10.246.29.67:10001/
```

```
Line: 3347
Char: 12
Error: Expected identifier, string or number
Code: 0
URL: http://10.246.29.67:10002/communication_iframe
```

```
Line: 9
Char: 7
Error: 'localStorage' is undefined
Code: 0
URL: http://10.246.29.67:10002/communication_iframe
```

```
Line: 233
Char: 38
Error: Exception thrown and not caught
Code: 0
URL: http://10.246.29.67:10002/communication_iframe
```

lloyd:
I'm going to ignore this issue for now, it looks like shane's pull request would cover it.  If it looks like that PR (issue #1392) will not land, please reassign this back to me and I'll craft a patch targeted at dev.

jrgm:
I still have two errors (either ie6 or ie7):

Line: 3347
Char: 12
Error: Expected identifier, string or number
Code: 0
URL: http://10.250.4.157:10002/communication_iframe

Line: 233
Char: 38
Error: Exception thrown and not caught
Code: 0
URL: http://10.250.4.157:10002/communication_iframe

However, I see from https://github.com/mozilla/browserid/pull/1411#issuecomment-5038350 that the presence of some errors is expected? Are the above a concern? The dialog does show the unsupported screen when popped.

shane-tomlinson:
@jrgm - I can verify there are errors in IE6/IE7 on myfavoritebeer.org from prod, but they are not the same as these.  Which branch are you testing from?

shane-tomlinson:
Also issued pull request to myfavoritebeer.org: https://github.com/lloyd/myfavoritebeer.org/pull/56

jrgm:
I was on train-2012.04.11

shane-tomlinson:
This was merged in to dev with PR-1506.  Closing.

jrgm:
So, GH-1530 is closed/verified as WONTFIX for myfavoritebeer, so I'll take @jbonacci's comment above and my testing as verified for stage train-2012.04.27

Issue #1317, opened 2012-03-14
Clean up the CSS.


shane-tomlinson:
It's got a lot of cruft and overrides of overrides.

Start with some true common CSS.

Make a stylesheet each for mobile, tablets, and desktop.

shane-tomlinson:
Link to issue #1101

jrgm:
Closing and verifying for train-2012.04.27 - see https://github.com/mozilla/browserid/issues/1101#issuecomment-5524916

Issue #1101, opened 2012-02-10
use flexible css to improve layout on tablets


skinny97214:
http://cl.ly/342G401g1b2l3t2G1Z3J

skinny97214:
mock of tablet layout

http://cl.ly/0e3J0F2O3V1n0W1k3E0d

jrgm:
(Closing this issue based on PR GH-1392 which is in train-2012.04.27)

Okay, so I've been using up-to-date versions of these OS/browser combinations:

OSX                    FF FF_Beta Chrome Opera Safari 
WinXP                  FF FF_Beta Chrome Opera Safari IE8 IE7 IE6
Win7                   FF FF_Beta Chrome Opera Safari IE9
Ubuntu                 FF FF_Beta Chrome Opera 
Android4.0/Nexus       FF10(XUL) Aurora Stock ChromeBeta 
Android4.0/Asus_Tablet FF10(XUL) Aurora Stock ChromeBeta 
Android2.3/Droid2      FF10(XUL) Aurora Stock 
IOS5/iPhone4s          Safari

Going through general usage and specific focus on @shane-tomlinson's notes to commit https://github.com/shane-tomlinson/browserid/commit/8329893507b1a2ed6b49d6ca39f52d2bc4a1fd01, I haven't had any major issues, but did note GH-1548 (which isn't a big deal). So, looks good to me but any other mobile device testing would be appreciated. Marking verified for stage train-2012.04.27 with the caveat that I'm a poor judge of UI/UX ;-)

PR #1392, opened 2012-04-05
Make the design more responsive to all screen sizes - mobile, desktop and tablet.

shane-tomlinson:
- Start with a slight refactor of the CSS and DOM.
- Vertical center content when content is smaller than the screen size.
- Constrain email addresses and show ellipsis if emails overflow box width.  Make sure width is updated when browser window size changes or user changes the screen direction of their device.
- If form/email address list is taller than available form area, show scroll bar on desktop or expand contents so that entire app must be scrolled on mobile.
- If dialog opened on a tablet or large desktop window, add a gradient to the left size of the white area that matches the gradient on the arrow.
- On mobile, footer should always be at the very bottom of the screen unless contents overflow screen height, where it will be pushed off the bottom of the screen.

Things to test and be aware of:
- Dialog should respond to various screen sizes when resizing browser window on desktop.
- IE8 does not support media queries so narrowing the browser window will not force the dialog into "mobile" mode.
- Contents in the "form area", wait, delay, or error screens should always be vertically centered.
- IE6/IE7 should still see the unsupported dialog with the correct layout (untested on checkin due to #1390)
- Wide emails should always be constrained to the screen width on mobile or box width on desktop.  The layout should respond to changes in device orientation.
- The box surrounding the emails should show a scroll bar on desktop/tablet if there are more emails the available screen height.  On mobile, the entire screen height should expand so that the entire app must be scrolled.
- On mobile, the footer should always be at the very bottom of the screen unless the form area takes up more vertical space than the device height.

issue #1101
issue #1317

### DIFF
--- a/lib/static_resources.js
+++ b/lib/static_resources.js
@@ -84,6 +84,7 @@ var dialog_js = und.flatten([
     '/dialog/resources/internal_api.js',
     '/dialog/resources/helpers.js',
     '/dialog/resources/state.js',
+    '/dialog/resources/screen_size_hacks.js',
 
     '/dialog/controllers/actions.js',
     '/dialog/controllers/dialog.js',
--- a/resources/static/css/common.css
+++ b/resources/static/css/common.css
@@ -46,6 +46,10 @@ header, section, footer {
   display: block;
 }
 
+ul, li {
+  list-style-type: none;
+}
+
 .sans {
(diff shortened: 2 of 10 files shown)


shane-tomlinson:
Merged manually in 20421047800a486edbeb4ac1ad41c0419a13f329

Issue #1395, opened 2012-04-05
Internal APIs changed in dev (was: The assertion object is not serializable (on development))


st3fan:
I'm using the following code to call BrowserID and then pass the assertion to our own backend for verification:

```
function doBrowserId() {
  navigator.id.getVerifiedEmail(function(assertion) {
    if (assertion) {
      $.ajax({
        url: "/browserid/verify",
        data: { 'assertion': assertion },
        dataType: "json",
        success: function(data, status, xhr) {
           // Not important what happens here
        },
        error: function(xhr, status, error) {
           // Not important what happens here
        }
      });
    } else {
        // Not important what happens here
    }
  });
}
```

This works ok in production but fails on development. The assertion is serialized as "[Object object]" instead of a proper JSON encoded dictionary. Since this is pretty much based on sample code that many people have in production, I wonder if this should break like this.

lloyd:
@st3fan - steps to reproduce?

st3fan:
Not sure how to document that. Login to browserid & serialize assertion?

benadida:
@st3fan you're doing this on dev.myfavoritebeer.org?

lloyd:
this is a regression in our internal APIs.  It won't affect a lot of people, but the people it will affect are important (looking at you, @st3fan).  I think it's easy to work around this, and don't think we should needlessly break contracts.  Here's the conclusion I reached on IRC in discussing with @st3fan, with the context preceeding it: http://irclog.gr/#show/irc.mozilla.org/identity/73222

given this, assigning to self and giving it 5 stars.  

I repeat, dev specific, internal api specific.

lloyd:
so I'm unhappy with our testing of this feature.  It seems like the only tests we have don't actually execute the code, and they don't catch this regression...

benadida:
@lloyd yeah, we haven't given the internal API enough love, agreed. We should, especially as we're about to add more.

lloyd:
The only way I can think of improving the situation is to simulate native embedding with a test specific iframe that mimics how pancake (and other native code) will use our internal API, then to script that with something like selenium...

For the purposes of this bug I'm going to deploy a blind fix on a branch/ephemeral VM and let @st3fan test it...

lloyd:
@benadida - @callahad reports that there are people interested in this feature.  so that's a +1 to formalizing it, documenting it, providing and example, and making sure we have tests that validate our API isn't inadvertently broken. 

[...]

PR #1409, opened 2012-04-09
repair internal.get API - it should return a bare assertion represented ...

lloyd:
...as a string, not an object - issue #1395

### DIFF
--- a/resources/static/dialog/resources/internal_api.js
+++ b/resources/static/dialog/resources/internal_api.js
@@ -14,6 +14,17 @@
       storage = bid.Storage,
       moduleManager = bid.module;
 
+  // given an object containing an assertion, extract the assertion string,
+  // as the internal API is supposed to return a string assertion, not an
+  // object.  issue #1395
+
+  function assertionObjectToString(assertion) {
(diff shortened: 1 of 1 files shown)


lloyd:
@st3fan - can you please verify that this fixes the internal API for you.  Just...

... replace browserid.org with `issue1395.hacksign.in` for both where you include javascript resources and where you send assertions for verification.

depending on your report, we'll merge this or not.

st3fan:
@lloyd I cannot test this on iOS from a UIWebView because iOS complains about the certificate for *.hacksign.in. Maybe the intermediate certificate is missing on your server?

lloyd:
@st3fan try again please?  I _think_ I fixed the SSL issue.

benadida:
on code alone, assuming @st3fan confirms, r+.

shane-tomlinson:
@lloyd - When the old API is officially completely removed, we should clean up the unit tests for the internal api, there are a lot of tests with the "silent: true" option.

st3fan:
@lloyd I tested against `issue1395.hacksign.in` and that works ok.

lloyd:
given @benadida's review, and @st3fan's confirmation, I'm merging this.

Issue #1406, opened 2012-04-09
dev: "Is this your computer" should not be asked during first sign-up


lloyd:
report: "When I click the confirmation e-mail, it tells me something along the lines of "You've been logged in, you can now go back to 123done.org". But when I go back to the 123done.org tab, I'm not logged in yet; I need to go find the popup and respond to the "Is this your computer?" question before it logs me in."

@shane-tomlinson - any ideas how to elegantly surpress the question during initial sign-up?

steps to reproduce:
1. open a clean profile or enter private browsing / incognito mode
2. click signin on 123done.org
3. go get a cup of coffee for more than 60s
4. now click the verification link in your email and enter a username and password
5. notice that rather than signing you in, the dialog asks you The Question

This sucks because the user has to find the dialog post verification, and this was not our design intent.

jrgm:
verified on stage from train-2012.04.11. The user is not asked the question on initial signup/signin. A signin 60 seconds 
subsequent to the first appearance of the user will cause the user to be asked 'is this your computer'.

Issue #1405, opened 2012-04-09
dev: error shown in dialog after confirming your computer


lloyd:
reported via email:

"I got this when I clicked "Yes" in response to "Is this your computer?" while signing up for the first time.
http://d.pr/9BLk"

reproduced via 123done.org

Zarel:
I was the one who reported this on the ML, so I might as well add myself to the issue. Hello!

I got this issue while signing up with a Gmail address.

lloyd:
successfully reproduced via STR in issue #1406

jrgm:
verified on ff/chrome/safar/ie8/ie9/android2.2(stock,FF)/android4.0(stock,FF) not getting this error with train-2012.04.11

PR #1410, opened 2012-04-10
upon first time signup, don't ask the user if this is their computer - i...

lloyd:
...ssue #1406 - issue #1405

### DIFF
--- a/lib/wsapi/user_creation_status.js
+++ b/lib/wsapi/user_creation_status.js
@@ -19,7 +19,7 @@ exports.process = function(req, res) {
   if (wsapi.isAuthed(req, 'assertion')) {
     db.userOwnsEmail(req.session.userid, email, function(err, owned) {
       if (err) wsapi.databaseDown(res, err);
-      else if (owned) res.json({ status: 'complete' });
+      else if (owned) res.json({ status: 'complete', userid: req.session.userid });
       else notAuthed();
     });
   } else {
--- a/resources/static/shared/network.js
+++ b/resources/static/shared/network.js
@@ -257,7 +257,7 @@ BrowserID.Network = (function() {
       get({
         url: "/wsapi/user_creation_status?email=" + encodeURIComponent(email),
         success: function(status, textStatus, jqXHR) {
-          if (status.status === complete && status.userid) {
+          if (status.status === 'complete' && status.userid) {
             setUserID(status.userid);
           }
           complete(onComplete, status.status);
--- a/resources/static/shared/network.js
+++ b/resources/static/shared/network.js
@@ -257,6 +257,9 @@ BrowserID.Network = (function() {
       get({
         url: "/wsapi/user_creation_status?email=" + encodeURIComponent(email),
         success: function(status, textStatus, jqXHR) {
+          if (status.status === complete && status.userid) {
+            setUserID(status.userid);
+          }
           complete(onComplete, status.status);
         },
         error: onFailure
--- a/resources/static/shared/storage.js
+++ b/resources/static/shared/storage.js
@@ -279,6 +279,10 @@ BrowserID.Storage = (function() {
   }
 
   function shouldAskUserAboutHerComputer(userid) {
+    // if any higher level code passes in a non-userid,
+    // we'll tell them not to ask, triggering ephemeral sessions.
+    if (typeof userid !== 'number') return false;
+
     // we should ask the user if this is their computer if they were
     // first seen over a minute ago, if they haven't denied ownership
     // of this computer in the last 24 hours, and they haven't confirmed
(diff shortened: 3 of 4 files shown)


lloyd:
hey @shane-tomlinson, can you review this one?

lib/wsapi/user_creation_status.js
```diff
@@ -19,7 +19,7 @@ exports.process = function(req, res) {
   if (wsapi.isAuthed(req, 'assertion')) {
     db.userOwnsEmail(req.session.userid, email, function(err, owned) {
       if (err) wsapi.databaseDown(res, err);
-      else if (owned) res.json({ status: 'complete' });
+      else if (owned) res.json({ status: 'complete', userid: req.session.userid });
```
benadida, line 22: is this the first time we're exposing userid to the client, or is this the userid we're already exposing to the client? Is the only other place it exists in the encrypted cookie?

lloyd, line 22: we're exposing the userid outside of our servers for the first time with the API change.  train-2012.03.28 did not expose it, train-2012.04.11 will expose it in a number of different API calls.  This userid is the same that is in the encrypted cookie, which is the same that is in the database.

benadida, line 22: and is that necessary? No easy workarounds?

lloyd, line 22: the motivation was to be able to keep track of a user's response to "is this your computer".  If a user logs into different sites using different emails, they should only have to answer the question once per device, not once per email address.

challenging myself a bit, we _could_ instead keep a collection of emails and update that based on the server's response to `list_emails`.  I'd have to think through this a bit to see how reliable it would be.

How strongly opposed are you to exposing a user's own id to him?

benadida, line 22: we're sure we're only exposing this to the user themselves? In that case I'm r+ for this, and I'll add an issue to investigate whether this is problematic.

shane-tomlinson, line 22: @lloyd, @benadida - storing the list of emails that has been seen on the current computer seems to me to be a larger data leak than a numeric user id.

resources/static/shared/network.js
```diff
@@ -257,6 +257,9 @@ BrowserID.Network = (function() {
       get({
         url: "/wsapi/user_creation_status?email=" + encodeURIComponent(email),
         success: function(status, textStatus, jqXHR) {
+          if (status.status === complete && status.userid) {
```
shane-tomlinson: complete is actually a function, so this check will always fail.

complete's intent is to be a shortcut for 
`if(callback) callback(args...)`

the shortcut is called using 
`complete(callback, args...)`

which compresses to a few fewer bytes than the former.

Instead of `(status.status === complete...`, this should be `(status.complete === "complete"...`

resources/static/shared/storage.js
```diff
@@ -279,6 +279,10 @@ BrowserID.Storage = (function() {
   }
 
   function shouldAskUserAboutHerComputer(userid) {
+    // if any higher level code passes in a non-userid,
+    // we'll tell them not to ask, triggering ephemeral sessions.
+    if (typeof userid !== 'number') return false;
```
shane-tomlinson, line 284: This passes for the new user case because of the comment I made in network.js - https://github.com/mozilla/browserid/pull/1410/files#r670265.  Since the user's userID is never set, shouldAskUserAboutHerComputer is called with an undefined userid, which returns false and prevents the question from being asked.

shane-tomlinson:
Outside of the comments I have already made:

`resources/static/test/mocks/xhr.js` should be updated for 'user_creation_status' to have a userid whenever the status is 'complete'.

In the long run we are going to want to remove the notion of a userid from network.js and put that into user.js.  

In /dialog/resources/state.js, we should change 

`if( storage.usersComputer.shouldAsk(network.userid())`

to use an encapsulation in user.js so that we can keep this logic out of site of the main client code.  These are out of the scope of this pull request though.  With the changes indicated to network.js, I give an r+.

lib/wsapi/user_creation_status.js
benadida, line 22: I'm actually not worried about the privacy leak (although it's a good point to bring up), I'm worried about the security implications of exposing internal identifiers. That can bite you at some point. However, if we're only exposing the user_id to that specific user, it's not nearly the same risk, so I'm good to go.

resources/static/shared/network.js
lloyd: nice catch!  as you mention in the other line comment, the reason it appears to work is because I defended against calling shouldAsk with undefined inside storage.js as well.

lib/wsapi/user_creation_status.js
lloyd, line 22: Opened a new issue to explore this: GH-1421

lloyd:
13:12 < stomlinson> lloyd: with the change to network.js, I give my r+ to 1410

Issue #1391, opened 2012-04-05
Communication frame does not check for min browser requirements


shane-tomlinson:
This causes exceptions to be thrown and not caught in IE6/IE7 or IE8+ in Quirks mode.

See issue #1390

lloyd:
Reassigning to @shane-tomlinson as I expect this is related to pull request #1392.  If I'm wrong, please reassign back to me and I'll fix in dev.

shane-tomlinson:
@lloyd - one problem with IE6/IE7 and IE8 in quirks mode is that the communication frame is doing a Channel.build, which checks for the existence of window.localStorage, window.postMessage and window.JSON.  These items do not exist in the older IEs and an exception is thrown.

Since the communication iframe was not included by default before, we did not run into this problem.  We would have seen it on myfavoriteshow, but it looks like nobody checked.

#1392 does not cover this.

I would take care of this by putting a try/catch around the Channel.build, but you often times have a more elegant solution. I am assigning back to you to get the notice, can you either take care of it or assign back to me?

There may be other problems on top of this, but I am fairly certain this is the main problem.

lloyd:
@shane-tomlinson cool!  thanks for the clarification, I'll fix it and offer up a PR for your review.

jrgm:
As noted in https://github.com/mozilla/browserid/issues/1390#issuecomment-5253093, I still have errors about 
communication_iframe on ie6/ie7

shane-tomlinson:
Closing, PR-1506 fixes this issue.

PR #1411, opened 2012-04-10
improve failure on unsupported browsers - catch exceptions and allow the...

lloyd:
... dialog to pop up with the 'unsupported browser' messaging.  issue #1390 and issue #1391 - NOTE: we don't go to the trouble of surpressing all errors on the console.log for unsupported browsers, as this would require patching upstream code (like node-bigint) and the errors are cosmetic, but don't prevent a clear display of what the problem is to the user.

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -962,37 +962,43 @@
 
     // this is for calls that are non-interactive
     function _open_hidden_iframe() {
-      if (!commChan) {
-        var doc = window.document;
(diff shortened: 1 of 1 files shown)


lloyd:
@shane-tomlinson - I went the try/catch route here, I couldn't think of anything more elegant, the remaining errors are innocuous, and @benadida is making changes in the next iteration of jwcrypto that might reduce the amount of work we'd have to do to keep IE6/7 consoles clean.

benadida:
r+

PR #1412, opened 2012-04-10
The list of supported locales in train-2012.03.28

mathjazz:


lloyd:
the commit is being merged and deployed into train-2012.03.28, you'll know it's there when you see https://diresworb.org/ver.txt display the sha 64a08ec.

We'll merge the full change history back into the public repository on wed.

PR #1415, opened 2012-04-11
Fixes the wait, delay, and error screens not showing on mobile devices.

shane-tomlinson:
issue #1402


lloyd:
code changes look fine.

next, followed STR in GH-1402.  was able to reproduce the issue against dev.  pushed this patch.  saw that the issue was resolved and I got clearer messaging about re-enabling my cookies.

re-enabled cookies on my ios device, was able to auth and log in just fine.  r+

Issue #1414, opened 2012-04-11
Fx, dev RP: security exception thrown if cookies disabled.


shane-tomlinson:
STR
1) Load up Firefox
2) Turn off cookies
3) Open up firebug
4) Open up the dev RP

```
uncaught exception: [Exception... "Security error" code: "1000" nsresult: "0x805303e8 (NS_ERROR_DOM_SECURITY_ERR)" location: "http://192.168.1.88:10001/ Line: 138"]
(?)()include.js (line 629)
include.js()include.js (line 8)
[Break On This Error]   

;WinChan = (function() {
```

shane-tomlinson:
In Nightly, the exception is:

```
The operation is insecure.
...email: (localStorage.loggedInUser === 'null') ? null : localStorage.loggedInUser
```

shane-tomlinson:
This manifests itself on *.myfavoritebooze.org as well.

jrgm:
verified fixed for firefox on stage train-2012.04.11

PR #1417, opened 2012-04-11
Fix Fx throwing an exception when accessing localStorage if cookies are disabled.

shane-tomlinson:
- In the example RP, instead of accessing localStorage directly, access it through a variable that is guaranteed to exist.
- In page_helpers.js, do all localStorage access through new functions in storage.js
- In storage.js, if localStorage is inaccessible, create a standin.

issue #1414

### DIFF
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -105,6 +105,15 @@ pre {
 <script src="https://browserid.org/include.js"></script>
 <script>
 
+try {
+  var storage = localStorage;
+}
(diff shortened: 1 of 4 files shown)


lloyd:
Read the code. Looks good to me.

Ran the tests, look clean to me.

Then figured that if this were to break it would be in persisting or clearing the last used email for a site.  did some local testing and it looks like the last used email address is bolded properly.  r+

Issue #702, opened 2011-12-05
BrowserID links should look clickable


ghost:
OS: All
Browser: All

Most BrowserID links are black and bold while very few of them (e.g. learn more) are blue. In general, links are not emphasized just by being bold, so the user might not realize what he sees is actually a link. The fact that there are some blue links in BID, makes it even less probable to the user that the black bold text is a link.

All the BID links should be blue (or some other color that makes it obvious they are links) both for consistency reasons and user friendliness.

shane-tomlinson:
Closing, a duplicate of issue #459.  @ioanab - I am in agreement.

ghost:
@shane-tomlinson - issue #459 is closed. I filed this new bug to make sure this issue is tracked and fixed. Please reopen this issue or 459 until it is fixed. Thanks

jrgm:
verified on stage train-2012.04.27. Links are consistently styled.

PR #1428, opened 2012-04-12
Visual updates to make secondary action links be more consistent, clean up CSS.

shane-tomlinson:
- Secondary links all are some shade of dark grey/black with a dotted underline.
- Change the font used on the "about" page to be the standard Helvetica Neue except for the header.
- Change the style of link in the TOS/PP to black text with a dotted underline.
- Loads more link updates!
- A lot of cleanup to remove duplicate `color: #549fdc` entries by using the action class.
- On the main site, put the action links to the far left of the main button.
- Reduce the amount of "forgot password" CSS by generalizing.
- Generalize the yellow banner on the main site.

issue #702

### DIFF
--- a/resources/static/css/common.css
+++ b/resources/static/css/common.css
@@ -44,6 +44,7 @@ html[xmlns] .cf {
 
 header, section, footer {
   display: block;
+  width: 100%;
 }
 
 ul, li {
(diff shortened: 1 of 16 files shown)


skinny97214:
Doo eeet!

PR #1443, opened 2012-04-16
fix case where test_db_connectivity would fail when browserid database d...

lloyd:
...oesn't exist.  issue #1246


ozten:
Looks great! r+ didn't run locally.

Issue #1442, opened 2012-04-16
API regression in dev and beta - options are *required*


lloyd:
STR:
1. invoke navigator.id.get(function() { });

notice error:

```
Uncaught TypeError: Cannot set property 'onclose' of undefined
```

This is the underlying cause of https://bugzilla.mozilla.org/show_bug.cgi?id=745829

lloyd:
fix merged into train-2012.04.11.4 and dev.  now deployed on the latter, to be deployed on the former some time RSN.

jrgm:
verified fixed for train-2012.04.11-8 on stage per https://bugzilla.mozilla.org/show_bug.cgi?id=745829#c2, and by invoking
navigator.id.get(function(){}) in the chrome/firefox browser.

PR #1444, opened 2012-04-16
fix regression where an undefined second argument to .get() would cause ...

lloyd:
...a javascript error - issue #1442

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1122,6 +1122,7 @@
       },
       // get an assertion
       get: function(callback, options) {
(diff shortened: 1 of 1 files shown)


lloyd:
if r+, this also needs to be merged into train-2012.04.11 - I can do that with post review.

resources/static/include_js/include.js
```diff
@@ -1122,6 +1122,7 @@
       },
       // get an assertion
       get: function(callback, options) {
+        options = options || {};
```
ozten, line 1125: Excellent.

PR #1447, opened 2012-04-17
implement updated observer API

lloyd:
given feedback on list, the observer API has been updated to match the docs, which themselves have been updated:  https://gist.github.com/2287552/662ed56ef172c4542bddf30a6cf13b76c746de38

### DIFF
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -144,7 +144,7 @@ function checkAssertion(assertion) {
 };
 
 navigator.id.experimental.watch({
-  email: (storage.loggedInUser === 'null') ? null : storage.loggedInUser,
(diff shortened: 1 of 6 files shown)


example/rp/index.html
```diff
@@ -144,7 +144,7 @@
 };
 
 navigator.id.experimental.watch({
-  email: (storage.loggedInUser === 'null') ? null : storage.loggedInUser,
+  loggedInEmail: (storage.loggedInUser === 'null') ? null : storage.loggedInUser,
```
lloyd, line 147: yeah, that's right, we're checking for the string 'null'.  When you use our sample RP and want to simulate a session where the user is logged out (to check out how events are surpressed) you can type 'null' into the session box.

IOW, if you type 'null' into the session box, then `null` will be passed to `.watch()` as the loggedInEmail argument.

Issue #1000, opened 2012-01-30
Feature: Set password on new account inside dialog


shane-tomlinson:
This is a tracking bug for an experiment for @skinny97214 

The create account flow is quite confusing - mix in unverified emails and there is a lot of room for "WTF" with regards to account completion.  The purpose of this experiment is to put the "set password" flow inside of the dialog.

shane-tomlinson:
@skinny97214 - if we do the set password flow inside the dialog for new users, are we going to do the same for reset-password?

lloyd:
I've deployed the latest on https://issue1000.hacksign.in and http://issue1000.myfavoritebeer.org

@shane-tomlinson I've installed your public key so you can update the server.

lloyd:
@shane-tomlinson `git remote add issue1000 app@issue1000.hacksign.in:git`

Issue #290, opened 2011-09-23
confirmation link should check proper auth


benadida:
When an email is associated with an account, the confirmation link should just confirm that pre-declared link, and only if the user is properly logged in. If the user is logged in as a different account, that should _not_ work. It also should _not_ work if the user is not logged in at all. In other words, the power of that confirmation link should be very small.

lloyd:
This will negatively impact usability.

benadida:
can you say a bit more? Happy to think through potential problems, but it's quite important that we limit the scope of what can be accomplished with that link.

shane-tomlinson:
@benadida - for the new user flow, the user will not be logged in and in fact will not even have a password until after they click the verification link in their email.

benadida:
@shane-tomlinson ahah, good point. So in that case we should just make sure it's the same session as the one that initiated the email addition.

lloyd:
Sorry to have been so terse, my point was that for cases where the browser in which the verification link is opened is not the same one that was used when the link was sent, the user visible failure mode will be lame.  In the case where it's an "add email", we'll tell the user "now please provide your password".  In the case where it's account creation, we'll say "please open this link in the other browser, the one you were using when you signed in".  Both of these are crappy UX, especially the latter.

Now, my intuition tells me that this type of event will be rare.  Usually people use one browser, usually it's properly set as their default browser.  Typically when signing in on their phone, people will also perform the confirmation on their phone.  I wish we had hard numbers (and at the very least we should log when these events occur and track them if we decide to lock this link down).

The other cost associated with this is we have to do the UX and make sure that we're regularly testing these flows.

None of this cost is huge, but it should be weighed against the security benefits of locking these things down. So let me turn it around:  Why do you feel this is so important?  What types of attacks are we preempting with this change and do you expect they'll be more common than legitimate usage that will be negatively impacted?

This is fun, let's type more.

For the **account creation case** It seems like the key attack here is one where an attacker cannot intercept your email at will, but happens to remotely capture an account setup email.  Having this link, they'd be able to complete the account creation, and the would be able to log in with that email address without having verified it.  

If an attacker can read your email at will, then none of these counter measures are effective.   

For the **add email case**, without these countermeasures an attacker who happens to intercept one of your confirmation emails will be able to complete the action you initiated - verifying the email.  They will not, however, be able to use this email to log in.  

If an attacker can read your email at will, then none of these counter measures are effective.

[...]

PR #1450, opened 2012-04-17
Issue 1000 in dialog password

shane-tomlinson:
This is a big hairy beast of a change aimed at pulling the initial set password out of the verification step and into the dialog.


resources/static/dialog/resources/helpers.js
```diff
@@ -97,28 +97,35 @@
       complete(callback, false);
     }
     else {
-      user.addressInfo(email, function(info) {
+      BrowserID.User.addressInfo(email, function(info) {
```
shane-tomlinson: Change this to just user.addressInfo

resources/static/pages/forgot.js
```diff
@@ -21,7 +21,8 @@ BrowserID.forgot = (function() {
     var email = helpers.getAndValidateEmail("#email");
 
     if (email) {
-      user.requestPasswordReset(email, function onSuccess(info) {
+      // XXX TODO - the fake password is to make tests pass.
```
shane-tomlinson: The XXX TODO wasn't enough?  Get the fake password out of there and use user input.  The forgot page will have to be set up to input the password.

resources/static/shared/modules/page_module.js
```diff
@@ -51,7 +51,10 @@ BrowserID.Modules.PageModule = (function() {
 
     start: function(options) {
       var self=this;
+      self.options = options || {};
+
       self.bind("form", "submit", cancelEvent(onSubmit));
+      // TODO - why is this here and not in pick_email?
```
shane-tomlinson: Move this to pick_email

resources/static/shared/modules/page_module.js
```diff
@@ -153,6 +156,7 @@ BrowserID.Modules.PageModule = (function() {
     submit: function() {
     },
 
+    // XXX maybe we should not get rid of this.
```
shane-tomlinson, line 157: Should or should not?

resources/static/test/cases/controllers/authenticate.js
```diff
@@ -177,27 +180,28 @@
...
     });
 
     controller.createUser(function() {
-      equal(handlerCalled, false, "bad jiji, user_staged should not have been called with invalid email");
+      equal(handlerCalled, false, "bad jiji, new_user should not have been called with invalid email");
       start();
     });
   });
 
+  /*
```
shane-tomlinson: What about these tests?  If they really are not needed, take them out.

skinny97214:
Can has URL to test it?

skinny97214:
Do eeet!

lloyd:
@skinny97214 https://github.com/mozilla/browserid/issues/1000#issuecomment-5234165

lib/db/json.js
```diff
@@ -175,13 +175,14 @@ function addEmailToAccount(userID, email, type, cb) {
   });
 }
 
-exports.stageUser = function(email, cb) {
+exports.stageUser = function(email, hash, cb) {
```
lloyd, line 178: As discussed previously, the interesting part here is that the _chooser_ of the password controls the account once the email is verified.  So unlike our current implementation, an accidental click give the attacker _full_ control. 

I think we need to make sure that unless it's the same device, the password is required to complete verification.

lib/db/json.js
```diff
@@ -250,7 +252,7 @@ exports.verificationSecretForEmail = function(email, cb) {
   }, 0);
 };
 
-exports.gotVerificationSecret = function(secret, hash, cb) {
+exports.gotVerificationSecret = function(secret, cb) {
```
lloyd, line 266: so at a higher level, the wsapi handlers will need to verify a password is provided if required.

lib/db/mysql.js
```diff
@@ -420,12 +431,12 @@ exports.userOwnsEmail = function(uid, email, cb) {
...
   secrets.generate(48, function(secret) {
     // overwrite previously staged users
-    client.query('INSERT INTO staged (secret, new_acct, existing_user, email) VALUES(?,FALSE,?,?) ' +
+    client.query('INSERT INTO staged (secret, new_acct, existing_user, email, passwd) VALUES(?,FALSE,?,?,?) ' +
                  'ON DUPLICATE KEY UPDATE secret=VALUES(secret), existing_user=VALUES(existing_user), new_acct=FALSE, ts=NOW()',
-                 [ secret, existing_user, new_email ],
+                 [ secret, existing_user, new_email, hash ],
                  function(err) {
                    cb(err, err ? undefined : secret);
                  });
```
lloyd, line 436: r+ on database changes.  we have some challenges here with migration.  aside from the basic issue of getting the mysql tables altered, all outstanding verification links will be broken - they won't have a password set.  we need to figure out how to handle this.

The easiest possible way to address this would be to delete URLs from the stage table at the time we do the migration, but this will have user impact.  We could mitigate this by improve the page displayed to the user when they validate an email with an invalid link (secret no longer exists in database).

hmm.

lib/wsapi/complete_email_addition.js
```diff
@@ -17,62 +17,18 @@ exports.args = ['token'];
...
-            db.updatePassword(uid, hash, function(err) {
-              if (err) {
-                logger.warn("couldn't update password during email verification: " + err);
-              }
-            });
-          });
-        } else {
-          res.json({ success: true });
-        }
+        wsapi.authenticateSession(req.session, uid, 'password');
```
lloyd: we cannot authenticate this session until we've checked that the user is either using the same browser that they used to start the verification process, or that they know the password.

One implementation tactic would be to have this WSAPI return a 401 if there's nothing on the session indicating that they're on the same browser, and then have the frontend prompt for the password and retry the wsapi.

lib/wsapi/complete_user_creation.js
```diff
@@ -11,50 +11,31 @@ logger = require('../logging.js').logger;
...
-        logger.error("can't bcrypt: " + err);
-        return res.json({ success: false });
+        logger.warn("couldn't complete email verification: " + err);
+        wsapi.databaseDown(res, err);
+      } else {
+        // FIXME: not sure if we want to do this (ba)
+        // at this point the user has set a password associated with an email address
+        // that they've verified.  We create an authenticated session.
+        wsapi.authenticateSession(req.session, uid, 'password',
+                                  config.get('ephemeral_session_duration_ms'));
```
lloyd: ditto here for the comment in complete email addition.

lloyd:
Ok.  I like this change, there are a few things that need to happen before an r+ from me.  I'm conflating a couple other issues into this one, but feel like it's warranted.
1. Implement checks on stage_\* functions, they must require a password if not on the same device - make required updates to verification page GH-290
2. improve user messaging in verification page for invalid links, missing tokens: GH-1495 & GH-1497 (this is to help with migration)
3. test the crap out of this.

I'm perfectly happy trying to nail all of this down today.  The only thing I'll need an assist with is review after I'm done and help with the frontend changes.  

sound resonable?

lloyd:
So for completeness, the other migration tactic would be to add custom UI to request a password upon verification if the verification record does not contain one (user staged before train deployment, user clicks after).

lloyd:
unit tests pass.

lloyd:
"this email looks new, so let's get you set up" doesn't make it clear why I need to type a password, nor what this password is.

How about: "To use this email, choose a new password you'll enter when you sign in with BrowserID:"

dunno...  The former misses an opportunity to make it clear why a password is needed, and what this password is.

tests/cert-emails-test.js
```diff
@@ -32,6 +32,8 @@ suite.addBatch({
   "staging an account": {
     topic: wsapi.post('/wsapi/stage_user', {
       email: 'syncer@somehost.com',
+      pass: 'fakepass',
+      // TODO: is this pubkey needed?
       pubkey: 'fakekey',
```
lloyd: no, this pubkey is bogus.  let's remove it...

tests/password-update-test.js
```diff
@@ -34,6 +34,7 @@ suite.addBatch({
   "account staging": {
     topic: wsapi.post('/wsapi/stage_user', {
       email: TEST_EMAIL,
+      pass: OLD_PASSWORD,
```
lloyd, line 37: we left in an unnec pass arg to complete_user_creation lower down in this file...  let's fix it...

lloyd:
reviewed backend tests, two commits.  r+ for that subset of changes.

lloyd:
Ok, I've attacked this branch a bit and made changes to the backend APIs.  here are the changes:
1. `complete_email_addition` requires that either you're authenticated as the same user who initiated the email verification, or that you've provided in an optional POST var the `pass` of that user.  This api will return a 401 error code if a password is required
2. `complete_user_creation` requires that either you're in the same browser as the initiator of the email verification, OR that you've provided the password chosen by the initiator.  This api will return a 401 error code if a password is required
3. `email_for_token` no longer returns 'needs_password', but instead it returns `must_auth` - which tells you whether you must provide a password to invoke one of the two apis above.

To port to the above apis on the front end, I think you call email_for_token on the verification landing page, and if must_auth is required you have to prompt the user for a password, then send that up to the proper api as a `pass` parameter.

@shane-tomlinson - I'm out of time, but I think this is everything you need to finish off the frontend changes (this is GH-290 we're working on here, for your reference).

The only other bit that's remaining is improving the user visible errors in the verification landing page.  Fielders choice whether you really want to block this PR on that, or instead make that a possible approach when we open up a 5 star issue after merging this to figure out our migration tactic.

lloyd:
so to be clear, after implementation of the frontend to handle re-authentication during verification on a different browser, and review of the commits I've added, I'm r+ on this.  merge.  go.  merge.  woohoo!

lloyd:
with e54da63 the last known issue on this branch is addressed.  tests pass, I did manual testing of all affected flows that I can think of.  r+

Issue #1445, opened 2012-04-16
400 Bad Request: requires authentication response for POST /wsapi/cert_key                                                                  


jrgm:
This one is hard to reproduce, but I've done it 6 or 7 times now. Following the STR below (I doubt it only happens this way,
but that's what I was doing). I get a 400 response if the requests to /wsapi/complete_email_addition, /wsap/email_addition_status, /wsapi/list_emails, /wsapi/cert_key are closely grouped in time.

  We are very sorry, there has been an error!
  Please close this window and try again.
    Action: Getting Assertion
    Network Info: POST: /wsapi/cert_key
    Response Code - 400
    Response Text: Bad Request: requires authentication
    Error Type: Bad Request: requires authentication

Steps to reproduce:
1. Go to http://beta.myfavoritebeer.org/ and click signin.
2. If already authenticated, click 'This is not me'.
3. Enter an @eyedee.me address you have never used before and complete the authentication.
4. Click logout on beer, and click signin again.
5. Enter a non-primary email address you have never used before, collect the email and load the add_email_address in the same browser. Enter a password twice for the user, but do not submit the form yet.
6. This is the hard part. I installed the HTTPFox addon so I could watch the requests for email_addition_status and time when I hit submit. If you time it right, you will get the 400. If not try again.    

Here's the sequence for hitting a 400 from HTTPFox (https://addons.mozilla.org/en-US/firefox/addon/httpfox).
(Columns are in order relative-timestamp, time-for-request, bytes-sent, bytes-received, method, result-code, mime-type, url)

00:06:56.515 0.039 916  248 GET  200 application/json https://diresworb.org/wsapi/email_addition_status?email=abcabdab%2Br27%40gmail.com
00:06:59.584 0.038 916  248 GET  200 application/json https://diresworb.org/wsapi/email_addition_status?email=abcabdab%2Br27%40gmail.com
00:07:02.640 0.048 916  248 GET  200 application/json https://diresworb.org/wsapi/email_addition_status?email=abcabdab%2Br27%40gmail.com
00:07:05.721 0.045 916  248 GET  200 application/json https://diresworb.org/wsapi/email_addition_status?email=abcabdab%2Br27%40gmail.com
00:07:08.324 0.073 997  299 GET  200 application/json https://diresworb.org/wsapi/email_for_token?token=lHeJdLLcKVTz2mqkQFICv9qza2EPzO3ER53g2PjtduAjaTH4
[...]

PR #1452, opened 2012-04-17
Hotfix1445

lloyd:
...ondary email address is being added - mitigates issue #1445 for the purposes of a hotfix to train-2012.04.11

### DIFF
--- a/lib/wsapi/complete_email_addition.js
+++ b/lib/wsapi/complete_email_addition.js
@@ -41,27 +41,32 @@ exports.process = function(req, res) {
       });
     }
 
-    db.gotVerificationSecret(req.body.token, req.body.pass, function(e, email, uid) {
+    // got verification secret's second paramter is a password.  That password
+    // will only be used on new account creation.  Because we know this is not
+    // a new account, we don't provide it.
+    db.gotVerificationSecret(req.body.token, "", function(e, email, uid) {
       if (e) {
         logger.warn("couldn't complete email verification: " + e);
         wsapi.databaseDown(res, e);
       } else {
         // now do we need to set the password?
         if (r.needs_password && req.body.pass) {
+          // requiring the client to wait until the bcrypt process is complete here
+          // exacerbates race conditions in front-end code.  We'll return success early,
+          // here, then update the password after the fact.  The worst thing that could
+          // happen is that password update could fail (due to extreme load), and the
+          // user will have to reset their password.
+          wsapi.authenticateSession(req.session, uid, 'password');
+          res.json({ success: true });
+
           wsapi.bcryptPassword(req.body.pass, function(err, hash) {
             if (err) {
               logger.warn("couldn't bcrypt password during email verification: " + err);
-              return res.json({ success: false });
+              return;
             }
             db.updatePassword(uid, hash, function(err) {
               if (err) {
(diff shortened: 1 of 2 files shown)


lib/wsapi/complete_email_addition.js
```diff
@@ -41,27 +41,32 @@ exports.process = function(req, res) {
...
+          // here, then update the password after the fact.  The worst thing that could
+          // happen is that password update could fail (due to extreme load), and the
+          // user will have to reset their password.
+          wsapi.authenticateSession(req.session, uid, 'password');
+          res.json({ success: true });
+
           wsapi.bcryptPassword(req.body.pass, function(err, hash) {
             if (err) {
               logger.warn("couldn't bcrypt password during email verification: " + err);
-              return res.json({ success: false });
```
ozten, line 54: Do we use statsd in production? Might be a good place to put a counter.

tests/primary-then-secondary-test.js
```diff
@@ -141,6 +141,24 @@ suite.addBatch({
 });
 
 
+// after a small delay, we can authenticate with our password
+suite.addBatch({
+  "after a small delay": {
+    topic: function() { setTimeout(this.callback, 1500); },
```
ozten, line 147: Did this new test consistenly repro the issue (before you added the fix?)

lib/wsapi/complete_email_addition.js
```diff
@@ -41,27 +41,32 @@ exports.process = function(req, res) {
...
+    // got verification secret's second paramter is a password.  That password
+    // will only be used on new account creation.  Because we know this is not
+    // a new account, we don't provide it.
+    db.gotVerificationSecret(req.body.token, "", function(e, email, uid) {
       if (e) {
         logger.warn("couldn't complete email verification: " + e);
         wsapi.databaseDown(res, e);
       } else {
         // now do we need to set the password?
         if (r.needs_password && req.body.pass) {
```
ozten, line 53: Since gotVerificationSecret's hash argument was an empty string, then this block never executes, right?

lib/wsapi/complete_email_addition.js
lloyd, line 54: that's a fantastic idea.

ozten:
Couple questions, but as this is a hotfix, everything looks good.

r+ feel free to merge.

http://travis-ci.org/mozilla/browserid/builds/1115597 Passed

For long term fix, can we change the ordering so that `db.gotVerificationSecret` is not called with an empty string? Although, this isn't horrible and I don't totally grok all the moving pieces, so maybe this patch is also good for the long term fix.

tests/primary-then-secondary-test.js
lloyd, line 147: no, this new test was required to keep the tests passing after I made the fix.

The issue is related to requests running in parallel, specifically `complete_user_creation` and `user_creation_status`.  

When we make a proper, 100% correct fix for this class of race conditions on dev we should introduce some tests that do these things in parallel some number of times and ensure that results are always consistent.

lib/wsapi/complete_email_addition.js
lloyd, line 53: no, this block will execute when the caller of the API has provided a password AND when in the database the user has no password set (which is the norm for accounts with only primary addresses).

PR #1456, opened 2012-04-17
Allowing underscore in domain and origin, fixes Issue#1454

ozten:
Allow `_` in various parts of domain and origin regexp.
Fixes Issue#1454


shane-tomlinson:
r+ - thanks for the update to isEmail.  Get that awsbox up!

Issue #1453, opened 2012-04-17
cancel link in verification screen for required email does not look like a link


shane-tomlinson:

jrgm:
verified on stage train-2012.04.27. For either the required email case, and the general case, the link functions correctly and is styled consisent with other link styling.

PR #1457, opened 2012-04-18
Fix the link color of the cancel button in the required email check email screen.

shane-tomlinson:
- A little bit of cleanup to simplify the check_registration controller - there is only a back button now.
- In the ejs template, remove the href and id attributes from the strings passed to localizers. Always use back for the id.
- Add a test to make sure the correct messages are triggered whenever pressing the back button.

issue #1453


lloyd:
was it your intention that upon click of the 'cancel' link the dialog would close?

shane-tomlinson:
@lloyd - yes, closing the dialog is the desired behavior.  In the required email case, if the user selects "cancel" after they have sent the email, they have no state to which they can return to and take an alternative path.

Issue #1460, opened 2012-04-18
Assertion should be generated AFTER the "is this your computer" screen.


shane-tomlinson:
Right now, we generate the assertion THEN ask "is this your computer?"  Since the assertion is only valid for 2 minutes, this introduces an opportunity for the assertion to be invalid by time it reaches the RP and verifier.

Link to #1446

jrgm:
@stomlinson is this distinct from GH-1446?

shane-tomlinson:
@jrgm - this is a different problem than GH-1446.  This issue manifested itself when the user waited on the "is this your computer"? screen for more than 2 minutes, the user's assertion was expired before it was ever passed to the RP.

jrgm:
Okay, I see now what I was missing when trying against current production (which does not have this fix).

Just noting these STR for posterity:
1) signup a new email address via myfavoritebeer.org
2) logout and wait >60 seconds
3) click sign in on myfavoritebeer.org and use that email address
4) When you get the screen with "Is this your computer". Wait >120 seconds.
5) Click yes or no

Against train-2011.04.11 (current prod) the login to myfavoritebeer.org does not succeed. With train-2011.04.27 on beta.myfavoritebeer.org and diresworb.org, the login will succeed.

jrgm:
So, I've been through the above STR on Win7/WinXP (IE9/IE8,Chrome,FF,Opera,Safari), OSX (Chrome,Safari,Opera,FF), Ubuntu (Chrome,FF,Opera), Android4.0 (Stock,FF,Aurora), Android2.2 (Stock,FF), IOS5 (Safari) and don't have a problem (and also no problems in general with signup/signin/the-question when not waiting out the 2 minutes in answering the question). verified for train-2012.04.27

PR #1465, opened 2012-04-19
Change the order of assertion generation to be after the user is asked "is this your computer?"

shane-tomlinson:
It started as a simple exercise.  It ended in a yak shave.
- This fixes a problem where it is possible that a user takes more than 2 minutes to answer, making the already generated assertion invalid.
- A lot of general cleanup of the state machine.
- Push all post email selection flows to a new state, "email_valid_and_ready" which decides whether to ask the user the computer ownership status.
- Assertions are generated after the "is this your computer" is asked.
- doEmailChosen is replaced with doGenerateAssertion - there were multiple calls to generate an assertion, not very DRY.
- Fix a problem in network.js where a user who completes user registration should be marked as authenticated.
- rename email_chosen.js to generate_assertion.js to clarify its purpose.

issue #1460

### DIFF
--- a/lib/static_resources.js
+++ b/lib/static_resources.js
@@ -97,7 +97,7 @@ var dialog_js = und.flatten([
     '/dialog/controllers/verify_primary_user.js',
     '/dialog/controllers/provision_primary_user.js',
     '/dialog/controllers/primary_user_provisioned.js',
-    '/dialog/controllers/email_chosen.js',
+    '/dialog/controllers/generate_assertion.js',
     '/dialog/controllers/is_this_your_computer.js',
 
     '/dialog/start.js'
--- a/resources/static/dialog/controllers/generate_assertion.js
+++ b/resources/static/dialog/controllers/generate_assertion.js
@@ -3,7 +3,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-BrowserID.Modules.EmailChosen = (function() {
+BrowserID.Modules.GenerateAssertion = (function() {
   "use strict";
 
   var bid = BrowserID,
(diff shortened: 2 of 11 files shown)


lloyd:
omg...

lib/static_resources.js
```diff
@@ -97,7 +97,7 @@ var dialog_js = und.flatten([
     '/dialog/controllers/verify_primary_user.js',
     '/dialog/controllers/provision_primary_user.js',
     '/dialog/controllers/primary_user_provisioned.js',
-    '/dialog/controllers/email_chosen.js',
+    '/dialog/controllers/generate_assertion.js',
     '/dialog/controllers/is_this_your_computer.js',
```
lloyd, line 101: verified removed controller wasn't references anywhere else

resources/static/dialog/controllers/generate_assertion.js
```diff
@@ -3,7 +3,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-BrowserID.Modules.EmailChosen = (function() {
+BrowserID.Modules.GenerateAssertion = (function() {
```
lloyd, line 6: I like the name change.  GenerateAssertion is more granular and clear to me

resources/static/dialog/resources/state.js
```diff
@@ -123,6 +123,10 @@ BrowserID.State = (function() {
     handleState("primary_user_provisioned", function(msg, info) {
       info = info || {};
       info.add = !!addPrimaryUser;
+      // The user is is authenticated with their IdP. Two possibilities exist
+      // for the email - 1) create a new account or 2) add address to the
+      // existing account. If the user is authenticated with BrowserID, #2
+      // will happen. If not, #1.
```
lloyd, line 129: +1 on addition of comments like this

lloyd:
this is a massive change.  the restructuring and renaming makes sense to me, and from a read of the code I'm comfortable with it, provided we put extra testing focus on it, which I'll message to QA.

The bug this fixes already hit someone in IRC in production, so I feel like this is worth the risk.

r+

Issue #968, opened 2012-01-24
When the user cancels a log in of an unauthenticated primary account, the email is still marked as most recently used


ghost:
STR:
1. Launch http://beta.myfavoritebeer.org.
2. Sign in with a BID with multiple emails from a primary (eyedee.me).
3. Select an email and log in.
4. Log out, the click on the Sign In button again.
5. Select another email address, then click on the Sign In button.
6. Click on the Cancel button.

The user is returned to the dialog with the emails list and the email selected at step 5 is marked as most recently used. The email selected at step 3 should be the marked one since this is the last email the user logged in with.

jrgm:
verified on stage train-2012.04.27 using FF 12 and an account with 3 primary eyedee.me addresses

PR #1466, opened 2012-04-19
Fix primary email being set as most recently used if the user cancels redirect to authenticate with primary.

shane-tomlinson:
- all emails are set for the origin in user.getAssertion.  Having it in pick_email meant that as soon as a user chose an address and the controller was closed, the email was marked as the most recently used when this may not always be the case.

issue #968


lloyd:
r+ - I reviewed the code and it looks good.

I ran locally and tested spawning dialog, partially adding primary user, canceling dialog, re-opening dialog, and verifying that the last used email setting was still correct.

Issue #1446, opened 2012-04-17
"is this your computer" can show when adding a new email address


shane-tomlinson:
STR:

1) Create a new account.
2) Wait 60 seconds
3) Open dialog, add a new secondary address.
4) Verify address after 60 more seconds.
5) Uh oh.  "Is this your computer?" is displayed.

jrgm:
verified on stage train-2012.04.27 for the above STR the 'is this your computer' is not displayed when adding the second
email address to the account. The question will be displayed the next time the dialog is used >60 seconds from then.

PR #1467, opened 2012-04-19
Fix a user being asked "is this your computer" if they just verified an email.

shane-tomlinson:
Note - I am issuing this pull request with the knowledge that it will need to be updated if PR 1465 is merged before this.  If this is merged before PR 1465, that pull request will need to be updated to include use of this new API.

issue #1446


lloyd:
r+ - read the code and love the approach.  verified that there was only one place where shouldask was invoked (very DRY, stomlinson).

testing: reproduced issue in old code.  merged new code.  followed STR, no question after verificaition.  then signed in again, got the question.  All as intended.

Issue #1481, opened 2012-04-23
IE6/IE7: myfavoritebooze.org errors on page load


shane-tomlinson:
Line: 16
Error: Exception thrown and not caught

shane-tomlinson:
Link to #1390

lloyd:
wontfix, booze will die

PR #1482, opened 2012-04-23
Hotfix 1390 fix ie6 ie7

shane-tomlinson:
- storage.js - change var storage = localStorage to storage = window.localStorage in case localStorage does not exist.
- include.js - check to ensure the browser is supported before getting a silent assertion or else an exception is thrown when trying to build the JSChannel.

issue #1390
issue #1481

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -956,12 +956,21 @@
         throw "navigator.id.get() requires a callback argument";
       }
 
+      var browserSupported = BrowserSupport.isSupported();
+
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@lloyd - Can you review this one?  This will have to be updated for your changes for the new API.  Before trying to get a silent assertion we have to check to make sure the browser is supported.

shane-tomlinson:
Since this API is being deprecated this week, closing this in favor of a pull request against train-2012.04.11

Issue #1485, opened 2012-04-23
[IE8] error dialog shows up with solid black background (i.e., not just dimmed)


jrgm:
This is specific to IE8

STR
1) disable cookies in IE8
2) go to https://diresworb.org/signup
3) Enter a never before seen email address and click 'verify email'
4) An error message is shown as expected 'Browserid requires cookies'

However, the rest of the screen is now all black, not just dimmed as in other browsers.

shane-tomlinson:
Label 5 star since this is a regression.

shane-tomlinson:
Closing - merged into train-2012.04.11 and current dev.

jrgm:
verified fix on stage for train-2012.04.11 with the latest round of updates

PR #1486, opened 2012-04-24
Hotfix1485 ie8 black background for IE8 when it shows an error message.

shane-tomlinson:
issue #1485


shane-tomlinson:
cherry picked this into dev in commit 1e351badbf9bc07e7adaff87941eb86c023b0347

Issue #1484, opened 2012-04-23
[Chrome] exceptions thrown on /signup page when cookies disabled 


jrgm:
This is basically already covered in GH-1418, but chrome throws 'Uncaught Error: QUOTA_EXCEEDED_ERR:
DOM Exception 22' in two places when using the /signup page on diresworb.org (at lines 123 and 180
of resources/static/shared/storage.js).

jrgm:
The reason I filed this bug as separate from GH-1418 is that this is a core problem with storage.js with cookies disabled in chrome.

shane-tomlinson:
Labeling a 5\* because this is a regression.

shane-tomlinson:
Closing, this issue is avoided entirely with PR 1489 which has been merged.

jrgm:
verified on stage train-2012.04.27 specifically for Chrome (this issue was just about Chrome). Loading /signup (or any other path at diresworb.org) shows no js exception in the console.

PR #1487, opened 2012-04-24
Hotfix 1484 chrome exception signup

shane-tomlinson:
Swallow exceptions that Chrome throws when writing to localStorage if cookies are disabled.

I am not fully happy with this solution as it is a bandaid instead of a true fix for the root cause. I am writing a fuller solution against the current dev to show the cookies disabled error screen as soon as the user lands on the majority of main site pages.

issue #1484.

### DIFF
--- a/resources/static/shared/storage.js
+++ b/resources/static/shared/storage.js
@@ -28,6 +28,14 @@ BrowserID.Storage = (function() {
     }
   }
 
+  function logError(msg) {
+    // Check for the existence of window.console for IE.  IE does not support console
+    // logging unless its dev tools are open.
+    if(window.console && console.log) {
+      console.log(msg);
+    }
+  }
+
   function storeEmails(emails) {
     storage.emails = JSON.stringify(emails);
   }
@@ -128,6 +136,7 @@ BrowserID.Storage = (function() {
     }
     catch(e) {
       // Chrome, Safari and FF will blow up here if cookies are disabled.
+      logError("error writing to storage: " + e);
     }
   }
 
@@ -187,6 +196,7 @@ BrowserID.Storage = (function() {
     } catch(e) {
       // Do nothing, Chrome, Safari and FF will blow up here if cookies are
       // disabled.
+      logError("error writing to storage: " + e);
     }
   }
 
@@ -195,6 +205,7 @@ BrowserID.Storage = (function() {
       var allInfo = JSON.parse(storage[namespace] || "{}");
       return allInfo[key];
     } catch(e) {
+      logError("error reading from storage: " + e);
       return undefined;
     }
   }
--- a/resources/static/shared/storage.js
+++ b/resources/static/shared/storage.js
@@ -29,10 +29,13 @@ BrowserID.Storage = (function() {
   }
 
   function logError(msg) {
-    // Check for the existence of window.console for IE.  IE does not support console
-    // logging unless its dev tools are open.
-    if(window.console && console.log) {
+    try {
       console.log(msg);
+    } catch(e) {
+      // Do nothing. The console.log is inside of a try/catch because some
+      // versions of IE do not have window.console unless the dev tools are
+      // open.  This ensures that any browser that does not support console.log
+      // will not cause any problems while we are trying to log errors.
     }
   }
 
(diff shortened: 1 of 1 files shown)


benadida:
should you maybe console.log something so it's not failing silently?

shane-tomlinson:
@benadida - good idea.  I will add console logging.

resources/static/shared/storage.js
```diff
@@ -175,14 +192,25 @@ BrowserID.Storage = (function() {
...
 
   function generic2KeyGet(namespace, key) {
-    var allInfo = JSON.parse(storage[namespace] || "{}");
-    return allInfo[key];
+    try {
+      var allInfo = JSON.parse(storage[namespace] || "{}");
+      return allInfo[key];
+    } catch(e) {
+      logError("error reading from storage: " + e);
+      return undefined;
```
ozten, line 212: Cool, the semantics of returning undefined look the same, assuming the old code it was possible to ask for a key that didn't exist in allInfo.

resources/static/shared/storage.js
```diff
@@ -28,6 +28,17 @@ BrowserID.Storage = (function() {
     }
   }
 
+  function logError(msg) {
+    try {
+      console.log(msg);
+    } catch(e) {
+      // Do nothing. The console.log is inside of a try/catch because some
```
ozten, line 35: This is fine. I'm surprised we didn't have something like this in the code already:

```
if (window.console && console.log ) console.log(msg);
```

resources/static/shared/storage.js
```diff
@@ -175,14 +192,25 @@ BrowserID.Storage = (function() {
...
-    allInfo[key] = value;
-    storage[namespace] = JSON.stringify(allInfo);
+    try {
+      var allInfo = JSON.parse(storage[namespace] || "{}");
+      allInfo[key] = value;
+      storage[namespace] = JSON.stringify(allInfo);
+    } catch(e) {
+      // Do nothing, Chrome, Safari and FF will blow up here if cookies are
+      // disabled.
+      logError("error writing to storage: " + e);
```
ozten, line 202: Can the app continue without this storage? If not, we should have a global flag somewhere to bail and show an error message to the user.

If this is just a performance enhancement or we have other workarounds, than this looks good.

ozten:
This is looking really good. One question before I'm ready to r+ or r-.

resources/static/shared/storage.js
shane-tomlinson, line 212: I could remove that return undefined, eh?

resources/static/shared/storage.js
shane-tomlinson, line 35: in PR-1489(https://github.com/mozilla/browserid/pull/1489), I am adding a helpers.log so we can have this sort of logic in a single location - https://github.com/mozilla/browserid/pull/1489/files#L1R70

resources/static/shared/storage.js
shane-tomlinson, line 202: @ozten - the app cannot continue without storage.  I agree with your concern here and created PR-1489 to globally address this for all pages on the main site (minus the splash page) - if cookies are disabled (making localStorage unavailable), then an error is displayed to the user immediately upon landing on the page.

shane-tomlinson:
@ozten, @benadida - The aim of this pull request is to prevent Chrome from excepting on the signin page whenever cookies are disabled.  What we do in the signin, signup, and forgot password pages is every time the user updates the email address, we write the email address to localStorage.  If cookies are disabled in Chrome, an exception is thrown.  With PR-1489 (https://github.com/mozilla/browserid/pull/1489), I avoid this situation entirely.  An error message is displayed to the user as soon as they land on the page if their cookies are disabled.

Since this was a hotfix that was aimed at fixing the current beta and since it has been decided to let this exception continue to occur, I feel a more complete solution should be used in place of this one.

Perhaps _all_ reads and writes to localStorage should occur through a set of accessors that contain try/catch blocks.  In this way, we could write console messages from a single place instead of sprinkling them throughout storage.js

ozten:
This sounds like a good cleanup idea! hotfix vs long term solution makes sense.

Issue #1418, opened 2012-04-11
signup/signin/forgot pages from main site should check for cookies


shane-tomlinson:
The signup/signin pages do not check for whether the user has their cookies enabled before allowing the user to sign in.

jrgm:
This should also include the /forgot page, of course.

shane-tomlinson:
@jbonacci - is the above related to #1499?  If so, are they separate bugs?

Using a new profile with cookies disabled in Chrome for OSX on diresworb.org/signin, I receive the following error message:

```
BrowserID requires cookies
Please close this window, enable cookies and try again
Action: Authenticating User
Network Info: POST: /wsapi/authenticate_user
Response Code - 403
Response Text: Forbidden: no cookie
Error Type: Forbidden: no cookie
```

Which is the same as I see on browserid.org:

```
BrowserID requires cookies
Please close this window, enable cookies and try again
Action: Authenticating User
Network Info: POST: /wsapi/authenticate_user
Response Code - 403
Response Text: Forbidden: no cookie
Error Type: Forbidden
```

shane-tomlinson:
@jbonacci - can you look at https://github.com/mozilla/browserid/pull/1489 - this makes it so the signup, signin, forgot, verify_email_address, and add_email_address pages all have a cookie check as soon as the user lands.  If the user's cookies are disabled, a message will be displayed as soon as they land on the page.  That means the user won't even have the opportunity to enter their email address in.  This fix should also fix #1484 - Chrome excepting when cookies are disabled.  @ozten merged PR-1489 into dev today (2012.04.25)

jrgm:
404? Not found? For which URL. Or did you mean 403?

jrgm:
So, I've used Win7/WinXP (IE9/IE8,Chrome,FF,Opera,Safari), OSX (Chrome,Safari,Opera,FF), Ubuntu Chrome,FF,Opera), Android4.0 (Stock,FF,Aurora), Android2.2 (Stock,FF), IOS5 (Safari) with cookies disabled and visiting /signin /signup etc. and popping the /sign_in dialog from an RP. 

They all do immediately show the 'requires cookies' screen immediately on load, except GH-1542 where it doesn' t show for stock browser on Android 4.0 (but does show for stock/Android2.2). 

I'll note in passing the that on mobile with Gecko browsers (either XUL or native) with the 'requires cookies' screen showing, touching links in the background show some visual and haptic response to the touch but don't actually have any effect. 

[...]

PR #1489, opened 2012-04-24
Add a cookie check to most main site pages.

shane-tomlinson:
- Add cookie check to signin, signup, forgot, add_email_address, verify_email_address.
- to user.js->checkAuthentication, call the callback with false if cookies are disabled.

issue #1418
issue #1484

### DIFF
--- a/resources/static/pages/start.js
+++ b/resources/static/pages/start.js
@@ -11,6 +11,7 @@ $(function() {
    */
 
   var bid = BrowserID,
+      helpers = bid.Helpers,
       pageHelpers = bid.PageHelpers,
       user = bid.User,
       dom = bid.DOM,
@@ -78,6 +79,11 @@ $(function() {
     }
     else if(token && path === "/verify_email_address") {
       bid.verifyEmailAddress(token);
+    }
+    else {
+      // Instead of throwing a hard error here, adding a message to the console
+      // to let developers know something is up.
+      helpers.log("unknown path");
     }
 
     user.checkAuthentication(function(authenticated) {
--- a/resources/static/pages/start.js
+++ b/resources/static/pages/start.js
@@ -20,9 +20,10 @@ $(function() {
       path = document.location.pathname,
       moduleManager = bid.module,
       modules = bid.Modules,
-      CodeCheck = modules.CodeCheck,
+      CookieCheck = modules.CookieCheck,
       XHRDelay = modules.XHRDelay,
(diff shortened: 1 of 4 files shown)


resources/static/pages/start.js
```diff
@@ -33,76 +34,85 @@ $(function() {
...
+    }
+    else if (path === "/signup") {
+      bid.signUp();
+    }
+    else if (path === "/forgot") {
+      bid.forgot();
+    }
+    else if (path === "/add_email_address") {
+      var module = bid.addEmailAddress.create();
+      module.start({
```
ozten, line 76: Should we add this catch-all back in?

shane-tomlinson, line 76: The diff got confused and made it look like there was a catch all case for the path.  The below catch all (line 88 on the left) matches the if(cookiesEnabled) (line 77 on the left) from above.  But, a catch all with a console log is a good suggestion to aid developers who add new pages but do not add a corresponding handler.  Adding now.

resources/static/pages/start.js
```diff
@@ -33,76 +35,90 @@ $(function() {
...
+        token: token
       });
     }
+    else if(token && path === "/verify_email_address") {
+      bid.verifyEmailAddress(token);
+    }
     else {
-      displayNonAuthenticated();
```
shane-tomlinson, line 89: This catch all case actually belongs with the above removed code around line 77 - it is the else statement that goes with the "if(cookiesEnabled)"

resources/static/pages/start.js
```diff
@@ -33,76 +35,90 @@ $(function() {
...
+      // Instead of throwing a hard error here, adding a message to the console
+      // to let developers know something is up.
+      helpers.log("unknown path");
```
shane-tomlinson, line 86: @ozten - here is the new and improved catch all.

resources/static/shared/helpers.js
```diff
@@ -67,6 +67,15 @@
     }
   }
 
+  function log(msg) {
+    try {
+      window.console.log(msg);
```
shane-tomlinson, line 72: Adding a helpers.log message that prints to the browser's console if the console is available.

resources/static/pages/start.js
ozten, line 86: Cool.

ozten:
Great work! r+

Issue #1464, opened 2012-04-19
[ie8/ie9] IE reporting a problem with test check_registration: user validation with mustAuth result


jrgm:
On both ie8(winxp) and ie9(win7) the browser detects an uncaught exception for
http://host:port/test/?filter=controllers%2fcheck_registration%3a%20user%20validation%20with%20mustAuth
The test itself reports as passed.

jrgm:
This is of course related to GH-1440

jrgm:
And just to be clear, the fix for GH-1440 for those STR does work in ie8 and ie9 (and other).

shane-tomlinson:
@jrgm - are you seeing the following:

```
SEC7111: HTTPS security is compromised by http://testmob.org/include.js 
index.html?filter=controllers%2fcheck_registration%3a user validation with mustAuth
```

jrgm:
I didn't notice that. In MS dev tools console? I'll look again.

jrgm:
This is on train-2011.04.11 branch

shane-tomlinson:
@jrgm - I am now seeing:

```
SCRIPT5022: authenticate triggered more than once 
helpers.js, line 28 character 9
SCRIPT5022: authenticate triggered more than once 
helpers.js, line 28 character 9
```

shane-tomlinson:
@jrgm - excellent catch, thanks!

jrgm:
Okay, so I don't have this error now with ie8 or ie9 and a local pull of the latest for train-2012.04.11. @stomlinson did you mean to close this, or has the fix not been merged back to dev branch yet?

jrgm:
Ah, so, duh, I see in https://github.com/mozilla/browserid/pull/1490#issuecomment-5308907 that this does not need to 
backported to dev. So I'm going to close and call it verified

PR #1490, opened 2012-04-24
issue 1464 ie exception check registration

shane-tomlinson:
Fix the broken test in check_registration.  It was starting the registration check twice, meaning there were two concurrent polls happening.

I am submitting this pull request against the current beta because that is where @jrgm reported the problem.  A hotfix is not needed because this is not a user facing problem.

issue #1464

### DIFF
--- a/resources/static/test/cases/controllers/check_registration.js
+++ b/resources/static/test/cases/controllers/check_registration.js
@@ -47,18 +47,18 @@
     controller.startCheck();
   }
 
-  asyncTest("user validation with mustAuth result", function() {
+  asyncTest("user validation with mustAuth result - callback with email, type and known set to true", function() {
     xhr.useResult("mustAuth");
     createController("waitForUserValidation");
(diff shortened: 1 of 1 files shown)


benadida:
@shane-tomlinson if this is not a user-facing problem, maybe it's worth rebasing this against dev and waiting 2 weeks to release? No need to increase risk on this week's production release, I think.

shane-tomlinson:
@benadida - I created this patch against the 2012.04.11 train because @jrgm has been checking out the code and running the tests in a local environment.  Doing so with the 2012.04.11 train will continue to throw the exceptions. The merge can wait for the next train, but I think this is a low risk merge because it modifies the unit tests and none of the user facing code.

benadida:
oh, I should have looked at the diff first. r+.

shane-tomlinson:
Thanks @benadida.  This does not need backported to dev as dev took a different approach to this test that already had it fixed.

PR #1494, opened 2012-04-24
Hotfix 1390 min browser reqs

shane-tomlinson:
@lloyd - can you review this?  With this change, I am no longer adding the communication_iframe to the page if the user's browser does not meet the minimum browser requirements.  I did this because IE6 and IE7 throw an exception from within the communication_iframe.  We can prevent the iframe from being added on browsers we know do not meet minimum requirements, saving the users a few KB and keeping this error from happening.

A second change is on line 1126, in the logout function.  Before sending a message to the commChan, check to see if it exists.

To fully test this:
- Check both IE6 and IE7 to ensure they no longer throw an exception when loading the dev RP.
- Check browsers we know support the protocol to ensure they still get their silent assertions.

Changes:
- include.js - check to ensure the browser is supported before setting up the communication_iframe.  Adding the iframe in unsupported browsers will cause an exception within the iframe.

issue #1390

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -958,10 +958,16 @@
       }
     }
 
-    var commChan;
+    var commChan,
+        browserSupported = BrowserSupport.isSupported();
 
     // this is for calls that are non-interactive
     function _open_hidden_iframe() {
+      // If this is an unsupported browser, do not even attempt to add the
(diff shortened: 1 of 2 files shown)


lloyd:
r+ - ultra low risk focused change.  I can't think of how this would break anything.

shane-tomlinson:
I am closing this pull request in favor of a new one aimed at the current dev.

PR #1506, opened 2012-04-27
Fix IE6 and IE7 throwing exceptions in the communication iframe.

shane-tomlinson:
Pull request to replace https://github.com/mozilla/browserid/pull/1494.  This is against the current dev instead of the past train.
- include.js - check to ensure the browser is supported before setting up the communication_iframe.  Adding the iframe in unsupported browsers will cause an exception within the iframe.

issue #1390

Conflicts:

```
resources/static/include_js/include.js
resources/static/shared/storage.js
```

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -958,10 +958,16 @@
       }
     }
 
-    var commChan;
+    var commChan,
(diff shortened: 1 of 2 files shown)


lloyd:
this looks exactly like PR # 1494, which I already reviewed.  No hidden fees.  like r+

Issue #1505, opened 2012-04-27
IE6/dev: unsupported dialog not showing error message.


shane-tomlinson:
STR:

1) Open IE6.
2) Go to dev.myfavoritebooze.org
3) Click sign in.
4) Unsupported dialog opens, but where is the text?

This is not occurring on either prod or beta.

Issue #1496, opened 2012-04-24
IE6/IE7: unsupported dialog shows unecessary scroll bar (and sometimes two!)


shane-tomlinson:

jrgm:
verified fixed on stage train-2012.04.11 for IE6 and IE7 (winxp).

shane-tomlinson:
Closing since @jrgm has already verified.

jrgm:
I was aware of the pull request to fix myfavoritebeer - https://github.com/lloyd/myfavoritebeer.org/pull/56 so I used 
beta.myfavoriteshow.org to bring up the dialog with cookies disabled on IE6 and IE7  -- there is no double scrollbar. 
There is a disabled scrollbar that looks acceptable to me.

PR #1507, opened 2012-04-27
IE6 and IE7 fixes for the unsupported dialog.

shane-tomlinson:
This actually fixes two issues.  The first is scroll bars showing in the unsupported dialog in IE6 and IE7.  The second is IE6 not showing the contents of the unsupported dialog.

issue #1496 - scroll bars
issue #1505 - IE6 not showing contents of unsupported dialog.


lloyd:
r+ - this is a small focused CSS change that targets only IE6/IE7 browsers

Issue #1508, opened 2012-04-27
after password reset verification, dialog does not complete and becomes unresponsive.


shane-tomlinson:

shane-tomlinson:
Note, this does not affect the current beta as of Fri 2012.04.27

shane-tomlinson:
Dup of #1440.  Closing.

PR #1510, opened 2012-04-27
Fix reset password completion not generating an assertion.

shane-tomlinson:
issue #1508

### DIFF
--- a/resources/static/dialog/resources/state.js
+++ b/resources/static/dialog/resources/state.js
@@ -241,9 +241,14 @@ BrowserID.State = (function() {
     });
 
     handleState("reset_password", function(msg, info) {
-      // reset password says the password has been reset, now waiting for
-      // confirmation.
       info = info || {};
+      // reset password says the password user has confirmed that they want to
+      // reset their password.  doResetPassword will attempt to call the
+      // call the create user wsapi call.  If the wsapi call is successful,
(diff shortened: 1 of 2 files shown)


resources/static/dialog/resources/state.js
```diff
@@ -241,9 +241,14 @@ BrowserID.State = (function() {
...
-      // reset password says the password has been reset, now waiting for
-      // confirmation.
       info = info || {};
+      // reset password says the password user has confirmed that they want to
+      // reset their password.  doResetPassword will attempt to call the
+      // call the create user wsapi call.  If the wsapi call is successful,
+      // the user will be directed to the "go verify your account" message.
+
+      // We have to save off the staged email address here for when the user
+      // verifies their account and user_confirmed is called.
```
lloyd, line 251: this comment has grammar problems

shane-tomlinson, line 251: @lloyd - _I_ have grammar problems.  I'll update the comment.

lloyd:
what am I missing here?  I see zero change in terms of code that will run in normal usage of the dialog.  Is this intended to be a comment + test _only_ change?

lloyd:
comfirmed with @shane-tomlinson that this PR and commits merged when 2012.04.11 was merged to dev fully address the issue.  this PR is zero code.  r+

Issue #1495, opened 2012-04-24
If no verification token is given, /verify_email_address does not show an error.


shane-tomlinson:
STR:

1) Open browser to https://browserid.org/verify_email_address

Expected:

Error message saying the token is invalid.

shane-tomlinson:
Closing as fixed with fca7da4

PR #1520, opened 2012-05-01
Start the verifyEmailAddress page controller even if there is no token.

shane-tomlinson:
To test:
- Make sure verification works if a valid token is given.
- Make sure error message is displayed if invalid token is given.
- Make sure error message is displayed if no token is given.

issue #1495

### DIFF
--- a/resources/static/pages/start.js
+++ b/resources/static/pages/start.js
@@ -79,7 +79,7 @@ $(function() {
         verifyFunction: "verifyEmail"
       });
     }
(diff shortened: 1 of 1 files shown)


ozten:
Makes sense,
tests pass http://travis-ci.org/#!/mozilla/browserid/builds/1234468

Rock it!
r+

benadida:
@ozten any reason you didn't merge this?

ozten:
I must have hit merge, cancel. Merged now.

PR #1533, opened 2012-05-02
upgraded to new jwcrypto API, including backend, frontend, and tests.

benadida:
plus conformance tests.


bin/verifier
```diff
@@ -88,7 +88,7 @@ app.post('/verify', function(req, resp, next) {
     } catch (e) {
       return resp.json({ status: "failure", reason: "Content-Type expected to be one of: " + want_ct.join(", ") }, 415);
     }
-    return resp.json({ status: "failure", reason: "need assertion and audience" }, 400);
+    return resp.json({ status: "failure", reason: "need assertion and audience"}, 400);
```
lloyd, line 91: awww.  I liked that space. (NOTE: this is funny)

lib/keysigner/ca.js
```diff
@@ -4,15 +4,18 @@
 
 // certificate authority
 
-var jwcert = require('jwcrypto/jwcert'),
-    jwk = require('jwcrypto/jwk'),
-    jws = require('jwcrypto/jws'),
+var jwcrypto = require('jwcrypto'),
+    cert = jwcrypto.cert,
```
lloyd, line 8: +1 on a single entry point

lib/keysigner/ca.js
```diff
@@ -4,15 +4,18 @@
...
     path = require("path"),
     fs = require("fs"),
     secrets = require('../secrets.js'),
     logger = require('../logging.js').logger,
     urlparse = require('urlparse');
 
+// load up the right algorithms
+require("jwcrypto/lib/algs/rs");
+require("jwcrypto/lib/algs/ds");
```
lloyd, line 17: this seems confusing and magical to me.  Why do you need to explicitly require algorithms to load them?  Could this possibly happen on demand?

lib/keysigner/ca.js
```diff
@@ -22,37 +25,58 @@ try {
...
 
-function certify(hostname, email, publicKey, expiration) {
+function certify(hostname, email, publicKey, expiration, cb) {
   if (expiration == null)
-    throw "expiration cannot be null";
-  return new jwcert.JWCert(hostname, expiration, new Date(), publicKey, {email: email}).sign(secret_key);
+    return cb("expiration cannot be null");
+
+  cert.sign(publicKey, {email: email},
+            {iss: hostname, iat: new Date(), exp: expiration},
```
lloyd: (nitpick warning)  unfortunate that `iss`, `iat`, which are arguably illegible protocol details must be known by the user of the api.

lib/keysigner/ca.js
```diff
@@ -22,37 +25,58 @@ try {
...
+    });  
 }
 
 // exports, not the key stuff
 exports.certify = certify;
 exports.verifyChain = verifyChain;
+exports.verifyBundle = verifyBundle;
 exports.parsePublicKey = parsePublicKey;
-exports.parseCert = parseCert;
 exports.PUBLIC_KEY = public_key;
```
lloyd, line 82: not all of this code exported is run in production.  there are some comments in the code that are disconcerting when you don't know that (it took me a while to figure out that `verifyChain` is only used by our own unit tests.  The comment that concerned me was _for now we only do our own browserid.org keys_.

not a regression, shouldn't block the party.

lib/static_resources.js
```diff
@@ -19,7 +19,8 @@ var common_js = [
   '/lib/jquery-1.7.1.min.js',
   '/lib/winchan.js',
   '/lib/underscore-min.js',
-  '/lib/vepbundle.js',
+//  '/lib/vepbundle.js',
```
lloyd: why comment?  delete!

lib/static_resources.js
```diff
@@ -118,7 +119,8 @@ exports.resources = {
     '/lib/jschannel.js',
     '/lib/winchan.js',
     '/lib/underscore-min.js',
-    '/lib/vepbundle.js',
+//    '/lib/vepbundle.js',
+    '/lib/bidbundle-min.js',
```
lloyd: ditto

lloyd:
tests fail.  both locally for me and travis agrees: http://travis-ci.org/#!/mozilla/browserid/builds/1231601

lib/verifier/certassertion.js
```diff
@@ -120,35 +114,33 @@ function verify(assertion, audience, successCB, errorCB) {
...
-    }, function(pk, principal) {
-      var tok = new jwt.JWT();
-      tok.parse(bundle.assertion);
-
+    }, function(err, certParamsArray, payload, assertionParams) {
+      if (err) return errorCB(err);
+        
       // audience must match!
-      var err = compareAudiences(tok.audience, audience)
+      var err = compareAudiences(payload.aud, audience)
```
lloyd: `.aud` is a bit of a readability loss by clients of the lib.

package.json
```diff
@@ -17,8 +17,8 @@
         "etagify": "0.0.2",
         "express": "2.5.0",
         "iconv": "1.1.3",
-        "jwcrypto": "0.1.1",
         "mustache": "0.3.1-dev",
+        "jwcrypto": "https://github.com/mozilla/jwcrypto/tarball/978614f01c17739b507",
```
lloyd: jwcrypto 0.2.0 is published, we can update to that.

resources/static/shared/network.js
```diff
@@ -47,7 +47,8 @@ BrowserID.Network = (function() {
     // seed the PRNG
     // FIXME: properly abstract this out, probably by exposing a jwcrypto
     // interface for randomness
-    require("./libs/all").sjcl.random.addEntropy(result.random_seed);
+    // require("./libs/all").sjcl.random.addEntropy(result.random_seed);
+    // FIXME: this wasn't doing anything for real, so commenting this out for now
```
lloyd, line 51: if we don't want to block the merge on this (technically not a regression), let's open an issue to track?

resources/static/shared/provisioning.js
```diff
@@ -69,8 +69,10 @@ BrowserID.Provisioning = (function() {
     });
 
     chan.bind('genKeyPair', function(trans, s) {
-      keypair = jwk.KeyPair.generate("DS", BrowserID.KEY_LENGTH);
-      return keypair.publicKey.toSimpleObject();
+      trans.delayReturn(true);
```
lloyd, line 72: author of jschannel is a doofus, he shoulda taken a page from vows, perhaps.

resources/static/shared/storage.js
```diff
@@ -111,9 +111,12 @@ BrowserID.Storage = (function() {
     storage.tempKeypair = null;
     if (raw_kp) {
       prepareDeps();
-      var kp = new jwk.KeyPair();
-      kp.publicKey = jwk.PublicKey.fromSimpleObject(raw_kp.publicKey);
-      kp.secretKey = jwk.SecretKey.fromSimpleObject(raw_kp.secretKey);
+
+      // WORK HERE
```
lloyd: let's improve this comment.  it's ambiguous and could be a merge blocker.

resources/static/shared/user.js
```diff
@@ -36,14 +33,14 @@ BrowserID.User = (function() {
         // if the certificate expires in less that 5 minutes from now.
         function isExpired(cert) {
           // if it expires in less than 2 minutes, it's too old to use.
-          var diff = cert.expires.valueOf() - serverTime.valueOf();
+          var diff = cert.payload.exp.valueOf() - serverTime.valueOf();
```
lloyd, line 36: ok, third times the charm, then I shut up.  Being a crypto idiot, this coule be the exponent of the payload, it could be an expiration date.  Now I'm done harping on this.

resources/static/shared/user.js
```diff
@@ -929,10 +925,12 @@ BrowserID.User = (function() {
      */
     syncEmailKeypair: function(email, onComplete, onFailure) {
       prepareDeps();
-      var keypair = jwk.KeyPair.generate("DS", bid.KEY_LENGTH);
-      setTimeout(function() {
-        certifyEmailKeypair(email, keypair, onComplete, onFailure);
-      }, 0);
+      jwcrypto.generateKeypair({algorithm: "DS", keysize: bid.KEY_LENGTH}, function(err, keypair) {
+        // FIXME: eventually we can get rid of setTimeout since it's already a callback
```
lloyd: can we get rid of it now, or is does generateKeypair invoke callback before function completion?

resources/static/shared/user.js
```diff
@@ -958,21 +956,24 @@ BrowserID.User = (function() {
...
-              var tok = new jwt.JWT(null, expirationDate, audience);
-
-              // yield!
-              setTimeout(function() {
-                assertion = vep.bundleCertsAndAssertion([idInfo.cert], tok.sign(sk), true);
-                storage.site.set(audience, "email", email);
-                complete(assertion);
-              }, 0);
+
+              jwcrypto.assertion.sign(
```
lloyd: followon from previous comment.  is assertion.sign() guaranteed to invoke its callback argument after it returns?  if not, then this is a regression that could cause the "javascript is taking too long" errors we've seen on slow browsers (read: IE8).  Given our reduction of key length, this might no longer be an issue, worth the comment tho.

my preference is to have all apis in jwcrypto guarantee that callbacks are not invoked before the function that they're passed to returns.

resources/static/shared/user.js
```diff
@@ -958,21 +956,24 @@ BrowserID.User = (function() {
...
+                {aud: audience}, {exp: expirationDate},
+                sk,
+                function(err, signedAssertion) {
+                  setTimeout(function() {
```
lloyd: followup to my followup, what's the utility of setTimeout here?

tests/verifier-test.js
```diff
@@ -790,42 +897,61 @@ suite.addBatch({
...
         wsapi.post('/verify', {
           audience: TEST_ORIGIN,
           assertion: assertion
         }).call(this);
       },
-      "to return a clear error message": function (err, r) {
+      "to succeed": function (err, r) {
         var resp = JSON.parse(r.body);
         assert.strictEqual(resp.status, 'okay');
         assert.strictEqual(resp.issuer, "example.domain");
```
lloyd, line 961: _awesome_ updates to test coverage.  I did a cursory sampling, but love how many new areas we're testing.

lloyd:
tl;dr; r-

Stuff that I think should block the merge:
1. tests must pass
2. all asynchronous functions exposed by jwcrypto should guarantee not to invoke callbacks before they return - and we should remove setTimeouts in higher level code.
3. comment which says "WORK HERE" should be updated by @benadida - it's either a critical oversight or a future improvement, i can't tell.
4. we should update to jwcrypto 0.2.0 in package.json - npm does horribly with URLs and 0.2.0 is published
5. future improvements that are called out in comments should probably be turned into issues at the time we merged so we can triage and they don't get lost.

Non-critical stuff that lloyd cares about but shouldn't block the merge:
1. ideally jwcrypto would abstract us from three letter abbreviations
2. remove comments that pertain to a previous version of the code
3. for keysigner, because the code is so high value, I'd like to see us _only_ have code that runs in production under lib/keysigner and move everything used in testing out to the unit tests.  makes security review easier and more focused.
4. `require("jwcrypto/lib/algs/rs");` to add supported algorithms is not intuitive to this reader, and occurs in our code in at least three places.  what happens if I forget to do that?

hope this is useful review!  holler where I can help implement to get this merged.

lib/keysigner/ca.js
benadida, line 17: I agree that this should be fixed, I don't have an immediately simple way to do this that doesn't cause browserify to bloat the bid-bundle, and the goal was to keep that minimal. Are you okay with my delaying a fix for this until the next rev?

lib/keysigner/ca.js
benadida: I tried initially to make those not map to the spec, but it turned out to be harder to code against, in my opinion. If you feel very strongly, I can try again.

lib/keysigner/ca.js
benadida, line 82: as it turns out, <tt>verifyChain</tt> was already there and not used, I noticed the same thing. I chose not to take it out yet. Agree that cleaning up dead code is something we should do. Do we have an easy way to detect dead code?

lib/static_resources.js
benadida: agreed, I'll fix this.

resources/static/shared/network.js
benadida, line 51: https://github.com/mozilla/browserid/issues/1538

resources/static/shared/user.js
benadida: that's not currently the case. Sounds like you consider this a blocker. Will work on this.

resources/static/shared/user.js
benadida: just leftover from when calls were synchronous.

resources/static/shared/user.js
callahad, line 36: +1 to @lloyd's comment. `.expires` or `.expiry` would be more immediately comprehensible. Though I missed the fact that the abbreviation maps to the spec at https://developers.google.com/payment-express/api-reference, which is a mitigating factor...

benadida:
comments 1, 2, 3, and 5 should be addressed. 4 I will address once jwcrypto PR is merged.
Also addressed non-critical issue 2.

Now working on non-critical issues 1 and 3. I don't think 4 will be addressed in this PR, but I'll try.

benadida:
ok, choosing to NOT address non-critical issue 3. because that issue was _not_ introduced by this PR. We should fix it, so here's an issue to track it:

https://github.com/mozilla/browserid/issues/1541

but no need to bundle it with this PR.

benadida:
non-critical issue 1 is now taken care of. I believe this is now ready for review. @lloyd have at it!

lib/keysigner/ca.js
lloyd, line 17: Yeah, this needent block merge

lib/keysigner/ca.js
lloyd, line 82: I know of no easy/automated way.

lloyd:
two remaining problems:
1. pull request won't merge cleanly, we need to update it.
2. tests still don't pass, neither locally nor on travis: http://travis-ci.org/#!/mozilla/browserid/jobs/1246959

If you haven't set up mysql locally, you wouldn't have seen this.

lloyd:
ok, if these tests pass: http://travis-ci.org/#!/mozilla/browserid/builds/1248922

and if @benadida reviews f45689f and approves, then I say r+, let's merge.

ozten:
Thanks for merging, delegation stuff looks like a good merge.

benadida:
ok, merging!

PR #1534, opened 2012-05-03
implementation of interaction data collection

lloyd:
This pull request is not ready to be merged yet.  At least the following features remain outstanding:
1. (should not block merge) population of `user_agent` with real values parsed out of the UA string.
2. (should not block merge) renaming events published via mediator to have a clear naming scheme that will be stable over time
3. (should not block merge) updating the data blob with missing fields
4. support "re-attaching" a current session after return from primary.


resources/static/shared/interaction_data.js
```diff
@@ -0,0 +1,132 @@
...
+    if (!sample) {
+      currentData = undefined;
+      return;
+    }
+
+    // set current time
+    currentData.timestamp = result.server_time;
+
+    // language
+    currentData.lang = $('html').attr('lang');
```
ozten: Hawt.

resources/static/shared/network.js
```diff
@@ -353,6 +353,28 @@ BrowserID.Network = (function() {
...
+     * @param {function} [onComplete] - Callback to call when complete.
+     * @param {function} [onFailure] - Called on XHR failure.
+     */
+    sendInteractionData: function(data, onComplete, onFailure) {
+      post({
+        url: "/wsapi/interaction_data",
+        data: {
+          // reminder, CSRF token will be inserted here by xhr.js, that's
+          // why this *must* be an object
+          data: data
```
ozten, line 376: Yo dawg, I herd u like data...

Sorry I named this data.

Maybe iadata? or blobs? or ???

resources/static/shared/interaction_data.js
lloyd: you know, we could do this one on the backend too.

resources/static/shared/interaction_data.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  // on all events, update event_stream
+  mediator.subscribeAll(function(msg, data) {
+    if (sample === false) return;
+
+    if (currentData) {
+      currentData.event_stream.push([ msg, new Date() - startTime ]);
+    } else {
+      var d = storage.current();
```
shane-tomlinson: @lloyd, when does this get invoked?  When can sampling be enabled but there not be currentData?  It looks like currentData is created as soon as this module is run, and cleared only when session context comes in and the user has not been selected for sampling.

resources/static/lib/hub.js
```diff
@@ -23,14 +24,32 @@ Hub = (function() {
...
+    });
+
+    return currID++;
+  }
+
   function fire(message) {
     var messageListeners = listeners[message];
 
     if(messageListeners) {
+      // XXX: deviation from upstream!  upstream code doesn't pass
```
shane-tomlinson, line 40: @lloyd - excellent note, I should have put that in there myself.  Perhaps instead of having a discrepancy with the upstream, I should just change upstream to pass the name of the message.

resources/static/shared/interaction_data.js
```diff
@@ -0,0 +1,132 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
shane-tomlinson: A note in general about this module.  I am converting it to be a PageModule - primarily for unit testing.  Having items as a page module means we get to start them several times in sequence, each with different options, as well as eventually allowing us to simulate how redirection to/from an IdP would work.

resources/static/shared/storage.js
```diff
@@ -379,6 +379,58 @@ BrowserID.Storage = (function() {
...
+    }
+    storage.interactionData = JSON.stringify(id);
+  }
+
+  function getAllInteractionData() {
+    try {
+      var id = JSON.parse(storage.interactionData);
+      return id.slice(1);
+    } catch(e) {
+      alert(e);
```
shane-tomlinson: Are these intended to be there only for development?

resources/static/shared/interaction_data.js
```diff
@@ -0,0 +1,132 @@
...
+      storage.clear();
+    }
+  }
+
+  // on all events, update event_stream
+  mediator.subscribeAll(function(msg, data) {
+    if (sample === false) return;
+
+    if (currentData) {
+      currentData.event_stream.push([ msg, new Date() - startTime ]);
```
shane-tomlinson: When is this data saved to localStorage?  More importantly, when is this saved to the backend - only upon re-invocation of the dialog?  If that is the case, we may lose a valuable data point if we have a potential user who tries BrowserID once, decides it is too difficult to use, and never signs in again.

resources/static/shared/network.js
```diff
@@ -353,6 +353,28 @@ BrowserID.Network = (function() {
     },
 
     /**
+     * post interaction data
+     * @method setPassword
+     * @param {string} password - new password.
+     * @param {function} [onComplete] - Callback to call when complete.
+     * @param {function} [onFailure] - Called on XHR failure.
+     */
+    sendInteractionData: function(data, onComplete, onFailure) {
```
shane-tomlinson, line 370: Can we get a happy case test for this?

resources/static/shared/interaction_data.js
ozten: This is accurate to what they were shown (instead of us re-parsing accept-language headers.

lloyd:
we need to, update this pull request and review and test shanes additions.  after that, it seems like we're merge ready

PR #1535, opened 2012-05-03
Reusing checkSupport to implement delegatesAuthority in lib/primary. 

ozten:
We need enhanced support in lib/certassertion.js for supporting domains that delegate their authority to another domain.

This is a blocker for the Vinz Clortho dev deployment.

To test the broken...

1) Go to http://dev.myfavoritebeer.org/
2) enter <LDAP_USERNAME>@id-dev.allizom.org
3) Note: Authentication takes you to intranet-dev.allizom.org
4) In HTTP Basic Auth dialog, enter 

Username: <LDAP_USERNAME>@id-dev.allizom.org
Password: <REAL LDAP PASSWORD>

5) Login appears to work

Expected
Assertion issued

Actual
BrowserID note that email domain doesn't match primary's domain and errors out

Dev manual testing - Use delegator.org as your email domain and setup a /etc/hosts hack to your local instance of Vinz Clortho (https://dev.clortho.mozilla.org/)

Fixes Issue#1527


lloyd:
tests fail, r- - http://travis-ci.org/#!/mozilla/browserid/builds/1233383

ozten:
Weird, passed for me last night, digging.

lib/primary.js
```diff
@@ -226,6 +226,18 @@ exports.getPublicKey = function(domain, cb) {
   });
 };
 
+// Does emailDomain actual delegate to the issuingDomain?
+exports.delegatesAuthority = function (emailDomain, issuingDomain, cb) {
+  exports.checkSupport(emailDomain, function(err, urls, publicKey) {
```
lloyd, line 231: wasn't immediately apparent that checkSupport itself handles delegation chains.  but it does.  great.

benadida:
looks like this commit is going to conflict with upgrade-jwcrypto. @lloyd @ozten if you're okay with this, I would suggest merging upgrade-jwcrypto first and then tweaking this PR so that @ozten gets a chance to use the new jwcrypto API and provide feedback.

ozten:
This PR is blocking testing of Vinz Clortho, so I'd rather land it, then add the new jwcrypto in a new branch. Happy to learn the new stuff and do the second step.

lloyd:
I'm r+ on ozten's portion.  ozten, passing this PR back to you. If you're happy with the forwarding timeouts I've implemented to fix unit tests, merge!

PR #1549, opened 2012-05-07
Issue#1546 Consolidating exception logging, removing delete

ozten:
Consolidating exception logging, removing delete which can cause problems in Chrome. [Partially?] Fixes  - Austin King

New pull request on branch issue1546-kpi-storage
http://git.io/mxFWIg

### DIFF
--- a/resources/static/shared/storage.js
+++ b/resources/static/shared/storage.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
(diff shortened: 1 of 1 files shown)


resources/static/shared/storage.js
```diff
@@ -439,8 +444,7 @@ BrowserID.Storage = (function() {
     try {
       storage.interactionData = JSON.stringify([]);
     } catch(e) {
-      delete storage.interactionData;
```
ozten, line 442: I'm not sure exactly what to do here, feel free to add a commit to this branch or leave a comment with exact syntax of how to properly clean this up. Otherwise, we'll remove it to avoid Chrome issues.

Issue #1552, opened 2012-05-08
IE7/dev: Exceptions thrown on main site


shane-tomlinson:
This is a regression from the current staging with the following properties:

```
1e49b9b update ChangeLog for train-2012.04.27
locale svn r104885
```

Line: 2622
Char: 1
Error: Syntax error
Code: 0
URL: http://192.168.1.88:10002/

Line: 10
Char: 3
Error: Object expected
Code: 0
URL: http://192.168.1.88:10002/

Line: 46
Char: 3
Error: Object doesn't support this property or method
Code: 0
URL: http://192.168.1.88:10002/

Opening the script debugger reveals errors:

provisioning.js

```
var jwcrypto = require("./lib/jwcrypto");
```

start.js:

```
if(path && (checkCookiePaths.indexOf(path) > -1) {
```

and jQuery.

shane-tomlinson:
@benadida - on line 2621 of lib/bidbundle, there is:

```
const jwcrypto = require("./jwcrypto"),
```

which is going to blow up all browsers.

Filing a separate bug about this.

PR #1553, opened 2012-05-08
Partial fix IE6 and IE7 blowing up on main site.  These browsers do not have Array.prototype.indexOf

shane-tomlinson:
issue #1552


shane-tomlinson:
This pull request depends on the error in jwcrypto being fixed - https://github.com/mozilla/jwcrypto/pull/16

lloyd:
r+ - merging

PR #1556, opened 2012-05-08
Fix the IE6 and IE7 black screens on the main site.

shane-tomlinson:
- css/style.css - where relevant and opacity's are used, use filters for IE6 and IE7.
- shared/storage.js - when clearing localStorage if the user is not authenticated, call removeItem for the emails.  The call to JSON.stringify was blowing up.

issue #1529

### DIFF
--- a/resources/static/css/style.css
+++ b/resources/static/css/style.css
@@ -21,6 +21,7 @@ noscript {
   right: 0;
   bottom: 0;
   left: 0;
+  filter: alpha(opacity=0); /* Needed for IE6 and IE7 on the main site */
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
This pull request depends on the error in jwcrypto being fixed - https://github.com/mozilla/jwcrypto/pull/16

lloyd:
as part of closing, I updated to jwcrypto as per 0.2.1, re @shane-tomlinson comment.  IE8 still broken, I'll review and merge issue # and open a 5 star if necc.

Issue #1098, opened 2012-02-10
[API] need logo of RP


skinny97214:
Using the logo of the RP in the popup assures users that they correctly clicked and are not being phished.

ozten:
Existing sources of site logos:
- `<link rel="shortcut icon"` will often have a 48x48 pixel .ico file.
- [Open Web Apps Manifest](https://developer.mozilla.org/en/Apps/Manifest)
- [iOS Webapp Icon Support](https://developer.apple.com/library/ios/#DOCUMENTATION/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html)

Android didn't have much official documentation, but supports iOS's method. 48x48 pixel png seems like the baseline.

It would be good to survey traditional some websites and apps in the wild and see what is available.

Notes: We should check a sites SSL cert. If invalid or the hostname doesn't match, then we **do not** show their logo and instead show a generic image.

skinny97214:
If there's no image, we should just show the RP name as we do currently.

ringe:
Maybe the /.well-known/browserid document could point to a logo as well?

ozten:
`/.well-known/browserid` is only relevant to email providers (aka BrowserID Primaries).

We'll want customizing this logo to be available to websites (RP) as well as Primaries (IdP).

skinny97214:
@lloyd What's the status of getting the logo of sites?

skinny97214:
It could look something like this http://cl.ly/2j0o3i2e1T0p0Q401s1C

jrgm:
verified for stage and awsbox train-2012.06.22 - either siteName or siteLogo or both can be speciifed, etc., and using siteLogo on a non-https RP will produce the "mproper usage of API: only https sites can specify a siteLogo" error in the dialog.

PR #1569, opened 2012-05-10
Put RP name and logo into the dialog.

shane-tomlinson:
@warner, @lloyd, @benadida - Can you guys give this a thorough review?

Important areas to check - 
controllers/dialog.js - I added a check to only allow the characters specified in RFC 3986 in a URL.
controllers/rp_info.js - the logo src and site name are both written using javascript instead of in templates.  If the dialog is in http mode, logos from http and https are allowed, if the dialog is in https mode, only logos from https are allowed.

I am still using logoURL as the field in navigator.id.request - this is so we can expand the use of this field to include fully qualified URLs as well as data URIs once we better understand the security implications of doing so.
# 

commit log
- Add a new controller and template - rp_info
- Adjust the screen size hacks to take into account the dynamic height "favicon"
- Add an action to start the rp_info service.

Secondly, be much more strict about input that comes from external sources.
- Only certain characters are allowed in URLs
- Only absolute paths are allowed for logoURL
- if serving and https dialog, logoURL must be https

issue #1098


lloyd:
@warner, as agreed, lets' double team this

resources/static/test/cases/controllers/rp_info.js
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  test("html stripped from name", function() {
+    createController({
+      name: "<div id='badjiji' onmouseover='alert(\"You have been owned\")')>" + RP_NAME + "</div>"
+    });
+
+    equal($("#rp_name").text(), RP_NAME, "rp_hostname filled in");
+  });
+
```
warner: I'd add one more test here, checking that rp_logo cannot have a `"` in it, I think that's the only way to break out of the `<img src="` context.

resources/static/dialog/controllers/rp_info.js
```diff
@@ -0,0 +1,99 @@
...
+
+      // use stripHTML to make sure that no HTML or Javascript get through to
+      // the user, we don't want to open the user up to being XSSed wby
+      // some goon inserting JS that steals the user's credentials.
+      var data = {
+        hostname: stripHTML(options.hostname) || null,
+        name: stripHTML(options.name) || null,
+        logoURL: options.logoURL ? encodeURI(options.logoURL) : null
+      };
+
```
warner: Since we're not actually using the contents of hostname/name/logoURL in the template (just their presence), maybe `data` should only contain booleans like `hasHostname` and `hasLogo` instead of the actual values? That'd remove the need for stripHTML here.

resources/static/dialog/controllers/rp_info.js
```diff
@@ -0,0 +1,99 @@
...
+      renderer = bid.Renderer,
+      dom = bid.DOM,
+      doc = document,
+      sc;
+
+  function stripHTML(strToClean) {
+    if (!strToClean) return;
+
+    var tmp = document.createElement("div");
+    tmp.innerHTML = strToClean;
```
warner: Is this safe? Won't any embedded `<script>` tags in `strToClean` execute at this point? I'm hoping browsers have some better function for this..

resources/static/dialog/controllers/rp_info.js
```diff
@@ -0,0 +1,99 @@
...
+      };
+
+      renderer.render("#rp_info", "rp_info", data);
+
+      // Because name and logoURL come from user content, they cannot fully be
+      // trusted. To avoid even the remote possibility of an XSS attack, we
+      // write the src and rp_name using Javascript methods so that adding DOM
+      // Event handlers is mitigated.
+      if (data.logoURL) {
+        dom.setAttr("#rp_logo", "src", data.logoURL);
```
warner: This is the most important part. Hm, in fact, the `setAttr` might be completely sufficient, no other sanitization necessary.

resources/static/dialog/controllers/rp_info.js
shane-tomlinson: @warner - For a long time, I was under the impression that scripts are not be run unless they are attached to the DOM using appendChild or insertChild - this is _almost_ correct.  IE9 does not behave this way if the script that is to be filtered out has the "defer=true" attribute set.

example:

```
stripHTML("<script defer="true" type="text/javascript">alert('XSS')</script>")
```

will print XSS in an alert box in IE9 (only)

whereas

```
stripHTML("<script type="text/javascript">alert('XSS')</script>")
```

will not.

Because of this, I am instead replacing the list of sensitive chars found in https://www.owasp.org/index.php/XSS_Prevention_Cheat_Sheet#RULE_.231_-_HTML_Escape_Before_Inserting_Untrusted_Data_into_HTML_Element_Content to their HTML encoded equivalents.

resources/static/dialog/controllers/rp_info.js
shane-tomlinson: This is a good suggestion to help keep the code maintainable.  Making this change.

Issue #1567, opened 2012-05-09
dialog 'POST /wsapi/interaction_data' gets response 'HTTP/1.1 413 Request Entity Too Large'


jrgm:
I was working with @jbonacci and @skinny using FF15 against dev.myfavoritebeer.org. While it worked for me ok, I noticed in network traces that I was getting a 413 response on /wsapi/interaction_data. The POST was trying to send >40KB of data (and we cap at 10K of course). I don't know how I got in this state though. I'll note that FF15/osx is my main browser and this process has been up since midnight Sunday. The local_timestamp's in the json data are for event streams on Monday and Tuesday (sorry, I've been focussed on stage obviously this week and only rarely looked at dev). I can give you the POST data to ponder over. 

Marking 5-star for the train after train-2012-04-27.

```
--- request ---
(Request-Line)     POST /wsapi/interaction_data HTTP/1.1
Host               dev.diresworb.org
User-Agent         Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:15.0) Gecko/15.0 Firefox/15.0a1
Accept             application/json;text/plain
Accept-Language    en-us,hu;q=0.9,fi;q=0.8,zh-cn;q=0.7,el;q=0.6,zh-tw;q=0.4,zh;q=0.3,de;q=0.2,en;q=0.1
Accept-Encoding    gzip, deflate
DNT                1
Connection         keep-alive
X-Requested-With   XMLHttpRequest
Content-Type       application/json; charset=UTF-8
Referer            https://dev.diresworb.org/sign_in
Content-Length     41349
Cookie             browserid_state_634063=bDA2BGPDihHCkgIQPMOQwpfDjMO3w6o.RjzDqsK9w5PChsKYwoTDnMOjGCrDojsDM8O2woXDkW4FHUA5wqvDjMOEw5N0w60fZnNNwrwdwpbCpwbCmMKwwoLDhAIGGAEsAMOAW8OAw71SA8KUGV1VWsO4esK9WsOBw5TCkDTDsnHCllFvAMKHe14owoV9.1336599578593.3600000.GMKHwrRgF3fDrAJjI8OBw75tbcKnc2ZZwrlXIcKfwo9nw5tNwrDCl3rCsBEw
Pragma             no-cache
Cache-Control      no-cache

--- response ---
(Status-Line)      HTTP/1.1 413 Request Entity Too Large
Server             nginx/1.0.14
Date               Wed, 09 May 2012 22:18:15 GMT
[...]

PR #1587, opened 2012-05-11
when the server returns a 413 error for posting of interaction data, cle...

lloyd:
...ar the data from local storage - issue #1567

### DIFF
--- a/resources/static/shared/modules/interaction_data.js
+++ b/resources/static/shared/modules/interaction_data.js
@@ -96,10 +96,21 @@ BrowserID.Modules.InteractionData = (function() {
   function publishStored(oncomplete) {
     var data = storage.get();
 
+    // XXX: should we even try to post data if it's larger than some reasonable
+    // threshold?
     if (data && data.length !== 0) {
       network.sendInteractionData(data, function() {
         storage.clear();
         complete(oncomplete, true);
+      }, function(status) {
+        // if the server returns a 413 error, (too much data posted), then
+        // let's clear our local storage and move on.  This does mean we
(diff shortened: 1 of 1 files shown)


resources/static/shared/modules/interaction_data.js
```diff
@@ -96,10 +96,21 @@ BrowserID.Modules.InteractionData = (function() {
...
     var data = storage.get();
 
+    // XXX: should we even try to post data if it's larger than some reasonable
+    // threshold?
     if (data && data.length !== 0) {
       network.sendInteractionData(data, function() {
         storage.clear();
         complete(oncomplete, true);
+      }, function(status) {
+        // if the server returns a 413 error, (too much data posted), then
```
ozten, line 106: This looks great.

I didn't add a second (larger) limitation to the backend, as I thought the existing one was fine and we can tune later.

resources/static/shared/modules/interaction_data.js
```diff
@@ -96,10 +96,21 @@ BrowserID.Modules.InteractionData = (function() {
...
+        // let's clear our local storage and move on.  This does mean we
+        // loose some interaction data, but it shouldn't be statistically
+        // significant.
+        if (status && status.network && status.network.status === 413) {
+          storage.clear();
+        }
+        complete(oncomplete, false);
```
ozten, line 113: Capitol!

resources/static/shared/modules/interaction_data.js
lloyd, line 106: I like this approach.  I think we should decrease the noise of the event stream, and fix bugs like this before opening it up

PR #1593, opened 2012-05-14
Adding timestamp server side to get an intial integration going.

ozten:
Per an email, our basic integration isn't putting data into KPIggybank properly.

Do we want to land this for this train, so stage records data?

I'm not sure if timestamp should be added client side, or server side.

This should get us going, but pointing my local dev to 

```
"kpi_backend_db_url": "http://23.20.153.45/wsapi/interaction_data",
```

I still don't see the data if I visit that URL and grep through the returned JSON.

### DIFF
--- a/lib/wsapi/interaction_data.js
+++ b/lib/wsapi/interaction_data.js
@@ -28,6 +28,10 @@ var store = function (kpi_json, cb) {
       kpi_resp = function (res) {
         logger.debug('KPI Backend responded ' + res.statusCode);
       };
+  // TODO - timestamp should be client or server side?
+  if (! kpi_json.timestamp) {
+    kpi_json.timestamp = new Date().getTime();
+  }
   if (!! config.get('kpi_backend_db_url')) {
 
     var post_data = querystring.stringify({


lloyd:
timestamp should be server side.  +1 r+

jedp:
I had just pushed to this branch - json_data is a list of dicts, so the timestamp attr needs to be written on each dict in the list.

lib/wsapi/interaction_data.js
```diff
@@ -28,6 +28,10 @@ var store = function (kpi_json, cb) {
       kpi_resp = function (res) {
         logger.debug('KPI Backend responded ' + res.statusCode);
       };
+  // TODO - timestamp should be client or server side?
+  if (! kpi_json.timestamp) {
+    kpi_json.timestamp = new Date().getTime();
```
jedp, line 33: I think this should be 

```
und.each(kpi_json, function (kpi) {
  if (! kpi.timestamp) {
    kpi.timestamp = new Date().getTime();
  }
});
```

kpi_json is getting sent as a list, no?

ozten:
Ah, I'll grab https://github.com/mozilla/browserid/commit/779da8444edf91a1ca1332448c846820ac2314c2

Issue #1592, opened 2012-05-14
train-2012.05.14 will break outstanding verification emails


lloyd:
This train includes a fix for issue #1000, which will prompt users for their password inside the dialog, as opposed to after they click the verification link from their email.

As it stands right now, at the time we roll the train, all outstanding verification links will not work properly.

We must assess exactly how they will fail, and determine the impact and amount of user confusion that will ensue.  Depending on that assessment we should fix it, or derail.

lloyd:
Dreaming up potential solutions:
1. make it so these users are prompted in the verification page for password.  
2. make it so signin succeeds, but users have to figure out reset password later 
3. invalidate verification links

lloyd:
@shane-tomlinson and I assessed the cost of #1 - 

```
10:38 < lloyd> stomlinson: so me sending down a new bit in email_for_token is trivial
10:38 < lloyd> stomlinson: 10ish lines, max

10:39 < stomlinson> lloyd: the front end is going to require some HTML, some CSS, and some JS, plus a test or two.
10:40 < stomlinson> CSS - ~ one line addition.  JS - ~ 10 lines.  HTML - ~20-30 lines - enough to take care of the validation password and all of its tooltips in add_email_address and verify_email_address
```

lloyd:
details of #1 captured here: https://id.etherpad.mozilla.org/41

lloyd:
implementation is complete, tests pass, manual testing of the features looks good to me.  Here are the steps required to fully test this fix in the staging environment:
1. deploy `train-2012.04.27`
2. open dialog in myfavoritebeer
3. add an email or create a new user (depending on the case you're testing, described below)
4. PAUSE javascript execution - if the dialog encouters an error while polling it will stop polling
5. deploy `train-2012.05.25` (or whatever train this goes out in)
6. RESUME javascript execution
7. click on the deployment link in your email
8. verify behavior is as expected.  see below.

## test cases

There are several different cases that need to be tested for this one.  they can all be done in parallel

### new acct / same browser

For this case, you try to sign into mfb with a new email address that isn't registered yet.

Desired outcome: You are prompted for password in the verification screen, and can later log in using that password.  Once you sign in, you're authenticated to the browserid service.

### new acct/ different browser

For this case, you try to sign into mfb with a new email address that isn't registered yet, and you perform the verification in a different browser than you start out in.

[...]

PR #1601, opened 2012-05-15
Issue1592

lloyd:
see issue #1592 for full context and suggested testing procedure


lloyd:
http://travis-ci.org/#!/mozilla/browserid/builds/1340399

PR #1603, opened 2012-05-16
Fix loadgen

lloyd:
These commits repair loadgen after the port to jwcrypto and the api changes.  when merged, this should close issue 1596.

I wanted to add tests of loadgen as part of this, but that's an improvement we'll save for another day.

anyone may review and merge this.  @ozten, @shane-tomlinson, @jparsons, @benadida

### DIFF
--- a/bin/load_gen
+++ b/bin/load_gen
@@ -329,44 +329,53 @@ if (args.m) outputActiveUserSummary(args.m);
 
 const userdb = require("../lib/load_gen/user_db.js");
 
-if (args.u) {
-  // parse args.u
-  var start, end;
-  try {
-    var r = args.u.split('/');
-    if (r.length != 2) throw "expected format ##/##";
-    start = parseInt(r[0], 10);
-    end = parseInt(r[1], 10);
-    if (start >= end) throw "first number must be smaller than the second";
-  } catch(e) {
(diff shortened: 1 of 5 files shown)


jedp:
I'd like to take this one - I'm still getting up to speed with jwcrypto and this would help me understand how all the pieces fit together.

lloyd:
take it away, @jedp

jedp:
Coolio.  Thanks

bin/load_gen
```diff
@@ -329,44 +329,53 @@ if (args.m) outputActiveUserSummary(args.m);
...
-    var r = args.u.split('/');
-    if (r.length != 2) throw "expected format ##/##";
-    start = parseInt(r[0], 10);
-    end = parseInt(r[1], 10);
-    if (start >= end) throw "first number must be smaller than the second";
-  } catch(e) {
-    console.log("your -u argument is poorly formated: " + e.toString());
+const lg_crypto = require("../lib/load_gen/crypto.js");
+
+lg_crypto.init(function(err) {
```
jedp, line 334: heh.  this is where the usefulness of github diffs breaks down a little.  Quite an extravagant wash of green and red for one small change!

PR #1610, opened 2012-05-18
Issue 1597 - ensure that kpi data always includes core metadata (lang, window size, etc.)

jedp:
I've explained my thinking in a comment on the commits.  It's not the most beautiful solution, I admit, but it keeps things moving forward.

Passes tests without modification to test suite.  (that was actually rather challenging)

When inspected in kpiggybank, resulting data looks good.

@shane-tomlinson was a huge help earlier today, spending a fair chunk of time so we could work out what was going on together.  Thanks also to @ozten for letting me talk through this with him.


jedp:
The gods toy with me.  They let the tests pass until I do a git push.  Then they make the tests fail.  Sigh.

Closing this until I can figure that out.

PR #1612, opened 2012-05-18
Issue 1597 interaction data

shane-tomlinson:
@jedp - Mind reviewing this one?

I did a serious yak shave, but I feel like the data model that is saved to localStorage is easier to understand.  I moved the localStorage component from storage.js into shared/models/interaction_data.js, hopefully simplifying as I went along.  Tests were beefed up to make sure we weren't clobbering any current data when we publish staged data.

We must wait until session_context to start writing data into localStorage because this is the first point at which we know whether the current session is allowed to publish collected data.  Because we only know whether we are allowed to publish a session's data once session_context completes, I kept the initial set of the timestamp, screen size, etc where it was.

If a session's session_context never returns, data for the session will not be saved - this is an acceptable data loss for user data privacy.

### DIFF
--- a/resources/static/shared/models/interaction_data.js
+++ b/resources/static/shared/models/interaction_data.js
@@ -25,7 +25,6 @@ BrowserID.Models.InteractionData = (function() {
     try {
       storage.interaction_data = JSON.stringify(data);
     } catch(e) {
-      console.log(e);
       storage.removeItem("interaction_data");
     }
   }
--- a/resources/static/shared/models/interaction_data.js
+++ b/resources/static/shared/models/interaction_data.js
@@ -1,0 +1,164 @@
+/*globals BrowserID: true */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+BrowserID.Models.InteractionData = (function() {
+  "use strict";
+
+  var bid = BrowserID,
+      storage = bid.getStorage(),
+      network = bid.Network,
+      complete = bid.Helpers.complete;
+
+  function getInteractionData() {
+    var interactionData;
+    try {
+      interactionData = JSON.parse(storage.interaction_data);
+    } catch(e) {
+    }
+
+    return interactionData || {};
+  }
+
+  function setInteractionData(data) {
+    try {
+      storage.interaction_data = JSON.stringify(data);
+    } catch(e) {
+      console.log(e);
+      storage.removeItem("interaction_data");
+    }
+  }
+
+  function push(newData) {
+    stageCurrent();
+
+    var interactionData = getInteractionData();
+    interactionData.current = newData;
+
+    setInteractionData(interactionData);
+  }
+
+  function getCurrent() {
+    var interactionData = getInteractionData();
+
+    return interactionData.current;
+  }
+
+  function setCurrent(data) {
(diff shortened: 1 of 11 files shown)


resources/static/test/cases/shared/modules/interaction_data.js
```diff
@@ -179,8 +166,35 @@
         });
       });
     });
-
   });
 
 
+  asyncTest("simulate failed starts - data not sent until second successful session_context", function() {
+    // simulate 3 dialog openings without session completing for any of them.
```
shane-tomlinson: This comment is wrong, there is only 1 failed dialog opening.

resources/static/shared/models/interaction_data.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    return interactionData || {};
+  }
+
+  function setInteractionData(data) {
+    try {
+      storage.interaction_data = JSON.stringify(data);
+    } catch(e) {
+      console.log(e);
```
jedp: remove console.log?

resources/static/shared/models/interaction_data.js
```diff
@@ -0,0 +1,164 @@
...
+    return interactionData.current;
+  }
+
+  function setCurrent(data) {
+    var interactionData = getInteractionData();
+    interactionData.current = data;
+    setInteractionData(interactionData);
+  }
+
+  function stageCurrent() {
```
jedp, line 53: i like this

resources/static/shared/modules/interaction_data.js
```diff
@@ -41,7 +41,9 @@ BrowserID.Modules.InteractionData = (function() {
     // session data must be published independently of whether the current
     // dialog session is allowed to sample data. This is because the original
     // dialog session has already decided whether to collect data.
-    publishStored();
+
+    model.stageCurrent();
+    publishStored.call(self);
```
jedp, line 46: elegantly done

resources/static/shared/models/interaction_data.js
```diff
@@ -0,0 +1,164 @@
...
+  // care if the post is a success or failure as this data is not
+  // critical to the functioning of the system (and some failure scenarios
+  // simply won't resolve with retries - like corrupt data, or too much
+  // data)
+  function publishStaged(oncomplete) {
+     var data = getStaged();
+
+    // XXX: should we even try to post data if it's larger than some reasonable
+    // threshold?
+    if (data && data.length !== 0) {
```
jedp, line 91: I like that this logic has been moved to a model

resources/static/test/cases/shared/modules/interaction_data.js
```diff
@@ -179,8 +166,35 @@
         });
       });
     });
-
   });
 
 
+  asyncTest("simulate failed starts - data not sent until second successful session_context", function() {
```
jedp, line 172: awesome.

resources/static/shared/models/interaction_data.js
jedp: my mistake - i see that this is redefined in static/shared/javascript-extensions.js

jedp:
This yak is very stylishly shaved!

I just want to double-check with you that the console.log in resources/static/shared/models/interaction_data.js is the way you want it.

shane-tomlinson:
@jedp - thanks for the review, updated to remove the console.log message as well as clarify the erroneous comment.  Mind merging this one in?

jedp:
with pleasure!

Issue #1557, opened 2012-05-08
If user types wrong password in verify email screen, error dialog is displayed


shane-tomlinson:
If the user types the wrong password in the verify_email_address screen, instead of getting a second chance to type their password, the user is shown the error dialog.  The user has to re-load the page to try again.

STR requires two browsers

1) In browser 1, open dev.myfavoritebeer.org.  Sign in to BrowserID with a new secondary account that must be verified.
2) In browser 2, open verification email.
3) Type incorrect password
4) Instead of a tooltip that displays "incorrect password", the user is shown an error screen.

jrgm:
verified for stage train-2012.05.25. A tooltip is show to the user in browser 2 when they enter a password that does not match the password that was used in staging the user. 

Also closing this bug, since the code is there and works correctly.

PR #1616, opened 2012-05-18
show a tooltip when the user types a bad password in the add_email_address and verify_email_address screens.

shane-tomlinson:
Previously an error message was displayed.

issue #1557


lloyd:
r+

Issue #1613, opened 2012-05-18
whitelist which fields are sent to backend in KPI data.


shane-tomlinson:
Right now we are sending things like local_timestamp which are used internally but should not be sent to the backend as they could be used to fingerprint users.

lloyd:
+1 - "id" concerns me too.  and generally we should make every addition to this blob require explicit action.  whitelist, not blacklist.  @stomlinson++

jrgm:
verified for train-2012.05.25 running on awsbox (kpi sampling not enabled in stage). The only keys that are sent by the client are  'event_stream', 'lang', 'screen_size', 'sample_rate' (also checked fully on IE8).

PR #1625, opened 2012-05-21
KPI data privacy: whitelist filter data; blur timestamp

jedp:
Server-side filtering of KPI data according to comments on the dev-identity list
- remove local_timestamp from KPI data
- blur timestamp, rounding down to nearest 10-minute interval

@ozten, do you mind taking a look?

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -130,7 +130,7 @@ var conf = module.exports = convict({
     format: 'number = 0.0',
     env: 'KPI_BACKEND_SAMPLE_RATE'
   },
-  kpi_backend_db_url: 'string = "http://localhost/wsapi/interaction_data"',
+  kpi_backend_db_url: 'string = "http://localhost:3000/wsapi/interaction_data"',
   bcrypt_work_factor: {
     doc: "How expensive should we make password checks (to mitigate brute force attacks) ?  Each increment is 2x the cost.",
     format: 'integer{6,20} = 12',
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -130,7 +130,7 @@ var conf = module.exports = convict({
     format: 'number = 0.0',
     env: 'KPI_BACKEND_SAMPLE_RATE'
   },
-  kpi_backend_db_url: 'string = "http://localhost:3000/wsapi/interaction_data"',
+  kpi_backend_db_url: 'string = "http://localhost/wsapi/interaction_data"',
   bcrypt_work_factor: {
     doc: "How expensive should we make password checks (to mitigate brute force attacks) ?  Each increment is 2x the cost.",
     format: 'integer{6,20} = 12',
--- a/lib/wsapi/interaction_data.js
+++ b/lib/wsapi/interaction_data.js
@@ -9,7 +9,8 @@ const coarse = require('../coarse_user_agent_parser'),
       querystring = require('querystring'),
       und = require('underscore'),
       urlparse = require('urlparse'),
-      wsapi = require('../wsapi.js');
+      wsapi = require('../wsapi.js'),
+      TEN_MIN_IN_MS = 10 * 60 * 1000;
 
 // Accept JSON formatted interaction data and send it to the KPI Backend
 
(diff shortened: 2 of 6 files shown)


lib/wsapi/interaction_data.js
```diff
@@ -28,11 +29,22 @@ var store = function (kpi_json, cb) {
...
+
+  // Out of concern for the user's privacy, scrub the data of identifying
+  // details before storing it.
+  //
+  // - Remove the local_timestamp: This was necessary for the client to keep
+  //   track of the start time for the event stream, but it should not be part
+  //   of the final record.
+  //
+  // - Round the server timestamp off to the nearest 10-minute mark.
+  und.each(kpi_json, function (kpi) { delete kpi.local_timestamp;
```
shane-tomlinson, line 35: @jedp - instead of doing this here, we can whitelist fields on the front end so that the data is never sent in the first place.  It should make for a smaller sent data set as well as less work to worry about on the backend.

Please see issue #1613

jedp, line 35: Yes, I saw that issue yesterday.  Thanks, I will update this branch and put the whitelist on the client side.

jedp:
Updated to address issue #1613

Publish only whitelisted interaction_data values

@shane-tomlinson if you have a moment, would you take a look?

resources/static/test/cases/shared/models/interaction_data.js
```diff
@@ -23,36 +23,36 @@
   });
 
   test("after push, most recently pushed data available through getCurrent, getStaged gets previous data sets", function() {
-    model.push({ foo: "bar" });
-    equal(model.getCurrent().foo, "bar",
+    model.push({ lang: "foo" });
```
jedp, line 26: I had to change all your 'foo' to 'lang', since 'foo' is not a whitelisted key :)

Out of respect for the canonical foo - bar - baz sequence, I have additionally changed all bar to foo, and baz to bar.  Because I care.

lib/configuration.js
```diff
@@ -130,7 +130,7 @@ var conf = module.exports = convict({
     format: 'number = 0.0',
     env: 'KPI_BACKEND_SAMPLE_RATE'
   },
-  kpi_backend_db_url: 'string = "http://localhost/wsapi/interaction_data"',
+  kpi_backend_db_url: 'string = "http://localhost:3000/wsapi/interaction_data"',
```
shane-tomlinson: @jedp - was the port meant to be checked in?

resources/static/shared/models/interaction_data.js
```diff
@@ -89,7 +90,22 @@ BrowserID.Models.InteractionData = (function() {
...
     // threshold?
     if (data && data.length !== 0) {
-      network.sendInteractionData(data, function() {
+
+      // Scrub the data we are going to send and let only a set of whitelisted
+      // keys through.  This will remove such values as local_timestamp, which
+      // we needed to calculate time offsets in our event stream, but which
+      // could be used to fingerprint users.
+      var filtered = [];
+      data.forEach(function(obj) {
```
shane-tomlinson: Unfortunately, [].forEach cannot be used right now - IE8 does not support this and we have not yet added it to javascript-extensions.js.

You can either add Array.prototype.forEach from MDN (https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Array/forEach) to javascript-extensions, use _.each, or use a plain old for loop.

resources/static/shared/models/interaction_data.js
```diff
@@ -89,7 +90,22 @@ BrowserID.Models.InteractionData = (function() {
     // XXX: should we even try to post data if it's larger than some reasonable
     // threshold?
     if (data && data.length !== 0) {
-      network.sendInteractionData(data, function() {
+
+      // Scrub the data we are going to send and let only a set of whitelisted
+      // keys through.  This will remove such values as local_timestamp, which
+      // we needed to calculate time offsets in our event stream, but which
+      // could be used to fingerprint users.
+      var filtered = [];
```
shane-tomlinson, line 98: Doing this here is exactly the suggestion I was going to make.  Nice.

resources/static/shared/models/interaction_data.js
jedp: Ah! ok.

lib/configuration.js
jedp: It was a deliberate action executed out of total ignorance.  Removing it.

resources/static/test/cases/shared/models/interaction_data.js
shane-tomlinson, line 26: @jedp - wahahaha!  I appreciate the thought here.

resources/static/shared/helpers.js
```diff
@@ -53,6 +53,16 @@
     return url;
   }
 
+  function whitelistFilter(obj, validKeys) {
+    var filtered = {};
+    for (var key in obj) {
+      if (validKeys.indexOf(key) !== -1) {
```
shane-tomlinson: @jedp - ahHA!  Here is the last offender, [].prototype.indexOf does not exist in IE8.  _.indexOf would work here.

jedp: och. seriously?  I glanced at that function last night thinking the enumeration was going to bum IE out, but the MS docs said it was ok.  Didn't even think indexOf would be a problem.  Thanks for finding that!

PR #1627, opened 2012-05-22
improve parameter-escaping

warner:
@lloyd: this does the trunk-side cleanup we discussed today, and ought to resolve friday's issue too (but needs specific testing for that). Once this lands we can modify #1569 to use the same style.


resources/static/dialog/controllers/dialog.js
```diff
@@ -86,6 +86,7 @@ BrowserID.Modules.Dialog = (function() {
     if (/^http/.test(url)) u = URLParse(url);
     else if (/^\//.test(url)) u = URLParse(origin + url);
     else throw "relative urls not allowed: (" + url + ")";
+    // encodeURI limits our return value to [a-z0-9:/?%], excluding <script>
```
lloyd, line 92: nice explicit comment re embedded js.

resources/static/dialog/controllers/dialog.js
```diff
@@ -111,7 +112,7 @@ BrowserID.Modules.Dialog = (function() {
       return this.get(origin_url, {}, success, error);
     },
 
-    get: function(origin_url, params, success, error) {
+    get: function(origin_url, paramsFromRP, success, error) {
```
lloyd, line 133: great to make it clear that the second argument can contain anything.  **evil** was more dramatic, your word choice here is more clear.

resources/static/dialog/controllers/dialog.js
```diff
@@ -120,23 +121,32 @@ BrowserID.Modules.Dialog = (function() {
       var actions = startActions.call(self, success, error);
       startStateMachine.call(self, actions);
 
-      params = params || {};
+      // Security Note: paramsFromRP is the output of a JSON.parse on an
+      // RP-controlled string. Most of these fields are expected to be simple
+      // printable strings (hostnames, usernames, and URLs), but we cannot
+      // rely upon the RP to do that. In particular we must guard against
+      // these strings containing <script> tags. We will populate a new
+      // object ("params") with suitably type-checked properties.
```
lloyd, line 150: fantastic prose.

resources/static/dialog/controllers/dialog.js
```diff
@@ -120,23 +121,32 @@ BrowserID.Modules.Dialog = (function() {
...
+      params = {};
```
lloyd: uh oh, houston.  is that the introduction of a global variable?  we need a `var` here.

lloyd:
r- due to two small but important issues:
1. a global variable in introduced.  I think we'll need four more chars here.
2. no tests from travis-ci run - there are two ways to run unit tests - #1 create a pull request from a branch in the mozilla repo, or #2 enable travis in your repo - http://about.travis-ci.org/docs/user/getting-started/

once these two things are done, let's get this merged into dev and other branches as well.

warner:
good eye! Fixed the global, also pushed this branch to mozilla/browserid in the "issue-1627" branch, I think that'll trigger a travis-ci build.

resources/static/dialog/controllers/dialog.js
shane-tomlinson, line 150: This newly created whitelist of fields does not include requiredEmail - this must be added to the whitelist or else we will regress in requiredEmail support.

Input checking was initially done in dialog/resources/state.js->"start" handler, the requiredEmail field was the first input we added.  When the code was added for TOS/PP, the implementer put the TOS/PP checks here but the original requiredEmail check was never moved.  To keep all these checks in one place, the requiredEmail check should be moved here as well.

shane-tomlinson:
@warner, @lloyd - I like the approach to this - I did not know underscore had the escape function.

Please see comment https://github.com/mozilla/browserid/pull/1627/files#r858967.  requiredEmail support needs to be added back in before this is merged.

lloyd:
nice catch @shane-tomlinson!  @bwarner - looks like the final addition will be accepting requiredEmail if it looks like a valid email address.  @shane-tomlinson - where's a good email validation routine?

lloyd

resources/static/dialog/controllers/dialog.js
warner, line 150: Whoops! I just pushed another patch that passes requiredEmail through (unchecked), and shuffles the try/catch section a little bit to anticipate some sort of validator being added to the requiredEmail assignment.

One question though: why didn't my tests fail when I regressed requiredEmail? Does that imply that we're missing test coverage for that feature? (I have phantomjs installed, but not mysql, so it's possible that requiredEmail is covered by a test that I didn't run)

In general, the "copy only what you expect to use" pattern might protect us from other surprises, especially since it looks like internal code adds properties to 'params' that _didn't_ come from the RP, in which case the RP might sneakily supply them itself.

warner:
That patch moves the requiredEmail handling.. @shane-tomlinson, is that what you were thinking of? I don't know how to get the tests to start passing again, but maybe you've got a quick fix for it. Or we could leave requiredEmail where it was, I'm cool with that too.

resources/static/dialog/controllers/dialog.js
shane-tomlinson, line 150: @warner - Test coverage is far from complete in several areas, it looks like dialog.js is one of those areas where tests are lacking.

warner:
@lloyd : how's this look now? (BTW I just pushed a copy to the 'issue-1627' branch, so we should get travis-ci coverage in a few minutes). I verified that the latest added test correctly fails when I comment out the `.requiredEmail=` set, so we've got coverage there now.

lloyd:
r+

Issue #1582, opened 2012-05-11
"Is this your computer" screen is not keyboard navigable.


shane-tomlinson:

fmarier:
In my desktop Firefox (12), it is kinda navigable but in a fairly painful way:
- I need to tab until the end of the page (then it wraps around)
- Eventually the focus ends up on the Yes button
- I can press Enter there and it works

shane-tomlinson:
@jbonacci - This was only merged into dev last week, so this has not made staging yet.  In dev, I'm not seeing the behavior with Safari that you.  I go from the "Yes" button to the "No" button.  If I tab out from there, I have to hit tab once to get out of the URL bar, then once through each of the bookmarks, but I get back in.

shane-tomlinson:
@jbonacci - I'm not sure where I was looking for the dates, this _is_ in stage.  I am still not seeing the Safari issues that you are writing about.

shane-tomlinson:
@jbonacci - is this still a problem?  Safari for OSX is showing no signs of a problem for me.

Also, read http://stackoverflow.com/questions/2292016/is-there-a-workaround-to-safaris-operas-bug-that-you-cant-tab-through-hyperli?answertab=votes#tab-top

shane-tomlinson:
Closing in favor of #869, which as @jbonacci points out, is more general than this.

PR #1630, opened 2012-05-22
Make the "is this your computer" screen keyboard navigable.

shane-tomlinson:
- For the buttons, use id's instead of classes.
- Focus the "yes" button when the screen is shown.

issue #1582


jedp:
Works beautifully

Issue #1608, opened 2012-05-17
dbwriter did not auto-reconnect to mysql master


fetep:
We had a networking incident that impacted scl2<->phx1 traffic, and the dbwriters in phx1 lost their connection to the mysql master (in scl2) and did not reconnect on their own.

lsof shows no :3306-related open FDs.  dbwriter.log doesn't have any clues.

lloyd:
do we have actual user visible impact associated with this?  the way the code is constructed, connections are lazy based on requests that require database writes.  It appears based on other writeups of the issue that we did in fact have attempted requests that failed.

it would be helpful to see the relevant subset of the dbwriter logs durring the period immediately preceding the outage to slightly after its resolution.  at least I'll be able to try to piece together what _didn't_ happen that should have.

A skim of the code in place (`lib/db/mysql_wrapper.js`) doesn't reveal any obvious regression.  it should handle this type of outage, and there should be information in logs.

fetep:
It wasn't re-connecting, and our nagios probes that theoretically should be causing a database connection were also failing (they hit /wsapi/ping)

ozten:
Based on @fetep's thorough writeup (via email) I think these are the STR

1) Startup services backed by MYSQL
2) Stop mysqld
3) Go through login flow manually (or potentially many times via load_gen or whatever)
4) Start mysqld
5) Go through login flow

To fix this, I'll need a way to reproduce our code failing to reconnect consistently.

ozten:
Manual testing stopping and starting mysql during various steps doesn't repro.

I see expected behavior: 
Error screen
browserid (25592): warn: database is down, cannot process request: Error: connect ECONNREFUSED
dbwriter (25581): warn: database connection down: Error: connect ECONNREFUSED
Recover once msyql is started.

Will read code more closely as well as look at using load_gen to repro.

fetep:
The actual scenario is losing our TCP link.  So if you can run your
mysqld on another machine, and add an iptables block that will force
existing and new connections to timeout, that would be best.

ozten:
I guess a custom node proxy be a simple route, which points at the real MySQL port sometimes and a non-MySQL port other times.

If I can't get that working, I'll try iptables or a second VM. Thanks Pete!

ozten:
Running this simple single-connection oriented proxy and things work as expected, after 12 operations we emit a timeout...

https://gist.github.com/2770854

ozten:
[...]

PR #1633, opened 2012-05-22
db.ping() must use the same application level query timing and reconnect...

lloyd:
...ion logic as other query types - issue #1608

### DIFF
--- a/lib/db/mysql_wrapper.js
+++ b/lib/db/mysql_wrapper.js
@@ -39,14 +39,15 @@ exports.createClient = function(options) {
         logger.warn("database connection down: " + e.toString());
       });
     },
-    ping: function(cb) {
-      if (stalled) {
-        process.nextTick(function() {
-          cb("database is intentionally stalled");
-        });
-      } else {
-        this.realClient.ping(cb);
-      }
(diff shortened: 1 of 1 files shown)


lloyd:
NOTE: if you were to diff -w this, it would not look nearly as scary.  I suggest this one for immediate merge into dev because it's a 5 star that exacerbated a production outage, and is low riskish.

lib/db/mysql_wrapper.js
```diff
@@ -39,14 +39,15 @@ exports.createClient = function(options) {
...
-          cb("database is intentionally stalled");
-        });
-      } else {
-        this.realClient.ping(cb);
-      }
+    ping: function(client_cb) {
+      // ping queries are added to the front of the pending work queue.  they are
+      // a priority, as they are used by load balancers that want to know the health
+      // of the system.
+      queryQueue.unshift({
```
ozten, line 46: Very cool. This makes sense.

jedp:
regarding diff -w, i just learned that you can add ?w=1 to the diff url and it ignores whitespace

ozten:
Great work getting actual STR by interviewing pete.
r+

Issue #1555, opened 2012-05-08
User has to type password 4 times when confirming on browser other than initiator


shane-tomlinson:
A new user has to type their password 4 times when confirming their email on a browser other than the initiating browser.  This could happen if a user signs up for BrowserID on a mobile device and confirms on a desktop device.

Passwords 1 and 2 are needed in the dialog before the verification email is sent.
Password 3 is required in browser 2 when verifying email
Password 4 is required in the dialog after verifying email but before assertion is generated.

@lloyd - you believed you had a backend fix for this?

shane-tomlinson:
@lloyd - checkin e54da630b4b5b8164137471ad5895011300e41a1 is the initial checking for this.  

This checkin does not handle the case where an account goes from being an account with only primary addresses to an account with its first secondary email and the email was validated in another browser.

STR:

1) Open 2 browsers.
2) In Browser 1, create an account with a primary address
3) In Browser 1, add the first secondary address to account, set password, go to "check your email page"
4) In Browser 2, open verification link
5) In Browser 1, user has to type password.

lloyd:
The pull request has been merged.  this issue is fixed.

jrgm:
Bah, sorry for not getting to this earlier in the week. The 2-browser new secondary and add-secondary-to-secondary work (only 3 passwords required, not a 4th in the dialog again). 

However, I still get a 4th challenge in the browser1 when adding a secondary to a primary-only account (per STR in https://github.com/mozilla/browserid/issues/1555#issuecomment-5873009). Reopening.

shane-tomlinson:
Link to #1451

shane-tomlinson:
Closing as dup of #1682.

PR #1636, opened 2012-05-23
Fix the user being asked 4 times for a password when adding a secondary email to a primary account and verifying email ownership in a new browser.

shane-tomlinson:
@lloyd - can you review this one?  

I took the same approach for the "email_staged" message as you did for the "user_staged" message and pass the password.  The password gets handed to the check_registration controller which will try to log the user in if "email_addition_status" returns mustAuth.
- When "email_staged" is triggered, trigger it with the password.
- Rename helpers.addSecondaryEmailWithPassword to helpers.addSecondaryEmail
- Add unit tests for helpers.addSecondaryEmail
- Yak shave the unit tests to clean them up so that they are easier to follow.

issue #1555

### DIFF
--- a/resources/static/dialog/resources/helpers.js
+++ b/resources/static/dialog/resources/helpers.js
@@ -111,12 +111,12 @@
     }
   }
 
-  function addSecondaryEmailWithPassword(email, password, callback) {
+  function addSecondaryEmail(email, password, callback) {
     var self=this;
 
     user.addEmail(email, password, function(added) {
       if (added) {
-        var info = { email: email };
+        var info = { email: email, password: password };
         self.publish("email_staged", info, info );
       }
       else {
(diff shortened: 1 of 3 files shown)


resources/static/dialog/resources/helpers.js
```diff
@@ -111,12 +111,12 @@
...
   }
 
-  function addSecondaryEmailWithPassword(email, password, callback) {
+  function addSecondaryEmail(email, password, callback) {
     var self=this;
 
     user.addEmail(email, password, function(added) {
       if (added) {
-        var info = { email: email };
+        var info = { email: email, password: password };
```
shane-tomlinson, line 119: @lloyd - this is the important part - pass the password along with the email.  The password makes its way to the "check_registration" controller who will use it to log the user in if the status to "email_addition_status" is "mustAuth"

lloyd:
r+

Issue #1580, opened 2012-05-11
(dev/next-train) dialog 'forgot password' screen missing some context about which email


jrgm:
STR:
1. go to http://myfavoritebeer.org/ or http://dev.myfavoritebeer.org/, and click 'Sign in'
2. enter an email secondary address that is previously known to (browserid.org|dev.diresworb.org)
3. when you get prompted for a password, click 'forgot password'
4. Here's the difference. In prod 'reset password' shows the email address that will be reset; in dev, it lacks that context (perhaps this is intended, but it just struck me as curious).

In prod:
![img](https://img.skitch.com/20120511-fax8k3mirckfhm35atytfberne.png)

In dev:
![img](https://img.skitch.com/20120511-x6minenkihwdy96qrxjm34pdkf.png)

skinny97214:
Agreed, we need a short bit of copy that says this is a password reset

jrgm:
Actually, I was thinking that showing the email address (as it was before) would be good (what email address am I resetting).

jrgm:
Actually, sorry, yeah, you're right. Both copy and what email should be there.

billder:
The tricky thing here is overly associating a specific email address with the password. I'd like for people to associate the password with Persona, and not with one specific email address.

shane-tomlinson:
closed by issue #1889.

PR #1640, opened 2012-05-24
Make the in dialog set-password flow work with requiredEmail.

shane-tomlinson:
- In set_password.ejs, only show the "cancel" button if not setting the initial password for a required email.
- Update required_email.js to handle the set-password in dialog flow.
- Clean up required_email.js so that it is easier to follow.
- Rename the "add_email_submit_with_secondary" message to "stage_email" - much clearer.
- Show a message whenever resetting the account password.

issue #1638
issue #1580

### DIFF
--- a/resources/static/dialog/controllers/required_email.js
+++ b/resources/static/dialog/controllers/required_email.js
@@ -6,8 +6,7 @@
 BrowserID.Modules.RequiredEmail = (function() {
   "use strict";
 
-  var ANIMATION_TIME = 250,
-      bid = BrowserID,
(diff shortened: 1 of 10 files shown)


shane-tomlinson:
@lloyd - Can you give this a review?

lloyd:
r+ - most of the risk here is within the required email flow which is already broken.

Tests pass: http://travis-ci.org/#!/mozilla/browserid/builds/1424894

manual testing ensues...

lloyd:
basic manual testing reveals no major regressions.  merging.

PR #1647, opened 2012-05-29
Issue1645 - remove dead code

lloyd:
the code_update handler was a way to shut down a process via localhost HTTP.  This is no longer neccesary, our deployment mechanisms have grown up.  This particular bit of code is not something that's very smart to leave lying around.

### DIFF
--- a/bin/browserid
+++ b/bin/browserid
@@ -157,17 +157,7 @@ wsapi.setup({
 // #9 - handle views for dynamicish content
 views.setup(app);
 
-function doShutdown(readyForShutdownCB) {
-  require('../lib/bcrypt.js').shutdown();
-  db.close(readyForShutdownCB)
-}
-
-// #11 - calls to /code_update from localhost will restart the daemon,
-// this feature is not externally accessible and is only used by
-// the update logic
-shutdown.installUpdateHandler(app, doShutdown);
-
-// #12 if the BROWSERID_FAKE_VERIFICATION env var is defined, we'll include
+// #10 if the BROWSERID_FAKE_VERIFICATION env var is defined, we'll include
 // fake_verification.js.  This is used during testing only and should
 // never be included in a production deployment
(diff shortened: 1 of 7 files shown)


bin/browserid
```diff
@@ -157,17 +157,7 @@ wsapi.setup({
...
-  db.close(readyForShutdownCB)
-}
-
-// #11 - calls to /code_update from localhost will restart the daemon,
-// this feature is not externally accessible and is only used by
-// the update logic
-shutdown.installUpdateHandler(app, doShutdown);
-
-// #12 if the BROWSERID_FAKE_VERIFICATION env var is defined, we'll include
+// #10 if the BROWSERID_FAKE_VERIFICATION env var is defined, we'll include
```
shane-tomlinson, line 160: @lloyd - I like the updating of the comment #'s - there are some others that look like they could use numbers or need re-numbered.  For instance, the numbers jump straight from 10 to 13, and there are some app.use calls in between, and statsd does not have a number.

docs/DEPLOYMENT.md
```diff
@@ -142,9 +146,8 @@ if [ "x$GL_REPO" == 'xbrowserid' ] ; then
     echo "generating production resources"
     cd $NEWCODE/browserid && ./compress.sh && cd -
 
-    # stop the servers
-    curl -o --url http://localhost:62700/code_update > /dev/null 2>&1
-    curl -o --url http://localhost:62800/code_update > /dev/null 2>&1
+    # XXX: stop the servers!  you should deliver SIGINT to each
```
shane-tomlinson, line 149: @lloyd - Should this comment be updated before merge?

shane-tomlinson:
@lloyd - tests pass, this all looks good to me - logically only dead code was removed (as the title says).  My only questions are comment related, otherwise r+.

Issue #1632, opened 2012-05-22
extend the session cookie on browserid.org/persona.org


skinny97214:
Make the session cookie 4 weeks.

(/me hides)

lloyd:
+1  - @benadida ?

benadida:
this is after the user has said this is their computer? If so, sounds good to me.

shane-tomlinson:
Link this to #1572

jrgm:
Verified on stage train-2012.06.08. 
- If I start with a clean profile and bring up the dialog from an RP, before I am authenticated I get a cookie with an embedded duration of 2419200000 (4 weeks) and an expires value of 4 weeks from now. 
- When I signin as a known secondary with a password, the cookie is changed to have embedded duration of 3600000 (1 hour), and an expires value of 4 weeks from now.  
- After waiting >60s and answering Yes to "is this your computer" the call to prolong_session returns a cookie with an embedded duration of 2419200000 (4 weeks) and an expires value of 4 weeks from now. 

However, I'll note the values for the same steps in current production train-2012.05.25 branch:
- before authenticated: embedded duration of 604800000 (1 week) and an expiry value of 2 weeks from now.
- after signin: embedded duration of 3600000 (1 hour) and an expiry value of 2 weeks from now.
- after Yes to 'is this your computer': embedded duration of 1209600000 (2 weeks) and an expiry value of 2 weeks from now.

So, it seems that the duration was only controlling the value for a non-authenticated session.

PR #1648, opened 2012-05-29
increase authenticated session duration to 4 weeks (was effectively 1 we...

lloyd:
...ek): issue #1632

NOTE:  The intention previously was to have duration set to 2 weeks.  Due to a bug in the code, it was effectively 1 week.  This patch fixes that bug too.

### DIFF
--- a/lib/wsapi.js
+++ b/lib/wsapi.js
@@ -145,12 +145,10 @@ exports.setup = function(options, app) {
   var cookieSessionMiddleware = sessions({
     secret: COOKIE_SECRET,
     cookieName: COOKIE_KEY,
-    duration: 7 * 24 * 60 * 60 * 1000, // 1 week
+    duration: config.get('authentication_duration_ms'),
     cookie: {
       path: '/wsapi',
       httpOnly: true,
-      // IMPORTANT: we allow users to go 1 weeks on the same device
-      // without entering their password again
       maxAge: config.get('authentication_duration_ms'),
       secure: overSSL
     }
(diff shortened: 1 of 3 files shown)


lib/wsapi.js
```diff
@@ -145,12 +145,10 @@ exports.setup = function(options, app) {
...
     secret: COOKIE_SECRET,
     cookieName: COOKIE_KEY,
-    duration: 7 * 24 * 60 * 60 * 1000, // 1 week
+    duration: config.get('authentication_duration_ms'),
     cookie: {
       path: '/wsapi',
       httpOnly: true,
-      // IMPORTANT: we allow users to go 1 weeks on the same device
-      // without entering their password again
       maxAge: config.get('authentication_duration_ms'),
```
shane-tomlinson, line 152: @lloyd - Looking over the client-sessions README.md file, I see the following comment about maxAge:

```
 60           // cookie expiration parameters
 61           // this gets updated on every cookie call,
 62           // so it's not appropriate for saying that the session
 63           // expires after 2 weeks, for example, since the cookie
 64           // may get updated regularly and push the time back.
 65           maxAge: 14 * 24 * 60 * 60 * 1000 // in ms
```

I know right now a session and a cookie have the same length, but since the cookie is valid for two weeks from its last write, the cookie is going to expire after the session.  Does this have any negative repercussions?

lloyd, line 152: so maxAge is what appears in cookie's expires headers, and duration is what is actually embedded inside the encrypted cookie. 

_both_ of these put an upper bound on session duration, and either may expire first with no ill effect.  If the maxAge/expires header expires first, then the browser stops sending the cookie, and you're not authenticated.  If the duration expires, then client-sessions sees that, and throws out the cookie, and you're not authenticated.

So I think we're covered here.

ozten, line 152: This makes sense and looks good to me.
r+

jrgm:
verified note in GH-1632

Issue #1631, opened 2012-05-22
Fail gracefully when user attempts to copy+paste their password


ajalgaonkar:
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:15.0) Gecko/15.0 Firefox/15.0a1

Steps to reproduce:
1. Go to https://browserid.org
2. Next try and create a new account or Click forgot password and then try to go through the reset password link.
3. Once in the screen for password confirmation, enter the value in the password field, "select all" and paste in the field.
4. Next paste the same in the confirm password field as well.
5. Try and submit with the current username and password combination.

Expected behavior:

Firefox does not allow characters to be pasted in the password field, hence Browserid should also deal with this gracefully.

Actual behavior:

Instead of Password characters being copied, "**********" get copied or the dots which represent the hidden password. Hence field becomes invalid and cannot accept password entry.
Once the System rejects the Password entry the the following error is seen:

Post:/wsapi/complete_user_creation
Response code - 400
Response Text: Bad Request: CSRF Violation
Error Type: Bad Request: CSRF Violation

Screen Shot:
http://postimage.org/image/heh0mw1e9/ 

Additional Notes:

jrgm:
Thanks for the bug report. This actually points out a very interesting bug.

I think this is actually a backend bug, not something that the frontend should try to deal with. I don't see why we should limit the characters that are entered into the password field (e.g., should a non-english speaker be limited to entering ASCII characters).

I think the real bug is in the forwarding of requests from the webhead to other services. In a local dev browserid, when I submit a password that has non-ascii characters, the webhead receives the request for `/wsapi/stage_user` and forwards that request to the dbwriter. A browser (FF, Chrome, IE9) sends the request with a 'Content-Length' equal to the **octet** size of the JSON body. However, the forwarded request has a 'Content-Length' equal to the **character** count of the JSON body. And so the JSON is truncated, cannot be parsed, and an undefined `req.body`, leading to `lib/wsapi.js` sending back '400 Bad Request: CSRF violation'. 

Nothing special about passwords here; it's about forwarding requests that contain 8-bit in the request body.

(Side note: Chrome and IE9 do not allow anything to copied from a password field (not even the masked value), but FF does. Arguably a bug in Firefox, but not this issue).

lloyd:
@jrgm - thanks for catching and diagnosing this.  You saved me a ton of time and made it easy to reproduce, find, and fix.  patch and pull request en route.

[...]

PR #1649, opened 2012-05-29
Support non-ascii passwords: when forwarding requests, Content-Length sh...

lloyd:
...ould be byte length, not number of unicode chars.  issue #1631

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,5 @@
 train-2012.06.08 (in progress):
-  *
+  * Support non-english passwords: issue #1631
 
 train-2012.05.25:
   * many KPI improvements: #1597, #1613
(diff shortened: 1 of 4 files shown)


jrgm:
I think you should add the same fix here: https://github.com/mozilla/browserid/blob/dev/lib/wsapi_client.js#L129-133
(I was writing a test for this, and I needed that (or alternatively, not setting content-length at all)).

lloyd:
There are two interesting things here.  YES we need tests, and yes that fix needs to be here too.

Are you writing unit tests now?  That's freaking fantastic if so, but I'm also willing to take the responsibility for this one.

Either way, I'll add the fix now.

jrgm:
Started looking more at the unit tests that are there, yes. I have a simple one for this flow (first ascii password, then same with a password in Russian), but if you want to do that test that's fine too.

lloyd:
how about this, check in the simple test you've got in a branch, I'll review and append to this PR?

lloyd:
@shane-tomlinson this thing is ready for review and merge at your leisure.

Issue #1624, opened 2012-05-21
Minified javascript missing license and link to source


fmarier:
https://browserid.org/include.orig.js has the following header:

```
/**
 * Uncompressed source can be found at https://browserid.org/include.orig.js
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```

but that header is missing from https://browserid.org/include.js which I think is the place where it was intended to show up.

lloyd:
we probably broke this when we reworked our compression many moons ago.  we should determine what if anything we want at the header of this file and repair it.  `scripts/compress` is a good place to look for the gunk that does the compressin'.

lloyd:
I dug into this.  When we moved to programmatically interfacing UglifyJS, we lost copyright preservation, as that code is not included in the UglifyJS library, but rather the command line program.  patch coming to add this functionality back to `scripts/compress*`

jrgm:
verified for stage train-2012.06.08 comment is preserved for /include.js. @shane-tomlinson has an open issue GH-1655 for whether license and reference to source code repo is needed in other minified prod. resources. The answer seems to be no, for the most part, with caveats.

PR #1650, opened 2012-05-29
restore include.js license header and links to un-minified source: issue...

lloyd:
... #1624


shane-tomlinson:
@lloyd - Should the license also be included in the other three production js resources - dialog.js, browserid.js and communication_iframe.js as well?  The license is not included in any of these.

lloyd:
@shane-tomlinson I don't know!  I think we should inquire with our legal buddies on whether we really need to ship out license information.  For this patch I just aimed to restore the core functionality in the compression scripts...

Issue #1558, opened 2012-05-08
ensure uncaught exceptions are logged


lloyd:
we should make sure uncaught exceptions are properly logged, reported, and monitored.

ozten:
Lloyd shared this scary blog post http://debuggable.com/posts/node-js-dealing-with-uncaught-exceptions:4c933d54-1428-443c-928d-4e1ecbdd56cb

ozten:
Webdev is standardizing on [Sentry](http://sentry.readthedocs.org/en/latest/developer/client/index.html#a-working-example) for server side crash aggregation and analysis.

lloyd:
we should get petef to set an alarm on .uncaught_exception statsd counters, which are implemented in 3d2ca1f

lloyd:
https://bugzilla.mozilla.org/759251

jrgm:
Confirmed with train-2012.06.08 on an awsbox and installing a sighup handler to generated uncaught exceptions (since I don't have  real way to force an uncaught exception).

PR #1651, opened 2012-05-29
rework uncaught exception handling: log an error, don't bring the proces...

lloyd:
...s down, bump a statsd counter: issue #1558

### DIFF
--- a/lib/statsd.js
+++ b/lib/statsd.js
@@ -29,3 +29,7 @@ if (statsd_config && statsd_config.enabled) {
 
   statsd = new StatsD(options["host"], options["port"]);
 }
+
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@lloyd - After reading the referenced article, it makes me wonder what the proper approach is to uncaught exceptions - is it only to make a log and continue?

lib/statsd.js
```diff
@@ -29,3 +29,7 @@ if (statsd_config && statsd_config.enabled) {
 
   statsd = new StatsD(options["host"], options["port"]);
 }
+
+process.on('uncaughtException', function(err) {
```
ozten, line 33: Nice!

Issue #1652, opened 2012-05-29
compress-worker regexp for embeded URLs in CSS does not handle multiple background-images per declaration


shane-tomlinson:
```
 18 body {
 19   background-color: #6a7b86;
 20   background-image: url("/i/marketplace-header.png"), url("/i/grain.png");      
 21   background-position: center top, center top;
 22   background-repeat: repeat-x, repeat;
 23   color: #fff;
 24 }
 25
```

```
[11:32][browserid(rebrand)] $ ./scripts/compress 
creating templates.js
built /production/include.js in 0.02s (up to date)
Cachify bailing on hash... no such file /i/marketplace-header.png"), url("/i/grain.png
{ [Error: ENOENT, no such file or directory '/Users/stomlinson/development/browserid/resources/static/i/marketplace-header.png"), url("/i/grain.png']
  errno: 34,
  code: 'ENOENT',
  path: '/Users/stomlinson/development/browserid/resources/static/i/marketplace-header.png"), url("/i/grain.png',
  syscall: 'open' }
Cachify bailing on hash... no such file /i/marketplace-header.png"), url("/i/grain.png
{ [Error: ENOENT, no such file or directory '/Users/stomlinson/development/browserid/resources/static/i/marketplace-header.png"), url("/i/grain.png']
  errno: 34,
  code: 'ENOENT',
  path: '/Users/stomlinson/development/browserid/resources/static/i/marketplace-header.png"), url("/i/grain.png',
  syscall: 'open' }
```

jrgm:
verified in dev for train-2012.06.08

PR #1654, opened 2012-05-29
Fix not being able to have multiple background images per CSS declaration.

shane-tomlinson:
@lloyd or @ozten - Could one of you fellows review this?
- Modify the RegExp so that the saved match for a URL cannot include the characters '")

issue #1652

### DIFF
--- a/scripts/compress-worker.js
+++ b/scripts/compress-worker.js
@@ -89,7 +89,9 @@ function compressResource(staticPath, name, files, cb) {
 }
 
 function cachify_embedded (css_src) {
-  return css_src.replace(/url\s*\(['"](.*)\s*['"]\s*\)/g, function (str, url) {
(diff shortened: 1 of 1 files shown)


lloyd:
r+

scripts/compress-worker.js
```diff
@@ -89,7 +89,9 @@ function compressResource(staticPath, name, files, cb) {
 }
 
 function cachify_embedded (css_src) {
-  return css_src.replace(/url\s*\(['"](.*)\s*['"]\s*\)/g, function (str, url) {
+  // RegExp is set up to handle multiple url's per declaration, which is
+  // possible for things like background-images.
+  return css_src.replace(/url\s*\(['"]([^\)'"]+)\s*['"]\s*\)/g, function (str, url) {
```
ozten, line 94: Very cool.

PR #1656, opened 2012-05-29
Rebrand

shane-tomlinson:
Time to start the great rebrand!


resources/static/dialog/css/popup.css
```diff
@@ -404,22 +428,19 @@ a.emphasize:hover {
...
+  margin-left: -100px;
+  min-width: 4em;
 }
 
-#your_computer_content p {
-  padding-bottom: 1em;
-  line-height: 1.3em;
-  margin-top: 2em;
-  margin-bottom: 2em;
-}
```
lloyd, line 425: that, sir.  is one PILE of css changes.  I'm going to review these by testing on different devices to ensure usability first, and beauty second.

resources/static/dialog/resources/helpers.js
```diff
@@ -15,12 +15,19 @@
...
-         body.delay(500).animate({ "opacity" : "0.5"}, 500);
-      });
+      var endWidth = bodyWidth - 10;
+
+      body.addClass("completing");
+      /**
+       * CSS transitions are used to do the slide effect.  jQuery has a bug
+       * where it does not do transitions correctly if the box-sizing is set to
+       * border-box and the element has a padding
+       */
```
lloyd, line 29: how does this work on IE8 & IE9 with CSS transitions?  http://caniuse.com/#feat=css-transitions

resources/views/dialog_layout.ejs
```diff
@@ -13,12 +13,12 @@
...
   <%- cachify_css('/production/dialog.css') %>
-  <link href="https://fonts.googleapis.com/css?family=Droid+Serif:400,400italic,700,700italic" rel="stylesheet" type="text/css">
-  <title><%= gettext('BrowserID') %></title>
+  <link href='https://fonts.googleapis.com/css?family=Open+Sans:300,300italic,400,400italic,700,700italic' rel='stylesheet' type='text/css'>
+  <title><%= gettext('Mozilla Persona') %></title>
 </head>
   <body class="waiting">
       <header id="header">
-         <h1><a class="home" target="_blank" href="/">BrowserID Home</a></h1>
+         <h1><a class="home" target="_blank" href="/">Mozilla Persona Home</a></h1>
```
lloyd: what guidance have we got on 'Mozilla Persona' vs 'Persona'?  Is there a rule on when to use what?

resources/views/privacy.ejs
```diff
@@ -4,29 +4,29 @@
 
 <div id="content" class="display_always">
     <div id="legal">
-      <h2>Privacy & BrowserID</h2>
+      <h1>Privacy & Persona</h1>
```
lloyd, line 7: Do we need review of the privacy policy, given that the text has changed?  This is mostly a replacement of BrowserID with Persona, so it seems low risk...

resources/views/signin.ejs
```diff
@@ -69,6 +71,6 @@
 </div>
 
 <noscript>
-  We're sorry, but to sign in to BrowserID, you must have Javascript enabled.
+  We're sorry, Persona requires Javascript to be enabled.
```
lloyd: Perhaps "Persona requires that JavaScript is enabled".

resources/views/signup.ejs
```diff
@@ -108,7 +106,7 @@
         </form>
 
         <div class="notification" id="congrats">
-            <p class="serif">Thank you for signing up with <strong>BrowserID</strong>. You can now use your <strong>BrowserID</strong> account to <em>Sign In</em> or <em>Sign Up</em> to websites all across the web!
+            <p>Thank you for signing up with <strong>Persona</strong>. You can now use your Persona account to <em>Sign In</em> or <em>Sign Up</em> to websites all across the web!
```
lloyd, line 109: Along with this rebrand, now we're ready to localize the main site, eh?

Shall we open up or revive a bug to go through and wrap strings post-merge?

resources/views/tos.ejs
```diff
@@ -43,7 +43,7 @@
       trade or resell the Services for any purpose, unless you have been specifically permitted to do so in writing by Mozilla, or
     </li>
     <li>
-      access (or attempt to access) the Services by any means other than the BrowserID site itself (at browserid.org), official Mozilla-branded software (Firefox, Firefox for Mobile, and Firefox Home, referred to in this document as “Firefox Clients”), or third party software that utilizes APIs authorized and provided by Mozilla (“Third Party Clients”), unless you have been specifically allowed to do so in writing by Mozilla.
+      access (or attempt to access) the Services by any means other than the Persona site itself (at browserid.org), official Mozilla-branded software (Firefox, Firefox for Mobile, and Firefox Home, referred to in this document as “Firefox Clients”), or third party software that utilizes APIs authorized and provided by Mozilla (“Third Party Clients”), unless you have been specifically allowed to do so in writing by Mozilla.
```
lloyd, line 46: Same comment for ToS as for privacy policy.  Shall we give our legal eagle friends a heads up so they can make sure that whatever external resources point to or re-publish this text also gets updated?

lloyd:
note during email verification there is a flicker of "last step, enter your password".  I have seen this on dev too, and don't think that it is a regression.  I'll file a bug on it after merge.

lloyd:
travis gives this one an r+ - tests pass: http://travis-ci.org/#!/mozilla/browserid/builds/1464365

resources/static/dialog/resources/helpers.js
shane-tomlinson, line 29: @lloyd - the reason for the transition is jQuery has a bug with its animate function if you try to animate an element's width that has it's box sizing set to border-box and a padding.  jQuery does not include the padding in the element's width, so it makes the arrow jump all over the place.  Instead, I went for a simpler solution to let browsers that support CSS transitions do the work themselves, and those that don't just sit there for a second and a half and close.  IE8/IE9 fall into this category.

resources/views/privacy.ejs
shane-tomlinson, line 7: @lloyd - We need to review both the TOS and the privacy policy for the "name/photo" feature already, so we can ask about this.

https://bugzilla.mozilla.org/show_bug.cgi?id=755732

resources/views/signup.ejs
shane-tomlinson, line 109: @lloyd - There are going to be a lot of text changes in the coming weeks with the copy-writer doing copy-write things.  I think we should wait until we get updated text before we inundate our localizers with two large sets of translations.

lloyd:
**UPDATE:**  This issue is not an issue.  IE8/9 doesn't send cookies to hostnames with underscores.  ignore this.

on IE9 and IE8 with factory settings I get cookie warnings:

Bad IE, no cookie

lloyd:
On iOS, on main site, after clicking "sign in", the "how it works", "developers" and "sign-in" links are displayed.

This isn't so confusing on the landing page, but after clicking sign in, these links take up 1/2 of the screen.  Could we hide these links on iOS for the signin/signup pages?

links take up 1/2 the screen on iOS

lloyd:
on iOS, after signing in, the buttons and "new to persona?" banner fill the whole screen with the account manager beneath the "fold".

suggest hiding the buttons here too (maybe _everywhere_ except splash on mobile?)  and less margin/padding on the yellow "first time" banner:

no space, or too much space

lloyd:
buttons should line up with text on mobile manage page:

align me, plz

lloyd:
sign up button overlaps footer on splash page on iOS:

overlap bad

lloyd:
k, super nitpicky, the "is this your computer" screen header on iOS seems a little too crowded.  maybe a bit of space on top?

crowded

lloyd:
signin dialog looks fugly on iOS.  seems like it should fill all horizontal space:

horizontal space

lloyd:
"Check Your Email" header on iOS should be vertically centered in the available space.  Right now it's nestled against the subsequent text

vertically center

resources/views/signup.ejs
lloyd, line 109: that makes sense.  +1 to waiting.

lloyd:
tested on IE9 in IE8 emulation mode.  no specific issues found.  The css transition isn't supported, but the fallback feels fine to me.

resources/static/dialog/resources/helpers.js
lloyd, line 29: confirmed.  no issue with IE for css transitions.

lloyd:
On picker, "this is not me" and "add an email address" buttons stack up rather than being side by side.

How can we fix this for english _and_ german?

buttons too big

lloyd:
I went through ALL of my comments above after the updates by stomlinson.  Here are two remaining paper cuts (mentioned above):
1. on iOS "add another email" and "this is not me" buttons stack on top of each other
2. they layout of "if you don't mind me asking, is this your computer" could use a little space.

other than these two issues, I think this might be ready for merge with one caveat.  We need help from @skinny97214 and @jbonacci or @jrgm reviewing the designs ASAP once landed.  Skinny for overall feedback, and our QA homies for, especially, vetting other, more verbose, languages.

PR #1657, opened 2012-05-29
Router layer

zaach:
The router layer simply forwards all requests at this point. `/verify` requests go to `verifier`, wsapi database writes to `dbwriter`, and everything else to `browserid`.


bin/browserid
```diff
@@ -34,18 +34,6 @@ app = express.createServer();
...
-  logger.error('missing required configuration - url for the keysigner (KEYSIGNER_URL in env)');
-  process.exit(1);
-}
-
-// verify that we have a dbwriter configured
-if (!config.get('dbwriter_url')) {
-  logger.error('missing required configuration - url for the dbwriter (DBWRITER_URL in env)');
-  process.exit(1);
-}
-
```
lloyd, line 48: I _really_ like that we're moving forwarding out of the `browserid` process.  the more we can move out of that process, the more performace wins we'll get out of this.

bin/router
```diff
@@ -0,0 +1,127 @@
...
+  });
+}
+
+// handle /wsapi writes
+wsapi.setup({
+  router_mode: true,
+  forward_writes: urlparse(config.get('dbwriter_url')).validate().normalize().originOnly()
+}, app);
+
+// Forward all leftover requests to browserid
```
lloyd, line 112: I'm a little uncomfortable with the blanket forwarding of all other requests to browserid.  This consumes some resources for all requests, even invalid ones.  I understand why you wrote it this way, and it makes sense - it's hard to enumerate ALL the valid apis.

Maybe something we could do here in the near future is to break all static file serving out into a seperate process, and that process is the catchall.  this moves some small amount of additional work out of browserid, and when we move to a CDN, we can eliminate that process from production.

At this point all valid incoming urls would be enumerable and the router could drop those not targeted at a supported wsapi...  hmmm.

lib/http_forward.js
```diff
@@ -55,6 +55,21 @@ exports.forward = function(dest, req, res, cb) {
...
+    }
+    if (pres.headers.hasOwnProperty('etag')) {
+      res.setHeader('ETag', pres.headers['etag']);
+    }
+    if (pres.headers.hasOwnProperty('x-frame-options')) {
+      res.setHeader('X-Frame-Options', pres.headers['x-frame-options']);
+    }
+    if (pres.headers.hasOwnProperty('location')) {
+      res.setHeader('Location', pres.headers['location']);
+    }
```
lloyd: maybe refactor this into:

```
[ 'an', 'array', 'of', 'headers' ].forEach(function(header) {
    if (pres.headers.hasOwnProperty(header)) res.setHeader(header, pres.headers[header]);
});
```

a nitpick, drops a little code perhaps.  groups the forwarded headers closer together.

lib/wsapi.js
```diff
@@ -140,7 +140,7 @@ exports.setup = function(options, app) {
   var bodyParser = express.bodyParser();
 
   // stash our forward-to url so different wsapi handlers can use it
-  exports.fowardWritesTo = options.forward_writes;
+  exports.forwardWritesTo = options.forward_writes;
```
lloyd, line 143: geeze.  nice catch.  a consistent misspelling!  go lloyd!

lib/wsapi.js
```diff
@@ -200,6 +200,9 @@ exports.setup = function(options, app) {
         return bodyParser(req, resp, function() {
           next();
         });
+      } else if (options.router_mode) {
+        // skip wsapi request, let browserid middleware handle forwards
+        return next();
```
lloyd, line 205: it would be nice to short-circuit this logic even sooner, because it seems like there's a lot of work being done on the router that isn't strictly necessary.

How much code would we remove from request processing if we were to implement something like a `wsapi.routeMap()` function that returned a map of read URLs and write URLs?

Then I wonder about when we start implementing anti dos stuff.  What's the best code organization?  Would it be nice to be able to write checks, then associate them with some number of api calls?  will all of our request checks run for all apis?

hmmm.

lloyd:
Overall I'm r+ on this change.  Please read through the suggestions above.  In short think maybe a little bit of restructuring could be done to wsapi to reduce the amount of code that runs and lay the foundation for the request validation we're going to build in, but I'm ok with doing that post-merge.

@ozten - wanna give this a look too, given that we're restructuring our processes here?  (adding another one)

lib/http_forward.js
zaach: +1

bin/router
zaach, line 112: Good point. Between static resources and wsapi requests, I wasn't sure if there were other requests that browserid handled. But if they're enumerable, a whitelist makes more sense, yeah. +1 static file serving process for now.

lib/wsapi.js
zaach, line 205: Yep, this should be more modular. Currently each module specifies if it writes to the database or not, so it seems we can build a map from that. But each module must be read and loaded, so maybe it would be better to explicitly declare that info somewhere?

ozten, line 205: We've run into this problem a few times, when trying to declarative capture concerns like these.

Maybe we have a library that wraps express' API?

```
app.get('/foo', function (req, resp) ...
```

could be come

```
routes.get('foo', {db_mode: 'ro', cache: 'NEVER'}, function (req, resp) ...
```

Then you can can get this list of routes as well as metadata about those routes.

This would capture metadata near the code that runs, would avoid repeating code.

Aside: A pythonic or java way to do this would be annotations

```
# cache(never)
# db_mode(read_only)
def foo (req, resp): ...
```

ozten:
This is great work @zaach. I like the idea of extracting a router from the codebase.

lloyd:
r+  I think the remaining issues can be addressed post merge.  

tests pass:  http://travis-ci.org/#!/zaach/browserid/builds/1470642

I'm clicking the big green button!

Issue #1662, opened 2012-05-30
static-resource-test.js is failing


shane-tomlinson:
STR

1) Run ./tests/static-resource-test.js

Expected:
Test passes

Actual:
✗ 

  All resources expand
    ✗ We get stuff
      » expected 16,
    got  10 (==) // static-resource-test.js:35
✗ Broken » 1 broken (0.003s)

shane-tomlinson:
@lloyd took care of this with c7e84298b24c015d759a527850141d6dfe3d90c0

jrgm:
belated verify that that test is not failing in local dev for train-2012.07.30

PR #1663, opened 2012-05-30
Fix the failing static-resource-test.

shane-tomlinson:
issue #1662


shane-tomlinson:
@lloyd already took care of this with checkin c7e84298b24c015d759a527850141d6dfe3d90c0

Issue #1619, opened 2012-05-18
From GH-1567 - consider not recording repetitive polling XHR requests 


jrgm:
From GH-1567, the most critical point (#3) in https://github.com/mozilla/browserid/issues/1567#issuecomment-5656953 is fixed. I'm splitting out the remaining 3 points as separate issues.

This is point #1 - "@lloyd - we're collecting a ton of data during polling that is not valuable. perhaps we should not be doing this. this would suggest surpressing xhr events (only capture xhr errors)".

For example, the event stream for a sequence of popping the dialog with one already authed user, and then begin adding a second secondary email address to the account looks like this:

[u'pick_email', 243]
[u'authentication_checked', 243]
[u'add_email', 2126]
[u'xhr_start', 8787]
[u'xhr_complete', 8916]
[u'add_email_submit_with_secondary', 8930]
[u'xhr_start', 14560]
[u'password_set', 14562]
[u'xhr_complete', 15010]
[u'xhr_start', 15019]
[u'email_staged', 15021]
[u'xhr_complete', 15109]
[u'xhr_start', 18112]     <-- the dialog is now in a polling loop of GET /wsapi/email_addition_status?email=...
[u'xhr_complete', 18197]  <-- every 3 seconds, it adds one xhr_start/xhr_complete
[u'xhr_start', 21200]
[u'xhr_complete', 21289]
[u'xhr_start', 24292]
[u'xhr_complete', 24376]
... and so on for as long as the dialog is open (until in a about 12 minutes the blob > 10K).

shane-tomlinson:
This will be fixed with #1614

jrgm:
Yep, verified in dev instance of train-2012.06.08, no more xhr_\* repitition. A typical event stream is now like

"event_stream": [ 
 [ "screen.rp_info", 2203 ], 
 [ "screen.authenticate", 2229 ], 
 [ "screen.set_password", 20728 ], 
 [ "user.user_staged", 31452 ], 
 [ "screen.check_registration", 31458 ], 
 [ "user.user_confirmed", 63940 ], 
 [ "generate_assertion", 63941 ], 
 [ "screen.generate_assertion", 63960 ], 
 [ "assertion_generated", 64252 ], 
 [ "window.unload", 66049 ] 
]

Issue #1614, opened 2012-05-18
KPI event stream names should follow <namespace>.<event_name> format


shane-tomlinson:
Right now we sending the internal message names that are sent to the mediator. This is not useful for the backend as the mediator is full of extraneous data and does not directly say which screens a user sees.

jrgm:
Confirmed in dev for train-2012.06.08. A typical event stream now looks like 

"event_stream": [  
 [ "screen.rp_info", 2203 ],  
 [ "screen.authenticate", 2229 ],  
 [ "screen.set_password", 20728 ],  
 [ "user.user_staged", 31452 ],  
 [ "screen.check_registration", 31458 ],  
 [ "user.user_confirmed", 63940 ],  
 [ "generate_assertion", 63941 ],  
 [ "screen.generate_assertion", 63960 ],  
 [ "assertion_generated", 64252 ],  
 [ "window.unload", 66049 ]  
]

PR #1664, opened 2012-05-30
Filter and namespace the interaction data.

shane-tomlinson:
@jedp - mind reviewing this?
- Set up a table that says which events are noteworthy
- Save the number of emails a user has.
- All screens are namespaced under the screen namespace
- User interactions are saved under the user namespace
- window related events are under the window namespace
- Modify the hub to call global listeners before exact match listeners - this keeps events in the correct order in the store - the previous way had a new screen show up in the KPI event log before the "screen.cancel" event that triggered its display.

issue #1614
issue #1619

### DIFF
--- a/resources/static/shared/modules/interaction_data.js
+++ b/resources/static/shared/modules/interaction_data.js
@@ -30,6 +30,43 @@ BrowserID.Modules.InteractionData = (function() {
       dom = bid.DOM,
       sc;
 
+  /**
+   * This is a translation table from a message on the mediator to a KPI name.
+   * Names can be modified or added to the KPI storage directly.
+   * A name can be translated by using either a string or a function.
+   *
+   * value side contains - purpose
+   * null - no translation, use mediator name for KPI name.
+   * string - translate from mediator name to string.
+   * function - function takes two arguments, msg and data.  These come
+   *   directly from the mediator.  Function returns a value.  If no value is
+   *   returned, field will not be saved to KPI data set.
+   */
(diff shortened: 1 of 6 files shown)


resources/static/shared/modules/interaction_data.js
```diff
@@ -30,6 +30,43 @@ BrowserID.Modules.InteractionData = (function() {
       dom = bid.DOM,
       sc;
 
+  /**
+   * This is a translation table from a message on the mediator to a KPI name.
+   * Names can be modified or added to the KPI storage directly.
+   * A name can be translated by using either a string or a function.
+   *
```
jedp, line 37: cool

resources/static/shared/modules/interaction_data.js
```diff
@@ -30,6 +30,43 @@ BrowserID.Modules.InteractionData = (function() {
...
+    generate_assertion: null,
+    assertion_generated: null,
+    emails_displayed: function(msg, data) { return "user.email_count:" + data.count; },
+    user_staged: "user.user_staged",
+    user_confirmed: "user.user_confirmed",
+    email_staged: "user.email_staged",
+    email_confirmed: "user.email_confrimed",
+    notme: "user.logout",
+  };
+
```
jedp, line 59: this is great.  i like how extensible and flexible you've made this.

jedp:
Thanks  - this is great.  And it looks like when change our minds about which events to track and at what granularity, this will make it very easy to do that.

Issue #1660, opened 2012-05-30
server timestamp no longer in KPI blob (with event_stream, lang, sample_rate, screen_size).


jrgm:
The data uploaded to /wsapi/interaction_data on the wire consists of blobs like:

```
{
  "event_stream": [ event, ... ],
  "lang": "en-us",
  "sample_rate": 1,
  "screen_size": {
    "height": 1050,
    "width": 1680,
  },
}
```

Is it intentional that the server `timestamp` to have been stripped from that blob? (It is clearly not on the whitelist in the code, but the code for interaction_data looks like it's (optionally) expecting it).

jrgm:
link to GH-1597 and GH-1613

jedp:
Hmm...  I can't reproduce this.

To answer your question, it's not intentional that that would happen, no.  The timestamp should be added on the server-side in `lib/wsapi/interaction_data.js`.

Are you finding this is happening in the dev branch?  When I run browserid locally, I get the timestamps.  Also, they're being recorded in the blobs at https://kpiggybank.hacksign.in/

jedp:
Ah!  Ok.  Reproduced.

We also lost `user_agent` somehow.  

I will fix.

jrgm:
Verified in a dev env (feature not enabled in stage). The "data" portion of the payload sent contains only the keys
lang, timestamp, event_stream, sample_rate, screen_size

PR #1670, opened 2012-05-31
Send server timestamp along with KPI data.

shane-tomlinson:
- Add tests to make sure all expected fields are present.

issue #1660

### DIFF
--- a/resources/static/test/cases/shared/models/interaction_data.js
+++ b/resources/static/test/cases/shared/models/interaction_data.js
@@ -87,7 +87,16 @@
 
         // When the interaction_data next completes, this will be the only data
         // that is pushed.
-        model.push({ lang: "bar", secret: "Attack at dawn!!!" });
+        var now = new Date().getTime();
+
+        model.push({
+          event_stream: [],
+          sample_rate: 1,
+          timestamp: now,
+          local_timestamp: now,
+          lang: "bar",
+          secret: "Attack at dawn!!!"
+        });
         model.stageCurrent();
 
         xhr.useResult("valid");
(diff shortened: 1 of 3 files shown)


resources/static/test/cases/shared/models/interaction_data.js
```diff
@@ -87,7 +87,16 @@
...
-        model.push({ lang: "bar", secret: "Attack at dawn!!!" });
+        var now = new Date().getTime();
+
+        model.push({
+          event_stream: [],
+          sample_rate: 1,
+          timestamp: now,
+          local_timestamp: now,
+          lang: "bar",
+          secret: "Attack at dawn!!!"
```
lloyd, line 98: what is this secret?  what's its utility and intended configuration in production?

resources/static/test/cases/shared/models/interaction_data.js
```diff
@@ -87,7 +87,16 @@
 
         // When the interaction_data next completes, this will be the only data
         // that is pushed.
-        model.push({ lang: "bar", secret: "Attack at dawn!!!" });
+        var now = new Date().getTime();
+
+        model.push({
+          event_stream: [],
+          sample_rate: 1,
+          timestamp: now,
```
lloyd, line 95: didn't we agree to round timestamp down to 5 minute granularity to make log correlation intractable, even for ourselves?  is this the work of a subsequent bug?  

We should do this in browserid before blobs leave for the piggy bank.

resources/static/test/cases/shared/models/interaction_data.js
shane-tomlinson, line 98: @lloyd - it is an extraneous field to check to make sure the whitelist is working properly.

resources/static/test/cases/shared/models/interaction_data.js
shane-tomlinson, line 95: @lloyd - @jedp took care of the rounding on the backend.  We can do it on the front end as well.

lloyd, line 95: I think the browserid router process is where we'd want to round, not in frontend code nor in kpi servers.

make sense?

PR #1680, opened 2012-06-04
Hotfix clearing expired certs are cleaned up

benadida:


resources/static/shared/user.js
```diff
@@ -696,8 +696,7 @@ BrowserID.User = (function() {
      * @param {function} [onFailure] - Called on error.
      */
     checkAuthenticationAndSync: function(onComplete, onFailure) {
-      network.checkAuth(function(authenticated) {
-        setAuthenticationStatus(authenticated);
+      User.checkAuthentication(function(authenticated) {
```
shane-tomlinson, line 699: @benadida - Is the intent here to cause stored emails to be deleted if the user is not authenticated?

shane-tomlinson:
@benadida, @lloyd - There are no tests for getSilentAssertion in resources/static/test/cases/shared/user.js.  Bad jiji.

If the intent of the change to checkAuthenticationAndSync(L699) is to cause stored emails to be deleted if the user is not authenticated, the test in user.js labelled 'checkAuthenticationAndSync with invalid authentication' should be updated to ensure this.

benadida:
@shane-tomlinson agreed, we should have more tests.

The intent is more than that: right now there's a pretty bad bug where, if you have an expired cert, it's still used to provide silent assertions, and those fail because the cert is expired. So RPs are seeing this, especially with the new watch() API. I'm trying to fix this ASAP.

shane-tomlinson:
@benadida - The overall intent I am good on, it is the particular line (699) that I was wondering about.  The overall net gain of changing line 699 is to check cookies as well as authentication.  If cookies are disabled, localStorage will not be touched but no assertion will be generated.  If cookies are enabled and the user is not authenticated, localStorage will be cleared.

benadida:
@shane-tomlinson so I figured that not checking cookies in checkAuthenticationAndSync -- when they are being checked in checkAuthentication -- was an oversight. But maybe it wasn't?

shane-tomlinson:
@benadida You are right that it was an oversight - good catch.  Since this is a hotfix, I am being more cautious than usual and want to understand the intent and the repercussions of each change.

After doing a visual inspection of everywhere checkAuthentication and checkAuthenticationAndSync are called, I am comfortable with this change.

benadida:
I'm going to close this PR since we won't be hotfixing this, and I'll merge it into dev instead.

PR #1681, opened 2012-06-04
Fix globals that were being unintentionally created in wsapi.js

zaach:
Some apparent copy/paste errors were causing global variables to be created in wsapi.js. This commit fixes the global creation and fixes modules that were relying on the implied globals.

### DIFF
--- a/lib/bcrypt.js
+++ b/lib/bcrypt.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const
+path = require('path'),
 computecluster = require('compute-cluster'),
 logger = require('../lib/logging.js').logger,
 bcrypt = require('bcrypt'),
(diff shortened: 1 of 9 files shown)


shane-tomlinson:
@zaach - On the surface, this PR looks great.  Unfortunately, I am getting errors while running ./scripts/test_backend.  Until the tests pass, r-.

```
[11:46][browserid(zaach_dev)] $ ./scripts/test_backend                          
tests/add-email-with-assertion-test.js
The "sys" module is now called "util". It should have a similar interface.
······error getting csrf token:  http error
✗✗
```

shane-tomlinson:
@zaach - When creating a new browserid directory and checking out a new branch with your changes, everything works as expected.  Digging deeper.

shane-tomlinson:
After removing the locales directory, putting the computer to sleep and re-waking it, the tests all passed.  I was unable to reproduce the above problem and merged this in.  @ozten saw the same problem when he updated.

jrgm:
verified in train-2012.06.08

PR #1687, opened 2012-06-05
Update production locales for train-2012.05.25

mathjazz:


ozten:
We don't want to lose locales. We should accept locales with less the 100% strings. We'll meet today to discuss.

Issue #1320, opened 2012-03-14
Mobile: manage screen width not responsive to device width


shane-tomlinson:
It's pegged at 300px.

shane-tomlinson:
9c8e6b680f0bfaa1f7085a59078b341ea74fcd88 takes care of this as well.

PR #1688, opened 2012-06-05
Make sure the manage screen looks right in both landscape and portrait mode on mobile devices.

shane-tomlinson:
issue #1320

To test this:
1) On your mobile device, sign in on the main site.
2) The manage screen will show - make sure it looks right in both portrait and landscape mode.
3) Make sure the TOS and Privacy pages look as expected.

### DIFF
--- a/resources/static/css/m.css
+++ b/resources/static/css/m.css
@@ -2,7 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-@media screen and (max-width: 936px) {
+/*
+ * 946 is the max width of the center area plus its padding
+ */
+@media screen and (max-width: 946px) {
   #wrapper {
     width: 100%;
   }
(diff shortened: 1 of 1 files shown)


resources/static/css/m.css
```diff
@@ -33,6 +45,9 @@
   }
 }
 
+/*
+ * 505 catches mobile devices in portrait mode
```
ozten, line 49: "mobile devices" - meaning typical phones, but not tablets?

We have a lot of different min-width. It might be good to standardize the 2 or three layouts we support and document them at the top of the file, then use these same values in the media queries. 620 and 505 seem like less popular breaking points, but I haven't done a lot of production media queries.

These are just knits, r+

shane-tomlinson, line 49: @ozten - I totally agree with you here.  The 620 and 505 are "magic" numbers that make the main site header look reasonable.  When the window reaches ~620 width, the right hand menu starts to drop below the left hand Persona icon.  Unless the menu is centered when the drop happens, it looks terrible.  505 is about the width where the centered menu becomes too wide and fills two lines, at this point I shrink the text, shrink the logo, and try to make things look appropriate again. - this comment would make good documentation in the CSS.

shane-tomlinson, line 49: 57ca1e5b2c496834fd7fe3058cda43cf215dce12 puts the above comment into the CSS for clarity for maintainers.

PR #1692, opened 2012-06-06
when obtaining an assertion silently, still make sure that expired certs are cleaned up

benadida:


shane-tomlinson:
@benadida - I have added a bunch of tests with respect to how network.cookiesEnabled, user.checkAuthentication, user.checkAuthenticationAndSync and user.getSilentAssertion should work.  

I do not know how to create an expired cert to test to make sure getSilentAssertion and checkAuthenticationAndSync are clearing the original cert and creating a new one.

lloyd:
r+ - http://travis-ci.org/#!/mozilla/browserid/builds/1545869

Issue #1693, opened 2012-06-06
new layout main site: even if content fits on screen, scroll bar appears.


shane-tomlinson:
![Main site showing scroll bar](https://img.skitch.com/20120606-epm8miw4r19pq1hued4qjspbt3.png)

jrgm:
verified on stage train-2012.06.22 -- the scrollbar does not appear on all other browser/os combinations, except for Opera on Windows.

Filed GH-1866 specifically for Opera on Windows

PR #1694, opened 2012-06-06
Fix the scroll bar appearing on the main site's index page if it is not needed.

shane-tomlinson:
To test:

1) On a desktop browser, open the main site index page.  If signed in, sign out.
2) Expand the window vertically so the center content, header and footer are all displayed.
3) See if there is a scroll bar.

issue #1693


lloyd:
r+

Issue #1540, opened 2012-05-04
clear password field if user changes email address in /sign_in dialog


jrgm:
1. Go to 123done.org and click 'Sign in with Persona'
2. If already authed, click 'This is not me...'
3. Enter secondary email address that is known to diresworb.org and hit next.
4. In the password field enter 4 random characters. 
5. User now "realizes" they entered the "wrong" email address (albeit one that is known to diresworb.org)
6. User goes back and fixes the email address to be the "correct" one and hits next.

Expected: Password field is shown and empty
Actual: Password field is shown but contains the four previously typed characters.

This obviously is a very rare scenario and I can see an argument for leaving the previously typed characters should be kept
and letting the user decide what to do.

lloyd:
blank it, shane!

jrgm:
verified in current dev; still waiting for stage but lgtm.

PR #1696, opened 2012-06-06
Clear the password if the user types a password then changes the email address.

shane-tomlinson:
Steps to check:

1) Open the BrowserID dialog.  If signed in, click "This is not me"
2) Enter a known secondary email address into the email field.
3) Begin typing a password.
4) Erase a character in the email address.  Finish the email address again and hit "enter"
5) Password field should be cleared.

issue #1540


lloyd:
r+, very focused minimal change, and I see nothing obvious wrong with the code.

PR #1698, opened 2012-06-06
/wsapi/interaction_data should see the User Agent

ozten:
Testing notes:
1) Open the dialog
2) Several times
3) Note console of browserid

Before - no User Agent in KPI blob.

Now - a User Agent in KPI blob.

I'm having trouble running unit tests locally.

Travis says good to go.
http://travis-ci.org/#!/mozilla/browserid/builds/1548334

### DIFF
--- a/lib/http_forward.js
+++ b/lib/http_forward.js
@@ -90,6 +90,10 @@ exports.forward = function(dest, req, res, cb) {
     preq.setHeader('If-None-Match', req.headers['if-none-match']);
   }
 
+  if (req.headers['user-agent'] && '/wsapi/interaction_data' === req.path) {
+    preq.setHeader('User-Agent', req.headers['user-agent']);
+  }
+
   // if the body has already been parsed, we'll write it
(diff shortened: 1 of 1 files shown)


lib/http_forward.js
```diff
@@ -90,6 +90,10 @@ exports.forward = function(dest, req, res, cb) {
     preq.setHeader('If-None-Match', req.headers['if-none-match']);
   }
 
+  if (req.headers['user-agent'] && '/wsapi/interaction_data' === req.path) {
+    preq.setHeader('User-Agent', req.headers['user-agent']);
+  }
+
```
jedp, line 96: sensible.  I like the whitelist approach for this.

jedp:
Looks good!  R+

jrgm:
verified in dev instance for stage train-2012.06.08. A valid user_agent is passed along in the payload to kpiggybank. (Will need to re-check this when this actually all goes live).

Issue #1699, opened 2012-06-06
i18n.js uses a global instance of `fs` module


zaach:
Tests fail after removing of global instances of modules (issue #1681), including `fs`, causing tests to fail when the `locale` directory is present.

jrgm:
verified on stage train-2012.06.08

PR #1700, opened 2012-06-06
require the `fs` module in i18n.js instead of relying on a global `fs`

zaach:
issue #1699 - Certain code paths in i18n.js only run when JSON locales are present, so this was not caught in #1681.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -18,7 +18,8 @@
 var logger = require('./logging.js').logger,
     Gettext = require('node-gettext'),
     path = require('path'),
-    util = require('util');
+    util = require('util'),
+    fs = require('fs');
 
 const BIDI_RTL_LANGS = ['ar', 'db-LB', 'fa', 'he'];
 


lib/i18n.js
```diff
@@ -18,7 +18,8 @@
 var logger = require('./logging.js').logger,
     Gettext = require('node-gettext'),
     path = require('path'),
-    util = require('util');
+    util = require('util'),
+    fs = require('fs');
```
ozten, line 22: Great work tracking this down. I'm looking through history, to try to see why it got deleted. Most likely an accident. It's not clear to me how L10n has been working...

zaach:
wsapi.js was creating a global instance of `fs`, which is why things broke when it was [removed](https://github.com/zaach/browserid/commit/47fd1b0ac116c2dfb9d06d6f2bb9066b0cac2976#L2R19).

Maybe would could add tests for globals?

ozten:
How would you do that?

I thought undeclared variables stay at the module scope in Node.js (instead of bleeding into global scope like window in Browsers). I'm probably wrong, how does that actually work and what would we test for?

jrgm:
@zaach, @ozten I was thinking of using something like node uses https://github.com/joyent/node/blob/master/test/common.js#L77-140

ozten:
That is really cool.

zaach:
jshint also has an option to detect globals, which can run statically.

Issue #1701, opened 2012-06-06
Clean up global variables that are being created and used in lib/


zaach:
jshint reveals [many instances](https://gist.github.com/2885056) of global variables being created and/or used by modules in the `lib/` directory.

jrgm:
These have been cleaned up in train-2012.06.22 (at least as far as jshint knows). Thanks @zaach!

PR #1703, opened 2012-06-06
Clean up remaining references to global or undefined variables in lib/

zaach:
see issue #1701


lloyd:
this is fantastic.  many a befuddling behavior or un-graceful error handling will go away with this change.

enthusiastic r+.

jrgm:
These have been cleaned up in train-2012.06.22 (at least as far as jshint knows).

Issue #1711, opened 2012-06-08
dev: title on main site is "Mozilla Person" not 'persona'


jrgm:
![img](https://img.skitch.com/20120608-nrgaxhdjskcpbr2r74tbwmw7f1.png)

lloyd:
merged.

jrgm:
verified with stage train-2012.06.08.

PR #1714, opened 2012-06-08
Change "Mozilla Person" in the title to "Mozilla Persona".

shane-tomlinson:
issue #1711


lloyd:
I dunno.  Mozilla Persons are pretty awesome...  r+

Issue #1706, opened 2012-06-07
Unsupported Browser page needs more styling


callahad:
Compare https://browserid.org/unsupported_dialog and https://dev.diresworb.org/unsupported_dialog

The content on the rebranded page should probably be similarly centered.

jrgm:
verified on stage train-2012.06.08 for ie6 and ie7

PR #1715, opened 2012-06-08
Better styling for the unsupported dialog for IE6 and IE7

shane-tomlinson:
issue #1706


lloyd:
css only, looks good to me. r+

PR #1722, opened 2012-06-08
made new API calls fail when not called on navigator.id object

benadida:


resources/static/include_js/include.js
```diff
@@ -1112,11 +1112,15 @@
 
     navigator.id = {
       request: function(options) {
+        if (this != navigator.id)
+          throw new Error("all navigator.id calls must be made on the navigator.id object");
```
ozten, line 1116: Is this enforcement okay with Soup or other non-shim usage of include.js ?

benadida, line 1116: it should be... they may be doing the "wrong" thing, but everyone should correct it.

@digitarald any thoughts?

lloyd:
r+ because this is _only_ on the new api and shouldn't break others in production

Issue #1713, opened 2012-06-08
router in tests showing 'error forwarding request: code=ECONNRESET'


jrgm:
I was running `npm test` in an osx clean clone of current `dev` and noticed these warnings. So, that's about the 10k limit on POST; `router` seems to have the same check, but still the test shows the full request forwarded to browserid webhead. Perhaps this does not matter in prod/stage, but it would good to look into this. /cc @zaach

```
tests/post-limiting-test.js
·····router (63136): error: error forwarding request: code=ECONNRESET
····
✓ OK » 9 honored (5.523s)
```

zaach:
Thanks @jrgm – I'll look into this.

lloyd:
fixed with above mentioned merge.  see also (duplicate) issue #1745

jrgm:
Verified stage train-2012.06.22. See https://github.com/mozilla/browserid/issues/1745#issuecomment-6639681

PR #1723, opened 2012-06-08
Enforce post size limit in the router only - issue #1713

zaach:
In cases where the content-length header isn't set, the router streams requests to the browserid process and destroys the request if it goes over the size limit. When both the router and browserid processes perform this check, an error occurs as both attempt to destroy the same connection.

### DIFF
--- a/bin/browserid
+++ b/bin/browserid
@@ -60,10 +60,6 @@ app.use(i18n.abide({
   disable_locale_check: config.get('disable_locale_check')
 }));
 
-// limit all content bodies to 10kb, at which point we'll forcefully
-// close down the connection.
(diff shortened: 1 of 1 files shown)


bin/browserid
```diff
@@ -60,10 +60,6 @@ app.use(i18n.abide({
   disable_locale_check: config.get('disable_locale_check')
 }));
 
-// limit all content bodies to 10kb, at which point we'll forcefully
-// close down the connection.
-app.use(express.limit("10kb"));
-
```
ozten, line 66: Wow, great find. Router to the rescue.

r+

lloyd:
r+  I can't think of a case where we would want to check in both places.

Issue #1240, opened 2012-03-02
mismatch of PP/TOS text and button text ('next' vs. 'sign in')


jrgm:
1) Go to beta.myfavoritebeer.org and click signin
2) click 'This is not me' if signed in
3) Enter either the email of an existing user or something new (adsfiowerui@example.com)

The text says "By clicking next ...", but the button is either 'sign in' or 'verify email'

![img](https://img.skitch.com/20120302-n2yqx5dggjiupmgfac9m4wsnmb.png)

lloyd:
@skinny97214 will give us a chart, @ozten will do a first pass, after we change the world we will do another pass.

skinny97214:
Only show on screens where the button is the last click. It shouldn't show next to the [next] button.

shane-tomlinson:
@skinny97214 - there is a problem with only showing the tos/pp message only at the last click.  If the user is logged in to their primary whenever they initially enter their email address into the email field, no "next" button is shown.  

A second source of weirdness is when a user goes to add an email to an existing account.  More on this to follow.

User signing in/up to BrowserID:

```
 primary is an address supported by a Primary IdP.
 secondary is an address that must be verified using BrowserID as the secondary authority.

* Email is already registered with BrowserID:
       - primary, user not logged in with IdP - show at the "next" button.
       - primary, user logged in with IdP - No "next" button is ever shown.
       - secondary - show when the password field and "sign in" buttons are shown.
* Email is not registered with BrowserID (new user):
       - primary, user not logged in with IdP - show at the "next" button.
       - primary, user logged in with IdP - No "next" button is ever shown.
       - secondary - show when the "verify email" button is shown.
```

If the user is signed in to BrowserID:

```
* at the email picker.
```

Add Email Screen - If new email is:

```
* secondary - verification screen is shown without a second button being displayed.
* primary, user is signed in with IdP - email is added without a second button being displayed.
* primary, user is not signed in with IdP - show with the "verify" button.
```

As you can see, with the add email flow, two of the 3 flows do not display a second button, and one does.  We only know which of the three states the user will be in once the user clicks the first button.  Should we add button clicks or show the tos/pp statement for all 3?

skinny97214:
Ok, we can show it on the first sign-in screen next to [next] 

... IFF we reduce when users see this. It shouldn't show if their local storage indicates they've already acknowledged the TOS. Nice to have: It also shouldn't appear if they site indicates to us the user already has an account.

[...]

PR #1727, opened 2012-06-11
Issue 1240 tos update

shane-tomlinson:
Show the BrowserID and RP TOS/PP at the correct time with the correct text.

To Test:

```
* BID TOS/PP is only displayed to new users.
* RP TOS/PP is only displayed to new users, users who are visiting a new site, or users who have just authenticated to BrowserID.
```


resources/static/dialog/css/popup.css
```diff
@@ -284,24 +284,24 @@ section > .contents {
 #favicon img {
     display: block;
     margin: 0 auto;
-    max-height: 128px;
-    max-width: 128px;
+    max-height: 100px;
+    max-width: 100px;
```
lloyd: this is an important change.  100x100 is our new siteLogo maxsize...

resources/static/test/cases/controllers/dialog.js
```diff
@@ -151,6 +151,7 @@
         catch(e) {
           // do nothing, an exception will be thrown because no modules are
           // registered for the any services.
+          console.log(e);
```
lloyd: intentional console.log() ?

shane-tomlinson: @lloyd - good eye - this console.log message is intentional since it is in the tests.

shane-tomlinson: @lloyd I removed this console.log message in 0a4c08e3b9747463593d6821138be48987a33abe

shane-tomlinson:
Link to issue #1240

lloyd:
r+ - tests clean, walkthrough with skinny looks good.  expect some instability but we'll knock it down as it comes up.

PR #1728, opened 2012-06-11
Issue 1695 host fonts

shane-tomlinson:
Host fonts locally.  For each font that is used, there are 3 fonts, one .woff, one .eot, and one .ttf.  .woff is for most modern desktop browsers.  .eot is for IE.  .ttf is for Android and iOS.

So that the number of fonts retreived is kept to a minimum, there are separate font .css files - one for common, one for the main site, and one for the dialog.

Fonts caching works correctly with connect-cachify.


resources/static/css/common.css
```diff
@@ -291,7 +291,6 @@ a.secondary[disabled], .submit_disabled a.secondary, .submit_disabled a.secondar
 }
 
 .headline-main, h1, h2, h3, h4 {
-  font-family: 'Open Sans', "Lucida Sans", "Lucida Grande", "Lucida Sans Unicode", Verdana, sans-serif;
```
lloyd, line 294: the only line in this patch I can't understand is this one.  Was this css overlaying some catchall lower down?  redundant?

shane-tomlinson, line 294: @lloyd - the "body" tag in "common.css" line 20 declares the same font face universally.

Issue #1376, opened 2012-04-03
Form does not always submit when hitting "enter" in password field in IE8


shane-tomlinson:
STR

1) Load up myfavoritebeer.org in IE8
2) Open dialog, if signed in, sign out.
3) In email field, type in known email address, press enter
4) In password field, type in password, press enter.
5) try pressing enter again.

Nothin'

Expected - form is submitted.

shane-tomlinson:
Took out the "fix" because it caused a double submit in Firefox and Chrome.  Grrr.

seanmonstar:
We could add a timeout if the key is enter, and cancel the timeout in submit, so that the timeout makes the submit happen in IE, and good browsers cancel the timeout and submit once.

Yeck, I feel the need to sacrifice a few kittens after having written that sentence.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #1729, opened 2012-06-11
Fix IE8 not submitting the password field on the enter key.

shane-tomlinson:
To test:

1) To avoid the "script taking too long" message in IE8, run <server>/test/index.html?filter=shared/modules/page_module
2) In IE8, open the dialog.  If logged in, click "this is not me".  
3) Enter in a known secondary email, press "enter".  Enter in password, press enter.  If fix is working, second "enter" will work.
4) Follow same steps in other browsers to make sure pressing "enter" in other browsers does not cause a double submit.

issue #1376


lloyd:
r+ - simple and clean.  only possible issues might be on other browsers that don't properly preventDefault?

Issue #1734, opened 2012-06-12
Verification screen flashes email address field before success message.


shane-tomlinson:
It's pretty ugly.

lloyd:
+1  this bugs the heck outta me

shane-tomlinson:
Link to #218

jrgm:
verified on stage train-2012.06.22 - there is no flash of the email address either signing up new, or adding a user; re-using an already registered verify link brings up the new style dialog.

PR #1735, opened 2012-06-12
Post Verification Redirect!

shane-tomlinson:
Steps to test:

1) When creating an account with secondary email and verifying in same browser, verification page is redirected to RP.

2) When adding a secondary email to existing account and verifying in same browser, verification page is redirected to RP.

3) When resetting the password of a secondary email and verifying in same browser, verification page is redirected to RP.

4) When creating an account with secondary email and verifying in different browser, verification page is _NOT_ redirected to RP.

5) When adding a secondary email to existing account and verifying in different browser, verification page is _NOT_ redirected to RP.

6) When resetting the password of a secondary email and verifying in different browser, verification page is _NOT_ redirected to RP.

issue #385

### DIFF
--- a/resources/static/dialog/controllers/dialog.js
+++ b/resources/static/dialog/controllers/dialog.js
@@ -186,17 +186,11 @@ BrowserID.Modules.Dialog = (function() {
 
         // returnTo is used for post verification redirection.  Redirect back
         // to the path specified by the RP.
-        var returnTo;
         if (paramsFromRP.returnTo) {
-          returnTo = fixupAbsolutePath(origin_url, paramsFromRP.returnTo);
-        }
-        else {
-          // Native implementations will be behind on the returnTo feature.  Until
-          // they are ready, set the returnTo to be the origin_url;
-          returnTo = origin_url;
+          var returnTo = fixupAbsolutePath(origin_url, paramsFromRP.returnTo);
+          user.setReturnTo(returnTo);
         }
 
-        user.setReturnTo(returnTo);
 
         if (hash.indexOf("#CREATE_EMAIL=") === 0) {
           var email = hash.replace(/#CREATE_EMAIL=/, "");
--- a/resources/static/dialog/controllers/dialog.js
+++ b/resources/static/dialog/controllers/dialog.js
@@ -184,6 +184,17 @@ BrowserID.Modules.Dialog = (function() {
           params.siteName = _.escape(paramsFromRP.siteName);
         }
 
+        var originHREF = paramsFromRP.originHREF;
+
+        // Native implementations will be behind on the originHREF feature.  Until
+        // they are ready, set the originHREF to be the origin_url;
+        if (!originHREF) originHREF = origin_url;
+
+        if(originHREF.indexOf(origin_url) !== 0) {
+          throw "originHREF/origin mismatch";
+        }
+        user.setOriginHREF(fixupURL(origin_url, originHREF));
+
         if (hash.indexOf("#CREATE_EMAIL=") === 0) {
           var email = hash.replace(/#CREATE_EMAIL=/, "");
           if (!bid.verifyEmail(email))
--- a/resources/static/dialog/controllers/dialog.js
+++ b/resources/static/dialog/controllers/dialog.js
@@ -184,16 +184,19 @@ BrowserID.Modules.Dialog = (function() {
           params.siteName = _.escape(paramsFromRP.siteName);
         }
 
-        var originHREF = paramsFromRP.originHREF;
-
-        // Native implementations will be behind on the originHREF feature.  Until
-        // they are ready, set the originHREF to be the origin_url;
-        if (!originHREF) originHREF = origin_url;
-
-        if(originHREF.indexOf(origin_url) !== 0) {
-          throw "originHREF/origin mismatch";
+        // returnTo is used for post verification redirection.  Redirect back
+        // to the path specified by the RP.
+        var returnTo;
+        if (paramsFromRP.returnTo) {
+          returnTo = fixupAbsolutePath(origin_url, paramsFromRP.returnTo);
         }
-        user.setOriginHREF(fixupURL(origin_url, originHREF));
+        else {
+          // Native implementations will be behind on the returnTo feature.  Until
+          // they are ready, set the returnTo to be the origin_url;
+          returnTo = origin_url;
+        }
+
+        user.setReturnTo(returnTo);
 
         if (hash.indexOf("#CREATE_EMAIL=") === 0) {
           var email = hash.replace(/#CREATE_EMAIL=/, "");
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1069,6 +1069,9 @@
       // don't do duplicative work
       if (commChan) commChan.notify({ method: 'dialog_running' });
 
+      // originHREF is used for post-email-verification redirect
+      options.originHREF = document.location.href;
+
       w = WinChan.open({
         url: ipServer + '/sign_in',
         relay_url: ipServer + '/relay',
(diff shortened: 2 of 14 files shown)


resources/static/dialog/controllers/dialog.js
```diff
@@ -184,6 +184,17 @@ BrowserID.Modules.Dialog = (function() {
           params.siteName = _.escape(paramsFromRP.siteName);
         }
 
+        var originHREF = paramsFromRP.originHREF;
```
shane-tomlinson: @lloyd - can we get this information from WinChan instead of having to pass it from include.js?  It would be cleaner.

resources/static/dialog/controllers/dialog.js
```diff
@@ -184,6 +184,17 @@ BrowserID.Modules.Dialog = (function() {
...
+
+        // Native implementations will be behind on the originHREF feature.  Until
+        // they are ready, set the originHREF to be the origin_url;
+        if (!originHREF) originHREF = origin_url;
+
+        if(originHREF.indexOf(origin_url) !== 0) {
+          throw "originHREF/origin mismatch";
+        }
+        user.setOriginHREF(fixupURL(origin_url, originHREF));
```
shane-tomlinson: Would it make sense to set the origin in user.js to use the originHREF instead of having a separate field?

resources/static/include_js/include.js
```diff
@@ -1069,6 +1069,9 @@
       // don't do duplicative work
       if (commChan) commChan.notify({ method: 'dialog_running' });
 
+      // originHREF is used for post-email-verification redirect
+      options.originHREF = document.location.href;
```
shane-tomlinson: This is explicitly set and _cannot_ be overridden by the RP.

lloyd:
we should expose a path in the api so an RP can override the location that a user is sent back to.  I suggest `returnTo` as a property name, and that we keep this as a path only to start.

/cc @benadida

resources/static/dialog/controllers/dialog.js
```diff
@@ -184,6 +184,20 @@ BrowserID.Modules.Dialog = (function() {
           params.siteName = _.escape(paramsFromRP.siteName);
         }
 
+        // returnTo is used for post verification redirection.  Redirect back
+        // to the path specified by the RP.
+        var returnTo;
+        if (paramsFromRP.returnTo) {
+          returnTo = fixupAbsolutePath(origin_url, paramsFromRP.returnTo);
```
lloyd: I like that this shows a user visible error if returnTo is not allowable.  this seems a reasonable way to surface the error to devs.

resources/static/dialog/controllers/dialog.js
```diff
@@ -184,6 +184,20 @@ BrowserID.Modules.Dialog = (function() {
...
+        }
+        else {
+          // Native implementations will be behind on the returnTo feature.  Until
+          // they are ready, set the returnTo to be the origin_url;
+          returnTo = origin_url;
```
lloyd: hmm.  that's odd.  would it be better to have _no_ redirect for native implementations until they implement it?

resources/static/pages/verify_secondary_address.js
```diff
@@ -105,6 +122,12 @@ BrowserID.verifySecondaryAddress = (function() {
 
       token = options.token;
       verifyFunction = options.verifyFunction;
+      doc = options.document || document;
+
+      redirectTimeout = options.redirectTimeout;
+      if (typeof redirectTimeout === "undefined") {
+        redirectTimeout = REDIRECT_TIMEOUT;
+      }
```
lloyd, line 152: skimming the code (I'll verify by deploying and testing), it seems like we display the page for a while before redirecting.  That seems right as a first step to _me_, but I think it's contrary to skinny's ultimate desire.  I think skinny ultimately wants _nothing_ from persona visible and the user to land directly on the RP.

I can think of how this might be accomplished with a display: none and fade-in if no redirect url is present (like in verify-on-different browser)...

resources/static/shared/storage.js
```diff
@@ -147,18 +146,18 @@ BrowserID.Storage = (function() {
     }
   }
 
-  function setStagedOnBehalfOf(origin) {
-    storage.stagedOnBehalfOf = JSON.stringify({
+  function setReturnTo(origin) {
```
lloyd: is this argument misnamed?  it's a _url_ now right?  not just an origin.

resources/static/shared/storage.js
```diff
@@ -147,18 +146,18 @@ BrowserID.Storage = (function() {
...
+    storage.returnTo = JSON.stringify({
       at: new Date().toString(),
       origin: origin
```
lloyd: ditto.

resources/static/shared/storage.js
```diff
@@ -147,18 +146,18 @@ BrowserID.Storage = (function() {
...
     });
   }
 
-  function getStagedOnBehalfOf() {
+  function getReturnTo() {
     var origin;
 
     try {
-      var staged = JSON.parse(storage.stagedOnBehalfOf);
+      var staged = JSON.parse(storage.returnTo || storage.stagedOnBehalfOf);
```
lloyd: nice catch for the transition.  maybe add a comment saying we can delete the || part in about a month.

resources/static/shared/user.js
```diff
@@ -273,24 +273,29 @@ BrowserID.User = (function() {
...
+    },
+
+    getReturnTo: function() {
+      return this.returnTo;
+    },
+
     /**
-     * Create a user account - this creates an user account that must be verified.
-     * @method createSecondaryUser
+     * Create a user account - this creates an user account that must be verified.  * @method createSecondaryUser
```
lloyd: maybe add an `\n` here for readability.

lloyd:
blocked on a couple nits above.  I'll deploy ephemerally and test, pending that and responses to the questions I consider this ready for merge.  We should bump the conversation about parameter naming on list once we merge: https://groups.google.com/group/mozilla.dev.identity/browse_thread/thread/fe1de8fbbd491232#

resources/static/shared/user.js
shane-tomlinson: Woah, how'd that \n get yanked out?

resources/static/pages/verify_secondary_address.js
shane-tomlinson, line 152: @lloyd - I made this timeout configurable primarily for testing.  I didn't want to slow the tests down for 5 seconds when I didn't need to, but it does give us extra flexibility to remove the timeout.

As to the second part of your concern - I have started that bit in a separate branch to address issue #1734 (verification screen flash) and would prefer to continue this suggestion in that branch.

resources/static/shared/storage.js
shane-tomlinson: @lloyd - good catch.

lloyd:
BLOCKER: when no returnTo param is supplied, we get an error in dialog.

visible here: http://pvr.123done.org/

lloyd:
r+ on the code and basic functionality. Two things we might want to consider before merge:
1. we should not rely on the dialog completing for the user to be automatically logged in. One case I hit twice while testing was that there was a delay while waiting for the email. So I closed the dialog, did something else, went back and checked my email. I got back to the RP flawlessly, but was not logged in. I think all we need is an authenticated session to persona's servers and to call storage.setLoggedIn(origin, email) in the verification page before redirect.
2. the redirect feels a little jarring. It says it will redirect me, and I wait for an uncomfortable amount of time, then it shoots me over... Some kind of progress meter here might help? I'm pretty used to a countdown "redirecting in 3 seconds" then change 3 to 2, 2 to 1, and go. dunno

lloyd:
r+

Issue #34, opened 2011-06-21
port to express


lloyd:
rather than using connect, why not use express?  it's a thin layer atop connect that provides additional interesting features.

lloyd:
done

PR #1739, opened 2012-06-12
Add BigTent as a Proxy IdP for a whitelist of domains

ozten:
We should land this branch in this train, if possible.

Patch adds two config variables: `bigtent_domains` and `bigtent_url`.
It augments wsapi address_info so that after primary check fails, it also checks bigtent_domains for proxyidp support.

It updates the verifier to allow proxyidp domains to certify a different domain than their own. This is a bug and will be fixed in [Issue #34](https://github.com/mozilla/browserid-bigtent/issues/34), before or after beta.

It updates the screen flows for a special case to proxyidp.

Pull Request caveats
1) No rebase
I spent over an hour rebasing and squashing to get it down to 1 commit, but was unable too. Lots of good work by @shane-tomlinson, so I it's good to keep the commit history there :)

2) Failing qunit tests
I appear to have broken the unit tests around requireEmail on the last merge from dev.
I found one place http://git.io/RK2Fzw and that fixed 50% of the failing qunit tests, but
I don't see the root cause of the other ones.

@shane-tomlinson can you look at http://travis-ci.org/#!/mozilla/browserid/jobs/1604258
and then
https://gist.github.com/2920556 for obvious merge mistakes?


lib/wsapi/address_info.js
```diff
@@ -47,6 +51,22 @@ exports.process = function(req, res) {
...
+      primary.checkSupport(bigtent, function(err, urls, publicKey, delegates) {
+        if (bt_done) {
+          return;
+        }
+        bt_done = true;
+        if (err || ! urls) {
+          logger.warn('error checking BigTent for IdP details: ' + err);
+          return httputils.serverError(res, "BigTent unavailable");
+        }
+        urls.type = 'primary';
```
shane-tomlinson: @ozten - for proxyidp's, the type was originally "proxyidp" and this differentiation allowed the front end to differentiate between a BigTent primary and a normal primary.  BigTent primaries were not shown the "you have to verify with your primary" with the URL screen.  Is the change to "primary" intentional or should this still be "proxyidp"?

lib/wsapi/auth_with_assertion.js
```diff
@@ -38,7 +38,8 @@ exports.process = function(req, res) {
       if (err) return wsapi.databaseDown(res, err);
 
       // if this is a known primary email, authenticate the user and we're done!
-      if (type === 'primary') {
+      if (type === 'primary' ||
+          type === 'proxyidp') {
```
shane-tomlinson, line 42: @ozten - still using "proxyidp" here.

lib/wsapi/add_email_with_assertion.js
```diff
@@ -23,6 +23,7 @@ exports.i18n = false;
 // authenticated.
 exports.process = function(req, res) {
   // first let's verify that the assertion is valid
+  logger.info('AOK PROXYIPD add_email_with_assertion assertion=' + req.body.assertion);
```
shane-tomlinson: @ozten - Do you want to take this out?

shane-tomlinson:
@ozten - I'm excited for this to land, awesome work.  Overall, things look good to me but there are a couple of changes that I think need made - noted in the code.

Additional concerns:
1) If the BigTent server is not running, right now there is a hard failure for users of gmail, yahoo, and hotmail - see https://github.com/mozilla/browserid/pull/1739/files#L7R65.  If BigTent is not available, should this instead fall back to checking gmail, yahoo and hotmail whether they are primaries, and if not, fall back again to being a secondary?

2) If gmail, yahoo or hotmail decided to turn on primary support without first consulting us, BigTent is going to override their wishes.  Before BigTent does its magic, should we first check to see if they have native primary support?

shane-tomlinson:
@ozten - Secondly, I have rebased this branch from the current dev.  All tests pass, on both the front and backends, but I cannot get the BigTent verification to happen - in the dialog I keep seeing:

```
Invalid email input to service function [:email]string-1
```

I have pushed a new branch (bigtent_with_dev) for you to check out.  The updated branch fixes some of the merge issues that crept in along the way.

lib/wsapi/add_email_with_assertion.js
ozten: Doh.

lib/wsapi/address_info.js
ozten: You're absolutely right. I have this set back to proxyidp locally.

It does make me wonder if it will break non-shim clients. If they test for auth and provision properties, they will be fine, but if they test for type === 'primary', they won't handle this case. CC @benadida @anantn

ozten:
Okay, pushed two commits:
1) Fixes for feedback, plus new BigTent config for pete
2) Cherry-picked Shane's unit test fix

ozten:
Lets move the discussion to https://github.com/mozilla/browserid/pull/1753

Issue #1741, opened 2012-06-13
move to persona.org URLs


lloyd:
production:
- browserid.org will redirect to login.persona.org for everything but verification requests
- login.persona.org will replace browserid.org
- static.login.persona.org will be the location all resources are served from
- verifier.login.persona.org will be the new verifier entry point

stage: 
- diresworb.org will redirect to login.anosrep.org for everything except verification requests
- login.anosrep.org will replace diresworb.org
- static.login.anosrep.org will be the location all resources are served from
- verifier.login.anosrep.org will be the new verifier entry point

dev:
- dev.diresworb.org will redirect to login.dev.anosrep.org for everything but verification redirects
- login.dev.anosrep.org will replace dev.diresworb.org
- dev.anosrep.org will redirect to login.dev.anosrep.org
- static.dev.anosrep.org will be where all resources are served from
- verifier.dev.anosrep.org will be the new verifier entry point

ephemeral deployments:
- *.personatest.org will be the url for everything

jrgm:
verified for stage train-2012.06.22 for notes above and in https://bugzilla.mozilla.org/show_bug.cgi?id=767708#c5, except for "static.login.anosrep.org will be the location all resources are served from". The resources are all 200 OK, but it appears js/css come from static.login.anosrep.org, while images/fonts come from login.anosrep.org. (GH-1911 not all static resources being served from static.login.anosrep.org)

shane-tomlinson:
@jrgm, @lloyd - Can we close this since the bulk of the work and rely on #1911 (not all static resources served from static.login) for the cleanup?

shane-tomlinson:
@lloyd - is this work now done?

lloyd:
thanks for finding this cruft, shane dawg.

PR #1743, opened 2012-06-13
Change browserid.org urls to persona.org, introduce static.login.persona...

lloyd:
....org for all static resources.

when we merge this into dev, we'll want to update all of our dev urls.

when we deploy this on stage and prod, we'll want to update our main domain keypair to invalidate outstanding certificates.

see issue #1741 for the planned URL structure in each environment.


lloyd:
BLOCKER:  https://github.com/mozilla/connect-cachify/pull/15 must be merged and published, and we must move onto the new version to fix `//` in urls

example/primary/provision.html
```diff
@@ -5,7 +5,7 @@
 
 <html>
 <head>
-<script type="text/javascript" src="https://browserid.org/provisioning_api.js"></script>
+<script type="text/javascript" src="https://static.login.persona.org/provisioning_api.js"></script>
```
shane-tomlinson: In PRIMARY_PROTOCOL.md (https://github.com/mozilla/browserid/pull/1743/files#L3R272), the docs inform users to get their provisioning_api.js file from `https://login.persona.org/provisioning_api.js`, but here it is fetched from `https://static.login.persona.org/provisioning_api.js`.  Do we want the general public retrieving theirs from static.login.\* as well?

docs/PRIMARY_PROTOCOL.md
```diff
@@ -269,7 +269,7 @@ user has successfully authenticated with the primary.
 To support browsers without native BrowserID support, the
 authentication page should include a javascript shim, hosted at:
 
-    https://browserid.org/authentication_api.js
+    https://login.persona.org/authentication_api.js
```
shane-tomlinson, line 272: How is caching implemented on `provisioning_api.js` and `authentication_api.js`?  Are these etag only?  If there are updates, will changes be propagated out to users?

lib/configuration.js
```diff
@@ -59,8 +59,18 @@ var conf = module.exports = convict({
...
-    env: 'URL'
+    format: 'string = "https://login.persona.org"',
+    env: 'PUBLIC_URL'
+  },
+  public_static_url: {
+    doc: "The publically visible URL from which static resources are served",
+    format: 'string',
+    env: 'PUBLIC_STATIC_URL'
+  },
+  public_verifier_url: {
```
shane-tomlinson, line 70: further down in the file(L205), there is `verifier_url`, does that serve a different purpose from this?  (as an aside, there is also a browserid_url specified, is that different from public_url?)

lib/configuration.js
```diff
@@ -274,18 +297,25 @@ conf.validate();
...
 module.exports.performSubstitution = function(app) {
-  if (conf.get('public_url') != 'https://browserid.org') {
+  if (conf.get('public_url') != 'https://login.persona.org' ||
+      conf.get('public_static_url') != 'https://static.login.persona.org' ||
+      conf.get('public_verifier_url') != 'https://verifier.login.persona.org')
+  {
     app.use(postprocess(function(req, buffer) {
-      return buffer.toString().replace(new RegExp('https://browserid.org', 'g'), conf.get('public_url'));
+      return buffer.toString()
+        .replace(new RegExp('https://login.persona.org', 'g'), conf.get('public_url'))
```
shane-tomlinson, line 310: When load tests are being run on stage, this substitution will take place for every request.  Is there any performance penalty in creating the regular expressions for each substitution?  This used to be a problem in browsers when RegExps were used in tight loops (DOM lookups, etc).  If so, you can cache off the regular expressions so they are only created once.

lib/wsapi.js
```diff
@@ -42,8 +42,8 @@ const COOKIE_SECRET = secrets.hydrateSecret('browserid_cookie', config.get('var_
 var COOKIE_KEY = 'browserid_state';
 
 // to support testing of browserid, we'll add a hash fragment to the cookie name for
-// sites other than browserid.org.  This is to address a bug in IE, see issue #296
-if (config.get('public_url').indexOf('https://browserid.org') !== 0) {
+// sites other than login.persona.org.  This is to address a bug in IE, see issue #296
+if (config.get('public_url').indexOf('https://login.persona.org') !== 0) {
```
shane-tomlinson, line 46: Nice, that should keep the cookie from being sent for static resources.  ++

resources/static/shared/user.js
```diff
@@ -349,8 +349,8 @@ BrowserID.User = (function() {
         if(authInfo.authenticated) {
           persistEmailKeypair(email, "primary", authInfo.keypair, authInfo.cert,
             function() {
-              // We are getting an assertion for browserid.org.
-              User.getAssertion(email, "https://browserid.org", function(assertion) {
+              // We are getting an assertion for persona.org.
+              User.getAssertion(email, "https://login.persona.org", function(assertion) {
```
shane-tomlinson, line 359: @shane-tomlinson - test to make sure signup/signin/add email works with a primary user.

resources/views/relay.ejs
```diff
@@ -6,6 +6,6 @@
 <head>
   <meta charset="utf-8">
   <title>Browser ID</title>
-  <script type='text/javascript' src='https://browserid.org/relay/relay.js'></script>
+  <script type='text/javascript' src='https://static.login.persona.org/relay/relay.js'></script>
```
shane-tomlinson: @lloyd - If resources on static.login.\* have far future cache headers the relay is updated, changes will not be automatically downloaded by clients.  Relay is a simple file that likely will not change - so it should be safe to do here.

example/primary/provision.html
lloyd: no.  this is a bug in my port.  I will fix it.

docs/PRIMARY_PROTOCOL.md
lloyd, line 272: yes, these must be verified every time and served with etag.  I don't know if we want to do a temporary redirect to a permaurl on cdn, or not.

lib/configuration.js
lloyd, line 70: yes.  <process_name>_url are _internal_ urls, how different processes can reach each other.

public_<purpose>_url are _public_ urls.  things that will actually get served outside of our data center to users.

lib/configuration.js
lloyd, line 310: load testing doesn't hit static resources.  so zero affect.  there _is_ a performance penalty for this substituion, but when we implement a CDN in dev, stage, and prod, it will happen so rarely (once per deployment per resource) that it doesn't matter.  and never in prod.

resources/views/relay.ejs
lloyd: you're right.  the correct thing to do here is to use cachify and embed perma-urls into the html document.  that way we remove an HTTP transaction in the cache-full case for IE.

Issue #1248, opened 2012-03-02
'/manage|/users|/primaries' in browserid/views.js redirect to unreachable internal hostname in stage/prod                                               


jrgm:
A very minor issue but these URL's redirect to the internal hostname (e.g., web1) of the host that handled the request.                     

Could either just remove the redirects since they don't work anyways, or fix them to use the load-balanced public hostname of the site. (I suppose option #3 is to have zeus handle them, but does anyone ever hit these URL anymore?).

jrgm:
Ping. With the router in place on stage train-2012.06.08, these URLs redirect to http://127.0.0.1:PORT/ which is even uglier than redirecting to a blackhole unreachable 10.x.x.x IP. 

I vote for a zeus rule to redirect them all to '/' for the next prod push, and then just remove the redirect code entirely (these URL haven't been reachable for months now). 

/cc @zaach, @jbonacci

lloyd:
reopened, the remaining work is the production redirect.

jrgm:
I know I'm the one who suggested to remove these redirects and add a zeus rule to handle these redirects. 

So great, the futile redirects in the code are removed. 

I was about to flie the ops bug for the zeus rule, and I know I suggested it, but I don't think it's worth it to add that overhead to zeus for URL that have been dead for months now. I'll close this bug tomorrow unless anyone feels strongly about adding the redirect /cc @fetep

lloyd:
+1 for close, I don't feel strongly about maintaining redirects.  @vthunder @callahad ?

callahad:
I don't feel strongly either...

jrgm:
Ok, closing. These will now be 404 if they are ever actually requested, which I doubt.

jrgm:
verified on stage train-2012.06.22

PR #1746, opened 2012-06-14
remove redirects for /manage, /users and /primaries - issue #1248

lloyd:


lib/browserid/views.js
```diff
@@ -179,11 +179,6 @@ exports.setup = function(app) {
 
   // REDIRECTS
   const REDIRECTS = {
-    "/manage": "/",
-    "/users": "/",
-    "/users/": "/",
-    "/primaries" : "/developers",
-    "/primaries/" : "/developers",
     "/developers" : "https://github.com/mozilla/browserid/wiki/How-to-Use-BrowserID-on-Your-Site"
```
shane-tomlinson: @lloyd - As a further change - the GitHub link is now empty content that displays a link to MDN - might as well update here as well - https://developer.mozilla.org/en/BrowserID/Quick_Setup is the MDN link.

Issue #1742, opened 2012-06-13
Run images through a lossless compressor.


shane-tomlinson:
Our background images take up more room than they should, Google Pagespeed says we can do better.

PR #1747, opened 2012-06-14
Run images through PageSpeed's lossless image compressor to shrink them.

shane-tomlinson:
issue #1742


lloyd:
I see no difference between these images other than size.  r+

PR #1753, opened 2012-06-14
Bigtent with dev

ozten:
@shane-tomlinson rules and did this rebase as well as fixed the unit tests.

Moving Pull Request #1739 party here...


config/local.json
```diff
@@ -11,6 +11,11 @@
   },
   "express_log_format": "dev_bid",
   "email_to_console": true,
-  "env": "local",
-  "kpi_backend_sample_rate": 1.0
+  "kpi_backend_sample_rate": 1.0,
+  "bigtent": {
```
ozten, line 16: @lloyd - How do I give this default information to `npm test`? I tried

```
process.env['BIGTENT'] = JSON.stringify({"gmail.com": "https://google.login.persona.org"})
```

But that doesn't pass muster. Is there a JSON file that is sourced for vows tests?

ozten:
@shane-tomlinson besides a default config for vows tests... what else is left before this can land on dev?

config/local.json
ozten, line 16: I've commited {} as the default config (in lib/configuration), so that there is a default, disabled bigtent.

shane-tomlinson, line 16: @ozten - Thanks for the update, I can get the server started but when I enter a big tent supported email into the dialog and am redirected to the BigTent IdP, I see:

```
Invalid email input to service function [:email]string-
```

shane-tomlinson, line 16: @ozten - there is an error in BigTent's routes.js.

On line 15,   

```
app.get('/proxy/:email', function(req, res, next){
```

Then you get the email on line 21 using: 

```
     var service = proxy.service(req.params.email);
```

emails are appended on to the URL as the email GET parameter, we do NOT search for :email and replace as this route is looking for.  The second gotcha is to avoid XSS, we encodeURIComponent each item, so `@` is converted to `%40`

```
/proxy/:email?email=set117%40gmail.com&return_to=http%3A%2F%2F192.168.1.88%3A10002%2Fsign_in%23ADD_EMAIL%3Dset117%40gmail.com
```

ozten, line 16: Sounds like this new branch is missing a patch. BrowserID should be using well known to find the auth url. Up until this week, there was a special rule that address_info used a hardcoded /proxy/:email, but that is no longer the case.

Making sure your bigtent is up to date...
If you do curl -v -k https://yourhostname:yourport//.well-known/browserid you should see 

```
"authentication": "/authentication",
```

If not, can you update and restart bigtent?

lib/wsapi/address_info.js
```diff
@@ -48,6 +53,20 @@ exports.process = function(req, res) {
     if (urls) {
       urls.type = 'primary';
       res.json(urls);
+    } else if (proxyidp.isProxyIdP(email)) {
+      var bigtent = proxyidp.bigtentHost(email);
+      primary.checkSupport(bigtent, function(err, urls, publicKey, delegates) {
+        if (bt_done) {
```
shane-tomlinson, line 60: Does the callback get called multiple times and must be aborted?  I'm not clear what this does.

lib/verifier/certassertion.js
```diff
@@ -134,7 +135,9 @@ function verify(assertion, audience, successCB, errorCB) {
       // that the email's domain delegated authority to the issuer
       var domainFromEmail = principal.email.replace(/^.*@/, '');
 
-      if (ultimateIssuer != HOSTNAME && ultimateIssuer !== domainFromEmail)
+      if (ultimateIssuer != HOSTNAME &&
+          ultimateIssuer !== domainFromEmail &&
+          ! proxyidp.isProxyIdP(principal.email))
```
shane-tomlinson, line 144: Does this leave room for an assertion for a domain labelled as a proxyIDP to come from a source that is neither BigTent, the HOSTNAME or the domainFromEmail?  

Say the ultimateIssuer is `nefarious.com` (somebody there created an assertion and passed it to us), the HOSTNAME is `persona.org` and the domainFromEmail is `gmail.com` - does that mean the assertion is validated?

lib/wsapi/address_info.js
ozten, line 60: It should not, but there was a production issue where it did have infinite recursion. 

That production failsafe `done` hasn't been cleaned up, so I reproduced the same failsafe. It's safe to leave it in and adds a second level of protection.

lib/verifier/certassertion.js
ozten, line 144: This is a great point. Is there an existing qunit or vows test, where we can test this?

Also [BigTent Issue #34](https://github.com/mozilla/browserid-bigtent/issues/34) was declared a blocker yesterday and relates to this addresses.
Gist: BigTent will chain it's certs with a BrowserID signed cert. Without this, bigtent would bleed into the protocol during verification.

config/local.json
```diff
@@ -11,6 +11,11 @@
   },
   "express_log_format": "dev_bid",
   "email_to_console": true,
-  "env": "local",
```
ozten, line 15: Why do we drop local? (Not that I know what env did)

shane-tomlinson, line 15: @ozten - it looks like it was moved to just below the "bigtent" declaration.

ozten, line 15: You can ignore me :)

I was confused by a link n IRC, which I had already responded too and then the link loaded this line (instead of what you were actually talking about)

lib/verifier/certassertion.js
shane-tomlinson, line 144: @ozten - I'm not seeing any obvious tests for this condition in tests/verifer-test.js - a test like this would have to be handled by vows.  The QUnit tests never interface with the backend, all the network requests there are stubbed in.  

@callahad - I hear you are heading this up now, could you explore this?  If there is not a vows test for this condition, we should get one.

fetep:
If I understand this change correctly, it's going to break anyone doing verification on their own.  I'd much prefer we do the chained cert support so we don't have to special case bigtent domains in every verifier implementation (thinking specifically of tokenserver, here).

benadida:
Broad design point: why tag users as having a new type, specifically proxyidp? Why are these not simply primaries with assumed authority delegations? This would make the code less repetitive and thus less likely to bitrot. I'd prefer we do it that way unless there's a clear reason why not.

shane-tomlinson:
@benadida - @ozten and I both want this change as well.  @ozten can better explain its original intent, I subsequently made use of this knowledge to skip the "you are now being redirected to xyz.com to verify your identity" screen. Xyz.com for big tent domains displays the big tent @ Mozilla server instead of the IdP the user is expecting. With the revised TOS/PP flow, we can remove this screen entirely. The downside to doing so is that the screen gives the user information that they are being redirected as well as offer them an opportunity to hit cancel and change an erroneous address.

Perhaps we could either take the screen out universally and inform the user they will be redirected from the authentication screen, but inform them in a way that does not specify the URL.

benadida:
@shane-tomlinson maybe we should separate those two concerns: if you want to special-case the UI when the IdP is *.login.persona.org, that's totally ok, of course. But the backend logic should consider those two the same, I think, unless there's good reason.

shane-tomlinson:
@benadida - I feel like we should not be special casing on the front end either, even for *.login.persona.org.  BigTent user's are redirected away from Persona without any warning whatsoever, an action users may not be expecting.

ozten:
@shane-tomlinson can you remove proxyidp from the BrowserID frontend code?

I'll remove it from the BrowserID backend code, once [Cert Chaining](https://github.com/mozilla/browserid-bigtent/issues/34) is implemented.

Thanks!

benadida:
@shane-tomlinson @ozten is this ready for re-review, or not quite yet?

lib/db/json.js
```diff
@@ -158,7 +158,7 @@ exports.userOwnsEmail = function(uid, email, cb) {
 
 function addEmailToAccount(userID, email, type, cb) {
   // validate 'type' isn't bogus
-  if ([ 'secondary', 'primary' ].indexOf(type) === -1) {
+  if ([ 'secondary', 'primary', 'proxyidp' ].indexOf(type) === -1) {
```
lloyd, line 177: why did this make it into the database layer?

ozten:
@benadida - @lloyd will tackle the last backend tweaks to compliment @shane-tomlinson's fine work.

He will also have a fresh pair of eyes for reviewing the patch and may make other tweaks.

lloyd:
+10 to @benadida's suggestion "simply primaries with assumed authority delegations"  

In reading the code I cannot see why we would not do it this way.

lib/configuration.js
```diff
@@ -210,6 +210,11 @@ var conf = module.exports = convict({
     format: 'string?',
     env: 'BROWSERID_URL'
   },
+  bigtent: {
```
lloyd, line 213: is it useful to have this codename throughout our codebase?  would something more boring, but clearer be better?  `shimmed_idps` `shimmed_primaries` ?

this is a nit, it should not block merge.

ozten, line 213: Feel free to change.

It's been useful while grepping, since we already use primary, proxy, shim, and idp for other features.

resources/static/dialog/js/misc/state.js
```diff
@@ -119,15 +117,15 @@ BrowserID.State = (function() {
        */
       info = _.extend({ email: self.newUserEmail || self.addEmailEmail || self.resetPasswordEmail }, info);
 
-      if(self.newUserEmail) {
+      if (self.newUserEmail) {
```
lloyd: this is a nit.  whitespace changes inside feature branches makes merges harder.  especially for something as big as this, we should avoid them.

lloyd:
I want to take ownership of this PR, do a review, test, and merge it down. 

But it needs to have latest dev merged.  The merge is too complex for me, as I lack an understanding of the affected frontend code.  I tried for 45 minutes, but gave up because I think I would have introduced subtle bugs that would have wasted a lot of our time.

@shane-tomlinson - can you perform the merge? (not too bad:  9 files changed, 312 insertions(+), 43 deletions(-))

Once done, I'll review all feedback, update as needed, and get this merged into dev.

My goal would be to land this code without regressing our basic feature set and with the bigtent redirects turned off.  I will focus 90% of my testing and review at this.

Once landed, we avoid the cost of maintaining the branch and can issue much smaller pull requests over the next week to refine the support.

lloyd:
@shane-tomlinson assigned to you, pass it back to me once dev is merged in.

If we cannot get this patch merged and polished within the next week, we slip bigtent from beta.  clearly not the end of the world, but we're giving it an earnest run.

lib/verifier/certassertion.js
shane-tomlinson, line 144: @lloyd, @callahad, @ozten - before this is merged, I am still concerned about this line.

resources/static/dialog/js/misc/state.js
shane-tomlinson: @lloyd - all those state.js changes should have never been part of the PR.  I reverted state to be the same as is found in dev with one additional line to allow the unit tests to complete.  A separate state.js refactor PR (#1975) has been issued.

lib/wsapi/address_info.js
```diff
@@ -48,6 +53,20 @@ exports.process = function(req, res) {
...
+          return;
+        }
+        bt_done = true;
+        if (err || ! urls) {
+          logger.warn('error checking BigTent for IdP details: ' + err);
+          return httputils.serverError(res, "BigTent unavailable");
+        }
+        urls.type = 'proxyidp';
```
shane-tomlinson, line 68: address_info should still return a type of "primary" here and not let this bleed through to the front end.

lib/wsapi/address_info.js
```diff
@@ -48,6 +53,20 @@ exports.process = function(req, res) {
...
+    } else if (proxyidp.isProxyIdP(email)) {
+      var bigtent = proxyidp.bigtentHost(email);
+      primary.checkSupport(bigtent, function(err, urls, publicKey, delegates) {
+        if (bt_done) {
+          return;
+        }
+        bt_done = true;
+        if (err || ! urls) {
+          logger.warn('error checking BigTent for IdP details: ' + err);
+          return httputils.serverError(res, "BigTent unavailable");
```
shane-tomlinson, line 66: I do not like throwing an error here, it is a terrible user experience.  If there is a problem with BigTent, the address should be treated as a secondary.  Right now if the BigTent server is unavailable for any reason (network outage, etc), the user is presented with an ugly error screen instead of gracefully degrading to a secondary.

lib/wsapi/auth_with_assertion.js
```diff
@@ -38,7 +38,8 @@ exports.process = function(req, res) {
       if (err) return wsapi.databaseDown(res, err);
 
       // if this is a known primary email, authenticate the user and we're done!
-      if (type === 'primary') {
+      if (type === 'primary' ||
+          type === 'proxyidp') {
```
shane-tomlinson, line 45: proxyidp should no longer be an assertion type, this should be chained certs instead.

lib/verifier/certassertion.js
ozten, line 144: Changing this line is the point of the TBD cert chaining work. @lloyd and I are meeting with Ben soon.

To land the patch, we can revert this line to production code.

lib/wsapi/address_info.js
ozten, line 66: I don't think we can degrade to secondary. That is a horrible UX.

Sometimes I'm asked for my gmail password by google.

Then, I'm asked to create a password.

Then, I'm asked for my gmail password by google a bunch.

(second network issue or whatever) Then I'm prompted with a different screen where I try my gmail password and it doesn't work.

Also, by collecting a password up front in our new flows, is that even technically possible?

benadida, line 66: @shane-tomlinson I don't think we can have a graceful degradation to secondary once primaries are in place, including BigTent. It would mean that secondaries can always trump primaries for trust, which is the opposite of what we want to do. If Google is down, then you can't auth with your google email. I think that's okay.

shane-tomlinson, line 66: @ozten - I politely disagree.  Having a fallback mechanism for the user in case of the Persona server being unable to connect with the BigTent server is WAY better than just throwing up an error screen.  A fallback isn't as clean/easy as I initially thought because of the variety of cases, but it is workable, and I think it would offer a better overall experience.

If address-info returns primary for users who are trying to add an email to their Persona account - all is good, the user may have to set a password on their account, but nothing worse. The next time a user accesses their Persona account, if BigTent is available, it takes over.

If address-info returns primary for users who have previously registered a BigTent address and are trying to authenticate while BigTent is unavailable - if the user has a password associated with their account, perhaps let them use it.  The user _has_ verified they own the account, though whether we want to honor that verification in this situation feels slightly dirty.  If the user does NOT have a password associated with the account, that's the trickiest of all.  Perhaps send them down the secondary flow to create a new account or if they have a session already, add the email to an existing account?

With the reset-password flow, we _do_ have the notion of an unverified email, so we can easily send them down the "go verify your ownership of this address" flow.  Perhaps that could work here.

It's not an easy change, but there is room for improvement over a hard error.

lloyd, line 66: This discussions says to me that when we release bigtent, we must have extremely high confidence in it.  There is no turning back, and errors affect 60% of our users.

--lloyd

shane-tomlinson, line 66: @benadida - how is a primary only account handled currently if there is no way to reach the IdP?  If a user is trying to sign in to Persona using an address that was at one point associated with an IdP, that IdP is _not_ in the cache of IdP auth/verify URLs, and that IdP is unavailable currently, what happens - does address_info return "secondary" or "primary" as the result?  

It looks like from the lib/wsapi/address_info.js and lib/primary.js, IF primary.checkSupport cannot get URLS and the email is known, it automatically returns secondary.  How is this case different?

shane-tomlinson, line 66: @lloyd - I agree, we also have to ensure that the BigTent servers are always available from the Persona servers, especially if they are run on distinct machines.  I'm less worried about the BigTent code than about network reliability.

benadida, line 66: so, this is why I would like us to think of BigTent as a delegated IdP. If you can't reach BigTent, then you fall back to secondary just like we do today. But if you _can_ reach BigTent, then providing a path by which a successfully reached IdP re-triggers secondary is, I think, not something we should do. If BigTent is reachable, and Google is down, then users can't log in using google email. I think that's okay.

We have to be very careful not to explode the number of paths for the user, if only for security reasons.

In general, our story for primary-reverting-to-secondary is likely something that will need a bit of work. I don't think we'll be able to predict every case ahead of time, though we can try.

shane-tomlinson, line 66: @benadida - I agree 100% with what you are saying, but that is not what we are doing here.  If you cannot reach BigTent and Google is not yet its own IdP, we currently have a hard failure mode.

The above digging around does bring up a corner case that I do not believe we are handling (not just for BigTent).

I have not verified this yet with a test, but from a code examination I believe a failing corner case is possible.  If a user creates an account using a primary email, their account has no password.  If their IdP is not available when the call to address_info is made, address info will return with "type: secondary, known: true".  The front end will then show the password field for the user to enter their password.  The account will have no password.  The user will be stuck in limbo without knowing why.

benadida, line 66: if you cannot reach BigTent, then I think it should behave like an absent IdP, and that means back to secondary. But if you _can_ reach BigTent and Google itself is down, I think it's okay for us to say that you can't log in with your google email at that point in time. Falling back to secondary then would likely make things really odd (and would seriously complicate BigTent to the point of making things a bit too brittle.)

(As for other corner cases... yes I'm sure we'll see a bunch :)

callahad, line 66: So that's just staying BigTent should act like a normal IdP? If there's a `well-known`, we're going to send you there, period. If there's not a `well-known`, we're going to send you to our fallback. Right?

ozten:
Neither dev nor stage are happy. I don't think this should block this branch landing.

I fixed the bigtent part of https://github.com/mozilla/browserid-certifier/issues/7

Filed https://github.com/mozilla/browserid/issues/2014 to capture what I'm seeing in dev.

shane-tomlinson:
@ozten - I am still seeing the problem in the certifier using yahoo addresses - http://pastebin.mozilla.org/1702393

lloyd:
closing this down, we're going to try a much more minimal approach: https://github.com/mozilla/browserid/pull/2019

This branch is not going anywhere, so we can cherry pick in stuff that's needed.

Issue #1518, opened 2012-05-01
"unknown path" displayed for "/about", "/TOS", and "/privacy"


shane-tomlinson:
Link to #1514

callahad:
(Not user-visible; these are errors that show up in the console)

jrgm:
verified train-2012.07.06 on awsbox deploy

PR #1754, opened 2012-06-14
Prevent the "unknown path" console message from being displayed on the "/tos" and "/privacy" pages.

shane-tomlinson:
To test:
1) Open your dev console
2) go to the main site's /tos and /privacy pages.
3) Check for "unknown path" in the console.

issue #1518


resources/static/pages/start.js
```diff
@@ -154,6 +154,9 @@ $(function() {
     else if (path === "/about") {
       var module = bid.about.create();
       module.start({});
+    } else if(path === "/tos" || path === "/privacy") {
```
seanmonstar: while I don't like have `else` on a new line, this commit breaks the style of all the rest of the `else if` clauses. It should also have a space between `if` and `(path`.

``` js
}
else if (path ...
```

shane-tomlinson: @seanmonstar - updated.

Issue #1757, opened 2012-06-14
Create new process for serving static or "cacheable" resources


zaach:
We'll offload static files and rendered views from the browserid process and serve them from a new "static" process. This allows the browserid process to deal with white-listed WSAPI requests only. Eventually, static resources will be served from CDN.

lloyd:
a strong +1

One note, I've been informed by QA and prod ops in no uncertain terms that before we add a new process we must consult them with an architecture overview, ideally before a line of code is written.  I will do this now, retro-actively for router.  For the static file server I think we need to respect their wishes and write it up.  I can do this, or you can @zaach...  Now /me gets to writing about router.

zaach:
Short summary:

> We have created a new "static" process for serving the main site and associated static resources in order to reserve the browserid process for more dynamic, WSAPI requests. The router will be reconfigured to forward only WSAPI requests to the browserid process. Requests for static resources and Persona site pages will now be forwarded to the static process. The heartbeat of the router must be modified to check if the static server is responding, in addition to the browserid process.

webheads
    • /usr/bin/node bin/static

webheads: /var/log/browserid/
    • static.log
    • static-metrics.json

zaach:
More [here](https://etherpad.mozilla.org/static-process).

lloyd:
open?  who is open?

Issue #1756, opened 2012-06-14
Old version of nodemailer introduces global `util`, causes buggy side effects


zaach:
Bitten again by globals! The `layout.ejs` template file was using a reference to a global instance of `util`, which broke things when resources were served from a new static process rather than the browserid process. I tracked down the global `util` to the nodemailer module.

We should update our version of nodemailer (0.1.18) and explicitly pass `util` as a value to the views. The problem seems to be fixed in nodemailer version [0.1.24](https://github.com/andris9/Nodemailer/tree/e834625a88e3e6f8986a1df5144809ca59e556d7).

lloyd:
verified mail is still delivered in dev environment.  No major regression.  yay.

jrgm:
I haven't been experiencing any email issues with stage so verifying for train-2012.07.06

PR #1758, opened 2012-06-14
update nodemailer and pass a local instance of `util` to views - issue17...

zaach:
...56

see issue #1756

Related: #1757


shane-tomlinson:
@zaach - Can you update this pull request with the latest dev so that we can merge it next week?

PR #1762, opened 2012-06-15
in the communication iframe, initialize ALL local storage values immedia...

lloyd:
...tely at startup to fix IE8 localStorage sync problems.  issue #1637

### DIFF
--- a/resources/static/shared/storage.js
+++ b/resources/static/shared/storage.js
@@ -62,6 +62,7 @@ BrowserID.Storage = (function() {
     _.each({
       emailToUserID: {},
       emails: {},
+      interaction_data: {},
       loggedIn: {},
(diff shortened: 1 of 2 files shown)


resources/static/shared/storage.js
```diff
@@ -49,11 +49,33 @@ BrowserID.Storage = (function() {
...
+      emailToUserID: {},
+      emails: {},
+      loggedIn: {},
+      main_site: {},
+      managePage: {},
+      returnTo: null,
+      siteInfo: {},
+      stagedOnBehalfOf: null,
+      usersComputer: {}
+    }, function(defaultVal, key) {
```
shane-tomlinson, line 73: Also add "interaction_data" which defaults to {}

Issue #1761, opened 2012-06-15
disable siteLogo from HTTP served sites.


shane-tomlinson:

lloyd:
siteLogo is disabled on 123done until we can get SSL certificates:

https://github.com/mozilla/123done/commit/3bb0d38335dd0fed80b76b92864bbc265c8c515c

jrgm:
verified against stage train-2012.06.22 with an awsbox converted to an rp. Also tried getting something not nice done with the params but didn't find a way, and tried out an oversize image, which was handled as expected by the CSS.

PR #1763, opened 2012-06-15
Only allow siteLogo's to be served from https RPs.

shane-tomlinson:
issue #1761


lloyd:
r+ on code.  Here's what a developer will see:

Need SSL

I think this is an incredibly discoverable error message, and it's important to fail fast and fail loudly during initial integration.  merging.

Issue #1497, opened 2012-04-24
invalid verification link error message is does not match other error messages


shane-tomlinson:
https://skitch.com/set117/8ixsu/box-simpleshare

Should we make this the same style as other error messages?

skinny97214:
Good catch. Yes, please make it match.

lloyd:
kinda related to GH-218

shane-tomlinson:
link to #1734

jrgm:
veriified stage train-2012.06.22 - re-using a token, or using an invalid token bring up the new style error message.

Issue #218, opened 2011-08-24
validating an email a second time causes undecipherable error.


shane-tomlinson:

shane-tomlinson:
6c9ff3ca7dcf1679be6de13f9a1ac93f7d386bd3 changes the error message to ask "Have you previously verified this address?"  

What is the possibility of the backend keeping track of these tokens and their status, if the token is already validated when the call is made to /wsapi/prove_email_ownership', we pass back a status saying "already validated" or something similar so that we can display an appropriate message to the user.

lloyd:
The nice thing about how things work right now is we can purge all record of the secret after verification, or periodically (purge tokens older than 2 days?).  I like the performance/scalability properties of the current implementation.

So right now we have two possible states that a "secret token" can be in: known (and valid) or not.

If we wanted to have three states: verified, unknown, unverified - then we would increase database requirements slightly.  Given we send about 4 emails per second at 1M active, and each of those emails requires a row, we'd have about a 400k row table if we always kept tokens around for 2 days.

That's manageable, but interesting enough that I'd like to get feedback from UX as to how important the changed feedback is.  @clarkbw ?

clarkbw:
If I'm understanding correctly the total experience seems like it's going to be the same either way, we need them to re-validate (if an old token was used) or they are all good to go (clicking on the link a second time).  

The difference there being that we could tell people a token was already used versus not knowing the difference between used or incorrect.

## incorrect token case
- link gets garbled by the users email client somehow

_Our design goal is mostly to help them verify via a new link or something else_
- landing page says "Hey clarkbw@example.com, we're still waiting for your confirmation.  Do you want us to <a>resend it</a>?"
  - Alternates are to have a text input where people could copy & paste the code sent in the email if the link isn't working, requires some additional text in the email

## previously used token
- user clicked and verified their email previously
- in trying to find their browser id information the user re-clicks the original verify email

_design goal is to get them to verify an email if they need to do that or get them to the manage page_

### needs to verify an email
- works similar to the incorrect token case

### all user emails are verified
- redirect to the main page with a quick flash of "you're already verified and good to go"

shane-tomlinson:
@skinny97214 - since we are going gang-busters on the user completion flow, want to weigh in on this?  Action on this has gotten lost.

PR #1765, opened 2012-06-15
Fix the flashes that occur when verifying an email address.

shane-tomlinson:
- Only show info once the account has been verified.
- Unify the styles for the "has this address already been registered" error with standard error dialog.
- Remove all the dead code that was left over from issue #1000 merge.

issue #1734
issue #1497
issue #218

### DIFF
--- a/resources/static/css/style.css
+++ b/resources/static/css/style.css
@@ -515,28 +515,8 @@ button.delete:active {
   padding: 0;
 }
 
-#signUpForm .red {
-  color: red;
-}
-
 #signUpForm .submit {
   height: 28px;
-}
-
-#signUpForm .error {
-  margin-top: 20px;
-  color: red;
-  background-color: rgba(255,0,0,0.25);
-  -ms-filter:progid:DXImageTransform.Microsoft.gradient(startColorstr=#3fff0000,endColorstr=#3fff0000);
-  filter:progid:DXImageTransform.Microsoft.gradient(startColorstr=#3fff0000,endColorstr=#3fff0000);
-  zoom: 1;
-  padding: 5px;
-  line-height: 16px;
-
-  -webkit-border-radius: 3px;
-     -moz-border-radius: 3px;
-       -o-border-radius: 3px;
-          border-radius: 3px;
 }
 
 #signUpForm > .siteinfo {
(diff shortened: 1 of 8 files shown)


resources/static/css/style.css
```diff
@@ -515,30 +515,10 @@ button.delete:active {
...
 
-#signUpForm .red {
-  color: red;
-}
-
 #signUpForm .submit {
   height: 28px;
 }
 
-#signUpForm .error {
```
shane-tomlinson, line 526: These styles were for the old style error messages.  The old style messages have been replaced with the newer unified error message style.

resources/static/pages/verify_secondary_address.js
```diff
@@ -55,38 +49,37 @@ BrowserID.verifySecondaryAddress = (function() {
 
   function submit(oncomplete) {
     var pass = dom.getInner("#password") || undefined,
-        vpass = dom.getInner("#vpassword") || undefined,
```
shane-tomlinson, line 58: This was transition code for issue #1000.  No longer needed.

resources/static/pages/verify_secondary_address.js
```diff
@@ -103,36 +96,30 @@ BrowserID.verifySecondaryAddress = (function() {
...
     user.tokenInfo(token, function(info) {
       if (info) {
         redirectTo = info.returnTo;
         email = info.email;
         showRegistrationInfo(info);
 
-        needsPassword = info.needs_password;
         mustAuth = info.must_auth;
-
-        if (needsPassword) {
```
shane-tomlinson, line 115: Transition code, no longer needed.

resources/static/css/style.css
lloyd, line 526: thanks for the pre-emptive explanation

resources/static/pages/verify_secondary_address.js
lloyd, line 58: ditto

lloyd:
PRs which address lotsa things at once are hard to review well.  the actual code for the flash here is small, and I'm comfortable with it all.  doing some quick testing, running unit tests, and if that all works, merging.

lloyd:
all manual testing looks great, no flash seen, no affect to permutations of new user / email addition in different browser / same browser.

One blocker found, frontend tests fail: pages/verify_secondary_address: valid token, no password necessary - verify user and show site info (1, 4, 5)

lloyd:
travis agrees that this is a blocker: http://travis-ci.org/#!/mozilla/browserid/jobs/1627308

lloyd:
r+ - better: http://travis-ci.org/#!/mozilla/browserid/builds/1628161

PR #1768, opened 2012-06-15
New /about page

sawyerh:
Implemented the new /about (How It Works) page, CSS, and about.js. This is also mobile-friendly.


resources/views/about.ejs
```diff
@@ -3,38 +3,47 @@
...
+            </article>
+
+            <article class="blurb flexible">
+                <div class="graphic first">
+                    <img src="i/flexible-graphic.png" alt="Use multiple email addresses">
+                </div>
+
+                <div class="info">
+                    <h1>Persona is flexible</h1>
+                    <p>Within Persona, your identity is your email address. You can use as many email addresses as you want, but you still only need one password.</p>
```
callahad, line 27: "You can use as many email addresses as you want, but you still only need one password" isn't totally accurate -- Gmail, Hotmail, Yahoo, and folks with addresses at native Identity Providers will all need to have one password per email address. I'm not sure how to say that in a user-friendly way, though :)

resources/views/about.ejs
```diff
@@ -3,38 +3,47 @@
...
+                </div>
+            </article>
+        </section>
+
+        <section class="privacy">
+            <h2 class="title">Real privacy.</h2>
+
+            <article class="blurb half first" style="min-height: 195px; ">
+                <h1>Persona is proudly non-profit for you</h1>
+                <p>Persona is developed by Mozilla, a not-for-profit company trusted throughout the Web community. Our goal is to create technologies that balance an open Web platform with people’s privacy.</p>
```
callahad, line 37: There's a balance between an open web and privacy? That implies giving up privacy would give me a more open web, or that a closed web would be the most private of all. This same wording showed up in the Persona Copy draft that Crystal mentioned on the list last week. Should we review that copy before merging it, or merge now, adjust later?

sawyerh, line 37: You'll want to ask Crystal about this. This was the approved copy that I was given :)

resources/static/css/m.css
```diff
@@ -54,6 +54,44 @@
...
+    width: 100%;
+    float: none;
+    min-height: 0 !important;
+  }
+  .blurb.half.first {
+    margin-right: 0;
+  }
+
+  .blurb {
+    display: -moz-box;
```
shane-tomlinson: Flex Box is not yet supported by IE or Opera (http://caniuse.com/#feat=flexbox), but the display still looks reasonable.

resources/static/pages/about.js
```diff
@@ -0,0 +1,31 @@
+/*globals BrowserID:true, $:true*/
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function($) {
+
+  // Homepage half blurbs should be same height when they're next to eachother
+  $(window).load(function(){
```
shane-tomlinson: I would like this to load the same way we load other page's code, using the code in start.js.

resources/views/about.ejs
```diff
@@ -3,38 +3,47 @@
...
+                <h1>Persona preserves your privacy</h1>
+                <p>Persona does not track your activity around the Web. It creates a wall between signing you in and what you do once you’re there. The history of what sites you visit is stored only on your own computer.</p>
+            </article>
+        </section>
+
+        <a href="https://developer.mozilla.org/en/BrowserID/Quick_Setup" class="developers"><img src="i/developers-link.png" alt="Persona for developers"><span>Implement Persona on your site </span>Developer guides and API documentation</a>
+    </div><!-- #dashboard -->
 </div>
 
+<%- cachify_js('/pages/about.js') %>
```
shane-tomlinson: Instead of doing this, we are going to want to include this in our static resources list so we minimize the number of requests.

resources/static/css/style.css
```diff
@@ -659,3 +591,149 @@ footer ul li:first-child a:hover {
...
+  padding-bottom: 30px;
+  margin: 0;
+}
+
+.blurb, a.developers {
+  -webkit-box-shadow: 0 1px 0 rgba(255, 255, 255, 0.13);
+  -moz-box-shadow: 0 1px 0 rgba(255, 255, 255, 0.13);
+  -ms-box-shadow: 0 1px 0 rgba(255, 255, 255, 0.13);
+  box-shadow: 0 1px 0 rgba(255, 255, 255, 0.13);
+  background: rgba(13, 28, 41, 0.1);
```
shane-tomlinson, line 614: A default background color should be added for IE8 or else the blurbs are not highlighted from the page.  IE8 without highlighting

resources/static/css/style.css
```diff
@@ -659,3 +591,149 @@ footer ul li:first-child a:hover {
...
+  font-size: 14px;
+  color: #fff;
+}
+
+.blurb {
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
+  -ms-box-sizing: border-box;
+  -o-box-sizing: border-box;
+  box-sizing: border-box;
```
shane-tomlinson: border-box is already set on *

resources/static/css/style.css
```diff
@@ -659,3 +591,149 @@ footer ul li:first-child a:hover {
...
+  zoom: 1;
+  margin-top: 30px;
+  padding: 30px;
```
shane-tomlinson, line 622: The blurb with the Persona Dialog could use some bottom margin/padding around the image.  Image needs bottom-margin

resources/static/css/style.css
```diff
@@ -659,3 +591,149 @@ footer ul li:first-child a:hover {
...
+}
+.blurb.half.first {
+  margin-right: 4%;
+}
+.blurb .info, .blurb .graphic {
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
+  -ms-box-sizing: border-box;
+  -o-box-sizing: border-box;
+  box-sizing: border-box;
```
shane-tomlinson: same as above, we have box-sizing: border-box on *

shane-tomlinson:
@sawyerh - Thanks for this pull request, it is certainly going to make things easier.  Great work on this so far, and thanks for removing the old about page's CSS.  Other than the comments I have already made, Firefox Mobile (both current and Beta) overflow the screen.

Fennec Overflow

On mobile in general, there needs to be a bit of cleanup.  We have a separate stylesheet for mobile devices - m.css.

iOS titles
iOS Margins

resources/views/about.ejs
callahad, line 37: Will do -- thanks for the awesome pull request!

resources/static/css/style.css
```diff
@@ -659,3 +591,149 @@ footer ul li:first-child a:hover {
...
+  -moz-transition: all 300ms ease;
+  -ms-transition: all 300ms ease;
+  transition: all 300ms ease;
+  display: block;
+  padding: 13px 15px;
+  line-height: 1.4;
+  text-align: center;
+}
+a.developers:hover {
+  background: rgba(13, 28, 41, 0.2);
```
shane-tomlinson, line 713: Should a default background be set here for IE?

resources/static/css/style.css
sawyerh, line 622: This is part of the design

sawyerh:
@shane-tomlinson The issue with the overflowing on Firefox Mobile is related to the use of `display: box`. It seems it doesn't respect the width value and also breaks the use of `border-box`. I couldn't find a way around this so I've removed the -moz- prefixed version of this for now (works fine in Webkit).

resources/static/css/style.css
sawyerh, line 713: Fixed in 77441f1

resources/static/css/style.css
sawyerh: Fixed in 77441f1

resources/static/css/style.css
sawyerh, line 614: Fixed in 77441f1

jrgm:
looks ok on mainstream desktop and iphone/safari, android2/{stock,ff}, android4/{stock,ff,chrome}. ie8 has defects noted in GH-1951 which is not part of train-2012.07.06

Issue #1770, opened 2012-06-18
Fix load_gen to correctly send args to /verify


jrgm:
I realized that the load_gen calls to /verify were ~silently failing because they were sending an assertion without the certs (correct my terminology if that's wrong). 

Sorry for having missed this earlier. load_gen itself was reporting all success due to a bug that cancelled out the first bug. It was just some chirping in the metrics log that led me to this. 

I have a PR to fix this (er, I think it fixes it). Also while I was there, I fixed the trivial issue GH-1278.

lloyd:
@jrgm please verify things are as you expect after the merge and reopen if anything needs attention!

jrgm:
verified

Issue #1278, opened 2012-03-08
trivial error formatting throw message in lib/load_gen/common.js:48


jrgm:
There's a extra '+' in this statement at lib/load_gen/common.js:48

```
        if (resp.code !== 200) throw "non-200 status: " + resp.code +
          + " - " + resp.body;
```

-> "can't certify key: non-200 status: 500NaNServer Error: can't contact keysigner"

Harmless really. Maybe wontfix.

lloyd:
thank you @jrgm.  this was incredibly annoying.

jrgm:
verified

PR #1771, opened 2012-06-18
fix load_gen to call /verify with the right args

jrgm:
- send the right assertion to /verify -- (lib/load_gen/crypto.js)
- fix two comparisons I noticed would always be false (one of which was hiding the /verify failure)
- fix a trivial error message formatting bug GH-1278

fixes GH-1770 and GH-1278

### DIFF
--- a/lib/load_gen/common.js
+++ b/lib/load_gen/common.js
@@ -47,7 +47,7 @@ exports.authAndKey = function(cfg, user, ctx, email, cb) {
           try {
             if (err) throw err;
             if (resp.code !== 200) throw "non-200 status: " + resp.code +
-              + " - " + resp.body;
+              " - " + resp.body;
             if (typeof resp.body !== 'string') throw cb("no response body");
             userdb.addCertToUserCtx(ctx, email, resp.body);
             cb();
@@ -75,7 +75,9 @@ exports.genAssertionAndVerify = function(cfg, user, ctx, email, audience, cb) {
     // the contents so much
     try {
       if (err) throw err;
-      if (!typeof JSON.parse(r.body) === 'object') throw 'bogus response';
+      if (typeof JSON.parse(r.body) !== 'object') {
+        throw 'response is not a JSON object: ' + r.body;
+      }
     } catch(e) {
       return cb(e.toString() + (r ? (" - " + r.body) : ""));
     }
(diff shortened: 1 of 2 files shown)


lib/load_gen/common.js
```diff
@@ -47,7 +47,7 @@ exports.authAndKey = function(cfg, user, ctx, email, cb) {
           try {
             if (err) throw err;
             if (resp.code !== 200) throw "non-200 status: " + resp.code +
-              + " - " + resp.body;
```
ozten, line 50: Nice catch!

lib/load_gen/common.js
```diff
@@ -75,7 +75,9 @@ exports.genAssertionAndVerify = function(cfg, user, ctx, email, audience, cb) {
     // the contents so much
     try {
       if (err) throw err;
-      if (!typeof JSON.parse(r.body) === 'object') throw 'bogus response';
+      if (typeof JSON.parse(r.body) !== 'object') {
```
ozten, line 78: Ah, ya.

lib/load_gen/common.js
```diff
@@ -98,7 +100,8 @@ exports.genAssertionAndVerify = function(cfg, user, ctx, email, audience, cb) {
         try {
           if (err) throw err;
           if (r.code !== 200) throw "non-200 status: " + r.code;
-          if (!JSON.parse(r.body).status === 'okay') throw "verification failed with: " + r.reason;
+          var body = JSON.parse(r.body);
+          if (body.status !== 'okay') throw "verification failed with: " + body.reason;
```
ozten, line 104: Scans much easier.

ozten:
Great work jrgm!
I haven't run loadgen locally.

r+

lib/load_gen/common.js
jrgm, line 104: @ozten That's the error that was hiding the first error of posting to /verify without certificates, since that throw was unreachable. 

BTW, the crypto.js change looks good to you too?

jrgm:
Also, seeing as we have a lot of PR's queued up, this one doesn't really need to land from train-2012.06.22. I can use my own branch to run the tests for this train.

lib/load_gen/common.js
ozten, line 104: Yes, r+.

lloyd:
ditto.  r+

Issue #1733, opened 2012-06-12
(mobile) debug region gives feedback to end user


jrgm:
STR:
1. load stage diresworb.org and tap once in the right place

Expected: nothing happens
Actual: region is highlighted (and in some mobile browsers also give click audio and/or haptic feedback).

It's hard to get it happen on an iphone, visible on ICS/ChromeBeta, and obvious on stock and fennec (ICS or 2.2)

jrgm:
Actually, it's also visible in some desktop browsers when tab-cycling in the content (chrome win/mac, ie9, FF/win)

benadida:
consider whether we disable in production.

jrgm:
verified on stage train-2012.06.22

PR #1772, opened 2012-06-18
Hotfix 1733 Remove the debug region 

shane-tomlinson:
Remove the debug region until a fuller solution to only enabling it for dev is developed.

issue #1733


lloyd:
r+

Issue #1777, opened 2012-06-18
set password screen tab indexes not correct.


shane-tomlinson:
After tabbing from the verify password field, the cursor is placed on the cancel button

lloyd:
verified tab order sucks on stage, and rocks on dev (after merge).  closing.

jrgm:
verified stage train-2012.07.06

PR #1781, opened 2012-06-18
Remove the extra tabindex that caused the tab ordering to be incorrect.

shane-tomlinson:
issue #1777


lloyd:
r+ - simple and straight forward.  merging and will verify functional change on dev post merge to close issue #1777

Issue #1780, opened 2012-06-18
Firefox: pressing enter key in email address on authentication page sometimes shows invalid address tooltip.


shane-tomlinson:
Likely introduced with PR-1729 for issue #1376

STR)

1) Load up BrowserID dialog
2) If logged in, click "this is not me"
3) In email field, partially enter an email address that has been used before.
4) From the auto-complete list, select an address, press enter.
5) Watch the tooltip be displayed.

shane-tomlinson:
The order of events for a key press was incorrect in the unit tests.  The unit tests specify keydown, keyup, then keypress when it is in fact keydown, keypress, and finally keyup.

In Firefox, when pressing "enter" in an auto-complete box, the "keypress" key is triggered but not the "keyup" field.  In page_module, submit should be done on keyup instead of keypress.

jrgm:
verified on stage train-2012.06.22 for FF.

PR #1782, opened 2012-06-18
Fix submit occuring when selecting an email address in Firefox from the autocomplete list.

shane-tomlinson:
Ephemeral deployment at http://enter-key.123done.org/

Testing:
From Firefox, check that it is possible to select an address from the autocomplete box without a tooltip showing.
From IE8, make sure it is still possible to authenticate to dialog by pressing "enter" key in the password field.

issue #1780


lloyd:
r+

Issue #1732, opened 2012-06-11
KPI timestamp reported in milliseconds


nmalkin:
One of the pieces of information stored in the [KPI backend](https://wiki.mozilla.org/Privacy/Reviews/KPI_Backend) is the timestamp of the interaction (rounded to the nearest 10-minute interval). Right now, this is [sent over](https://github.com/mozilla/browserid/blob/9e52664b74a0f95d554e50bc71e77cba3bd73af0/lib/wsapi/interaction_data.js#L37-39) (and [stored](https://github.com/jedp/kpiggybank/blob/5ab4f6009fc630a6410314bf435a38090cbd11cc/lib/db.js#L39)) in _milliseconds_ since the Unix epoch. Though this is standard for JavaScript, we may want to — for standardization — convert this to Unix time (i.e., _seconds_ since epoch).

shane-tomlinson:
@nmalkin - This is a good suggestion - in the end we actually want the timestamp stored in 10 minute resolutions - see PR #1670, more specifically https://github.com/mozilla/browserid/pull/1670#discussion_r909140 where @lloyd suggests the rounding happen in the router.

benadida:
review whole data flow /cc @benadida @shane-tomlinson @jedp

nmalkin:
While #1784 rounded the timestamp to the nearest 10 minutes, the original issue (this one) was not addressed: the timestamp is still in milliseconds. @shane-tomlinson, do we have any plans to change it to seconds, or are we sticking with milliseconds?

jrgm:
Verified in dev where kpi in on (data_sample_rate:1 in session_context). The server timestamp is rounded to the nearest 10 minute block. 

Notes: 
- this issue was opened about using epoch seconds not milliseconds and morphed into moving the logic for rounding the timestamp sent to /wsapi/interaction_data from the server to the client. The latter is done; @nmalkin if you feel strongly about epoch seconds, open a new issue (I'm not going to morph this back again ;-) ). 
- this is a rounding (not a modulo truncation) to the nearest 10 minute slot so e.g., both 19:45:00 and 19:54:59 fall in the 19:50:00 slot (/wsapi/interaction_data was doing truncation).

shane-tomlinson:
@jrgm - I can just as easily round down, is there one which would be of more use than the other?

[...]

PR #1784, opened 2012-06-18
Round the timestamp to the nearest 10 minutes.

shane-tomlinson:
issue #1732


lloyd:
r+ - one misspelling in a comment, will fix and merge.

PR #1790, opened 2012-06-19
disable email field after pressing next

### LINKED ISSUE #1661: locking the email field between pressing next and password field showing up
This is a very minor suggestion but it would make the login form a bit nicer to fast (impatient?) typists:
1. click signin
2. type your email
3. press enter
4. type your password
5. press enter

The problem occurs between steps 3 and 4. Depending on how soon the password field appears (and steals the focus), part of your password will end up in the email field (and be visible to anybody looking at your screen).

If we were to lock (i.e. make read-only) the email field after you press enter and before the password field steals the focus, then this problem would go away.

seanmonstar:
re-enabled after request finishes

Checking in Android Browser, the keyboard goes away automatically when you submit the form (pressing Go in the keyboard). Also, re-enabling it does not make the Android keyboard re-appear. So the current Android flow isn't altered at all.

fixes #1661

/cc @shane-tomlinson

### DIFF
--- a/resources/static/pages/signin.js
+++ b/resources/static/pages/signin.js
@@ -48,7 +48,9 @@ BrowserID.signIn = (function() {
         email = helpers.getAndValidateEmail("#email");
 
     if(email) {
+      dom.setAttr('#email', 'disabled', 'disabled');
       user.addressInfo(email, function(info) {
+        dom.removeAttr('#email', 'disabled');
         addressInfo = info;
 
         if(info.type === "secondary") {
--- a/resources/static/dialog/controllers/authenticate.js
+++ b/resources/static/dialog/controllers/authenticate.js
@@ -43,6 +43,7 @@ BrowserID.Modules.Authenticate = (function() {
 
     if (!email) return;
 
+    dom.setAttr('#email', 'disabled', 'disabled');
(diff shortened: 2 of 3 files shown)


resources/static/pages/signin.js
```diff
@@ -79,7 +81,10 @@ BrowserID.signIn = (function() {
 
           complete(oncomplete);
         }
-      }, pageHelpers.getFailure(errors.addressInfo, oncomplete));
+      }, function onError() {
+          dom.removeAttr('#email', 'disabled');
```
shane-tomlinson: @seanmonstar - Once the error screen is displayed, it is similar to the "blue screen of death" and doesn't go away until the user reloads the page.  So we don't really need this part here.

seanmonstar: Ah ok, pulling out...

shane-tomlinson:
This is a good solution, works well in the desktop and mobile browsers I've tested.  What do you think about adding this also to the signup screen and in the dialog's authentication screen?

seanmonstar:
Sure thing, I'm still learning where all these things can happen.

seanmonstar:
@shane-tomlinson added in the dialog email and signup page emails

Issue #1791, opened 2012-06-19
storage.retrieveTemporaryKeypair() is broken and possibly unused


fmarier:
The following code in /resources/static/shared/storage.js throws an exception:

```
function prepareDeps() {
  if (!jwcrypto) {
    jwcrypto = require("./lib/jwcrypto");
  }
}
```

It should of course be fixed like I did in 4ed942cfef0d3accf45807cd545dfd27864a9770.

It was never caught by our tests because this is the test for it:

```
test("retrieveTemporaryKeypair", function() {
  // XXX needs a test
});
```

Now, I did a quick grep of the code and could not find any uses of it in the codebase. So it looks like dead code that should probably be removed.

shane-tomlinson:
@fmarier - store/retrieveTemporaryKeypair were taken out in b0534666e5fb16909544697bf0bb3153a88555df because it was not used.

fmarier:
Ah! it got cleaned up the day after I rebased my branch onto the latest dev. Serves me right for not keeping up with the latest code :)

PR #1792, opened 2012-06-19
Fix a broken require for jwcrypto

fmarier:
Not sure why we never caught this because it throws an exception
as soon as you call prepareDeps().

This is related to GH-1791.

### DIFF
--- a/resources/static/shared/storage.js
+++ b/resources/static/shared/storage.js
@@ -37,12 +37,6 @@ BrowserID.Storage = (function() {
     if (window.console && console.error) console.error(msg);
   }
 
-  function prepareDeps() {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@fmarier - with checkin b0534666e5fb16909544697bf0bb3153a88555df, prepareDeps is never called.  Instead of fixing it as you propose here, it should just be removed as it is dead code.

fmarier:
Ok, I've added a commit which removes that unnecessary function.

I will actually be using it for keywrapping, but I can easily introduce it back into the file when I merge that code in.

lloyd:
r+

PR #1796, opened 2012-06-19
Issue 1733 debug region

shane-tomlinson:
Enable the development menu in the local development and AWS configurations only.  Update views.


lloyd:
r+ - clear focused change, I see no risk.

Issue #1767, opened 2012-06-15
Heartbeat of router should reflect health of browserid and static processes


zaach:
The router/browserid/static processes will live on the same box, so the health of the router process should reflect the health of those other processes. If they go down, it has nothing to route to.

/cc @lloyd

lloyd:
I think this is the behavior right now.  router forwards 'ping' wsapi to browserid, who checks db conn, and responds.  We should make this more explicit.

zaach:
@lloyd How about an options parameter on `setup` with a `dependencies` property:

```
heartbeat.setup(app, {dependencies: [dbwriter_url, browserid_url, static_url]});
```

Each `dependency_url+'/__heartbeat__'` is checked before the ok is given.

lloyd:
I like dependencies.  I also think we should add a `custom` property so we can do things like check the database:

```
heartbeat.setup(app, {
    dependencies: [dbwriter_url, browserid_url, static_url],
    custom: function(cb) {
      //  first argument to cb is an err value, null or undefined on success.
      db.ping(cb);
    });
```

jrgm:
@zaach is this work all done now? Can this one be closed.

zaach:
Yep. Closing.

PR #1800, opened 2012-06-19
Add "deep check" flag to heartbeat

zaach:
issue #1767

### DIFF
--- a/lib/heartbeat.js
+++ b/lib/heartbeat.js
@@ -10,6 +10,8 @@ url = require('url');
 // the path that heartbeats live at
 exports.path = '/__heartbeat__';
 
+const checkTimeout = 10000;
+
 // a helper function to set up a heartbeat check
 exports.setup = function(app, options, cb) {
   var dependencies = [];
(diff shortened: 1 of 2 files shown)


lib/heartbeat.js
```diff
@@ -4,20 +4,74 @@
...
 const
 urlparse = require('urlparse'),
-logger = require('./logging.js').logger;
+logger = require('./logging.js').logger,
+url = require('url');
 
 // the path that heartbeats live at
 exports.path = '/__heartbeat__';
 
+const checkTimeout = 10000;
```
lloyd, line 13: an explicit 10s timeout for checking heartbeats is nice.  This allows us to test and in cases where nodes we depend on are not reachable down, we're not running a bunch of simultaneous http requests.

lib/heartbeat.js
```diff
@@ -25,29 +79,39 @@ exports.setup = function(app, cb) {
...
   else if (typeof url !== 'object') throw "url string or object required as argumnet to heartbeat.check";
   if (!url.port) url.port = (url.scheme === 'http') ? 80 : 443;
 
   var shortname = url.host + ':' + url.port;
 
-  require(url.scheme).get({
+  var req = require(url.scheme).get({
     host: url.host,
     port: url.port,
     path: exports.path
```
lloyd, line 98: we assume all of our deps will mount the heartbeat handler at the same location as we do.  fair enough.

lib/heartbeat.js
```diff
@@ -25,29 +79,39 @@ exports.setup = function(app, cb) {
...
-  }, function (e) {
+    else {
+      logger.error("non-200 response from " + shortname + ".  fatal! (" + res.statusCode + ")");
+      cb(false, "response code " + res.statusCode);
+    }
+  });
+  req.on('error', function (e) {
     logger.error("can't communicate with " + shortname + ".  fatal: " + e);
-    cb(false);
+    cb(false, e);
```
lloyd, line 108: typically the convention I've tried to use throughout the source is callbacks always take an `err` param as their first argument.  this breaks that convention.  I do this because that's what node does throughout its api.

lloyd:
r- - two things I'd like to see here:
1. unit tests
2. always return errors as first argument.

otherwise great.

lib/heartbeat.js
```diff
@@ -25,29 +79,39 @@ exports.setup = function(app, cb) {
...
+  });
+  req.on('connect', function (socket) {
+    socket.setTimeout(checkTimeout);
+    socket.on('timeout', function() {
+        req.abort();
+        cb(false, "timeout");
+    });
```
lloyd, line 115: this doesn't do what we intend.  if you kill SIGSTOP the dependent process, then 'connect' is never raised and the timeout is 120s.  I think what we want is a guaranteed maximum time for the check as perceived by the application.

lib/heartbeat.js
```diff
@@ -25,29 +79,39 @@ exports.setup = function(app, cb) {
...
+  var req = require(url.scheme).get({
     host: url.host,
     port: url.port,
     path: exports.path
   }, function (res) {
     if (res.statusCode === 200) cb(true);
-    else logger.error("non-200 response from " + shortname + ".  fatal! (" + res.statusCode + ")");
-  }, function (e) {
+    else {
+      logger.error("non-200 response from " + shortname + ".  fatal! (" + res.statusCode + ")");
```
lloyd, line 102: not sure these should be .error() level messages.  That one we reserve for codepaths that should never be executed in production.  I'd hope that the failing dependee process would have .error() messages that tell us why its failing, OR if it's related to the hardware or environment outside of our APP, we'll have warnings in our logs, and the absence of errors will point ops to the rest of the deployment infrastructure.  suggest .warn() for heartbeat failures.

lib/heartbeat.js
```diff
@@ -4,20 +4,74 @@
...
+    function ok(yeah, msg) {
+      res.writeHead(yeah ? 200 : 500);
+      res.write(yeah ? 'ok' : 'bad');
+      if (msg) res.write(msg);
+      res.end();
+    }
+
+    // check all dependencies if deep
+    if (deep) {
+      for (var i = 0; i < count; i++) {
```
lloyd, line 71: this is broken in the case there are no dependencies

lloyd:
closing down this pull request and opening a new one with a new pile of commits that addresses the issues above.

Issue #1731, opened 2012-06-11
include jshint in unit tests


lloyd:
We should create a backend unit test that ensures a clean jshint run, so we don't regress on all the cleanup @zaach did for us

shane-tomlinson:
@lloyd - I'd like to do this on the front end too, but there is a lot of work that needs done.  I do this quite frequently on files I touch, cleaning as I go, but there are a lot of files that need treated yet.

For both the front end and the back end, we should come up with universal jshint config files so your jshint run does not yield different results from my jshint run.

zaach:
It looks like you can supply the jshint module a custom configuration when using it from node, which would be useful if we only want to unit test for the more critical syntax mistakes and have a separate, more stylistically opinionated configuration for IDEs. Though it would be a good idea to commit common .jshintrc files for front-end and back-end code.

jrgm:
verified for train-2012.06.22

PR #1802, opened 2012-06-20
add jshint checking for globals as a unit test

lloyd:
issue #1731, issue #1785

This is zaach's contribution plus a small commit to verify that jshint is installed and running.


lloyd:
r+ on zaach's work, my addition is trivial.  Because this is _just_ a new unit test, I don't see a reason to delay merge.

Issue #1803, opened 2012-06-20
backend unit tests take too long


lloyd:
It takes at least 10s per test (that starts up our software) on my 13" mbp to run backend unit tests.  there are about 35 distinct tests.  If there were an easy way to optimize this, we should.  

One fallout of this is that we're blowing past our allotted time on travis more and more and getting false failures.

PR #1804, opened 2012-06-20
optimize backend unit tests by starting all daemons in parallel, and the...

lloyd:
...n starting router.  issue #1803

on my dual coare 13" mbp this reduces total runtime of backend tests from 8m51.564s to 7m10.813s - that's a 20% speedup for a very safe and isolated change in just the test harness.  This also decreases the time from typing `npm start` to actually being able to hit the local server.

review and merge at will, plz.


zaach:
I noticed a small speed-up on a faster MBP, but looks good! r+

lloyd:
zaach: what kinda speedup do you see? you gots 4 cores?

zaach:
2.5 GHz Intel Core i7

It went from about 2m50s to 2m43s.

lloyd:
i need a new computer!

zaach:
Haha, +1

Issue #18, opened 2011-06-15
implement "forgot password" flow


lloyd:
because it's easy to forget your password!

PR #1806, opened 2012-06-20
rename loggedInEmail to loggedInUser with backcompat support.  issue #18...

lloyd:
...05

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1027,13 +1027,25 @@
 
       _open_hidden_iframe();
 
+      // back compat support for loggedInEmail
+      if (typeof options.loggedInEmail !== 'undefined' &&
+          typeof options.loggedInUser !== 'undefined')
+      {
+        throw "you cannot supply *both* loggedInEmail and loggedInUser";
+      }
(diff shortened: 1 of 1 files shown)


resources/static/include_js/include.js
```diff
@@ -1027,13 +1027,25 @@
...
+      if (typeof options.loggedInEmail !== 'undefined' &&
+          typeof options.loggedInUser !== 'undefined')
+      {
+        throw "you cannot supply *both* loggedInEmail and loggedInUser";
+      }
+      else if(typeof options.loggedInEmail !== 'undefined')
+      {
+        options.loggedInUser = options.loggedInEmail;
+        delete options.loggedInEmail;
+      }
```
callahad, line 1044: Looks good! Pedantry -- shouldn't this be using same-line curly braces to match the rest of this file?

callahad:
Should we add a unit test for this? Otherwise, r+.

benadida:
@callahad is right, we need tests, I guess this means we wait until early next week to merge, that's probably ok.

lloyd:
oh damn.  This means we can't really call the .watch() api stable when we roll 2012.06.22 on july 4th.

Issue #1730, opened 2012-06-11
number_sites_logged_in not in KPI whitelist


nmalkin:
KPIs sent to the server are checked against a [whitelist](https://github.com/mozilla/browserid/blob/1a2c215461b48cb74f5ea486aa0358b015c8cf71/resources/static/shared/models/interaction_data.js#L13), but number_sites_logged_in (one of our KPIs) is not in it.

ozten:
To finish of the KPI Backend, we need to populate all the fields of the KPI JSON blob that have been spec'd out.

Spec:
https://wiki.mozilla.org/Privacy/Reviews/KPI_Backend#Type_of_data_stored:

So we also need  number_emails, new_account, email_type, etc. Anything that is in the spec and missing from the blob.

ozten:
@shane-tomlinson is it okay if I assign this to you? We need your magick powers.

benadida:
@shane-tomlinson downgrade if this isn't blocking.

shane-tomlinson:
closed with b1f30fb

jrgm:
verified with an awsbox for train-2012.07.06

The keys in KPI_WHITELIST are sent with reasonable values:

```
email_type
event_stream
lang
new_account
number_emails
orphaned
sample_rate
screen_size
sites_signed_in
sites_visited
timestamp
```

One note, `new_account` is only sent when a new account is created. Filed as GH-2061 but may be acceptable as is.

ozten:
@nmalkin just discovered the number_sites_logged_in was split into sites_signed_in and sites_visited.

This will be documented at https://etherpad.mozilla.org/kpi-dashboard-privacy-review-update-1-num-sites-logged-in

Issue #1667, opened 2012-05-30
Orphaned dialog KPI


ozten:
@jbonacci was wondering if it would be easy to capture orphaned dialogs as a KPI?

I don't know if this would be easy or fit with our model of the state machine.

lloyd:
what does orphaned mean?  A dialog that was spawned, got lost, and the user never got back to?

These dialogs should will automatically closed when the site is reloaded.

Perhaps the signature of an orphan is getting to a certain state ("go check your email"), then a long delay, followed by a close event?

It would be worth having nathan spend some time working with the current event streams to see if we can reliably detect this.

shane-tomlinson:
@lloyd - sure thing.  I misunderstood @jbonacci's want here - my take on an "orphan" differs.  At the moment, we do not collect KPIs from the main site, only the dialog.  I am defining an orphan as any dialog that closes without generating an assertion.

https://github.com/mozilla/browserid/pull/1807/files#L3R79

self.success is set to true if the user is redirected to their IdP to authenticate (state.js:L198) or if an assertion was generated (state.js:L324)

shane-tomlinson:
@seanmonstar added some awesomeness in PR #1882 to make orphaned work as @nmalkin expects.  Closing.

PR #1807, opened 2012-06-20
Getting a bunch of new KPIs ready

shane-tomlinson:
- 'number_emails'
- 'sites_signed_in'
- 'sites_visited'
- 'orphaned'
- 'new_account'
- 'email_type'

issue #1667
issue #1730


resources/static/dialog/resources/state.js
```diff
@@ -97,6 +107,10 @@ BrowserID.State = (function() {
     handleState("new_user", function(msg, info) {
       self.newUserEmail = info.email;
 
+      // Add new_account to the KPIs *before* the staging occurs allows us to
+      // know when we are losing users due to the email verification.
+      mediator.publish("kpi_data", { new_account: true });
```
ozten, line 112: Nice. Yes, we have to publish data pretty early so we don't have nay blind spots. Good call.

resources/static/dialog/resources/state.js
```diff
@@ -146,9 +160,15 @@ BrowserID.State = (function() {
         redirectToState("primary_user_ready", info);
       }
       else {
-        // We don't want to put the provisioning step on the stack, instead when
-        // a user cancels this step, they should go back to the step before the
-        // provisioning.
+        user.isEmailRegistered(email, function(known) {
+          if (!known) {
+            mediator.publish("kpi_data", { new_account: true });
```
ozten, line 165: Interesting, yep.

resources/static/shared/models/interaction_data.js
```diff
@@ -10,7 +10,21 @@ BrowserID.Models.InteractionData = (function() {
...
+        'lang',
+        'screen_size',
+        'sample_rate',
+        'timestamp',
+        'number_emails',
+        'sites_signed_in',
+        'sites_visited',
+        'orphaned',
+        'new_account',
+        'email_type'
```
ozten, line 25: Hawt!

resources/static/shared/models/interaction_data.js
```diff
@@ -97,13 +111,7 @@ BrowserID.Models.InteractionData = (function() {
...
       var filtered = [];
       _.each(data, function(obj) {
-        filtered.push(whitelistFilter(obj, [
-          'event_stream',
-          'lang',
-          'screen_size',
-          'sample_rate',
-          'timestamp'
-        ]));
+        filtered.push(whitelistFilter(obj, KPI_WHITELIST));
```
ozten, line 114: Much cleaner.

resources/static/shared/modules/interaction_data.js
```diff
@@ -132,6 +131,12 @@ BrowserID.Modules.InteractionData = (function() {
     self.samplesBeingStored = true;
   }
 
+  function onKPIData(msg, result) {
+    var currentData = this.getCurrent();
+    _.extend(currentData, result);
```
ozten, line 159: Is there any danger of overwriting data here? Would sites_signed_in ever exist in both `currentData` and `result`, to where we'd need to add up their values, instead of overwriting `currentData.sites_signed_in`?

resources/static/shared/user.js
```diff
@@ -745,6 +745,31 @@ BrowserID.User = (function() {
...
     /**
+     * Authenticate the user with the given email and assertion.  This will sync
+     * the user's addresses.
+     * @method authenticateWithAssertion
+     * @param {string} email
+     * @param {string} assertion
+     * @param {function} [onComplete] - Called on completion with status. true
+     * if user is authenticated, false otw.
+     * @param {function} [onFailure] - Called on error.
+     */
```
ozten, line 756: Nice docs.

resources/static/test/cases/shared/storage.js
```diff
@@ -167,5 +174,14 @@
...
 
+  test("setLoggedIn, getLoggedIn, loggedInCount", function() {
+    var email = "testuser@testuser.com";
+    storage.setLoggedIn(TEST_ORIGIN, email);
+    equal(storage.getLoggedIn(TEST_ORIGIN), email, "correct email");
+
+    storage.setLoggedIn("http://another.domain", email);
+    equal(storage.loggedInCount(), 2, "correct logged in count");
+  });
+
```
ozten, line 185: Nice tests!

ozten:
Just one question before r+...

Otherwise, Code looks great. I didn't run locally. I can't find test output on Travis CI.

resources/static/test/cases/shared/models/interaction_data.js
```diff
@@ -95,7 +95,12 @@
           timestamp: now,
           local_timestamp: now,
           lang: "bar",
-          secret: "Attack at dawn!!!"
```
jedp, line 98: sigh.

resources/static/test/cases/shared/modules/interaction_data.js
```diff
@@ -246,4 +246,16 @@
     });
   });
 
+  asyncTest("kpi_data message adds fields to current kpi_data", function() {
+    createController();
+    network.withContext(function() {
+      mediator.publish("kpi_data", { number_emails: 1 });
+      var current = controller.getCurrent();
+      testHelpers.testObjectValuesEqual(controller.getCurrent(), {
```
jedp, line 253: did you mean to use the new var 'current' instead of 'controller.getCurrent()' here?  (or remove line 253?)

resources/static/shared/storage.js
```diff
@@ -213,10 +217,19 @@ BrowserID.Storage = (function() {
 
     if (siteInfo) {
       delete siteInfo[key];
+
+      // If no more info for site, get rid of it.
+      if (!_.size(siteInfo)) delete allSiteInfo[site];
```
jedp, line 222: Does the flow always go through siteCount first?  Is there any danger of allSiteInfo being undefined at this point?

resources/static/test/cases/shared/storage.js
jedp, line 185: Agreed!  Nice tests, and nice style of tests.

resources/static/test/cases/shared/modules/interaction_data.js
shane-tomlinson, line 253: @jedp - good catch.

resources/static/shared/storage.js
shane-tomlinson, line 222: @jedp - there is no danger of siteInfo being undefined, line #218 checks `if (siteInfo)` before branching into here.

resources/static/shared/modules/interaction_data.js
shane-tomlinson, line 159: @ozten - Great question - all addition/subtraction/mutations are expected to have happened before they are saved to the KPI data store.  Instead of doing any addition or subtraction here, we are setting explicit values.  

sites_signed_in is set once on window_unload (https://github.com/mozilla/browserid/pull/1807/files#L3R77).  This ensures all sites for the currently signed in user are counted, including if the user has just signed in to a site for the first time.

This does leave a hole in coverage for sites_signed_in if a user opens a dialog, sees that somebody else was logged in, and clicks "this is not me".  The original signed in user will not have the number of sites signed in saved.

lloyd:
afaict we need one more commit on this PR to address the "new var 'current' instead of 'controller.getCurrent()' here".

I'll re-read, but I can't see any other major issues.  I'll add a patch, test and merge in about 2 hrs if there's no response before then.

lloyd:
with the last patch, r+ - checking tests and merging..

lloyd:
http://travis-ci.org/#!/shane-tomlinson/browserid/builds/1708667

PR #1809, opened 2012-06-20
fix #error gradient on Android

### LINKED ISSUE #1721: android: cookies disabled center area color mismatch.
Android Screenshot

seanmonstar:
it's gradient doesn't go all the way down either. We must keep
the backgrounds on #wait, #delay, and #error, since they are
needed to cover the #formWrap background.

the same background of #error is added to #content, so that
even though #error doesn't extend all the way down, the gradient
still does, and it looks correct on Android again.

fixes #1721

/cc @shane-tomlinson

### DIFF
--- a/resources/static/css/common.css
+++ b/resources/static/css/common.css
@@ -357,7 +357,7 @@ footer .help {
   font-weight: bold;
 }
 
-#wait, #delay, #error {
+#content, #wait, #delay, #error {
   background-color: #dadee1;
   background-image: url("/i/grain.png"), -moz-linear-gradient(top, #dadee1, #c7ccd0);
 }
--- a/resources/static/css/common.css
+++ b/resources/static/css/common.css
@@ -357,9 +357,14 @@ footer .help {
   font-weight: bold;
 }
 
-#content, #wait, #delay, #error {
+#wait, #delay, #error {
   background-color: #dadee1;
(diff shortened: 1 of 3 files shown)


resources/static/css/common.css
```diff
@@ -357,7 +357,7 @@ footer .help {
   font-weight: bold;
 }
 
-#wait, #delay, #error {
+#content, #wait, #delay, #error {
   background-color: #dadee1;
   background-image: url("/i/grain.png"), -moz-linear-gradient(top, #dadee1, #c7ccd0);
```
shane-tomlinson: It looks like we need some gradients for webkit, opera, and ms here too.  How'd I miss that?

shane-tomlinson:
@seanmonstar - I'm testing this now, will have more to say.  I think the solution may be even simpler yet, but I want to check.

resources/static/css/common.css
seanmonstar: Oh yea, I didn't notice that either. Does the gradient override the grain.png? Or is the png transparent?

seanmonstar:
@shane-tomlinson curious what you have found.

seanmonstar:
Yea, that looks right as well.

shane-tomlinson:
Re-opened as PR-1842.  Closing this one.

PR #1810, opened 2012-06-21
add notification to add_email dialog after submit

### LINKED ISSUE #1433: 'Add new email address' screen should show user notification while getting 'email_info' being called.
This would be similar to how it is done on the sign in page, otherwise the user is left in the dark as to what is occurring.

seanmonstar:
fixes #1433

/cc @shane-tomlinson

### DIFF
--- a/resources/static/dialog/controllers/add_email.js
+++ b/resources/static/dialog/controllers/add_email.js
@@ -7,17 +7,38 @@ BrowserID.Modules.AddEmail = (function() {
   "use strict";
 
   var bid = BrowserID,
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
Works a treat (British for it works perfectly).  I'm wondering if it is time to pull out showHint/hideHint into PageModule so that it can be shared.  showHint would require a change in function signature to take the hints to hide.

seanmonstar:
I wondered similarly, but I still find the PageModule structure to be a bit of voodoo until I read more into it.

Issue #1793, opened 2012-06-19
/forgot, /verify_email_address page are cachable - email, token are not embedded in page.


shane-tomlinson:

lloyd:
headers now look right (must re-validate every time, but cachable and clients may use ETags):

```
HTTP/1.1 200 OK
x-powered-by: Express
strict-transport-security: max-age=2592000; includeSubdomains
content-type: text/html; charset=utf8
content-length: 5926
vary: Accept-Encoding,Accept-Language
cache-control: public, max-age=0
etag: "1c1bc4b3738f9840e9f4d5bc1fd30c02"
x-frame-options: DENY
connection: close
```

All unit tests pass and manual testing works.  case closed.

jrgm:
verified stage train-2012.07.06 in chrome and firefox - initial load on an empty cache is 200 OK with 'cache-control: public, max-age=0' and an etag: and subsequent GET of the same URL returns 304 Not Modified with either /verify_email_address or /forgot (as already noted above and the previous train did not do (200 OK on every load)).

PR #1817, opened 2012-06-21
Make the forgot and verify_email_address pages cachable.

shane-tomlinson:
issue #1793


lloyd:
@shane-tomlinson - So at some point we changed these templates so they no longer included dynamic content?  That's fantastic.  /cc @zaach

lloyd:
r+, enthusiastically.  This changes makes all of our views cachable and removes all per-user data from our DOMs.  love it.

Issue #1778, opened 2012-06-18
Remove no longer needed class names in templates - serif and sans


shane-tomlinson:
These are remnants from the last design and should be removed to shrink the content shipped down the wire.

lloyd:
I see no visible regression in our dev site after merging this pull request.  closing.

PR #1818, opened 2012-06-21
Remove the no longer used "serif" and "sans" class names.

shane-tomlinson:
issue #1778


lloyd:
confirmed neither of these classes are referenced in our CSS.  thanks for making time to simplify and clean up, @stomlinson!

r+

jrgm:
Havent't noticed any visual issues in stage train-2012.07.06. Verified. Ish.

Issue #1548, opened 2012-05-05
[win7/Chrome] /sign_in dialog can be resized so the screen is (essentially) blank


jrgm:
This is specific to Chrome on my Win7 VM. I can't get anything similar with other browsers. I have managed to duplicate this on OSX and Ubuntu using Chrome, but it is extremely difficult to find the right size to trigger it on those windowing systems. 

However, on Win7/Chrome it is really easy to do this. (By easy, I mean "if you are trying to do this"; this bug may be nothing because most reasonable users would just resize it again to get the contents showing again).

Anyways, there's a fair size range of sizes where Chrome lays out the /sign_in content with nothing usable visible.

STR
1. Start Chrome and use an RP to bring up the dialog
2. Begin resizing the dialog up and to the left 
OR
1. Start Chrome, bring up dialog, and start Developer Tools
2. In the console, enter window.resizeTo(630,340) which is in the sweet spot for my win7 box.

Chrome version: 18.0.1025.168 m on Windows 7 (fully updated).

![img](https://img.skitch.com/20120505-c9qcsirwr3kqunrkafphmta13s.png)

lloyd:
I can reproduce on stage, but no longer on dev.  fixified.

jrgm:
can't reproduce on train-2012.07.06 on awsbox deploy

PR #1819, opened 2012-06-21
Fix chrome showing a blank screen on resizing to exactly 640px.

shane-tomlinson:
issue #1548

### DIFF
--- a/resources/static/dialog/resources/screen_size_hacks.js
+++ b/resources/static/dialog/resources/screen_size_hacks.js
@@ -12,7 +12,10 @@
         signInEl = $("#signIn");
 
     selectEmailEl.css("position", "static");
-    if($(window).width() >= 640) {
+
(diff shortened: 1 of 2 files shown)


resources/static/dialog/resources/screen_size_hacks.js
```diff
@@ -12,7 +12,10 @@
         signInEl = $("#signIn");
 
     selectEmailEl.css("position", "static");
-    if($(window).width() >= 640) {
+
+    // The mobile breakpoint is 640px in the CSS.  If the max-width is changed
+    // there, it must be changed here as well.
+    if($(window).width() > 640) {
```
lloyd, line 18: aha.  a one char bug.  This makes sense - javascript calculations included the boundary condition while CSS did not.

lloyd:
r+

PR #1823, opened 2012-06-21
update nodemailer and pass a local instance of `util` to views - issue17...

zaach:
...56

rebase of #1758


zaach:
Fixes #1756. Ready to be merged next week.

lloyd:
r+ - because we're updating nodemailer here, we should get QA to test mail delivery failures in the stage environment as part of testing the train this goes out on.  /cc @jbonacci @jrgm

lloyd:
correct!  this is for train 2012.07.06 - targeting production on 7/18.  not for the current train on stage.

PR #1843, opened 2012-06-26
Issue 1760 dep required email

shane-tomlinson:
The initial steps to deprecate requiredEmail
- If console.log is available, log a message.
- Ignore requiredEmail if it reaches the dialog.
- Remove requiredEmail from the example RP.


fmarier:
The diff looks fine to me. So r+ I guess.

callahad:
Clicked the little green "merge" button. This should be live in production on 2012-07-18 via the 2012-07-06 train.

Issue #1849, opened 2012-06-27
Round server timestamp DOWN.


shane-tomlinson:
link to #1732

jrgm:
verified on awsbox running train-2012.07.06

PR #1850, opened 2012-06-27
Round the server timestamp DOWN to the nearest 10 minutes.

shane-tomlinson:
issue #1849

### DIFF
--- a/resources/static/shared/modules/interaction_data.js
+++ b/resources/static/shared/modules/interaction_data.js
@@ -124,9 +124,9 @@ BrowserID.Modules.InteractionData = (function() {
 
     // server_time is sent in milliseconds. The promise to users and data
     // safety is the timestamp would be at a 10 minute resolution.  Round to the
-    // nearest 10 minute mark.
+    // previous 10 minute mark.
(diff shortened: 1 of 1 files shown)


resources/static/shared/modules/interaction_data.js
```diff
@@ -124,9 +124,9 @@ BrowserID.Modules.InteractionData = (function() {
 
     // server_time is sent in milliseconds. The promise to users and data
     // safety is the timestamp would be at a 10 minute resolution.  Round to the
-    // nearest 10 minute mark.
+    // previous 10 minute mark.
     var TEN_MINS_IN_MS = 10 * 60 * 1000,
-        roundedServerTime = Math.round(result.server_time / TEN_MINS_IN_MS) * TEN_MINS_IN_MS;
+        roundedServerTime = Math.floor(result.server_time / TEN_MINS_IN_MS) * TEN_MINS_IN_MS;
```
ozten, line 129: This will put all KPI into 10 minute chunks, associating them to the previous 10 minute interval.

r+

PR #1852, opened 2012-06-27
A huge reorganization of the client side directories. 

shane-tomlinson:
General structure is:

common/
      js
      css
      i

dialog/
      js
      css
      i

pages/
      js
      css
      i

I have not changed the structure of the following directories:
auth_with_idp
communication_iframe
i18n
relay


seanmonstar:
This looks much more sensible to me.

jrgm:
marking as verified but not really checking anything here.

PR #1858, opened 2012-06-28
updates to primary flow so that the DOM API is a true abstraction.

benadida:
this uses sessionStorage in a way that should be IE8 safe, but I don't have an easy way to test IE8 right now so creating this pull request to make that easier.


resources/static/test/mocks/window.js
```diff
@@ -15,6 +15,7 @@ BrowserID.Mocks.WindowMock = (function() {
 
   function WindowMock() {
     this.document = new DocumentMock();
+    this.sessionStorage = {};
```
shane-tomlinson, line 18: @benadida - Nice update to the mock!

resources/static/dialog/js/modules/dialog.js
```diff
@@ -196,24 +196,15 @@ BrowserID.Modules.Dialog = (function() {
...
-            throw "invalid #CREATE_EMAIL= (" + email + ")";
-          params.type = "primary";
-          params.email = email;
-          params.add = false;
-        }
-        else if (hash.indexOf("#ADD_EMAIL=") === 0) {
-          var email = hash.replace(/#ADD_EMAIL=/, "");
-          if (!bid.verifyEmail(email))
-            throw "invalid #ADD_EMAIL= (" + email + ")";
+        if (hash.indexOf("#AUTH_RETURN") === 0) {
```
shane-tomlinson, line 199: @benadida - I really like this simplification of the previously janktastic code.

resources/static/authentication_api.js
```diff
@@ -34,11 +33,11 @@
     };
 
     navigator.id.completeAuthentication = function(cb) {
-      window.location = getParameterByName('return_to');
+      window.location = 'https://login.persona.org/sign_in#AUTH_RETURN';
```
shane-tomlinson, line 36: Coding in the return URL this way makes _much_ more sense.

example/primary/sign_in.html
```diff
@@ -33,37 +33,30 @@
...
       .success(function(r) {
-        window.location = getParameterByName('return_to');
+        navigator.id.completeAuthentication();
       });
   });
 
   $("#cancel").click(function(e) {
     e.preventDefault();
-    window.location = getParameterByName('return_to');
+    navigator.id.cancelAuthentication();
```
shane-tomlinson: @benadida - should this be navigator.id.raiseAuthenticationFailure - when cancelling from the sample IdP, this throws the following exception:

```
TypeError: navigator.id.cancelAuthentication is not a function
http://10.246.29.67:10005/sign_in.html?email=shane.tomlinson%40example.domain
Line 59
```

resources/static/authentication_api.js
```diff
@@ -34,11 +33,11 @@
...
     };
 
     navigator.id.raiseAuthenticationFailure = function(reason) {
-      window.location = getParameterByName('return_to');
+      window.location = 'https://login.persona.org/sign_in';
```
shane-tomlinson: The url should include the #AUTH_RETURN hash.  What happens in the error case is the user will return to the dialog, the dialog will see the user is not authenticated to the IdP, but it will insert the previously used email.  Without the #AUTH_RETURN, the user will not have the email address automatically inserted after a failure.

benadida: Should we really try to provision anyways? That seems unfortunate if we know we don't need to. But I guess we can fix that later.

benadida:
@shane-tomlinson took your comments into account, returned to a different anchor for cancellation that gives us a hook to optimize later.

PR #1860, opened 2012-06-28
No longer clear the stored returnTo on window.unload.

shane-tomlinson:


lloyd:
r+!  If a user closes the dialog before checking their email, they should not "break" redirection.

jrgm:
Works for me. Causing returnTo to be set with a new or adding email to account does nto remove the returnTo and redirection happens when using the verification link.

Issue #1826, opened 2012-06-21
Font too large when displaying rp_hostname without a siteName or siteLogo


jrgm:
If the RP does not supply either siteName or siteLogo, we display rp_hostname as H2. When a siteName is given rp_hostname is an H3.

H2 is too large from a hostname (which is usually much longer than a siteName (e.g., www.mozilla.org vs. Mozilla)).

![img](https://img.skitch.com/20120621-rre5abs8ndrmsu4ws2yknpjbja.png)

lloyd:
hey skinny, heres a thing

skinny97214:
Follow smartell's mock 

http://f.cl.ly/items/2d3E30203E3f1d293N2W/persona-login-old.png

jrgm:
verified on stage train-2012.07.06.

PR #1865, opened 2012-06-28
Change the font size of the RP name and logo as per skinny's suggestions...

shane-tomlinson:
....

issue #1826

### DIFF
--- a/resources/static/dialog/css/style.css
+++ b/resources/static/dialog/css/style.css
@@ -301,9 +301,8 @@ section > .contents {
 }
 
 #favicon h3 {
-    font-size: 19px;
-    margin-top: 0;
     font-size: 14px;
+    margin-top: 0;
 }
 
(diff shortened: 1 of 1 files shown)


resources/static/dialog/css/style.css
```diff
@@ -297,11 +297,13 @@ section > .contents {
 
 #favicon h2 {
     margin: 10px 0 0 0;
+    font-size: 18px;
 }
 
 #favicon h3 {
     font-size: 19px;
     margin-top: 0;
+    font-size: 14px;
```
seanmonstar: `font-size` is declared twice right here.

shane-tomlinson: @seanmonstar - exactly why we have code reviews.  Nice catch.  Updating.

seanmonstar:
cant be automatically merged, but otherwise r+

PR #1868, opened 2012-06-29
Remove duplicate copy of the db password setting

fmarier:
The same lines appear a few lines up from here.


lloyd:
very nice cleanup.  thank you.  r+

PR #1876, opened 2012-06-29
when serving non-minified resources, substitute urls in templates so onc...

lloyd:
...e can locally test links embedded in templates (like persona tos/pp)

### DIFF
--- a/bin/browserid
+++ b/bin/browserid
@@ -144,6 +144,12 @@ app.use(function(req, res, next) {
   next();
 });
 
+// if we're not serving minified resources (local dev), then we should add
+// .ejs to the mime table so it's properly substituted.  issue #1875
+if (!config.get('use_minified_resources')) {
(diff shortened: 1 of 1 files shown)


bin/browserid
```diff
@@ -144,6 +144,12 @@ app.use(function(req, res, next) {
   next();
 });
 
+// if we're not serving minified resources (local dev), then we should add
+// .ejs to the mime table so it's properly substituted.  issue #1875
+if (!config.get('use_minified_resources')) {
+  express.static.mime.types['ejs'] = 'text/html';
```
ozten, line 150: Very cool. Didn't know how to do this before :)

ozten:
r+

Tests pass except for a npm install timeout: http://travis-ci.org/#!/mozilla/browserid/builds/1738189

Please merge if qunit tests pass locally.

Issue #1867, opened 2012-06-29
ActiveX not not available with Windows 8 Metro


stuartparmenter:
In micrajax.js, in getXHRObject(), it checks for window.ActiveXObject before window.XMLHttpRequest which causes bad things to happen when trying to use BrowserID within a Metro app.  Changing the ordering of these fixes the problem and is consistent with Microsoft (http://blogs.msdn.com/b/ie/archive/2011/08/31/browsing-without-plug-ins.aspx).  I have verified this locally.

PR #1877, opened 2012-06-29
Reverse the order of checking for window.XMLHttpRequest and window.Activ...

shane-tomlinson:
...eXObject when creating the XHR object.

Fix for Windows Metro apps.

issue #1867


benadida:
looks good, r+ and merging.

Issue #1827, opened 2012-06-21
KPI data should acknowledge successful sign-in even after dialog is closed


nmalkin:
**Scenario:**
Dialog opens, new user enters email, creates password, verification email is sent. User _closes dialog_, goes to email client, clicks on verification link. Email is verified and user is logged in.

**The problem:**
When the KPI data is sent over (the next time the dialog is opened), the data contains no indication of whether the user succeeded or failed at logging in. (Our most important metric!)

(A similar problem affects other metrics, like users resetting their password or adding a second email address.)

**Proposed solution:**
When the dialog next opens, before sending over the KPI data, we should check
1. if the last event in the event stream is _user.user_staged_  
   This means the last thing we did is send a verification email
2. if the user is authenticated.

If they are, that means the user succeeded at signing in, and we modify the payload to acknowledge that.

**How?**
I think the cleanest solution is to change the _[orphaned](https://github.com/mozilla/browserid/pull/1807/files#L3R79)_ flag introduced in PR #1807. Because the dialog was closed before an assertion was generated, it would normally be _true_. But if we change it to _false_, we'll be able to tell that, even though the dialog was closed before an assertion was generated (this can be verified from the event stream), the user did manage to log in.

**Limitation**
If the user switches browsers (creates an account in one, verifies email in another, and/or logs in a different one), this method won't work. To account for this, we would probably have to wait until later in the flow to send interaction data or store some information about the user on the server. Both solutions have troubling implications, so, given that this case is generally unlikely, we should probably ignore it for now.

nmalkin:
Thanks, Sean!

seanmonstar:
@nmalkin is that correct? passing `{ orphaned: false}`, so you know it was orphaned and then adopted, or should I just delete the `orphaned` property out of it?

nmalkin:
That's what I had in mind. I think it's clearer if it's set to false than if it's missing altogether.

PR #1882, opened 2012-06-29
Improve orphaned KPI reporting

seanmonstar:
If a dialog was orphaned previously, but the user was able to confirm their email address and hence be logged in, update the KPI to say that `orphaned` is false.

See #1827 for more.

### DIFF
--- a/resources/static/common/js/modules/interaction_data.js
+++ b/resources/static/common/js/modules/interaction_data.js
@@ -28,6 +28,7 @@ BrowserID.Modules.InteractionData = (function() {
   var bid = BrowserID,
       model = bid.Models.InteractionData,
       network = bid.Network,
+      storage = bid.Storage,
       complete = bid.Helpers.complete,
       dom = bid.DOM,
       sc;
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@seanmonstar, @nmalkin - I like the simplicity of this approach but it fails in two cases:

1) If the user logs in through the main site before they open the dialog again, orphaned will be set to false.
2) If the user is adding an email address to an already authenticated account, never finish the verification process but open the dialog again, orphaned will be set to false since they are already authenticated.

#1 I think we can live with, #2 is an issue that needs addressed.

It is opening a can of worms, but perhaps we need to hook up the KPI infrastructure on the main site and if /wsapi/complete_user_verification or /wsapi/complete_email_verification are called, we look at the previous dialog session's KPI data.  If "orphaned" is true, set it to "false"?

seanmonstar:
Regarding #2, I thought the user.staged KPI only meant he was a new user
needing confirmation. It looked like there were other KPIs for adding a new
email.
On Jul 2, 2012 4:15 AM, "Shane Tomlinson" <
reply@reply.github.com>
wrote:

resources/static/common/js/modules/interaction_data.js
```diff
@@ -98,13 +98,46 @@ BrowserID.Modules.InteractionData = (function() {
...
+    // then scrub out the previous sessions data.
+
+    var self = this;
+
+    // if we were orphaned last time, but user is now authenticated,
+    // lets make the assumption that their action worked out okay, and
+    // remove the orphaned flag
+    // See https://github.com/mozilla/browserid/issues/1827
+    var current = model.getCurrent();
+    if (current && current.orphaned) {
```
seanmonstar, line 132: needs to check for user_staged also

nmalkin:
> Regarding #2, I thought the user.staged KPI only meant he was a new user
> needing confirmation. It looked like there were other KPIs for adding a new
> email.

Not directly, I think, but we can look in the event stream and see that `screen.add_email` is in it. But @shane-tomlinson is right, `user.email_staged` is emitted in both cases.

Here's how I think we can deal with this without having to hook into the main site:

If we see `user.email_staged` but not `screen.add_email`, we know this is a new user. (There are other clues as well.) In that case, we proceed as we do right now (set _orphaned_ to _false_ if we're currently authenticated).

If we see `screen.add_email` _and_ `user.email_staged`, we know this is a returning user. Also in the event stream will be an event `user.email_count:1`; so we know how many emails they had registered during the previous interaction. And we can check how many emails they have now. If the number went up, then the process succeeded, and we can set _orphaned_ to _false_ again.

nmalkin:
> user.email_staged is emitted in both cases.

**Retracted.** `user.user_staged` is emitted for a new user; `user.email_staged` is emitted when an email is added.

So, as long as we look at the last event (and not just the _orphaned_ flag), we should be fine, right?

seanmonstar:
I was just about to type that up. Yea, I'm adding in additional tests for the 2 possible orphaned states.

seanmonstar:
I'm still trying to wrap my head around the flow: for adding an email, if the user closes the dialog, and then confirms the email by following the link, the verification page will have pushed the additional email into localStorage? So the check is `if (storage.getEmailCount() > current.number_emails) {` ?

Or does this require another request to the server before it can be determined?

seanmonstar:
@shane-tomlinson hows that look?

shane-tomlinson:
Thanks for this PR @seanmonstar.  

I have verified these situations:

1) new account, close dialog & no verification before dialog is opened again - orphan = true
2) new account, close dialog  & verify before dialog is opened again - orphan = false
3) new account, verify account before dialog close - orphan = false

4) add email, close dialog & no verification before dialog is opened again - orphan = true
5) add email, close dialog & verify before dialog is opened again - orphan = false
6) add email, verify account before dialog close - orphan = false

It looks good to me, but I am wondering if we can simplify the new account branch further by doing the exact same check as the "add email" branch by _only_ checking whether one of the two events is on the stream and whether the current email count is > email count stored when the dialog closed.

If that potential simplification does not work, I am an r+ on this.

seanmonstar:
@shane-tomlinson so for the user_staged event, you're suggesting to not `checkAuth`, and instead check the email count? Would a brand new user have had the email count increased also?

If so, then yea, this could be one branch: if orphaned and email is higher, then success.

shane-tomlinson:
@seanmonstar -   yeah, for some reason I thought when we verified a user/email address, we synced all the addresses.  But there was never a need to sync because the next time a user opens the dialog, their addresses will be synced.  So, your approach for the KPI data is correct.  This gets an r+ from me.

Issue #1883, opened 2012-06-30
jshint "Bad option: 'jQuery'"


jrgm:
@shane-tomlinson I was working on something else and was stealing your jshint advice lines and noticed that jshint was saying "Bad option: 'jQuery'" when I ran it. It seems that should be 'jquery: true' not 'jQuery: true' (at least for current jshint 0.7.1 version).

callahad:
I've submitted a pull request for the jshint docs, since they inconsistently refer to `jQuery` in one place: https://github.com/jshint/site/pull/52

Edit: Oh, hey, I'm an idiot. I should read things more better. And not try to fix them at 6 AM local.

PR #1891, opened 2012-07-01
Universally change jQuery->jquery when used as jshint option.

shane-tomlinson:
issue #1883


lloyd:
r+ - comment only change

Issue #1895, opened 2012-07-02
"email verification" copy on main site needs updated to match dialog.


shane-tomlinson:
This affects sign up and forgot password.

PR #1897, opened 2012-07-02
Update the main site "verify your email address" copy to match that of the dialog.

shane-tomlinson:
issue #1895

### DIFF
--- a/resources/static/pages/css/style.css
+++ b/resources/static/pages/css/style.css
@@ -428,6 +428,16 @@ button.delete:active {
           border-radius: 5px;
 }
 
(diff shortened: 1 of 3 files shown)


seanmonstar:
these templates don't need to use `gettext`?

seanmonstar:
this looks fine, r+.

@jrgm @lloyd this was supposed to hop into this train?

jrgm:
@seanmonster yeah, ship this for this train.

jrgm:
@seanmonstar See above what I said to @seanmonstar ;-)

Issue #1898, opened 2012-07-02
too much button flicker!


shane-tomlinson:
The submit button on the authentication screen flashes several times whenever the information is being retrieved for the email address.  It looks ugly.

jrgm:
verified stage train-2012.07.20

PR #1900, opened 2012-07-02
Add a small delay after an xhr request is complete before submit_disable...

shane-tomlinson:
...d class is removed from the body.  This helps to reduce the flicker users see when email addresses are being checked and there are consecutive XHR requests.

issue #1898


seanmonstar:
What am I looking for? What state does this happen?

When I go to dev.123done.org, and try to login, the first screen asks for my email. I enter it, it gets `addressInfo`, and then moves to the password screen shortly after (perhaps 0.5-1s). It doesn't look like flicker to me.

shane-tomlinson:
@seanmonstar - thanks for checking this out.

After you enter your password and hit enter, the submit button flickers a couple of times and looks pretty ugly, especially on slower devices.

resources/static/common/js/modules/xhr_disable_form.js
```diff
@@ -15,8 +15,17 @@ BrowserID.Modules.XHRDisableForm = (function() {
...
+      self.subscribe("xhr_complete", function() {
+        // Add a small delay between the time the XHR is complete and when the
+        // submit_disabled class is actually removed.  This helps reduce the
+        // amount of flicker the user sees if one XHR request completes and
+        // another one starts immediately afterwards.
+        // See https://github.com/mozilla/browserid/issues/1898
+        setTimeout(function() {
+          dom.removeClass("body", "submit_disabled");
+          self.publish("submit_enabled");
+        }, 100);
```
seanmonstar: is there any concern of multiple xhr_complete events triggering, and causing multiple setTimeouts to be setup, when only one is needed?

seanmonstar:
I went through the sign in process on my old Android phone, and didn't really notice flicker, so I still can't tell what exactly this is fixing, but I can just review the code and assume the flicker happens less.

Issue #1629, opened 2012-05-22
Reset password flow 


shane-tomlinson:
From a conversation that @skinny97214 and I had, setting the password after a reset password should be on the main site.

@skinny97214 - can you weigh in on this?

Link this to #1580, #1578, #730, #1474

skinny97214:
Here's a diagram of the flow. Everything happens in the dialog and email.

http://cl.ly/160P0w1e051T293t0x2d

shane-tomlinson:
@skinny97214 - One of the original reasons for putting the password in the dialog before sending the verification email is because we cannot re-focus the dialog in all browsers on all devices.  Re-focusing the dialog here post verification link click will suffer from the same problem.

As an alternative, we can put the set password back on the main site and redirect to the RP after the user sets the password.

shane-tomlinson:
We just merged.  Any remaining problems/changes can be opened as separate issues.

Issue #759, opened 2011-12-15
email reset / password recovery is not good


lloyd:
No one is happy with how password reset works, we need to re-think it considering both security and UX.

benadida:
UX + security agreement (sort of cause @lloyd not on phone) that blast all emails but keep account consolidated is the right approach. Ben will write up.

ghost:
Currently, when performing the steps from @jbonacci 's last comment, the email used at step 4 is separated into a new account with the new password. The other emails stay in the old account.

If you decide to keep this behavior, the user will have to be notified about what happened since his impression will be that the emails from the old account were simply lost. Perhaps display a warning when he clicks the Forgot Password link and wait for a confirmation before sending the verification email?

PR #1913, opened 2012-07-04
New reset password flow

lloyd:
This pull request implements the new password reset flow described here:

https://groups.google.com/group/mozilla.dev.identity/browse_thread/thread/fe03a79a905311d3#

this would address issue #759 but not (yet) issue #1629


lib/wsapi/stage_reverify.js
```diff
@@ -0,0 +1,72 @@
...
+            if (err) return wsapi.databaseDown(res, err);
+
+            var langContext = wsapi.langContext(req);
+            
+            // store the email being reverified
+            req.session.pendingReverification = secret;
+            
+            res.json({ success: true });
+            // let's now kick out a verification email!
+            email.sendAddAddressEmail(req.body.email, req.body.site, secret, langContext);
```
shane-tomlinson: @lloyd - is the reverification email supposed to be the same as the "add email to account" email?  If so, should we take out the confirm page, complete_reverify wsapi and front end code to call both of these?

lib/db.js
```diff
@@ -101,7 +102,10 @@ exports.onReady = function(f) {
 [
   'stageUser',
   'stageEmail',
-  'gotVerificationSecret',
+  'completeCreateUser',
+  'completeAddEmail',
+  'completePasswordReset',
+  'completeReverify',
```
lloyd: this change adds 5 new database functions, and removes 1.

lib/db/mysql.js
```diff
@@ -72,6 +73,7 @@ const schemas = [
     "user BIGINT NOT NULL," +
     "address VARCHAR(255) UNIQUE NOT NULL," +
     "type ENUM('secondary', 'primary') DEFAULT 'secondary' NOT NULL," +
+    "verified BOOLEAN DEFAULT TRUE NOT NULL, " +
```
lloyd, line 76: this feature has a database change.

lib/static/views.js
```diff
@@ -178,6 +178,17 @@ exports.setup = function(app) {
...
   });
 
+
+  app.get("/reset_password", function(req,res) {
+    renderCachableView(req, res, 'add_email_address.ejs', {title: 'Reset Password'});
+  });
+
+  app.get("/confirm", function(req,res) {
+    renderCachableView(req, res, 'add_email_address.ejs', {title: 'Confirm Email'});
+  });
```
lloyd, line 190: we added two new views with this change.

lib/wsapi/complete_reverify.js
```diff
@@ -0,0 +1,61 @@
...
+          logger.warn("couldn't complete email verification: " + e);
+          wsapi.databaseDown(res, e);
+        } else {
+          wsapi.authenticateSession(req.session, uid, 'password');
+          res.json({ success: true });
+        }
+      });
+    };
+  });
+};
```
lloyd: two new wsapis have been added to support this change.

lib/wsapi/cert_key.js
```diff
@@ -24,15 +24,22 @@ exports.process = function(req, res) {
...
-      if (err) {
-        logger.error("error forwarding request to keysigner: " + err);
-        httputils.serverError(res, "can't contact keysigner");
-        return;
-      }
+    // secondary addresses in the database may be "unverified".  this occurs when
+    // a user forgets their password.  We will not issue certs for unverified email
+    // addresses
+    db.emailIsVerified(req.body.email, function(err, verified) {
+      if (!verified) return httputils.forbidden(res, "that email requires (re)verification");
```
lloyd, line 31: now emails in the database can be unverified.  if they are, we don't issue certificates for them.

lib/wsapi/email_for_token.js
```diff
@@ -19,75 +20,66 @@ exports.args = ['token'];
...
-      // table, then we require a password be fetched from the user upon verification.
-      // these checks are temporary and should disappear in 1 trains time.
-      function needsPassword() {
-        // no password is set neither in the user table nor in the staged record.
-        // the user must pick a password
-        res.json({
-          success: true,
-          email: email,
-          needs_password: true
-        });
```
lloyd, line 65: code cleanup of transitional code for issue #1592

lib/wsapi/stage_email.js
```diff
@@ -11,11 +11,7 @@ email = require('../email.js'),
 sanitize = require('../sanitize'),
 config = require('../configuration');
 
-/* First half of account creation.  Stages a user account for creation.
- * this involves creating a secret url that must be delivered to the
- * user via their claimed email address.  Upon timeout expiry OR clickthrough
- * the staged user account transitions to a valid user account
- */
+/* Stage an email for addition to a user's account.  Causes email to be sent. */
```
lloyd, line 14: cleaning up copy paste errors in comments

lib/wsapi/stage_reverify.js
lloyd: So I think that we should send the same email for add email address and for reverify email address.  the landing page should be the same.  But I think we should use 'confirm' instead of 'add'.

This is a tiny thing, but I like confirm because it's more general and applies to both email addition and email reverification...  I'll add a change for your review.

lloyd:
Ok @shane-tomlinson - I'm passing this one back to you.  Here is what I've done:
- collapsed the add_email into confirm
- merged with latest dev
- done a full review of code, line by line.
- verified unit tests
- manually tested:
  - new acct creation
  - addition of email to acct
  - account consolidation (addition of email address that is associated with existing acct)
  - password reset
  - log in with acct after password reset
  - email re-verification after password resest
  - all of the above with verification on different browser
  - many of the above with some primary accts thrown in

Here's what I've found:

**Non-Blockers that should be made into issues upon merge:**
- The text on the reverify screen could be much more instructive.  This dialog screen is blasted in your face right when you expect to be logged in, and there's potentially lots of time between your reset password and usage of a different secondary email on your account.  This is a copy refinement bug.
- verification in a different browser could use better copy as well.  This is orthogonal to our changes here.

I have also a **Blocker** (how do you apply the blink tag in markdown?):
- If you try to complete password reset on a different browser, it fails with a blankish screen for the user.  complete_reset returns a 401 and you're stuck.  What should happen is you should be prompted for a password - the same password that you selected after clicking "i forgot my password".

so this has an r+ from me, with the exception of the Blocker.  I suggest we fix that and merge down so we have maximal time to do dev driven testing, and we avoid painful merges in the future.

lloyd:
STR for the &lt;blink&gt;**Blocker**&lt;/blink&gt;:
1. create an acct
2. in browser A, reset the password for the email on the acct
3. get that verification link from your email
4. assure that browser B is signed out
5. in browser B, load the verification link

Notice a blankish screen with a 401 error from `complete_reset`

shane-tomlinson:
@lloyd - I can cause several separate problems. To cause the problem you see above, I have to follow a different set of STRs

If I follow the STRs listed above, I correctly get the "enter your password" section in Browser B.  However, adding on to your STR, I get a breakage on an account with three addresses when verifying the third address in Browser B.

==== modified step 1 ====
1. create an acct with 3 secondary email addresses
2. in browser A, reset the password for the email on the acct
3. get that verification link from your email
4. assure that browser B is signed out
5. in browser B, load the verification link
6. in browser B, enter password.

==== new STRs ====
7. in browser A select a unverified address
8. in browser B, load the verification link.
9. in browser A, select the last unremaining unverified address.
10. in browser B, email_for_token gets the following response:
{"success":false,"reason":"missing password for user"}
11. Error screen is shown

I'm not sure where that missing password is coming from.

I also get a breakage under the following STRs
1. in browser A, create an account with a secondary email address.
2. in browser B, sign in to same account.
3. in browser A, reset the password for the email on the acct
4. get that verification link from your email.
5. in browser B, load the verification link.
6. in browser B, email_for_token gets the following response:
   {"success":true,"email":"set117@yahoo.com","must_auth":false}
7. in browser B, complete_reset gets the following response:
   Authentication Required: Provide your password

In step 6, email_for_token should have returned `"must_auth": true`

lloyd:
I'll take a look at this on the plane, hopefully have a fix to commit ~7 hours 
From now.   Unless I find wifi sooner.

--lloyd

lloyd:
I will try to reproduce manually, then create unit tests, then fix.

--lloyd

lib/wsapi/complete_reverify.js
shane-tomlinson: @lloyd complete_reverify is never used since the add email and reverify email got collapsed into "confirm".  The confirm page calls /wsapi/complete_email_addition.

lib/wsapi/email_reverify_status.js
```diff
@@ -0,0 +1,33 @@
...
+exports.args = ['email'];
+exports.i18n = false;
+
+exports.process = function(req, res) {
+  var email = req.query.email;
+  
+  // For simplicity, all we check is if an email is verified.  We do not check that
+  // the email is owned by the currently authenticated user, nor that the verification
+  // secret still exists.  These checks would require more database interactions, and
+  // other calls will fail in such circumstances.
```
shane-tomlinson, line 25: @lloyd - I am trying hard to think of a case where this _should_ return a "mustAuth" and the user needs to enter their password on the client, but so far I am not thinking of any.  Do you know of any?

What about this situation:
1) The user has an account with three addresses, one primary and 2 secondaries.
2) The user resets their password and verifies the first email address.
3) The user logs out of Persona
4) The user logs in to persona using their primary address.
5) The user selects the unverified secondary.
6) I _believe_ at this point, the user should have to enter their password, but then it will ship them over to the "verify your address" screen.
7) I'm not sure how this really works currently.

lib/db/mysql.js
```diff
@@ -342,45 +354,112 @@ exports.gotVerificationSecret = function(secret, cb) {
...
+        if (err) return cb('could not set user\'s password');
+        addEmailToUser(o.existing_user, o.email, 'secondary', cb);
+      });
+    } else {
+      addEmailToUser(o.existing_user, o.email, 'secondary', cb);
+    }
+  });
+};
+
+exports.completeReverify = function(secret, cb) {
```
shane-tomlinson: @lloyd - is this ever actually called since the confirm page calls /wsapi/complete_email_addition?

lib/wsapi/complete_reverify.js
lloyd: I fixed this, patches coming...

--lloyd

lib/wsapi/stage_reverify.js
lloyd: commit added to PR.

lib/db/mysql.js
lloyd: see latest patch, I've merged complete_email_addition and complete_reverification into one.

lib/wsapi/email_reverify_status.js
lloyd, line 25: this is the status call that the dialog polls.  this is not the the call that will return mustAuth.  email_for_token is where mustAuth is returned.

lloyd:
afaik, there re no open issues on this pull request.

My latest commit (3c9eef4), addresses the following issues:
1. problems related to email throttling and reset password (basically, if you caused a reverify email to be sent twice through the ui you'd get breakage with no feedback)
2. the problem you report above w.r.t. "no password for this user" - this was a bug in the json persistence layer, pre-existing, and has been fixed.
3. collapsing redundant and unused calls into one.  removing dead code.

@shane-tomlinson I'm r+ on this PR.  pending your final review and any other issues I got, I say let's merge and we'll focus some dev driven testing on it.

lib/wsapi/email_reverify_status.js
shane-tomlinson, line 25: @lloyd - both password_reset_status and user_creation_status return "mustAuth", they are both polling functions as well, which is why I am wondering here.

PR #1914, opened 2012-07-05
Node 0.8.x support

lloyd:
teensy patch, doesn't break 0.6, lets devs work with 0.8 if they want to and lets us start exploring it.  (really all this does is eliminate deprecation warnings)

### DIFF
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -16,6 +16,13 @@ logger = require('../logging.js').logger,
 configuration = require('../configuration.js'),
 temp = require('temp');
 
+// existsSync moved from path in 0.6.x to fs in 0.8.x
+if (typeof fs.existsSync === 'function') {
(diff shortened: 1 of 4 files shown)


zaach:
+1

lloyd:
@jedp notes that I overlooked the fact that 0.6 fs has no .existsSync - doh!  I swore I tested that!  he's taking the PR and will patch

so r-

jedp:
Jeeeeesus!  Sorry about the rebase mess.  It seemed I needed some of those earlier updates.  But oi, sorry.

Here's my bit - just a wrapper for `existsSync`!  Commit 09f65ff

lloyd:
r+ on jedp's update to this patch.  tested on 0.6 and 0.8 - no deprecation output on console from unit tests.

Issue #1905, opened 2012-07-03
Dialog displaying wrong translations


lloyd:
In our dev environment we're now extracting strings on every push, with the intention of making it so we can always meaningfully debug translations against dev.

The problem is english translations are off.

![translations off](https://img.skitch.com/20120703-ch8hi7jp3wmxqf61176h8bqn2i.png)

Notice that the fields here say "reset password" rather than "create password" and "verify password".  @ozten, what is the intent of the en_US .po files, and how do we manage them in production?  

Here is what I'm doing to extract strings on every push: https://github.com/mozilla/browserid/blob/dev/scripts/awsbox/post_deploy.js

ozten:
@mathjazz can you help with this issue?

In general - Mozilla web properties have en_US po files. What you see are the painpoints that other locales hit when we change or modify strings. It's kind of a nice warning signal as to what we'll have to give other locales a heads up on.

@mathjazz will know, but probably removing fuzzy strings will fix this.

lloyd:
I think all we need to do to fix this is to re-initialize the .po files for en_US and extract all strings.  I don't have permissions to do this.  Can someone who does have permissions help?  

This issue affects production and makes our dialog look silly.  see also #1675.

lloyd:
or better yet, just `rm -rf locale/en_US/`

mathjazz:
@lloyd If I read correctly, you're now trying to ignore en-US PO/JSON files and display original strings instead? Looks good to me.

lloyd:
@mathjazz we need a fix for this in the train we are rolling on monday, `train-2012.06.22`.

Can you _delete_ the en_US po files from svn?

I will make a patch to fix dev and ensure that en_US translations are not used from our side, but that patch is too risky to go in this train.

@jrgm - would you be ok with having @mathjazz delete the .po files before we generate strings today?

lloyd:
fwiw, locally I checked out svn, deleted en-US, ran everything, and translations in en_US look great, other locales display as expected.

lloyd:
fixed in dev.

rolling out to stage: https://bugzilla.mozilla.org/show_bug.cgi?id=767708#c26

will be in production monday.

closing down this issue.

jrgm:
This particular example looks fine on stage train-2012.07.06. Verifying but will continue to look for oddities in the next train.

Issue #1054, opened 2012-02-04
exception in l10n code during first page load after VM provisioning


lloyd:
```
Error: EINVAL, Conversion not supported.
    at GettextDomain.parseHeaders (/home/app/node_modules/node-gettext/index.js:105:22)
    at GettextDomain.parseHeaders (/home/app/node_modules/node-gettext/index.js:131:21)
    at GettextDomain.handle_strings (/home/app/node_modules/node-gettext/index.js:73:14)
    at GettextDomain.load_tables (/home/app/node_modules/node-gettext/index.js:66:14)
    at new GettextDomain (/home/app/node_modules/node-gettext/index.js:25:14)
    at Gettext.addTextdomain (/home/app/node_modules/node-gettext/index.js:221:28)
    at Object.handle (/home/app/code/lib/i18n.js:117:29)
    at next (/home/app/node_modules/connect/lib/http.js:201:15)
    at Object.logger [as handle] (/home/app/node_modules/connect/lib/middleware/logger.js:155:5)
    at next (/home/app/node_modules/connect/lib/http.js:201:15)
```

Steps to reproduce:
1. provision a new vm
2. deploy code to it (seen with sha c0479ca)
3. load the main page 
4. witness above trace
5. notice it goes away after reload

This may be related to recent provisioning changes, haven't dug in.

fetep:
this is also happening in stage, so we'll need to merge this fix.

fetep:
stack trace:

2012-02-06 07:48:43.447595500 Error: EINVAL, Conversion not supported.
2012-02-06 07:48:43.447599500     at GettextDomain.parseHeaders (/opt/browserid/node_modules/node-gettext/index.js:105:22)
2012-02-06 07:48:43.447602500     at GettextDomain.parseHeaders (/opt/browserid/node_modules/node-gettext/index.js:131:21)
2012-02-06 07:48:43.447605500     at GettextDomain.handle_strings (/opt/browserid/node_modules/node-gettext/index.js:73:14)
2012-02-06 07:48:43.447608500     at GettextDomain.load_tables (/opt/browserid/node_modules/node-gettext/index.js:66:14)
2012-02-06 07:48:43.447676500     at new GettextDomain (/opt/browserid/node_modules/node-gettext/index.js:25:14)
2012-02-06 07:48:43.447679500     at Gettext.addTextdomain (/opt/browserid/node_modules/node-gettext/index.js:221:28)
[...]

PR #1918, opened 2012-07-05
Fix english strings - never use translations for en_US

lloyd:
This is a fix for issue #1905, but not the only neccesary fix.  This should go into dev and there is no rush.  The patch fixes several papercuts related to our "debugging" locale and i18n development / debugging in general.  Also it ignores en_US as a translated locale in many places in the code, as we do not maintain an English translation of our English strings..


.gitignore
```diff
@@ -7,5 +7,6 @@
 /npm-debug.log
 /resources/static/build
 /resources/static/production
+/resources/static/i18n
```
lloyd, line 10: static/i18n is now completely generated, no files are checked into tree.

resources/static/common/js/gettext.js
```diff
@@ -6,10 +6,10 @@
 (function() {
   "use strict";
 
-  function Gettext(params) {
+  function Gettext() {
       return {
         gettext: function (msgid) {
-          if (json_locale_data && json_locale_data["client"]) {
+          if (window.json_locale_data && json_locale_data["client"]) {
```
lloyd, line 12: this code was designed to work if json_locale_data was not defined, but threw an exception when it was removed.  In development we were always downloading a really old client.json file, and for strings it did not know about the code would work ok.  now it works when json_locale_data is not present.

resources/static/common/js/gettext.js
```diff
@@ -33,11 +33,7 @@
       };
   };
 
-  var params = {
-        "domain" : "client",
-        "locale_data" : json_locale_data
-  };
-  var gt = new Gettext(params);
```
lloyd, line 40: dead code!

scripts/merge_po.sh
```diff
@@ -21,12 +21,22 @@ done
...
 if hash podebug >/dev/null; then
+
+    # our debug locale has a tendency to be characterized as 'CHARSET' rather than
+    # UTF-8, this hack works around the problem.
+    # see issue #1054
+    for file in locale/templates/LC_MESSAGES/*.pot ; do
+        mv $file $file.old
+        sed 's/CHARSET/UTF-8/g' $file.old > $file
+        rm -f $file.old
+    done
```
lloyd, line 32: this code works around issue #1054 which was manually fixed once, but keeps popping up and prevents our "debug locale" from working.  it's ugly, but it works and I cannot think of a better solution.

scripts/run_locally.js
```diff
@@ -39,7 +39,12 @@ process.env['HTTP_PROXY'] = HOST + ":10006";
 process.env['HOST'] = HOST
 
 // use the "local" configuration
-process.env['CONFIG_FILES'] = path.join(__dirname, '..', 'config', 'local.json');
+var configFiles = [];
+if (process.env['CONFIG_FILES']) {
+  var configFiles = process.env['CONFIG_FILES'].split(',');
+}
+configFiles.push(path.join(__dirname, '..', 'config', 'local.json'));
+process.env['CONFIG_FILES'] = configFiles.join(',');
```
lloyd, line 47: when running locally, don't stomp user specified config files!  This is especially useful when debugging l10n stuff, which is why it's bundled with this PR.  (yeah right, I just got lazy)

lloyd:
@ozten and @mathjazz - would you mind reviewing this for me?

lloyd:
so @mathjazz has cleaned up en_US from locale/, that's good.  I still think this pull request is very useful, but there is no real urgency on its review and merge.

mathjazz:
@ozten, could you please take a look at this one? I don't really feel I'm capable of reviewing this code without taking a closer look to our i18n infrastructure. But if you're in the same situation, I'll take care of it.

resources/static/i18n/zh_TW/messages.json
```diff
@@ -1,195 +0,0 @@
...
-         null,
-         "（更好的登入機制）"
-      ],
-      "Browser ID": [
-         null,
-         "Browser ID"
-      ]
-   }
-}
-;
```
ozten, line 195: I think removing .json string files is a great idea. We'll want to make @fetep aware of this, as he may be able to improve how he updates l10n preview, since these language files are no longer in the main browserid rpm.

scripts/compress
```diff
@@ -42,7 +42,12 @@ Object.keys(all).forEach(function(resource) {
   // and is bundled into the Big Minified Piles Of Resources we ship.
   // Here we sub the former with the latter.
   var ix = all[resource].indexOf('/common/js/templates.js');
-  if (ix != -1) all[resource].splice(ix, 1, '/build/templates.js');
+  if (ix !== -1) all[resource].splice(ix, 1, '/build/templates.js');
+
+  // remove all i18n en_US resources, they are unnecessary - issue #1905
+  ix = all[resource].indexOf('/i18n/en_US/client.json');
```
ozten, line 48: This will make us less aware of different classes of l10n problems. We won't know about them until localizers complain. If we have to maintain the en_US .po and .json files, then we'll see how string changes affect others, earlier.

scripts/merge_po.sh
```diff
@@ -21,12 +21,22 @@ done
 
 # Optionally auto-localize our test locale db-LB
 if hash podebug >/dev/null; then
+
+    # our debug locale has a tendency to be characterized as 'CHARSET' rather than
+    # UTF-8, this hack works around the problem.
+    # see issue #1054
+    for file in locale/templates/LC_MESSAGES/*.pot ; do
+        mv $file $file.old
+        sed 's/CHARSET/UTF-8/g' $file.old > $file
```
ozten, line 30: We should integrate i18n-abide with browserid.

Now that I"m looking, this was already a feature of that module
https://github.com/mozilla/i18n-abide/blob/master/bin/extract_po.sh#L11

Moving away from node-gettext is new and we should do it systematically in i18n-abide so all Node based projects get that update.

ozten:
r+ with the already stated caveats.

lloyd:
the testing point is a really good one, as is rolling these changes into i18n-abide and moving onto that.

with respect to our awareness of l10n problems, we were not leveraging the en_US.po files for awareness, and we were just shipping broken strings into production.

I'm thinking though that we can introduce unit tests that deeply excercise the process of going from `.json` files to properly translated resources - we can test both our server and client side translation mechanisms in an automated fashion.

I wonder where the right place to test the pipeline of going from .po files to .json files is?

lloyd:
opened issue to address the testing concern: https://github.com/mozilla/browserid/issues/1969

with that, merging down.

Issue #1920, opened 2012-07-05
from skinny: slide the arrow off the screen.


shane-tomlinson:

jrgm:
verified stage train-2012.07.20

PR #1922, opened 2012-07-05
Slide the arrow all the way off the screen.

shane-tomlinson:
In desktop browsers other than IE, when selecting an email address, force the arrow to slide all the way off the screen.

issue #1920


seanmonstar:
Nice, looks better.

Issue #1923, opened 2012-07-05
/signup page - inconsistent "Verify Password" v. "Repeat Password"


jrgm:
On the signup page the label is 'Verify Password' and the placeholder is 'Repeat Password' . Shouldn't they be the same, or is this specifically intended to read this way. (Actually maybe just Password for the placeholder). If it's intended to be this way just close this issue. 

![img](https://img.skitch.com/20120705-c84an889kquc6bitgr4igim8k5.png)

shane-tomlinson:
@jrgm - this affects the forgot password screen as well.

jrgm:
yeah, I just realized /forgot had the same issue. r+ to both.

jrgm:
verified train-2012.07.06 stage /signup and /forgot verify password matches

PR #1924, opened 2012-07-05
Change the placeholder for the verification password from "Repeat Passwo...

shane-tomlinson:
...rd" to "Verify Password" to match the dialog.

issue #1923


jrgm:
r+ shipit

Issue #1916, opened 2012-07-05
front-end tests timing out at /dialog/views/error.ejs


jedp:
getting this:

```
router (15324): info: forwarding request: /dialog/views/error.ejs -> http://127.0.0.1:10007/dialog/views/error.ejs
'waitFor()' timeout
✗ · · 

  frontend unit tests
    ✗ pass!
      » expected 0,
    got  1 (===) // vows.js:91
✗ Broken » 6 honored ∙ 1 broken (180.043s)

```

zaach:
Seems to work for me:

```
router (88508): info: forwarding request: /dialog/views/error.ejs -> http://127.0.0.1:10007/dialog/views/error.ejs
'waitFor()' finished in 65519ms.
Tests completed in 65352 milliseconds.
1202 tests of 1202 passed, 0 failed.
· · · ✓ OK » 7 honored (69.607s)
```

jrgm:
wfm phantomjs on osx

jedp:
It's intermittent for me on Linux.  It works sometimes, times out sometimes.  I just re-ran front-end tests twice.  First time they passed; second time they timed out. :(

jedp:
@zaach - whoah - 69 seconds?  These always take at least 170 sec on my quad i7 2.8GHz with Linux.  Dang!

jrgm:
Offtopic but with a vmware Ubuntu 11.10 running on physical mac quad i7 2.3 I get:

Mozilla/5.0 (X11; Linux i686; rv:16.0) Gecko/16.0 Firefox/16.0
Tests completed in 30112 milliseconds.
1202 tests of 1202 passed, 0 failed.

179 seconds? Huh?

jedp:
Nor I. :)

PR #1926, opened 2012-07-05
Replace setTimeout with event emitter/handler in error-display test

jedp:
In issue #1916, I reported that front-end tests were mysteriously timing out some of the time.  I think the problem may have been in a setTimeout in error-display.

I've updated error-display to use an event listener instead, and that appears to have solved my problem.


resources/static/test/cases/common/js/error-display.js
```diff
@@ -18,17 +18,20 @@
     }
   });
 
-  asyncTest("can initialize and open the error display", function() {
+  asyncTest("can initialize and open the error display", function openErrorDisplay() {
+    bid.DOM.bindEvent(window, "resize", resized_once);
```
jedp: this event was already being emitted by the error module.  Harmless via phantomjs, but as I'm typing this, I wonder whether people could be resizing their browser when running interactively in 10002/test right when this test was being run?  Seems unlikely, but also not completely water-tight.

scripts/phantomrunner.js
```diff
@@ -78,8 +78,8 @@ page.open(phantom.args[0], function(status){
 
                         var failingItems = node.querySelectorAll(".fail");
                         var failingItemsCount = failingItems.length;
-                        for(var i = 0; i < failingItemsCount; i++) {
-                          var failingItem = failingItems.item(i);
+                        for(var j = 0; j < failingItemsCount; j++) {
```
jedp, line 81: happened to notice this while poking around

resources/static/test/cases/common/js/modules/extended-info.js
```diff
@@ -0,0 +1,33 @@
...
+    teardown: function() {
+      $("#error").hide();
+    }
+  });
+
+  asyncTest("can initialize and open the extended info", function openExtendedInfo() {
+    $("#error").show();
+    var errorDisplay = ExtendedInfo.create();
+    errorDisplay.start({ target: "#error" });
+    errorDisplay.open(function() {
```
shane-tomlinson, line 26: @jedp - Here is the change I made in the tests - instead of depending on the "resize" event to be called, I pass a callback to the open function which is called once the extended info is open.

shane-tomlinson:
@jedp - I see your simple pull request and I raise you a yak shave.  I wasn't super thrilled with the way the entire error-display stuff worked, so this is a good time to update.  

Instead of an ErrorDisplay only being used for one specific screen, I decided it makes more sense to have a generic "ExtendedInfo" module.  This module is created whenever any screen is shown, right now it is only shown for the error screen.

All "screens" (dialog, wait, form and delay) can have an "ExtendedInfo" screen now.  Since each screen has a target, the ExtendedInfo looks for the "openMoreInfo" and "moreInfo" elements.  If openMoreInfo is clicked, it slides "moreInfo" down.

This all came from making "open" take a callback instead of an event.

resources/static/common/js/modules/extended-info.js
```diff
@@ -0,0 +1,46 @@
...
+
+      /**
+       * XXX What a big steaming pile, use CSS animations for this!
+       */
+      $(extendedInfoEl).slideDown(function() {
+        // The expanded info may be partially obscured on mobile devices in
+        // landscape mode.  Force the screen size hacks to account for the new
+        // expanded size.
+        dom.fireEvent(window, "resize");
+        complete(oncomplete);
```
jedp, line 38: w00t!

resources/static/dialog/views/error.ejs
```diff
@@ -35,11 +35,11 @@
   <% } %>
 
   <% if(typeof action !== "undefined" || typeof network !== "undefined") { %>
-    <a href="#" id="openMoreInfo">
+    <a href="#" class="openMoreInfo">
```
jedp, line 38: I like this more generic approach.

resources/static/common/js/screens.js
```diff
@@ -13,21 +13,43 @@ BrowserID.Screens = (function() {
...
-        this.visible = true;
+
+        // extendedInfo takes care of info that is on a screen but hidden by
+        // default.  When the user clicks the "open extended info" button, it
+        // is displayed to them.
+
+        if (self.extendedInfo) {
+          // sometimes a screen is overwritten and never hidden.  When this
+          // happens, old extendedInfos need to be torn down.
+          self.extendedInfo.stop();
```
jedp, line 29: Good catch; appreciate the comment here

resources/static/common/js/modules/extended-info.js
```diff
@@ -0,0 +1,46 @@
...
+
+
+  var Module = bid.Modules.PageModule.extend({
+    start: function(config) {
+      var self=this;
+
+      self.checkRequired(config, "target");
+      self.target = config.target;
+
+      var openerEl = self.openerEl = $(".openMoreInfo", self.target);
```
jedp, line 20: Yes, I really like this generic approach.

jedp:
You are the original hairdresser to Mongolian livestock.

This is great.  I really like this new approach, both for its new flexibility and its testability.  I was definitely not feeling good about my event-listener hack; it felt too fragile, and I was going to reject it myself this morning, but then I found this instead!

r+

Issue #1325, opened 2012-03-16
Scrollbar is not obvious in landscape mode; 'Sign in' button not visible


krupa:
Stock browser/ Android 2.2

Steps to reproduce:
1. Hold the phone in landscape mode (This is important since things work fine in Portrait mode).
2. Load https://marketplace-dev.allizom.org
3. Log in using a BrowserID account which has multiple email ids associated with it.
4. Select an email id with which you want to log in with.

expected behavior:
There is a 'Sign in' button for people to click

observed behavior:
In landscape mode, the 'Sign in' button is below the fold and users need to scroll to be able to view this. However, the scrollbar is not obvious and is shown only when people try to scroll.

sawyerh:
@skinny97214: 

Proposed design solution:
- Currently: http://d.pr/i/4JvG and http://d.pr/i/jqnG
- Proposed solution: http://d.pr/i/WIWB (at least for landscape)

skinny97214:
Can we shorten the input enough to slide in the [next] button to its right?

sawyerh:
@skinny97214 Something more like this (http://d.pr/i/b13C)? Are you aware of any button labels that are longer than 4 characters (next, done, back)?

skinny97214:
Yah, that's more of what I was thinking. We should check with Matej to see
what the worst case localization is. I do prefer this one if it can work in
all locales. Can it wrap to the line below if it exceed a certain width?
Need some schmancy detection I'm guessing.

sawyerh:
@skinny97214 - If the problem we're trying to solve is the button not being immediately visible, the small button won't always work for screens that have multiple fields or screens that have a lot of text above the fields (example: http://d.pr/i/NExy). This is an alternative to the original solution I posted - http://d.pr/i/31ys

I'm not sure if the button being "below the fold" is a major issue though — people tend to know how to scroll the window. Plus, both solutions proposed won't work for screens in portrait mode when we only have 320px to work with. The button would have to flow below the fields again.

skinny97214:
@sawyerh I agree the "below the fold" isn't the worst issue. I do think it's ideal if it's above in landscape, but not 100% required for every screen given all our other constraints. My prediction is that if it looks horrible in landscape, most users will rotate to portrait. Incidentally, I'm moving toward having only one password field, so that page will get better on mobile.

The blue arrow is promising but I'm not convinced it's actionable enough. I wouldn't be comfortable launching this without user testing it. It also doesn't support cancel.

[...]

PR #1949, opened 2012-07-06
Tweak dialog mobile sizes. Prevent iOS from making text huge.

sawyerh:


lloyd:
/me leaves this to @seanmonstar or @shane-tomlinson

shane-tomlinson:
@sawyerh - is this in reference to an issue?

shane-tomlinson:
@sawyerh - I am holding off on merging this, there are several issues that need addressed.  Making one layout work for both desktop and mobile is a pain.

resources/static/dialog/css/m.css
```diff
@@ -27,8 +28,9 @@
   }
 
   .inputs > li > label {
-    font-size: 18px;
-    margin-bottom: 10px;
+    font-size: 16px;
+    margin-bottom: 5px;
+    overflow: visible;
```
shane-tomlinson: @sawyerh - The `overflow: hidden` causes long email addresses to show the ellipsis instead of having the email addresses run off the end of the screen.  By setting it to visible, long email addresses go off the end again.

Perhaps a better place to put the overflow: hidden and text-overflow: ellipsis is on '#selectEmail > .inputs > li > label'  Or maybe we need a new class for email address labels.

resources/static/dialog/css/m.css
```diff
@@ -142,6 +148,7 @@
     position: relative;
     top: 0;
     bottom: 0;
+    padding-bottom: 20px;
```
shane-tomlinson, line 160: This makes a dark band appear in the footer.  The top half is dark, the bottom half is light.

sawyerh:
@shane-tomlinson This was for issue #1325. I'll address the issues you mentioned and recommit.

resources/static/dialog/css/m.css
```diff
@@ -98,12 +107,18 @@
...
+    display: block;
+  }
+
   .pickemail #signIn .table {
       padding-top: 0;
   }
 
   #signIn .vertical {
       position: static;
+      display: block;
```
sawyerh, line 121: Removing the `display: table` and `display: table-cell` from `.table` and `.vertical` fixes an issue where very long emails would push the content off the screen instead of cutting off the email with an ellipsis (screenshot of bug: http://d.pr/i/OFfn)

resources/static/dialog/css/m.css
sawyerh, line 160: This is fixed by setting the `height: auto` on `html, body`

sawyerh:
@shane-tomlinson Things should look better now. I also caught a bug with long emails pushing the content off the screen (http://d.pr/i/OFfn), which should now be fixed.

resources/static/dialog/css/m.css
shane-tomlinson, line 121: @sawyerh - the table/table-cell allows the content to be vertically centered.

shane-tomlinson:
@sawyerh - thanks for the update.  I am having a closer look at this.  Removing the table/table-cell display removes the vertical centering of the forms on larger mobile screens.  All form fields are pushed to the top of the screen leaving a large space at the bottom.

Forcing the forms to be at the top of the screen does have an advantage with virtual keyboards - it eliminates the problem on Android where the email field would be obscured when the VKB is displayed.  Giving this more thought.

shane-tomlinson:
@sawyerh - great work with this.  

I appended a small change for additional cleanup (4cfcead3f357d4d794cfb5beb9ba804a56fe6016) to this PR and merged manually.  The only major difference is I keep the contents vertically centered.  I want to experiment on some large screen mobile devices before allowing the form content float to the very top.  I worry on devices such as the Galaxy SIII, it will be difficult to select email addresses with your thumb if the device is held in the right hand.

Merged with checkin f8c8e4734d256e2cafa0b6a2fb1855df7b705e99

r+

PR #1953, opened 2012-07-06
make sign up button display inline-block

### LINKED ISSUE #1885: win7/opera 11.64 - main site layout issue with split signup button
I'm splitting out this issue from GH-1856 where initially I thought both the font and signup button layout issue were just for Opera. The font issue also affects other browsers on windows; I've only seen the split signup button on Opera/windows. 

So this issue is about the "split button" layout. 

img

seanmonstar:
fixes #1885


seanmonstar:
@shane-tomlinson This fixed the issue, but there might be a better way. It still looks correct in other browsers and with the width wider.

Reduce screen width to see issue.

Issue #2, opened 2011-04-16
IP should check for primary support upon email addition


lloyd:
While the user is adding a new email, if the email is from a host who natively supports BrowserID, then the IP should query host-meta to determine the URL of the UI for that support.  That UI should be embedded into the window spawned by the IP so that the user can log in.

Part of this is smoothing out the spec to which primaries will author auth code (postMessage?  size?  etc).

lloyd:
this is implemented in dev along with primary support.

PR #1954, opened 2012-07-06
setting a timeout for provisioning of 20s

### LINKED ISSUE #1570: Certificate provisioning doesn't time out
When using primary IdPs on various demo sites which use diresworb I've seen a hang during provisioning.

It should not hang - if IdPs don't respond the way we expect (e.g., because they are using a shim that talks to a different domain), it should time out and fail.

seanmonstar:
if `teardown` doesn't happen within 20 seconds, provisioning will `fail` with a timeoutError.

I couldn't find where these `msg` and `code` lineup with anything in the UI. However, #1570 doesn't say anything about the UI. That's in another issue.

fixes #1570

### DIFF
--- a/resources/static/common/js/provisioning.js
+++ b/resources/static/common/js/provisioning.js
@@ -57,6 +57,7 @@ BrowserID.Provisioning = (function() {
     // positives if the user is on a slow connection.
     // the timeout should only happen if the provisioning site doesn't
     // want to provision for us.
+    // see https://github.com/mozilla/browserid/pull/1954
     function iframeOnLoad() {
       if (timeoutID) {
         clearTimeout(timeoutID);
--- a/resources/static/common/js/provisioning.js
+++ b/resources/static/common/js/provisioning.js
@@ -53,6 +53,10 @@ BrowserID.Provisioning = (function() {
     iframe.setAttribute('src', args.url);
     iframe.style.display = "none";
 
+    // start the timeout once the iframe loads, so we don't get false
+    // positives if the user is on a slow connection.
+    // the timeout should only happen if the provisioning site doesn't
+    // want to provision for us.
     function iframeOnLoad() {
       if (timeoutID) {
         clearTimeout(timeoutID);
--- a/resources/static/common/js/provisioning.js
+++ b/resources/static/common/js/provisioning.js
@@ -52,6 +52,24 @@ BrowserID.Provisioning = (function() {
     var iframe = document.createElement("iframe");
     iframe.setAttribute('src', args.url);
     iframe.style.display = "none";
+
+    function iframeOnLoad() {
+      if (timeoutID) {
+        clearTimeout(timeoutID);
+      }
+      // a timeout for the amount of time that provisioning is allowed to take
+      timeoutID = setTimeout(function provisionTimedOut() {
+        fail('timeoutError', 'Provisioning timed out.');
+      }, MAX_TIMEOUT);
+    }
+
+    if (iframe.addEventListener) {
+      iframe.addEventListener('load', iframeOnLoad, false);
+    } else if (iframe.attachEvent) {
+      iframe.attachEvent('onload', iframeOnLoad);
+    }
+    // else ruh-roh?
+
     document.body.appendChild(iframe);
 
     var chan = Channel.build({
@@ -97,10 +115,6 @@ BrowserID.Provisioning = (function() {
       successCB(keypair, cert);
     });
 
-    // a timeout for the amount of time that provisioning is allowed to take
-    timeoutID = setTimeout(function provisionTimedOut() {
-      fail('timeoutError', 'Provisioning timed out.');
-    }, MAX_TIMEOUT);
   };
 
   return Provisioning;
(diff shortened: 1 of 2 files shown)


lloyd:
my only fear here is ultra high latency connections.  I was at such a place on friday, and I think I might hit a false positive in such a case...

Any ideas on a simple approach that will allow us to handle such scenarios gracefully?  waiting for the RP's iframe to load in some cross browser fashion?

r/ (that's like, between an r+ and r-.  I'm feeling indecisive).

seanmonstar:
So, what happens currently? Does the iframe not load at all, or does the iframe load, and the server says that it can't talk to dev.direwsword.org or whichever, and so doesn't do that provisioning?

If the iframe always loads, but then the page loaded won't communicate, I could perhaps do something with the iframe's `readyState`...

lloyd:
1: currently if things are slow, the user just waits and things complete.

2: currently if the IdP is broken, the user just waits and the screen never changes.

I feel a little bit like we fix #2 and possible break #1.

@seanmonstar - as a first fix, if you could wait untiil the iframe is loaded (give it 15s to load?) then start the timer after it loads, maybe that would be a bit more robust?  

don't feel too strongly here, except it'd be a shame if a commit designed to improve user feedback prevented users from using things :/

what do you think?

seanmonstar:
@lloyd I can make the timeout start from `iframe.onload`

lloyd:
love it.  that's enough for an r+ from me.

reliable on ie8?

seanmonstar:
@lloyd attachEvent onload works in IE. added commit using this.

here's a jsfiddle of just that part: http://jsfiddle.net/Aajc3/

lloyd:
pending silly little comment update, as far as I'm concerned, you're free to merge @seanmonstar.

Attention @jbonacci & @jrgm:  This change makes it so that when you log in with a primary, there is a maximum allowed time for the primary to respond.  Ideally we'd have a way to _really_ crank up latency to a realistic worst case and see if we accidentally hit this limit.  Perhaps it's sufficient to test logging in with eyedee.me emails from the worst wifi you know, and via mobile in a crappy coverage area...

this will likely go into train-2012.07.20

resources/static/common/js/provisioning.js
```diff
@@ -47,6 +52,24 @@ BrowserID.Provisioning = (function() {
     var iframe = document.createElement("iframe");
     iframe.setAttribute('src', args.url);
     iframe.style.display = "none";
+
```
lloyd, line 55: let's include a comment here which describes the rationale behind waiting for iframe load?  yeah, I know, it's a nit.

seanmonstar:
Do you mean remove the comment that makes you giggle?

On Tuesday, July 10, 2012, Lloyd Hilaiel wrote:

> pending silly little comment update, as far as I'm concerned, you're free
> to merge @seanmonstar.
> 
> Attention @jbonacci & @jrgm:  This change makes it so that when you log in
> with a primary, there is a maximum allowed time for the primary to respond.
>  Ideally we'd have a way to _really_ crank up latency to a realistic worst
> case and see if we accidentally hit this limit.  Perhaps it's sufficient to
> test logging in with eyedee.me emails from the worst wifi you know, and
> via mobile in a crappy coverage area...
> 
> this will likely go into train-2012.07.20
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/mozilla/browserid/pull/1954#issuecomment-6897713

## 

Sean McArthur
http://seanmonstar.com

seanmonstar:
Oi, never mind. I got the email late.

On Tuesday, July 10, 2012, Sean McArthur wrote:

> Do you mean remove the comment that makes you giggle?
> 
> On Tuesday, July 10, 2012, Lloyd Hilaiel wrote:
> 
> > pending silly little comment update, as far as I'm concerned, you're free
> > to merge @seanmonstar.
> > 
> > Attention @jbonacci & @jrgm:  This change makes it so that when you log
> > in with a primary, there is a maximum allowed time for the primary to
> > respond.  Ideally we'd have a way to _really_ crank up latency to a
> > realistic worst case and see if we accidentally hit this limit.  Perhaps
> > it's sufficient to test logging in with eyedee.me emails from the worst
> > wifi you know, and via mobile in a crappy coverage area...
> > 
> > this will likely go into train-2012.07.20
> > 
> > ---
> > 
> > Reply to this email directly or view it on GitHub:
> > https://github.com/mozilla/browserid/pull/1954#issuecomment-6897713
> 
> ## 
> 
> Sean McArthur
> http://seanmonstar.com

## 

Sean McArthur
http://seanmonstar.com

lloyd:
I should have reference the comment AFTER writing it.  I blame the internets!

jrgm:
You can induce packet latency on osx/bsd with the following (adjust 1000ms to what you want; 1000ms will get 2000ms packet RTT):

```
sudo ipfw add pipe 1 tcp from any to any 80,443 out
sudo ipfw add pipe 2 tcp from any 80,443 to any in
sudo ipfw pipe 1 config delay 1000ms
sudo ipfw pipe 2 config delay 1000ms
sudo ipfw list
# sudo ipfw -q flush; # deletes all rules (can be more specific if you need).
```

But that's from bsd client to RP and "browserid.org", and doesn't cover "browserid.org" to the primary IdP (except when running local dev). I can look into the current state of iptables/netem/ipfw on linux for an awsbox test env.

seanmonstar:
@lloyd added a comment for your enjoyment

lloyd:
@seanmonstar this time it didn't make me giggle.  but I sure do like it.

r+

Issue #1958, opened 2012-07-07
backend unit tests failing on tests/static_resources.js


jrgm:
Travis says (and is right)

✓ OK » 1 honored (0.008s)
1892tests/stalled-mysql-test.js
1893······································· 
1894✓ OK » 39 honored (13.250s)
1895tests/static-resource-test.js
1896✗ 
1897
1898
1899  All resources expand
1900    ✗ We get stuff
1901      » expected 12,
1902    got  9 (==) // static-resource-test.js:35
1903✗ Broken » 1 broken (0.005s)
1904
1905npm ERR! browserid@0.0.1 test: `./scripts/test`

I think https://gist.github.com/3063839 is a fix (it's really a one-line, but I also reformatted the really long line).

lloyd:
fixed in dev and train-2012.07.06 on browserid_private.  thanks for the find, and the fix.

jrgm:
verified on the other repo train-2012.07.06

Issue #1937, opened 2012-07-06
signup of primary on main site /signup fails (404 on static.dev.anosrep.org/lib/winchan.js) 


jrgm:
STR
1. Go login.dev.anosrep.org and click sign in.
2. enter &lt;something&gt;@eyedee.me and click the sign in button (where <something> can be either an existing or unknown address to eyedee.me).
3. screen transitions to show a verify button. Click the button.

Expected: dialog comes up with /authenticate_with_primary, loads /lib/winchan.js and /auth_with_idp/main.js and screen moves to eyedee.me/browserid/sign_in.html?email=...

Actual: dialog to /authenticate_with_primary comes up and is blank

Notes:
1. httpfox shows request to https://static.dev.anosrep.org/lib/winchan.js is 404
2. error console shows Error: WinChan is not defined, Source File: https://static.dev.anosrep.org/auth_with_idp/main.js, Line: 5
3. curl -v https://dev.anosrep.org/lib/winchan.js returns 301 to https://login.dev.anosrep.org/lib/winchan.js but that returns 404
4. this does not happen in current stage or in production
5. I am boring even myself

shane-tomlinson:
@jrgm - This issue goes deeper than the winchan problem.  Once the winchan include is fixed, there are problems re-starting the dialog once the user has authenticated with the IdP.  Digging in.

shane-tomlinson:
@benadida - The sessionStorage change from PR #1858 breaks authenticating to an IdP from the main site.

https://github.com/mozilla/browserid/blob/dev/resources/static/authentication_api.js#L36-40 shows that completeAuthentication and reaiseAuthenticationFailure are hard coding the return URLs back to the normal authentication dialog, which initially seemed like a good idea.  The problem is when authentication to an IdP occurs from the main site, the IdP is supposed to redirect back to https://login.persona.org/authenticate_with_primary - this is why we had the return_to parameter on the URL.

shane-tomlinson:
Marked 5\* since this is a regression.

benadida:
I'll get on this today.

benadida:
Now... this looks like it's not covered in tests. @shane-tomlinson any chance you can help me there?

lloyd:
fixed with PR #1955

jrgm:
works in dev for the original STR. Will need to reconfirm when we get to stage.

jrgm:
verified in train-2012.07.30 and in prior. Primary signin works.

PR #1959, opened 2012-07-07
fix issue GH-1958; add '/production/authenticate_with_primary.js' to ite...

jrgm:
...ms to exclude and reformat the single line array

Seems right to me and back/back_mysql/front tests pass locally, but I'm only pattern matching from the commit to fix GH-1937 https://github.com/mozilla/browserid/commit/a3e88e7815cc38d531ad1a9e8d8842c64f344d4c

### DIFF
--- a/tests/static-resource-test.js
+++ b/tests/static-resource-test.js
@@ -23,7 +23,15 @@ suite.addBatch({
       var res = resources.resources;
       assert.ok(files['/production/dialog.css'].length >= 3);
       // Get ride of non-localized asset bundles
-      ['/production/communication_iframe.js', '/production/include.js', '/production/dialog.css', '/production/browserid.css', '/production/ie8_main.css', '/production/ie8_dialog.css', '/production/relay.js', '/production/html5shim.js'].forEach(
+      ['/production/communication_iframe.js',
+       '/production/include.js',
+       '/production/dialog.css',
+       '/production/browserid.css',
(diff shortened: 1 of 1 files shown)


seanmonstar:
looks correct. that test is making sure assets that need to be localized are available correctly. this new file doesnt need it, I assume. r+

tests/static-resource-test.js
```diff
@@ -23,7 +23,15 @@ suite.addBatch({
...
-      ['/production/communication_iframe.js', '/production/include.js', '/production/dialog.css', '/production/browserid.css', '/production/ie8_main.css', '/production/ie8_dialog.css', '/production/relay.js', '/production/html5shim.js'].forEach(
+      ['/production/communication_iframe.js',
+       '/production/include.js',
+       '/production/dialog.css',
+       '/production/browserid.css',
+       '/production/ie8_main.css',
+       '/production/ie8_dialog.css',
+       '/production/relay.js',
+       '/production/html5shim.js',
+       '/production/authenticate_with_primary.js'].forEach(
```
lloyd, line 34: thank you for using newlines.  I like newlines.

lloyd:
concur with @seanmonstar - r+

jrgm:
Hmm. @lloyd did you intend to not pick up this fix in train-2012.07.06?

jrgm:
Ignore previous; I was on the wrong tree.

Issue #1917, opened 2012-07-05
why do we need node-gettext?


lloyd:
node gettext requires iconv, which is the slowest building native dependency we have.  The frontend uses .json files, while the backend reads directly from .mo files.

I wonder.

because our i18n flow generates both `client.json` (strings needed on the client) files, and `messages.json` (strings needed on the server), could we use the `messages.json` strings (instead of binary .mo files) on the server and have a simpler i18n process?

The reasons I'm interested in exploring this:
1. building iconv@1.1.3 accounts for **65%** of the time it takes to `npm install`
2. travis-ci often kills our tests cause they take too long (see #1)
3. this version of iconv crashes on node 0.8 - not important now, but will become important in the future
4. there will probably be a performance improvement
5. it probably will simplify our code

I think we could move off of node-gettext at very low engineering cost and remove this costly and fragile dependency, but it's worth noting that iconv@1.2.1 builds 3x faster... I don't know if there are breaking changes there...

Now if we chose to use a pure javascript implementation of gettext, would backing it with mofiles be better than backing it with the json files that we're already generating?

seanmonstar:
if this is the only reference to iconv, then that would remove a big blocker to installing on Windows (#1769)

lloyd:
fwiw, I tried the update to iconv@1.2.1 and it makes our build about 20% faster.  meh.  PR 1962 is moar awesome.

Issue #1675, opened 2012-06-01
second password field in flow is labelled 'reset password'


jrgm:
1. Create a secondary email address on browserid
2. Clear all profile data in a browser
3. Go to an RP click 'sign in' to bring up the dialog. 
4. Enter the known secondary email address and then click password. 

Expected: 
Prompted to enter a new password twice 

Actual: 
Prompted to enter a new password twice, but the second one is labelled 'reset password', which seemed to me and @tonychung  to be a confusing way to label that field. (Reset? What's a reset password? Is this supposed to be the same as the first, or is it something different?). 

![img](https://img.skitch.com/20120601-gq6d59w6c6q1n4jgi3ag9gm56n.png)

jrgm:
actually, this is shown for either forgot password, or in signing up a new account. Funny how I looked right past it until Tony mentioned it. :-/

jrgm:
Ping. Any chance of getting this phrase fixed for the pending train. Just "Password" would be better than 'reset password'. 
See http://irclog.gr/#show/irc.mozilla.org/identity/113028 for @AutomatedTester's comments.

jrgm:
By the way, this is how dev currently appears, with focus tabbed to the 'done' button. 

![img](https://img.skitch.com/20120703-xe5345td314pus217gpyegdkp2.png)

lloyd:
this issue is the same as GH-1905 - it is a problem with our english language "translations" we need to get this sorted as it's visible in stage too.

lloyd:
fixed in dev.

rolling out to stage: https://bugzilla.mozilla.org/show_bug.cgi?id=767708#c26

will be in production monday.

closing down this issue.

PR #1962, opened 2012-07-07
don't use node-gettext

lloyd:
Rather than directly reading binary files in our server at runtime to localize our views, we can read json files that we generate from those binary files already in a batch step when we extract strings during our localization process.

This change removes our largest dependency, cuts our `npm install` time way down which reduces our travis test time to 18 minutes down from about 28 minutes - more importantly it fixes the frequent errors we see on travis because our build takes too long.  Also, this removes one of the blockers for developers to be able to work on windows (like the @seanmonstar).  I expect it will shave a couple minutes off our dev deploy time too, once merged.  I'll bet you two US dollars it makes our servers a bit more efficient, too.

On the downside, we don't get some fancy features of node-gettext that we were not using anyway.

issue #1917


lib/i18n.js
```diff
@@ -55,39 +54,41 @@ exports.abide = function (options) {
...
+    var l = (lang === options.debug_lang ? 'db_LB' : localeFrom(lang));
+
+    // ignore .json files for en-US
+    if (lang == 'en-US') return;
+
+    try {
+      // populate the in-memory translation cache with client.json, which contains
+      // strings relevant on the server
+
+      // XXX: these files should be json.  not javascript.
```
lloyd, line 73: I chose to make a minimal change to start, but this is ugly.  changing it I believe will require working with @mathjazz

lib/i18n.js
```diff
@@ -55,39 +54,41 @@ exports.abide = function (options) {
...
+      var json_locale_data; // for jshint
+      eval(fs.readFileSync(path.join(json_dir, l, 'messages.json')).toString());
+      translations[l] = json_locale_data.messages;
```
lloyd, line 76: we cache the entire json file in memory. notta soa biga.

lib/i18n.js
```diff
@@ -55,39 +54,41 @@ exports.abide = function (options) {
...
-        if (mo_cache[l].json_exists && l == default_locale) {
-          // mo files aren't critical... carry on
-          if (! options.disable_locale_check) logger.warn(msg);
-        } else {
-          logger.error(msg);
-          throw msg;
-        }
+      // verify that client.json is present
+      if (!existsSync(path.join(json_dir, l, 'client.json'))) {
+        throw 'client.json';
```
lloyd, line 80: this is just a sanity check, and it's not strictly necc.  it's an attempt to die loudly if it appears that strings weren't properly extracted

lib/i18n.js
```diff
@@ -119,27 +120,18 @@ exports.abide = function (options) {
...
-      req.ngettext = identity;
+    var gt;
+
+    if (translations[locale]) {
+      gt = function(sid) {
+        return (translations[locale][sid] ? translations[locale][sid][1] : sid);
+      };
+    } else {
+      gt = function(a) { return a; }
     }
```
lloyd, line 131: NOTE:  I did _not_ implement ngettext - we aren't using it anywhere.  It is not too hard to implement if we do want to start using it to handle phrases that talk about some number of thingamabobbers.

package.json
```diff
@@ -16,11 +16,9 @@
         "ejs": "0.4.3",
         "etagify": "0.0.2",
         "express": "2.5.0",
-        "iconv": "1.2.1",
         "mustache": "0.3.1-dev",
         "jwcrypto": "0.3.2",
         "mysql": "0.9.5",
-        "node-gettext": "0.1.1",
```
lloyd, line 23: these are my favorite two lines!

lib/i18n.js
seanmonstar, line 76: I have no idea how much memory that would take up, but would that mean that less used languages would be taking up a meaningful amount of memory that could be used to cache other data?

lib/i18n.js
seanmonstar, line 80: how would you feel about a slightly more descriptive sanity check?

`throw new Error('client.json file doesn't exist in ' + path.join(json_dir, l));`

seanmonstar, line 80: nevermind me, I didn't see the `catch` underneath that logs a better message.

package.json
seanmonstar, line 23: I don't know if you use it in there, but nodemailer will try to use iconv if it exists (but it works fine without it).

ozten:
Although I recognize the benefit of getting rid of iconv, I have the following concerns:
- Optimizing for build time instead of runtime performance
  *\* we can fix the build / CI environment
- Throwing away trusted, working code for new code
- We may not need the fancy features of node-gettext today, but we may need them tomorrow and they'd be already built
  *\* If all of our strings are interpolated client side in the future, this isn't a valid point, as we had to roll our own client side gettext library.

As stated in PR 1918, please update i18n-abide with the decisions you make in this pull request, so we have a consistent platform.

lloyd:
thanks for weighing in here @ozten.

**runtime performance**:  My intuition tells me that this is a runtime improvement in terms of both memory and CPU.  (note @seanmonstar before we were keeping files in memory as well).  But screw _intuition_, I'll generate numbers and attach them to the PR.  This code is exercised every time an email is sent (something we could optimize), so it's worth a little rigor.

**trusted working code** - because this code really provides us with a tiny bit of functionality that is simple to exhaustively test, and we shave 10 lines of code and reduce the complexity of our system with this change, I think it's worth it.

**fancy features** - because we do run gettext on the client side, what you said.  

The thing I like most about this change is that we make the contract between persona the product and the translators of that product much simpler.  Basically, our software on both client and server needs json files that contain translated strings in UTF-8.  We don't care how we get those.  The increased symmetry between client and server and reduced contract between persona the product and those who translate it are compelling to me.  (and I'm guilty here of sticking to what I know here, this is a pattern I've applied in the past and when the translation mechanism underneath us changed completely, we were insulated and barely felt it)

lloyd:
well, I forgot to mention that decorating strings with `gettext( )` is a requirement for extraction to work, and part of that contract...  that's common to both a future where this change is merged, and one where it is not.

ozten:
Cool. The contract "with those translating it" is SVN and po files, but I agree having the same code for client gettext and server side gettext makes maintenance easier.

lloyd:
yeah, you're right, I'm wrong.  I guess what I'm actually trying to do is separate persona the product from the build system that turns `.po` files from svn into `.json` files that it eats.  So I guess really I'm saying it's `.mo` files that can become an implementation detail of that build system, and in fact, local/compile-mo.sh becomes obsolete straightaway.

lib/i18n.js
```diff
@@ -55,39 +54,41 @@ exports.abide = function (options) {
...
+
+  var debug_locale = localeFrom(options.debug_lang);
 
   options.supported_languages.forEach(function (lang, i) {
-    var l = localeFrom(lang),
-        default_locale = localeFrom(options.default_lang);
+    var l = (lang === options.debug_lang ? 'db_LB' : localeFrom(lang));
+
+    // ignore .json files for en-US
+    if (lang == 'en-US') return;
```
shane-tomlinson, line 67: @lloyd - @ozten pointed out in another PR that he was not comfortable with not using translations for en-US - does that hold here?

lib/i18n.js
```diff
@@ -55,39 +54,41 @@ exports.abide = function (options) {
...
+  var json_dir = path.resolve(
+          path.join(__dirname, '..'),
+          path.join(options.i18n_json_dir));
+
+  var debug_locale = localeFrom(options.debug_lang);
 
   options.supported_languages.forEach(function (lang, i) {
-    var l = localeFrom(lang),
-        default_locale = localeFrom(options.default_lang);
+    var l = (lang === options.debug_lang ? 'db_LB' : localeFrom(lang));
```
shane-tomlinson, line 64: @lloyd - is "debug_lang" supposed to be defined in config/production.json?  What happens when somebody translates Swiss-Italian?

lib/i18n.js
shane-tomlinson, line 76: @lloyd - the entire cache is only populated once per server start?

lloyd:
performance change:

Before, with node-gettext: Requests per second:    170.62 [#/sec](mean)

After, without node-gettext: Requests per second:    184.13 [#/sec](mean)

So this patch is runtime performance win, but again, we can and will cache this stuff so this isn't a really useful metric.

(to get these numbers, I did: ab -n 10000 -H 'Accept-Language: de' http://127.0.0.1:10002/

lib/i18n.js
```diff
@@ -55,39 +54,41 @@ exports.abide = function (options) {
...
+    // ignore .json files for en-US
+    if (lang == 'en-US') return;
+
+    try {
+      // populate the in-memory translation cache with client.json, which contains
+      // strings relevant on the server
+
+      // XXX: these files should be json.  not javascript.
+      var json_locale_data; // for jshint
+      eval(fs.readFileSync(path.join(json_dir, l, 'messages.json')).toString());
```
seanmonstar, line 75: I just noticed this `eval` here. Do we need to do that? nodejs can `require` json files just fine.

`var json_data = require(path.join(json_dir, l, 'messages.json'));`

lib/i18n.js
jedp, line 131: I'm surprised we don't use ngettext.  I would have expected us to be counting thingamabobbers like email accounts and identities etc.  True, it shouldn't be hard to implement if we ever need it.

lib/i18n.js
lloyd, line 76: according to memwatch, it costs us  "size": "930.64 kb" of memory to cache all of our locales.

lib/i18n.js
lloyd, line 75: this eval sucks, I agree.  Because the script that generates the messages.json file is in svn next to our translation files, and because it generates JS and not JSON, I thought we would fix this in a subsequent pass.

I also think we should move scripts that read from SVN and generate our json should live in our repository...  That we are free to change how we consume stuff without requiring coordination with external teams.

lib/i18n.js
lloyd, line 67: I added an issue for this: https://github.com/mozilla/browserid/issues/1969

I think the concern that we need to be testing our extraction works right is completely valid.  

I think how we were doing it (using en-US po files in production and having the way our users see our software different from the way the devs do), was broken - evidenced by broken strings making it out to users in issue #1675

I'm the one being the opinionated dogmatic guy about this, so I take responsibility for adding that testing!

lloyd:
```
17:58 < lloyd> ozten: how u feeling about PR 1962?
17:59 < ozten> lloyd: I trust you (and others that weighted in) to make the right call
17:59 < ozten> I personally think we've got bigger fish to fry
18:01  * zcarter would enjoy a fish fry
```

lib/i18n.js
seanmonstar, line 75: So it generates a JS file with a .json extension? That's less than good.

lloyd:
I have talked about this PR with anyone who will listen.  I even talked it through out loud with my dog.

I remain convinced that the symmetry between how client and server access translated strings is a good thing.  Reducing our dependencies on external software, where reasonable, is a good thing (in this case we shave lines of code and simplify our string compilation phases by removing the external dependency).  Making our deployments twice as fast is also pretty nice, it means travis starts working again (this is the fastest and cheapest way to make that happen), it means ephemeral deployments spin up faster, it means dev deploys faster.

Finally, because this change is completely within i18n.js, it is easy to revert.

My intention is to merge this tomorrow in the early afternoon.  Please speak up before then if there's a good reason not to, of if there are changes required in the patch!

Issue #1783, opened 2012-06-18
IE8: Unit tests do not complete when slow script warning disabled.


shane-tomlinson:
STR)

1) Open an IE8 VM
2) Disabled the "A script on this page is causing Internet Explorer to run slowly" error by following the steps in http://support.microsoft.com/kb/175500
3) Reload IE
4) Start the unit tests.  They do not complete.

jrgm:
verified I get a clean no error run on winxp/ie8 with awsbox instance of train-2012.07.20 (and of course with the registry hack to allow long scripts).

PR #1966, opened 2012-07-09
Fix up controllers/rp_info front end tests in IE8.  Only the assertion timeout check fails now.

shane-tomlinson:
Testing:

Please see the issue for instructions to disable the "this script is taking too long" warning.

@lloyd - there is only one test failing in IE8 now, that is the check to make sure an assertion's expiration time is valid.  In IE8, it is off by a couple of seconds, could you have a look at this?

issue #1783

### DIFF
--- a/resources/static/test/cases/common/js/user.js
+++ b/resources/static/test/cases/common/js/user.js
@@ -53,6 +53,7 @@ var jwcrypto = require("./lib/jwcrypto");
 
       // expiration date must be within 5 seconds of 2 minutes from now - see
       // issue 433 (https://github.com/mozilla/browserid/issues/433)
+      // An IE8 VM takes about 7 seconds to generate an assertion.
       var diff = Math.abs(expires - nowPlus2Mins);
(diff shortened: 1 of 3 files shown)


lloyd:
so a long time ago when we knocked down keylength, we didn't do the same thing in our frontend tests.  The only browser where that will show up is IE8. Fixed, commit added.

lloyd:
BTW: it is no longer neccesary to disable the "this script is taking too long" message. and in fact, I don't think we should do that.  it would prevent us from realizing our users are seeing that in prod.

lloyd:
verified that all tests pass on IE8.  r+

NOTE: the only thing I'm seeing is 1/8 test runs fail on this test:

http://10.0.0.20:10002/test/?filter=shared%2Ftooltip%3A%20show%20long%20tooltip%2C%20takes%20about%205%20seconds

IE8 appears to not be waiting the full 4500ms in this test, and setTimeout appears to be invoking its callback EARLY.  10mins of googling showed nothing up.  The fix would be to wrap setTimeout, but I'm just going to ostrich this for now and merge the PR, as its much better than it was!

PR #1975, opened 2012-07-10
state machine refactor.

shane-tomlinson:
1) remove requiredEmail left overs.
2) remove TOS/PP leftovers that were there for requiredEmail.
3) make global state variables instance state variables.
4) update/remove tests that tested requiredEmail.


lloyd:
reassigning to @seanmonstar 

Shane, there's a lot of balls up in the air here.  Is this patch still relevant?

shane-tomlinson:
@lloyd - This patch is still relevant - it removes dead code paths that were used for requiredEmail as well as make all state variables members of the class instead of being static to all instances.

resources/static/dialog/js/misc/state.js
```diff
@@ -117,17 +110,7 @@ BrowserID.State = (function() {
     handleState("authentication_checked", function(msg, info) {
       var authenticated = info.authenticated;
 
-      if (requiredEmail) {
```
shane-tomlinson, line 120: we can never get a requiredEmail option from the frontend now, this is no longer needed.

resources/static/dialog/js/misc/state.js
```diff
@@ -166,23 +149,6 @@ BrowserID.State = (function() {
       // know when we are losing users due to the email verification.
       mediator.publish("kpi_data", { new_account: true });
 
-      _.extend(info, {
```
shane-tomlinson, line 169: requiredEmail is no longer used, cancelable defaults to true in modules/set_password.  This set cancelable to true as long as requiredEmail was not set.  since requiredEmail is no longer set, depend on the default.

The TOS/PP was only shown on the set password screen for requiredEmail.  Since requiredEmail is no longer used, these options can be ignored.

resources/static/dialog/js/misc/state.js
```diff
@@ -243,43 +208,26 @@ BrowserID.State = (function() {
...
 
     handleState("primary_user_unauthenticated", function(msg, info) {
-      // a user who lands here is not authenticated with their identity
-      // provider.
       _.extend(info, {
-        add: !!addPrimaryUser,
-        email: email,
-        requiredEmail: !!requiredEmail,
-
-        // In the requiredEmail flow, a user who is not authenticated with
```
shane-tomlinson, line 262: TOS/PP info was only displayed to requiredEmail users who never signed in at this particular site.  Since requiredEmail is not needed, TOS/PP no longer needs to be displayed here.

resources/static/dialog/js/misc/state.js
```diff
@@ -243,43 +208,26 @@ BrowserID.State = (function() {
...
         idpName: info.idpName || URLParse(info.auth_url).host
       });
 
-      if (primaryVerificationInfo) {
-        primaryVerificationInfo = null;
-        if (requiredEmail) {
-          startAction("doCannotVerifyRequiredPrimary", info);
-        }
-        else if (info.add) {
+      if (self.primaryVerificationInfo) {
```
shane-tomlinson, line 228: Change from a "static" pseudo-global primaryVerificationInfo variable to an object instance.

resources/static/dialog/js/misc/state.js
```diff
@@ -469,26 +418,6 @@ BrowserID.State = (function() {
       user.passwordNeededToAddSecondaryEmail(function(passwordNeeded) {
         if(passwordNeeded) {
           self.addEmailEmail = info.email;
-
-          _.extend(info, {
```
shane-tomlinson, line 473: This calls doSetPassword, the same comments apply as https://github.com/mozilla/browserid/pull/1975/files#r1216393

resources/static/dialog/js/misc/state.js
```diff
@@ -511,18 +440,17 @@ BrowserID.State = (function() {
...
       options = options || {};
 
-      controller = options.controller;
-      if (!controller) {
+      self.controller = options.controller;
+      if (!self.controller) {
         throw "start: controller must be specified";
       }
 
-      addPrimaryUser = false;
```
shane-tomlinson, line 521: These no longer need to be reset since they are now instance variables.

resources/static/dialog/js/misc/state.js
```diff
@@ -243,43 +208,26 @@ BrowserID.State = (function() {
     });
 
     handleState("primary_user_provisioned", function(msg, info) {
-      info.add = !!addPrimaryUser;
       // The user is is authenticated with their IdP. Two possibilities exist
       // for the email - 1) create a new account or 2) add address to the
-      // existing account. If the user is authenticated with BrowserID, #2
+      // existing account. If the user is authenticated with Persona, #2
       // will happen. If not, #1.
+      info = info || {};
```
seanmonstar, line 215: why the default for an empty object here, but not above at L190?

Issue #1526, opened 2012-05-01
code cleanup: validate vs sanitize


lloyd:
We have two code libraries for checking input parameters, one interfaced programmatically, one interfaced declaratively.

We should move all of our input validation into a single file, and make input validation clear and concise in each implemented API.  One way to go about this would be to change:

```
exports.args = ['oldpass','newpass'];
```

for instance, to 

```
exports.args = [
 { name: 'oldpass',
   type: 'password'
   required: true
 }, 
 { name: 'newpass',
   type: 'password', 
   required: true
 }
];
```

Basically I suggest we extend the parameter validation of the wsapi framework we've got to perform type specific validation (emails, passwords, origins, etc), and make the way that individual apis interface it declarative - moving validation out of `exports.process`.

/cc @ozten @shane-tomlinson @benadida @fmarier 

Now, if we were really in a mood to yak shave we could build wsapi.js into a reusable framework for restful api development.  But my razor is a little dull right now.

lloyd:
how do you guys feel about depending on https://github.com/chriso/node-validator ?

/cc @ozten @shane-tomlinson @benadida @fmarier @warner @zaach

benadida:
looks good on quick overview, your call!

ozten:
Same feedback here. Maybe sanity test performance before adding it extensively. Has a good amount of followers and forks.

lloyd:
good point, @ozten.

I expect this will be a really cheap change to make - and I need it for some UX changes.  I'll investigate in a branch, get some numbers on perf, and if that looks good, serve up a PR to you two.

warner:
seems reasonable. What I always look for is annotation at a code boundary which tells the reader what assumptions are safe for them to make about the incoming data. So an `exports.args` isn't a bad way to go. The place where these schemes usually get hung up is on the type language: how to distingush between strings, strings that are really interpreted as email addresses, strings that are really interpreted as unconfined HTML, etc. But fundamentally it's all about correctly telling the programmer that it's safe to ignore what goes on outside the walls, because this type-checking will always be enforced.

fmarier:
I like your approach. The lack of consistency in input validation / sanitization is one of the things I picked up in my review of the WSAPI code.

lloyd:
I have started this work in the `api_param_validation` branch, I should be able to complete it today and then do a quick performance comparison...  please have a peek and criticize as I go !

PR #1979, opened 2012-07-11
perform rigorous validation on all API parameters, cleanup redundancy in...

lloyd:
... sanitize.js and validate.js - issue #1526

Signed-off-by: Lloyd Hilaiel lloyd@hilaiel.com


lib/wsapi.js
```diff
@@ -97,12 +97,6 @@ function authenticateSession(session, uid, level, duration_ms) {
   }
 }
 
-function checkPassword(pass) {
-  if (!pass || pass.length < 8 || pass.length > 80) {
-    return "valid passwords are between 8 and 80 chars";
-  }
-}
-
 function langContext(req) {
```
lloyd, line 100: this is now built into validate.js

lib/wsapi.js
```diff
@@ -271,7 +265,9 @@ exports.setup = function(options, app) {
 
       // set up the argument validator
       if (api.args) {
-        if (!Array.isArray(api.args)) throw "exports.args must be an array of strings";
+        if (Array.isArray(api.args)) {
+          console.log("WARNING: you should update", operation, "it uses unvalidated arguments");
+        }
```
lloyd: this check is actually unnecesary, it should be removed - we verify that api.args is an object inside validate.js

lib/wsapi/address_info.js
```diff
@@ -21,19 +21,19 @@ logger = require('../logging.js').logger;
...
 
 const emailRegex = /\@(.*)$/;
 
 exports.process = function(req, res) {
   // parse out the domain from the email
-  var email = url.parse(req.url, true).query['email'];
-  var m = emailRegex.exec(email);
-  if (!m) {
-    return httputils.badRequest(res, "invalid email address");
-  }
```
lloyd, line 35: this is representative of the change.  no more ad-hoc validation of parameters inside wsapi implementation.  the argument specification in exports.args expresses expected type.

lib/wsapi/address_info.js
```diff
@@ -21,19 +21,19 @@ logger = require('../logging.js').logger;
...
+  var m = emailRegex.exec(req.params.email);
+
   // Saftey value for production branch only
+  // (lth) ^^ what does this mean? ^^
```
lloyd, line 36: I didn't mean to add that, but I still don't know the answer.

lib/wsapi/cert_key.js
```diff
@@ -27,12 +31,14 @@ exports.process = function(req, res) {
...
     // addresses
-    db.emailIsVerified(req.body.email, function(err, verified) {
+    db.emailIsVerified(req.params.email, function(err, verified) {
       if (!verified) return httputils.forbidden(res, "that email requires (re)verification");
 
       // forward to the keysigner!
       var keysigner = urlparse(config.get('keysigner_url'));
       keysigner.path = '/wsapi/cert_key';
+      req.body = req.params;
+      console.log('bid params', req.params);
```
lloyd: console.log should be removed.

lib/wsapi/cert_key.js
```diff
@@ -27,12 +31,14 @@ exports.process = function(req, res) {
...
     // a user forgets their password.  We will not issue certs for unverified email
     // addresses
-    db.emailIsVerified(req.body.email, function(err, verified) {
+    db.emailIsVerified(req.params.email, function(err, verified) {
       if (!verified) return httputils.forbidden(res, "that email requires (re)verification");
 
       // forward to the keysigner!
       var keysigner = urlparse(config.get('keysigner_url'));
       keysigner.path = '/wsapi/cert_key';
+      req.body = req.params;
```
lloyd, line 48: we should add a comment explaining what this line does and why it's necessary.

lib/wsapi/complete_reverify.js
```diff
@@ -0,0 +1,66 @@
...
+          logger.warn("couldn't complete email verification: " + e);
+          wsapi.databaseDown(res, e);
+        } else {
+          wsapi.authenticateSession(req.session, uid, 'password');
+          res.json({ success: true });
+        }
+      });
+    };
+  });
+};
```
lloyd: this file should not exist, this is a merge error.

lib/wsapi/update_password.js
```diff
@@ -48,22 +43,22 @@ exports.process = function(req, res) {
         return res.json({ success: false });
       }
 
-      logger.info("updating password for email " + req.session.userid);
-      wsapi.bcryptPassword(req.body.newpass, function(err, hash) {
+      logger.info("updating password for user " + req.session.userid);
```
lloyd, line 46: fixing an orthogonal bogus log message.

package.json
```diff
@@ -32,6 +32,7 @@
         "uglifycss": "0.0.5",
         "underscore": "1.3.1",
         "urlparse": "0.0.1",
+        "validator": "0.4.9",
```
lloyd, line 35: this change comes with one new dependency.

lloyd:
@fmarier this pull request is probably best reviewed by you, there are some comments already inline for issues found, we'll address those shortly.  curious to hear your thoughts!  /cc @benadida

lib/validate.js
```diff
@@ -14,29 +14,104 @@
 
 const
 logger = require('./logging.js').logger,
-httputils = require('./httputils.js');
+httputils = require('./httputils.js'),
+check = require('validator').check;
+
+var types = {
```
lloyd, line 20: these are the various types that can come in through our api.  please review carefully here (you too @benadida) to see if the constraints are as tight as they can be, and no tighter.

bin/keysigner
```diff
@@ -74,13 +74,16 @@ try {
   process.exit(1);
 }
 
+
+
 // and our single function
-app.post('/wsapi/cert_key', validate(["email", "pubkey", "ephemeral"]), function(req, resp) {
+app.post('/wsapi/cert_key', validate({ 'email': 'email', 'pubkey': 'pubkey', 'ephemeral': 'boolean' }), function(req, resp) {
   var startTime = new Date();
+  console.log(req.params);
```
fmarier: this console.log should be removed

lib/validate.js
```diff
@@ -14,29 +14,104 @@
...
+  email: function(x) {
+    check(x).isEmail();
+  },
+  password: function(x) {
+    check(x).len(8,80);
```
fmarier, line 25: (I realize that the original code was using this value, so I'm not criticizing this pull-request.)

Is there a reason why we limit the password to 80 characters as opposed to say 255?

lib/validate.js
```diff
@@ -14,29 +14,104 @@
...
+  assertion: function(x) {
+    check(x).len(50,10240).regex(/[0-9a-zA-Z~_-]+/);
+  },
+  pubkey: function(x) {
+    check(x).len(50,10240);
+    JSON.parse(x);
+  },
+  origin: function(x) {
+    // allow single hostnames, e.g. localhost
+    if (typeof x !== 'string' || !x.match(/^https?:\/\/[a-z\d_-]+(\.[a-z\d_-]+)*(:\d+)?$/i)) {
```
fmarier, line 42: I'm not sure how deep we want to go with this one, but I'll just mention http://blog.gnukai.com/2010/06/fqdn-regular-expression/ and https://metacpan.org/module/Data::Validate::Domain here for reference in case we want to improve this regexp.

lib/validate.js
```diff
@@ -14,29 +14,104 @@
...
     if (req.method === "POST") {
-      params_in_request = req.body;
+      reqParams = req.body;
     } else {
-      params_in_request = req.query;
+      reqParams = req.query;
     }
 
+    // clear body and query to prevent wsapi handlers from accessing
+    // un-validated input parameters
```
fmarier, line 74: Awesome!

fmarier:
I like this change. The API is well-designed and the extra burden on developers is minimal. The whitelist approach also gets a big +1 from me.

lib/validate.js
lloyd, line 25: No, I don't think there's any good reason for this limitation.  If you feel that it should be bumped up, let's open a separate issue.

lib/validate.js
lloyd, line 42: I will open a separate issue for refining the regex.  It is not a regression from before, so shouldn't block this merge, but heck, we should improve it.

lloyd:
thanks for the review @fmarier!  with your approval, merging to dev and opening a new issue to track improving the origin regex.

benadida:
a post-mortem comment to say that I'm _really_ happy about this change and this approach, which exactly fits super-best practices for input checking

lloyd:
@benadida :D

PR #1996, opened 2012-07-12
Fix cookie check cookie made into a session cookie in IE8

shane-tomlinson:
IE8 doesn't know how to handle max-age when setting cookies turning the "cookie check cookie" into a session cookie.  Using expires=<date> as well.  Manually expire the cookie after its existence is checked for.

1982

### DIFF
--- a/resources/static/common/js/network.js
+++ b/resources/static/common/js/network.js
@@ -715,14 +715,32 @@ BrowserID.Network = (function() {
       withContext(function() {
         var enabled;
         try {
-          // set a test cookie with a duration of 1 second.
           // NOTE - The Android 3.3 and 4.0 default browsers will still pass
           // this check.  This causes the Android browsers to only display the
           // cookies diabled error screen only after the user has entered and
           // submitted input.
-          // http://stackoverflow.com/questions/8509387/android-browser-not-respecting-cookies-disabled/9264996#9264996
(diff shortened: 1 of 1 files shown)


resources/static/common/js/network.js
```diff
@@ -715,14 +715,32 @@ BrowserID.Network = (function() {
...
+          // only supports expires.
+
+          // expire the cookie in 2 seconds so there is no chance of the
+          // session expiring while this check is occuring.
+          var maxAgeSeconds = 2,
+              expiryDate = new Date();
+          expiryDate.setTime(expiryDate.getTime() + (maxAgeSeconds * 1000));
+
+          document.cookie = "__cookiesEnabledCheck=1; max-age="
+            + maxAgeSeconds + "; expires=" + expiryDate.toGMTString();
```
seanmonstar: why set expires in the future? We can just set the cookie, if it can be read, `enabled = true`, and then set expires to `-1` (or anything older than right now).

shane-tomlinson: Good call @seanmonstar on the future expiration.  an expires of -1 turns the cookie into a session cookie.  To delete the cookie, I set the cookie to expire in the past.

lloyd:
@seanmonstar can haz r+ and green button?

seanmonstar:
Nicely simplified.

PR #1997, opened 2012-07-12
Update the cookies disabled copy

shane-tomlinson:
issue #1167


jedp:
What she said.  The "easier" doesn't make sense if it's not possible to sign in without cookies.

Issue #2001, opened 2012-07-12
wsapi fails with extra parameters are not allowed: email


ringe:
I am trying to log in to http://dev.myfavoritebeer.org/ using mama@rin.no and mormor@eyedee.me

In both cases the login window returns the following

```
Action: Authenticating with Assertion
Network Info: POST: /wsapi/auth_with_assertion
Response Code - 400
Response Text: {"success":false,"reason":"extra parameters are not allowed: email"}
Error Type: {"success":false,"reason":"extra parameters are not allowed: email"
```

HAR dump: http://pastie.org/4243218
HAR Viewer: http://www.softwareishard.com/har/viewer

As you can see in the dump, I tried to manually POST to the wsapi by removing the email element, but this did not work. Anything I later did to set the "ephemeral" value to some boolean gave the same error, too.

shane-tomlinson:
I believe this is caused by the super strict input parameter validation - PR #1979

jrgm:
verified stage train-2012.07.20. email param is not needed.

PR #2004, opened 2012-07-12
Remove the extra "email" parameter sent to auth_with_assertion.

shane-tomlinson:
issue #2001


jedp:
r+

jrgm:
verified stage train-2012.07.20. Was never actually used.

Issue #1989, opened 2012-07-11
helpers.extend is dupe of _.extend


seanmonstar:
If `underscore` is always a hard requirement (it's even used inside the helpers.js), then there is no need to keep 2 versions of the same function, that does the same thing.

shane-tomlinson:
@seanmonstar - agreed.  I added helpers.extend before we decided _ was a hard req.  Will issue a PR to remove this.

jrgm:
code change. tests pass train-2012.07.20

PR #2012, opened 2012-07-12
remove helpers.extend

shane-tomlinson:
We can universally use _.extend instead.

issue #1989


lloyd:
+1

seanmonstar:
yay, tests pass and everything

PR #2019, opened 2012-07-12
Minimalist implementation of support in core for Proxy IDP (a.k.a. "BigTent")

lloyd:
With the exception of potential window resizing issues, this PR is intended to implement the basic required support for proxied support of popular IDPs.

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -226,7 +226,8 @@ var conf = module.exports = convict({
   },
   proxy_idps: {
     doc: "A mapping of domain names to urls, which maps popular email services to shimmed IDP deployments.",
-    format: 'object { } *?'
+    format: 'object { } *?',
+    env: 'PROXY_IDPS' // JSON text, i.e. {"yahoo.com":"yahoo.login.persona.org"}
   }
 });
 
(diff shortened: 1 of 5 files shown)


benadida:
r+ pending tests.

ozten:
@shane-tomlinson You'll want to look at this PR and think about what is essential on https://github.com/mozilla/browserid/pull/1753 but missing from this PR.

lloyd:
yeah, @shane-tomlinson, I'd _really_ like you to vet this.

Ben and I did a bull-in-a-china-shop thing here and tried to do the tiniest bit possible to make this go.  Our theory is that we do yahoo first.  Here are the frontend bits that we hacked out with the claims:
- resizing - because we're doing yahoo first, it just works for now.  We will figure out how and where to put the resizing logic in a subsequent PR.
- custom language/screens - The default IDP screen looks pretty darn good.  Do you think there's a need to have different language?

Is there anything else that @benadida and I chopped out that you feel is critical?

lloyd:
ben's given an r+, we've got tests, I've aired this out with @callahad and @ozten - merging, slightly worried that @shane-tomlinson will hit me, and that I cannot run away because he's now faster than me.

http://travis-ci.org/#!/mozilla/browserid/builds/1874037

shane-tomlinson:
@lloyd -

```
custom language/screens - The default IDP screen looks pretty darn good. Do you think there's a need to have different language?
```

We have no custom languages or screens.  @skinny97214 asked me to remove the "you are being redirected to your IdP" screen because it shows the BigTent URL, which is unacceptable because it will lead to user confusion.  "If I am signing in at Yahoo, why does it say bigtent.mozilla.org?"

Issue #2007, opened 2012-07-12
redirect timeout is sometimes first displayed with < 5 seconds.


shane-tomlinson:
This can happen if the CPU is busy and the fade in of the redirection text takes a long time.  Zac from automated testing even has a video where it first displays with 1 second.

stephendonner:
Is it this video: https://saucelabs.com/jobs/a7c57aabbab84aef95a961a45cb254fa# ?  It's from this report, if so: http://qa-selenium.mv.mozilla.com:8080/job/browserid.dev.123done.linux.firefox.10/312/HTML_Report/

stephendonner:
Confirmed with @zacc that this is responsible for our dev.123done build failures -- can we get this bumped up in stars/priority?  Realize you guys are slammed with regressions and prod rollouts, but want to ensure that our automation is kept humming along, to help!  Thanks.

zacc:
Video, starts at 0:28: http://qa-selenium.mv.mozilla.com:8080/job/browserid.dev.123done.windows.7.firefox.11/290/HTML_Report/?

stephendonner:
Or, for those output the VPN, and more directly: https://saucelabs.com/jobs/386925d6baa042a3ac864586d9be4ac0

seanmonstar:
@shane-tomlinson I wonder if it's because of a possible network call also.

At https://github.com/mozilla/browserid/blob/dev/resources/static/pages/js/verify_secondary_address.js#L104, the `updateRedirectTimeout` countdown gets setup, and if `mustAuth` is false because it's the same browser, then `submit` is called.

In submit, at L55, is a call to `user[verifyFunction]()`, and only once that returns, is the confirmation page shown, at L61.

It could be that the animation doesn't fade it in, but it could also be that the network request is taking a moment. I'll see about making them start at the same time.

PR #2024, opened 2012-07-13
Make redirect countdown and redirect in sync

seanmonstar:
This moves the countdown and the redirect timeout into the same timeout, and they don't start until after the animation has faded in, and the possible network request has finished.

issue #2007


shane-tomlinson:
Thanks for the update @seanmonstar - looks good.  I like the cleanup and the testing.

r+

Adding some JSHint comments and manually merging.

PR #2026, opened 2012-07-13
expire sessions when the password is changed

warner:
This series of patches causes client sessions to cease being valid when the user changes their password. Any outstanding sessions must re-authenticate.

Some notes to reviewers:
- There are several cleanup/refactoring patches before the actual functional change. These were needed to make the real change easy to read.
- The term "superSessionToken" derives from the idea that each client session (expressed as a cookie, with a session duration and authentication level) is actually part of a "super session", containing multiple normal sessions. The super session lasts from one password reset to the next (or until an explicit "log me out everywhere" action, which looks very similar). Normal sessions are not allowed to extend beyond the end of their super session.
- This adds a DB fetch at the start of each wsapi call, which is regrettably unavoidable if this feature is to work correctly. It also adds a DB fetch in any path which updates the session cookie. In the long run (i.e. after 1.0), I think the DB code should be changed to make exactly one fetch at the start of the wsapi operation (retrieving the whole user row, even if some pieces are unused), and then maybe do one write at the end of the operation. This would at least avoid doing multiple reads for a single function.
- stalled-mysql-test.js was modified to match the new reality (all wsapi operations require a DB, so some previous successes are now expected failures). I had to modify wsapi_client.js to pass the 503 from /wsapi/session_context back to the caller, instead of throwing a (sync) error that killed the test. I'd appreciate careful study here, to make sure I didn't screw anything up.
- I also changed test/lib/wsapi.js to make it possible to pass in an alternate context, so the test for this feature could simulate multiple parallel sessions (changing the password in one should cause the other to stop working).
- If a session stops working because the password was changed elsewhere, we want the client to go back to the login screen and ask for the new password. My test merely checks that the expired session gets a 403, and I don't know if that's enough to test the desired behavior, or if clients might get an ugly error and _not_ be redirected back to the login screen. (since this is a wsapi call, and not a normal page fetch, I'm optimistic).

### DIFF
--- a/lib/db/mysql.js
+++ b/lib/db/mysql.js
@@ -65,7 +65,7 @@ const schemas = [
   "CREATE TABLE IF NOT EXISTS user (" +
     "id BIGINT AUTO_INCREMENT PRIMARY KEY," +
     "passwd CHAR(64)," +
-    "lastPasswordReset BIGINT" +
+    "lastPasswordReset TIMESTAMP DEFAULT 0 NOT NULL" +
     ") ENGINE=InnoDB;",
 
   "CREATE TABLE IF NOT EXISTS email (" +
@@ -89,7 +89,7 @@ const schemas = [
     ") ENGINE=InnoDB;",
 ];
 
-function now() { return new Date().getTime().toString(); }
+function now() { return Math.floor(new Date().getTime() / 1000); }
 
 // log an unexpected database error
 function logUnexpectedError(detail) {
@@ -371,7 +371,7 @@ exports.completeCreateUser = function(secret, cb) {
 
     // we're creating a new account, add appropriate entries into user and email tables.
     client.query(
-      "INSERT INTO user(passwd, lastPasswordReset) VALUES(?,?)",
+      "INSERT INTO user(passwd, lastPasswordReset) VALUES(?,FROM_UNIXTIME(?))",
       [ o.passwd, now() ],
       function(err, info) {
         if (err) return cb(err);
@@ -451,7 +451,7 @@ exports.addPrimaryEmailToAccount = function(uid, emailToAdd, cb) {
 exports.createUserWithPrimaryEmail = function(email, cb) {
   // create a new user acct with no password
   client.query(
-    "INSERT INTO user(lastPasswordReset) VALUES(?)",
+    "INSERT INTO user(lastPasswordReset) VALUES(FROM_UNIXTIME(?))",
     [ now() ],
     function(err, info) {
       if (err) return cb(err);
@@ -515,7 +515,7 @@ exports.checkAuth = function(uid, cb) {
 
 exports.lastPasswordReset = function(uid, cb) {
   client.query(
-    'SELECT lastPasswordReset FROM user WHERE id = ?',
+    'SELECT UNIX_TIMESTAMP(lastPasswordReset) AS lastPasswordReset FROM user WHERE id = ?',
     [ uid ],
     function (err, rows) {
       cb(err, (rows && rows.length == 1) ? rows[0].lastPasswordReset : undefined);
@@ -524,7 +524,7 @@ exports.lastPasswordReset = function(uid, cb) {
 
 exports.updatePassword = function(uid, hash, invalidateSessions, cb) {
   var query = invalidateSessions ?
-    'UPDATE user SET passwd = ?, lastPasswordReset = ? WHERE id = ?' :
+    'UPDATE user SET passwd = ?, lastPasswordReset = FROM_UNIXTIME(?) WHERE id = ?' :
     'UPDATE user SET passwd = ? WHERE id = ?';
   var args = invalidateSessions ? [ hash, now(), uid ] : [ hash, uid ];
   client.query(query, args,
@@ -591,7 +591,7 @@ exports.cancelAccount = function(uid, cb) {
 
 exports.addTestUser = function(email, hash, cb) {
   client.query(
-    "INSERT INTO user(passwd, lastPasswordReset) VALUES(?)",
+    "INSERT INTO user(passwd, lastPasswordReset) VALUES(FROM_UNIXTIME(?))",
     [ hash, now() ],
     function(err, info) {
       if (err) return cb(err);
--- a/lib/db/mysql.js
+++ b/lib/db/mysql.js
@@ -13,7 +13,7 @@
  *    +--- user -----------------+      |*int    id       |
  *    |*int    id                |<-----|*int    user     |
  *    | string passwd            |      |*string address  |
- *    | string lastPasswordReset |      | enum   type     |
+ *    | int    lastPasswordReset |      | enum   type     |
  *    +--------------------------+      | bool   verified |
  *                                      +-----------------+
  *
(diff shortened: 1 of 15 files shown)


benadida:
I think the term supersessiontoken is going to be confusing to anyone looking at the code. Let's make it more functional, e.g. password_changed_at.

benadida:
I'm okay with the extra DB fetch though I agree that we should aggressively work towards one query per call only. (And we already have one query per call, I think.)

benadida:
@lloyd should take a look at the change in stalled-mysql-tests for sure.

warner:
yeah, I was using "passwordChangeTime" before switching to "superSessionToken". Lloyd told me that skinny will be looking to add a "log me out everywhere" button, which would be implemented by just bumping this token, but not actually changing the password. So I used a more neutral name to avoid winding up with something that sounded weird when seen in the context of that code.

I'm happy to change it back, though.. I agree that a "super session" is not an immediately obvious concept.

benadida:
ah interesting, I see your reasoning now.

How about.... lastLoginReset, or something like that?

lib/wsapi.js
```diff
@@ -177,60 +234,44 @@ exports.setup = function(options, app) {
...
+      // then let this request fall through
+      if (operation !== 'fake_verification' || !process.env['BROWSERID_FAKE_VERIFICATION'])
+        return httputils.badRequest(resp, "no such api");
+    }
+
+    // perform full parsing and validation
+    return cookieParser(req, resp, function() {
+      bodyParser(req, resp, function() {
+        cookieSessionMiddleware(req, resp, function() {
+          checkExpiredSession(req, resp, function() {
```
lloyd, line 273: hmm.   don't we _only_ need to check whether the session is expired if a user is logged in?  not all of our apis require authentication.  one such case is the `authenticate` call.

lib/db/mysql.js
```diff
@@ -65,7 +64,8 @@ if (conf.get('env') === 'test_mysql' && process.env['STALL_MYSQL_WHEN_PRESENT'])
 const schemas = [
   "CREATE TABLE IF NOT EXISTS user (" +
     "id BIGINT AUTO_INCREMENT PRIMARY KEY," +
-    "passwd CHAR(64)" +
+    "passwd CHAR(64)," +
+    "lastPasswordReset BIGINT" +
```
lloyd: any reason that we would not use the `TIMESTAMP` field of mysql?  like we do for staged.ts?  It shaves 4 bytes and is symmetric.  we don't get MS resolution, but I don't think we care, right?

lib/db/mysql.js
```diff
@@ -9,14 +9,13 @@
...
- *    | string passwd |    \- |*int user      |
- *    +---------------+       |*string address|
- *                            | enum type     |
- *                            | bool verified |
- *                            +---------------+
+ *                                      +--- email -------+
+ *    +--- user -----------------+      |*int    id       |
+ *    |*int    id                |<-----|*int    user     |
+ *    | string passwd            |      |*string address  |
+ *    | string lastPasswordReset |      | enum   type     |
```
lloyd: this diagram is out of date, it's not a string anymore.

lib/wsapi/authenticate_user.js
```diff
@@ -64,59 +64,69 @@ exports.process = function(req, res) {
...
+                                      : config.get('authentication_duration_ms')
+                                    }, function(err) {
+            if (err)
+              return wsapi.databaseDown(res, err);
+            res.json({ success: true, userid: uid });
+
+            // if the work factor has changed, update the hash here.  issue #204
+            // NOTE: this runs asynchronously and will not delay the response
+            if (config.get('bcrypt_work_factor') != bcrypt.get_rounds(hash))
+              updateHash(req, res, uid, hash);
```
lloyd, line 80: very nice restructuring.  that indenting was out of hand.

lib/wsapi/update_password.js
```diff
@@ -55,13 +55,24 @@ exports.process = function(req, res) {
           return res.json({ success: false });
         }
 
-        db.updatePassword(req.session.userid, hash, function(err) {
+        var passwordChanged = (req.params.oldpass != req.params.newpass);
+        db.updatePassword(req.session.userid, hash, passwordChanged,
+                          function(err) {
```
lloyd, line 60: hmm.  because this is a user initiated password update, is there any good reason to not reset their active sessions if they're resetting their password to the same thing?  I feel like going the other way would be safer - always resetting sessions on password update.

tests/password-update-test.js
```diff
@@ -149,6 +173,12 @@ suite.addBatch({
     "fails as expected": function(err, r) {
       assert.strictEqual(JSON.parse(r.body).success, false);
     }
+  },
+  "using the other (expired) session": {
+    topic: wsapi.post('/wsapi/prolong_session', {}, context2),
+    "fails as expected": function(err, r) {
+      assert.strictEqual(r.code, 403);
+    }
```
lloyd, line 199: beautiful tests.

lib/wsapi.js
warner, line 273: sure.. is there a good way to tell when it's needed? I figured a hardcoded list of operation names wouldn't be a good idea..

lib/db/mysql.js
warner: true, we don't really care, although unit tests are likely to fail if the timestamp resolution can't distinguish between two calls in rapid succession, so milliseconds are easier. (if we get full second resolution, the test will probably need an artificial 2s stall to avoid false negatives).

lib/wsapi/update_password.js
warner, line 60: the check is actually there to accomodate the "bcrypt work factor has changed so we're going to re-hash your password for you in the background" case, which delegates to update_password (with old==new). The unit test for that case expected the sesion to remain valid, plus it seemed a bit gratuitous to expire the user's session immediately after they log in (which is when the work-factor-has-changed check occurs) when they aren't actually changing their password.

lib/wsapi.js
lloyd, line 273: yeah, I think we can just say if `req.session.userid` is truey (an integer), then there must exist in the sesssion a lastPasswordReset token that matches the current one in the database.  if `req.session.userid` is falsey, the session is not yet authenticated and we can skip the db hit.

lloyd, line 273: in fact, I'm rather curious what is actually going on right now for un-authenticated api calls.  req.session.userid would be undefined and we'd be passing that into `db.lastPasswordReset()`

warner, line 273: it's probably returning `undefined`, which would get "stored" in the session (but of course assigning `undefined` to a property is a nop). My intention with that code was to skip the comparison if the DB returns 0 or `undefined`, since that would mean the DB row pre-dates this code (and I didn't want to gratuitously expire sessions when this code first deploys). Somehow the check for that disappeared while I was merging/rebasing this.. I'll add it back.

warner, line 273: hrm, bypassing the check for unauthenticated cookies changes the behavior of stalled-db situations, causing one of the stalled-mysql-test.js tests to fail.. looking into that now

warner, line 273: ok, updated the test to match, should be green now

lib/wsapi/update_password.js
lloyd, line 60: ok, that makes sense.

jrgm:
verified train-2012.08.17

Issue #1655, opened 2012-05-29
Tracking: license may be needed on all served production resources.


shane-tomlinson:
Email already sent to Jishnu and licensing@mozilla.org to find out more information.  Link to #1624

Resources that may need licenses:
dialog.js, browserid.js and communication_iframe.js

shane-tomlinson:
From Gerv:

```
No, it doesn't need to contain the license.

MPL 2 section 1.13 defines "Source Code" as "the form of the work
preferred for making modifications". Section 1.6 defines "Executable
Form" as "any form of the work other than Source Code Form", which
includes minified JS. Section 3.2 b) says you can distribute the
Executable Form under whatever terms you want.

However, 3.2 a) says that when you do that, you need to make the source
available, and tell people where to get it. We already make source
available, so that's not a problem. In terms of telling people, a HTML
or JS comment with the words "Source: <SCM URL>" would be the minimum
required. You can have one comment in the top-level HTML page pointing
at the browserid repo; you don't need one per file. Ideally, the SCM URL
would contain the revision number of the deployed build, rather than
being a generic URL.

You could, of course, go a bit further and be more explicitly
open-sourcey by having a small "Source" or "For Developers" or similar
link at the bottom of the entry page or main page.

Gerv
```

jrgm:
verified stage train-2012.07.20 The top of e.g., /sign_in or login.anosrep.org/ says:

```
<!DOCTYPE html>
<!-- This Source Code Form is subject to the terms of the Mozilla Public
   - License, v. 2.0. If a copy of the MPL was not distributed with this
   - file, You can obtain one at http://mozilla.org/MPL/2.0/.
   - Source: https://github.com/mozilla/browserid/
   - Current Commit: https://github.com/mozilla/browserid/commit/78b9df3
-->
```

PR #2031, opened 2012-07-13
Add a link to the repo and current source in the top level of each served up HTML resource.

shane-tomlinson:
- in each top level layout or ejs template, include a link to a partial that contains the license.
- The bottom of the license contains a link to the repo and the current SHA's commit.

issue #1655


shane-tomlinson:
@lloyd - would you mind reviewing this to see if my approach is valid?

lloyd:
@shane-tomlinson love to!

lloyd:
I love this change.  Pending a bit of testing I'm r+

Issue #1825, opened 2012-06-21
Emit additional events during authentication of returning users


nmalkin:
Right now, after the _screen.authenticate_ event (emitted for both new and returning users), no events are emitted for returning users who are signing in until either they succeed (_generate_assertion_) or give up (_window.unload_). This means that we have no effective way of distinguishing the scenario "new user opens dialog, freaks out, closes it" and "returning user opens dialog, tries five passwords that don't work, gets frustrated and leaves".

To better understand the flow of users through the system, and where we are losing people, the following events can be added:
- _authenticate.enter_password_ - emitted when, after we recognize the user as returning, we prompt them for their password
- _authenticate.password_submitted_ - emitted when the user, having entered their password, attempts to sign in (e.g., by clicking the _sign in_ button)

[The names are merely suggestions.]

Optionally, we can have the following events:
- _authenticate.password_success_ - when we recognize the password entered as correct
- _authenticate.password_fail_ - when the password is incorrect (and we tell the user that)

Though nice to have, they are not necessary, because the same information can be gleaned from other events (e.g., if we see _generate_assertion_ that means the authentication succeeded, and if we don't...).

PR #2033, opened 2012-07-13
Add new authentication related KPIs for @nmalkin

shane-tomlinson:
new KPIs in the event stream:
- authenticate.enter_password
- authenticate.password_submitted
- authenticate.password_success
- authenticate.password_fail

Added a couple of new testHelper functions, expectedMessage and unexpectedMessage

issue #1825


nmalkin:
Great, thanks!

Issue #2022, opened 2012-07-13
Use npm-seal to verify dependencies in production


zaach:
We can use npm-seal until npm supports it and we switch to node 0.8.x

[![seal](https://github.com/zaach/npm-seal/raw/master/npm-seal.png)](https://github.com/zaach/npm-seal)

jedp:
I love that little guy

zaach:
@lloyd There's an issue with having  npm-shrinkwrap.json around during development: we use a different set of dependencies for production and development, but the same shrinkwrap in both. We _could_ list all of the dependencies in the shrinkwrap, but then they would be installed in production as well.

Another alternative would be to generate/move/copy the shrinkwrap files before deploying, and only check the seal in production, post-installation. Thoughts?

lloyd:
@zcarter and I are hacking on a new magic npm proxy thing called lockdown.  So far it looks promising and seems like it might be a lower friction way to have confidence that what we develop on is what we actually deploy, and that everyone stays on the same dependencies for local dev, and we don't break doze.

stay tuned.

saying this is a beta2 req, but really would like to see this in beta1 :ok_hand:

ozten:
This sound like Awesomesauce that hardens to form a magick shell!

jrgm:
Closing this as npm-lockdown is now in the tree.

PR #2035, opened 2012-07-13
Add sealed-npm-shrinkwrap.json and a postinstall script to run `seal che...

zaach:
...ck` on our dependencies - issue #2022

When new dependencies are added sealed-npm-shrinkwrap.json must be regenerated.
This can be done by running `npm install; npm shrinkwrap` then `seal g`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,10 +37,11 @@
         "vows": "0.5.13",
         "awsbox": "0.2.12",
         "irc": "0.3.3",
-        "jshint": "0.7.1"
+        "jshint": "0.7.1",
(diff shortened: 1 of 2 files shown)


zaach:
Encountering tons of false positives in practice. Downloaded tarballs are a different size locally than on aws.

lloyd:
@zaach zoiks!  that was unexpected.  to what do you attribute this?

zaach:
@lloyd Perhaps some variance at the OS or library level. npm seems to have the correct shasums in its cache, so I'll try to extract those first.

Issue #1998, opened 2012-07-12
dev/staging: /cookies_disabled content collapsed on mobile


shane-tomlinson:
STR

1) In Firefox mobile or Android Native browser, open: https://login.anosrep.org/cookies_disabled

![Screen Shot of collapsed content](https://img.skitch.com/20120712-1cy5gfswmkwryqdh3jcnp38bg2.png)

shane-tomlinson:
The unsupported_dialog suffers from this problem as well.

shane-tomlinson:
This got merged into dev with checkin 684b785a58b8e5328ee1f0652acff8fead2f048b.  It's still not pretty, but this is back to the state that prod was in.

jrgm:
checked on ICS Chrome, stock, ff14-ff17, iOS5 safari, 2.2 stock ff14-ff15 and looks fine on all except nightly ff17 ICS which doesn't seem to layout the right size for any site I visit.

PR #2037, opened 2012-07-13
Show content for cookies_disabled and unsupported_dialog on mobile devic...

shane-tomlinson:
...es.
- This is still not pretty, but it is a safe for hotfix into staging. This makes the unsupported dialog and cookies disabled screens look equivalent to the current prod.


seanmonstar:
for issue #1998

Issue #2011, opened 2012-07-12
dev: if string is not localized, no string is displayed.


shane-tomlinson:
It should at least display the english version. I believe this started with the removal of node-gettext.

Issue #2006, opened 2012-07-12
it-ch/db-LB not working on ephemeral deployments when defined in aws.json


shane-tomlinson:

ozten:
My latest deployment lacks `podebug` from the [translate-toolkit](http://translate.sourceforge.net/wiki/toolkit/podebug) project.

That is what builds the db-LB language files. Maybe that is the issue? Did this work before with ephemeral deployments?

Issue #1970, opened 2012-07-09
enable debug 'locale' in our dev env


lloyd:
One should always be able to verify correct string markup via our devenv via a simple process (like switching browser languages).

previously I had tried to enable this, but it requires the addition of software that is not available via yum (translate-tooklit) to be run  on the server at deployment time.

We should figure this out and fix the feature so our debug locale is a more useful tool.

lloyd:
note, I don't think it's useful to enable locales in our dev env either, because translations only start after we freeze a train.  curious for thoughts on that one.

shane-tomlinson:
I think at least having db-LB is very useful, right now I am trying to add gettext to the entire main site but am having a difficult time figuring out how to enable the debug locale.

Issue #1963, opened 2012-07-08
debug local doesn't properly translate our strings


lloyd:
the "debug locale" we have does not properly process our substitution markers (`%s` and `%(name)`), nor does it handle html inside translatable strings at all.  Here is an example:

INPUT: `Please close this window, <a %(cookieLink)>enable <b>super dooper %(persona)</b> cookies</a> and try again`
EXPECTED: `uıaƃa ʎɹʇ pua <a %(cookieLink)>sǝıʞooɔ <b>%(persona) ɹǝdoop ɹǝdns</b> ǝʅqauǝ</a> ´ʍopuıʍ sıɥʇ ǝsoʅɔ ǝsaǝʅԀ`
ACTUAL: `uıaƃa ʎɹʇ pua >a/<sǝıʞooɔ >q/<)auosɹǝd(% ɹǝdoop ɹǝdns>q< ǝʅqauǝ>)ʞuı⅂ǝıʞooɔ(% a< ´ʍopuıʍ sıɥʇ ǝsoʅɔ ǝsaǝʅԀ`

PR #2039, opened 2012-07-13
i18n fixes

lloyd:
This change refines the way we debug our i18n implementation and does the following:
1. resolves a regression in gettext() - 5 star issue #2011
2. makes it so that for local development OR awsbox OR dev we've you can always switch to the debug locale (`it-CH`) and it always works - issue #1970 and issue #2006
3. translates strings with embedded HTML properly: issue #1963


bin/browserid
```diff
@@ -61,7 +61,7 @@ app.use(i18n.abide({
   supported_languages: config.get('supported_languages'),
   default_lang: config.get('default_lang'),
   debug_lang: config.get('debug_lang'),
-  locale_directory: config.get('locale_directory'),
+  translation_directory: config.get('translation_directory'),
```
ozten, line 64: Minor nitpick:
`locale` was chosen to maintain consistency with the Playdoh framework. The benefit is that a Mozilla community member who has been hacking on webdev projects can more easily jump into the Identity stack.

lib/i18n.js
```diff
@@ -38,33 +39,29 @@ var translations = {};
...
     default_lang: 'en-US',
-    locale_directory: 'locale'
   }));
  *
  * Other valid options: gettext_alias, ngettext_alias
  */
 exports.abide = function (options) {
-
-  if (! options.gettext_alias)        options.gettext_alias = 'gettext';
-  if (! options.ngettext_alias)       options.ngettext_alias = 'ngettext';
```
ozten, line 49: ngettext RIP

lib/i18n.js
```diff
@@ -101,14 +98,10 @@ exports.abide = function (options) {
...
 
-    if (lang && lang.toLowerCase && lang.toLowerCase() == debug_lang) {
-        lang = 'db-LB'; // What? http://www.youtube.com/watch?v=rJLnGjhPT1Q
-    }
-
     resp.local('lang', lang);
 
     // BIDI support, which direction does text flow?
-    lang_dir = BIDI_RTL_LANGS.indexOf(lang) >= 0 ? 'rtl' : 'ltr';
+    lang_dir = ((BIDI_RTL_LANGS.indexOf(lang) >= 0) || debug_lang == lang) ? 'rtl' : 'ltr';
```
ozten, line 104: Cool cleanup, I like getting debug locale out of the `BIDI_RTL_LANGS`.

lib/i18n.js
```diff
@@ -122,9 +115,15 @@ exports.abide = function (options) {
 
     var gt;
 
-    if (translations[locale]) {
+    if (lang.toLowerCase() === debug_lang) {
+      gt = gobbledygook;
```
ozten, line 119: Gobbledygood is very cool! Works on node or in the browser, so we can use it everywhere.

resources/static/common/js/gettext.js
```diff
@@ -6,17 +6,25 @@
...
 
+  var bid = BrowserID,
+      dom = bid.DOM;
+
   function Gettext() {
       return {
         gettext: function (msgid) {
+          if (window.Gobbledygook &&
+              dom.getAttr('html', 'lang') === 'db-LB') {
+            return window.Gobbledygook(msgid);
```
ozten, line 17: My first reaction is that this is fragile and hacky. It's not that bad though.

tests/i18n-tests.js
```diff
@@ -90,12 +96,97 @@ suite.addBatch({
...
+    "return a non-translated string" : function(err, r) {
+      assert.strictEqual(200, r.code);
+      assert.strictEqual(r.body.trim(), "This is not translated");
+    }
+  }
+
+});
+
+// and let's stop them servers
+start_stop.addShutdownBatches(suite);
```
ozten, line 188: Great tests!

tests/static-resource-test.js
```diff
@@ -57,7 +57,7 @@ suite.addBatch({
       });
       // Fragile - filename with :locale...
       // When fixing this test case... console.log(res[Object.keys(res)[0]]);
-      var localeIndex = 8;
+      var localeIndex = 9;
```
ozten, line 60: hmm...

ozten:
unit tests failing according to Travis. (travis' fault? I didn't run tests locally)
http://travis-ci.org/#!/mozilla/browserid/builds/1859748

r+ once tests pass.

lib/i18n.js
lloyd, line 49: yeah.  I personally will add an ngettext implementation to front and back when we need it.  _promise_

bin/browserid
lloyd, line 64: I have no qualms naming the config param `locale`.  I only made this change because we no longer use the locale/ directory in our software (where the `.po`s and `.mo`s live).  

What's your preference?

lib/i18n.js
lloyd, line 119: :D  the parsing of HTML is pretty hokey, but seems good enough for all the strings we currently have.

resources/static/common/js/gettext.js
lloyd, line 17: yeah.  I would like to improve this.  The only role the locale `db-LB` now serves is to convey to the front end that we're running a debug locale.  If it weren't for this, we could just use the debug locale straightaway.

One way we could make this better would require jsxgettext.  if we had that, then rather than including and running `gobbledygook` on the front end, we could dynamically generate a client.json for the debug locale upon request...  That way we'd still get the property that the debug locale is always up to date, and we'd excercise a bit more of the frontend code.

Dunno.  What do you think?  /cc @zaach

tests/static-resource-test.js
lloyd, line 60: yeah man.  goofy.  every time you add a resource to static resources, you gotta increment this number...  these tests could be made more awesome.  meh.

lloyd:
I'll merge pending a happy travis!

lloyd:
happy! http://travis-ci.org/#!/mozilla/browserid/builds/1860950

bin/browserid
ozten, line 64: Just a knit with a reason. Roll whichever way.

Issue #1111, opened 2012-02-14
Only the headers and footers of the diresworb.org pages are localized 


ghost:
OS: Windows 7
Browser: Firefox 11.0
Language: French

STR:
1. Load https://diresworb.org/ in the browser.
2. Browse through its pages (home, sigin, sign up etc).
3. Sign in and repeat step 2.

Only the headers and footers of the pages are localized (the rest of the strings are in English).

ozten:
Great point @ioanab. This is by design, we'll use this bug to track work for the l10n-rebrand milestone.

ghost:
The following messages displayed on https://diresworb.org/ pages aren't localized either:
- message displayed when trying to sign in with a wrong password;
- message displayed when trying to sign in with an email not associated to BID;
- message displayed when trying to sign up with an email already associated to BID;
- message displayed when trying to reset the password for a non-existent user;
- error messages displayed when trying to change the password (not entering one of the passwords, entering an erred old password);
- messages displayed when removing an email;
- message displayed when canceling an account.

shane-tomlinson:
@ioanab - can you take screen shots of the offenders in Skitch or other alternative?  Is this in the dialog or on the main site - the main site is only partially translated at the moment, which is "as expected" for the moment.

ghost:
@shane-tomlinson - this is all on the main site (server). I will take the screen shots and add corresponding links in here when done.

shane-tomlinson:
@ioanab - thanks, the screen shots really really help me out!  Great work on the bug finding btw.

ghost:
thanks, @shane-tomlinson. Here are the screen shots for the cases listed above (both pages and messages) http://bit.ly/x0pYJw. Please let me know if I missed any.

lloyd:
this is a duplicate of issue #1862

PR #2047, opened 2012-07-16
localize main site

lloyd:
issue # 1862

@shane-tomlinson I saw you started this work!  I rebased to current dev which has fixified the l10n debugging language.  I will review your code, verify all strings are extracted for the main site, and merge if I can safely do so.

### DIFF
--- a/resources/static/common/js/lib/ejs.js
+++ b/resources/static/common/js/lib/ejs.js
@@ -152,9 +152,16 @@ EJS.Scanner = function(source, left, right) {
          double_left: 		left+'%%',
          double_right:  	'%%'+right,
          left_equal: 		left+'%=',
+         // set - Persona addition. The backend understands <%-, which acts
+         // identical to the frontend's <%=.  <%= on the backend escapes
+         // characters to their HTML code equivalents.  For unit testing, we
+         // write backend templates on the front end, so we have to be able to
+         // process <%-.  Creating an alias here.  Using it wherever
+         // left_equal is found.
+         left_dash: 		left+'%-',
          left_comment: 	left+'%#'})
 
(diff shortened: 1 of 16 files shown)


lloyd:
@jbonacci it's sunday, what are you doing commenting on pull requests?

nice find, I closed issue #1111 down as a dupe.

lloyd:
I found four issues, and pushed four patches to this branch.

@shane-tomlinson I am r+ here, but cannot merge because frontend tests are broken.  it looks like our client side ejs implementation does not support `<%-` !!  I briefly looked at migrating to https://github.com/visionmedia/ejs which looks like it will run client side.

If you can resolve the tests, let's merge this dude and have a fully translated product for our beta launch!

resources/static/common/js/lib/ejs.js
```diff
@@ -152,9 +152,16 @@ EJS.Scanner = function(source, left, right) {
          double_left: 		left+'%%',
          double_right:  	'%%'+right,
          left_equal: 		left+'%=',
+         // set - Persona addition. The backend understands <%-, which acts
+         // identical to the frontend's <%=.  <%= on the backend escapes
+         // characters to their HTML code equivalents.  For unit testing, we
+         // write backend templates on the front end, so we have to be able to
+         // process <%-.  Creating an alias here.  Using it wherever
+         // left_equal is found.
```
lloyd, line 160: this approach makes sense and is good enough for our unit tests. thanks for the comment!  Do we ever use <%- on templates that are evaluated on the frontend?  where we're outputting html?

lloyd:
r+ / tests pass / http://travis-ci.org/#!/mozilla/browserid/builds/1876915

Issue #2032, opened 2012-07-13
dev: If user enters < 8 char password, /authenticate_user returns 400 error


shane-tomlinson:
If a user enters an incorrect < 8 character password, /wsapi/authenticate_user returns a 400 error

```
{"success":false,"reason":"pass: Error: String is not in range"}
```

if the user enters an incorrect > 8 character password, /wsapi/authenticate_user returns a 200 with the following:

```
{"success":false,"reason":"password mismatch for user: set117@yahoo.com"}
```

lloyd:
hmmm.  This is by design.  Does the frontend not check the user's password?  We don't allow < 8 chars /me is slightly confused.

shane-tomlinson:
@lloyd - this is a recent change to the API.  When a user is authenticating, we have always sent _all_ passwords to the backend regardless of their length.  We only check the length when a user is creating their password.

jrgm:
verified stage train-2012.07.20. Too short password does not go to network and just shows tooltip for password mismatch. Incorrect password >= 8 chars gets the 'password mismatch' json response shown above, and the password mismatch tooltip.

PR #2053, opened 2012-07-16
Before authenticating, check password length.

shane-tomlinson:
issue #2032


lloyd:
r+ - seems we do less work here and get the user a response faster.  tests pass.

Issue #2055, opened 2012-07-16
Link to Mozilla Support from mainpage contains "en-US"


JasnaPaka:
On the main page (https://login.persona.org/) there is a link to https://support.mozilla.com/en-US/kb/what-browserid-and-how-does-it-work. This is bad because link contains "en-US" in URL. Please remove "en-US". This article is translated to many languages. Thanks.

shane-tomlinson:
@JasnaPaka - Thanks for filing this issue!  PR #2056 fixes this.  It should be reviewed and merged this week and should make its way to prod in a couple weeks.

jrgm:
verified stage train-2012.07.20 I get the correct language for help from sumo with en or fr

PR #2056, opened 2012-07-16
Remove en-US from the SUMO link

shane-tomlinson:
issue #2055


lloyd:
r+ - tested redirect with english and non-english browser lang prefs.

jrgm:
verified stage train-2012.07.20 I get the correct language for help from sumo with en or fr

PR #2057, opened 2012-07-16
Remove extraneous params in load_gen requests

zaach:
For compatibility with new api validation.


lloyd:
r+ - who knew our validation would cause the cleanup of so many messes?

jrgm:
verified with load_gen against stage train-2012.07.20

PR #2058, opened 2012-07-16
moved node-statsd to optionalDependency

### LINKED ISSUE #1751: windows: node-statsd won't npm install
node-statsd package.json has an `os` property, that excludes windows from the party.

I don't know if a statsd daemon would even run on Windows, but I do see in `lib/statsd.js` that config turns it on or off, so it would be possible for a windows box to run the server with the config set to off.

seanmonstar:
also prevents error from require node-statsd if it wasn't installed,
but logs error if config says it should be enabled.

fixes #1751

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -127,7 +127,7 @@ var conf = module.exports = convict({
   statsd: {
     enabled: {
       doc: "enable UDP based statsd reporting",
-      format: 'boolean = true',
+      format: 'boolean = false',
       env: 'ENABLE_STATSD'
     },
     host: "string?",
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
This gets an r+ from me, @lloyd what do you think of this?

lloyd:
optionalDependencies was introduced in npm 1.1.0-beta-10, here:
https://github.com/isaacs/npm/issues/995#commits-ref-28aa53d

now, what node version does that correspond to?  I think we're running 0.6.12 in production.

If we verify that we've got it in production, find the lowest node version that supports it and update package.json to require that version, I'm good with this change.

lloyd

seanmonstar:
0.6.12 has npm 1.1.4: https://github.com/joyent/node/tree/v0.6.12/deps

0.6.7 was the first release to have at least npm 1.1.0-beta-10: https://github.com/joyent/node/tree/v0.6.7/deps

lloyd:
R+ if package.json targets 0.6.7

--lloyd

PR #2060, opened 2012-07-16
Allow assertions issued by person to be used to authenticate.  This make...

lloyd:
...s it possible for "proxy idps" to work without the implementation details leaking out into others verifier implementations.

### DIFF
--- a/lib/primary.js
+++ b/lib/primary.js
@@ -12,7 +12,8 @@ http = require('http'),
 logger = require('./logging.js').logger,
 urlparse = require('urlparse'),
 jwcrypto = require("jwcrypto"),
-config = require("./configuration.js");
+config = require("./configuration.js"),
+secrets = require("./secrets.js");
 
 // alg
 require("jwcrypto/lib/algs/rs");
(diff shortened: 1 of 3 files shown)


tests/lib/primary.js
```diff
@@ -24,18 +24,18 @@ User.prototype.setup = function(cb) {
...
-    
+
     var expiration = new Date();
     expiration.setTime(new Date().valueOf() + 60 * 60 * 1000);
-    
+
     jwcrypto.cert.sign(self._keyPair.publicKey, {email: self.options.email},
                        {expiresAt: expiration, issuer: self.options.domain, issuedAt: new Date()},
-                       {}, g_privKey, function(err, signedCert) {
+                       {}, self.options.privKey || g_privKey, function(err, signedCert) {
```
benadida, line 35: what is the point of this change?

lib/primary.js
```diff
@@ -274,9 +283,10 @@ exports.verifyAssertion = function(assertion, cb) {
   }
 
   var getRoot = function(issuer, next) {
-    // issuer cannot be the browserid
+    // allow assertions rooted in certs issued by us.  this occurs in the proxy_idp case
+    // where we sign assertions for other domains.
     if (issuer === HOSTNAME) {
-      next("cannot authenticate to browserid with a certificate issued by it.");
+      next(null, PUBLIC_KEY);
```
lloyd, line 289: the alternative implementation tactic would be to always go through the getPublicKey call.  This would be problematic, however, because that code checks for authentication and provisioning urls.  this was the approach that affected the least number of lines, but we should consider how to simplify and improve in the future.

benadida, line 289: I like this short-circuiting.

tests/lib/primary.js
lloyd, line 35: so that we can test in our unit tests that auth_with_assertion works for proxy_idp issued certs.  if you look in this file, test/lib/primary.js is written right now assuming all primary users will come from our sample primary.

benadida:
r+ I'm going to merge this. @warner can you check this asynchronously?

ozten:
r+

Issue #2059, opened 2012-07-16
don't select any email address in the picker for a new site


lloyd:
if we have no data regarding which email address a user has used for a specific site that they are logging into, we should not have anything selected by default.

```
15:46 < skinny> lloyd we shouldn't be selecting anything by default
15:46 < skinny> lloyd unless they only have one email
15:47 < skinny> lloyd anything else is yak shaving
```

jrgm:
verified train-2012.08.17 on https://train0817.personatest.org and https://train0817.123done.org, and a little fiddling with localStorage. WIll check when I get back the other sites working with ephemeral.

PR #2067, opened 2012-07-17
Do not select an email address if no email is associated with a site.

shane-tomlinson:
- Still focus the first address if no address is selected.
- Show a tooltip if the user has not selected an email address.
- Adding test helper functions to test for element focus and check.

issue #2059


lloyd:
assigning @seanmonstar to this one.

seanmonstar:
r+

Issue #2068, opened 2012-07-17
frontend: use named constants instead of hard coding password min and max length


shane-tomlinson:
Setting up to fix #1992

PR #2069, opened 2012-07-17
Instead of hard coding 8 and 80, use named constants.

shane-tomlinson:
issue #2068


lloyd:
r+ - extremely low risk, nice cleanup.  and as always, a test rich patch.

PR #2075, opened 2012-07-17
not BrowserID, but Persona

lloyd:
fix text in manage page


jrgm:
verified on awsbox train-2012.07.20

Issue #2080, opened 2012-07-18
limit length of URLs and paths passed to dialog


shane-tomlinson:
IE8 limits URLs to 2083 total bytes with 2048 bytes in the path.

See http://support.microsoft.com/kb/q208427

jedp:
Generously re-assigning from me to @seanmonstar in the hope that he actually has IE8 somewhere

seanmonstar:
You are too kind.

PR #2081, opened 2012-07-18
Set a max length of URLs and URL paths.

shane-tomlinson:
issue #2080


seanmonstar:
Mostly looks good. Are errors that are thrown usually strings? If that's how the rest are, then fine, follow the pattern. If not, can we throw an `Error`?

shane-tomlinson:
@seanmonstar - unfortunately, we have been throwing strings.  I have wanted to make this conversion to new Error and it's subtypes, but have not yet.

refactorage time.

PR #2085, opened 2012-07-18
Update copy for forgot password on main site to match dialog.

shane-tomlinson:
issue #2023


seanmonstar:
Copy looks fine, but the issue suggests changed the flow completely.

Issue #2087, opened 2012-07-18
main site confirm screen - password field not automatically focused


shane-tomlinson:
STR

1) In Browser A, open dialog to email picker
2) In Browser A, add a new secondary email
3) In Browser B, make sure user is logged out.
4) In Browser B, open verification email
5) In Browser B, password field should be focused.

seanmonstar:
It's worth stating that this requires a browser that doesnt support `<input autofocus/>`, such as IE.

jrgm:
verified train-2012.08.17 - works everywhere but ie8 (on winxp), which I punted out to GH-2395

PR #2089, opened 2012-07-18
Focus the password field on the confirmation page if the user has to enter their password.

shane-tomlinson:
issue #2087


seanmonstar:
r+

Issue #2040, opened 2012-07-13
browserid metrics - referer not passed through


jrgm:
Someone on IRC (Pedro) was mentioning that the browserid-metrics log had bad info. I think the issue is that the referer header is not passed through to browserid, so when it outputs the metrics the rp is 'undefined'. My fault since this is visible on the stage versions of that file.  Example:

```
{"level":"info","message":"{\"rp\":\"undefined://\",\"ip\":\"127.0.0.1\",\"type\":\"signin\",\"at\":\"Fri, 13 Jul 2012 11:22:10 GMT\"}"}
```

jrgm:
This is also now https://bugzilla.mozilla.org/show_bug.cgi?id=773877

In addition to referer:, we need to pass user-agent: and x-forwarded-for and maybe x-originating-ip.

jrgm:
verified stage train-2012.07.20. The metrics are now in /var/browserid/log/router-metrics.json and look like this again:

```
{"level":"info","message":"{\"browser\":\"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:14.0) Gecko/20100101 Firefox/14.0.1\",\"rp\":\"http://beta.123done.org\",\"ip\":\"x.x.x.x, x.x.x.x\",\"type\":\"signin\",\"at\":\"Mon, 30 Jul 2012 04:42:47 GMT\"}","timestamp":"2012-07-30T04:42:47.825Z"}
```

PR #2092, opened 2012-07-18
move metrics to router: avoids headers not being forwarded and caching -...

zaach:
... issue #2040

### DIFF
--- a/bin/router
+++ b/bin/router
@@ -13,6 +13,7 @@ urlparse = require('urlparse'),
 express = require('express');
 
 const
+metrics = require('../lib/metrics.js'),
 wsapi = require('../lib/wsapi.js'),
 config = require('../lib/configuration.js'),
 heartbeat = require('../lib/heartbeat.js'),
@@ -146,6 +147,10 @@ wsapi.routeSetup(app, {
 
 //catch-all
 app.use(function(req, res, next) {
+
+  // log metrics
+  if (req.url === '/sign_in') metrics.userEntry(req);
+
   forward(
     static_url+req.url, req, res,
     function(err) {
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -56,6 +56,9 @@ function setupLogger() {
     mkdir_p(log_path);
 
   var filename = path.join(log_path, configuration.get('process_type') + "-metrics.json");
+  if (process.env.METRICS_LOG_FILE) {
+    filename = process.env.METRICS_LOG_FILE;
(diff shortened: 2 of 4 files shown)


bin/router
```diff
@@ -146,6 +147,10 @@ wsapi.routeSetup(app, {
 
 //catch-all
 app.use(function(req, res, next) {
+
+  // log metrics
+  if (req.url === '/sign_in') metrics.userEntry(req);
```
lloyd, line 152: nice, note to readers, this change will fix our under-reported metrics.

@zaach - should we log even if get data is present?

lib/static/views.js
```diff
@@ -91,7 +91,6 @@ exports.setup = function(app) {
   // this should probably be an internal redirect
   // as soon as relative paths are figured out.
   app.get('/sign_in', function(req, res, next ) {
-    metrics.userEntry(req);
```
lloyd, line 94: can we remove the require('metrics') line at the top of this file now?

lib/metrics.js
```diff
@@ -56,6 +56,9 @@ function setupLogger() {
     mkdir_p(log_path);
 
   var filename = path.join(log_path, configuration.get('process_type') + "-metrics.json");
+  if (process.env.METRICS_LOG_FILE) {
+    filename = process.env.METRICS_LOG_FILE;
+  }
```
lloyd, line 61: great approach to make this thing testable.

tests/metrics-header-test.js
```diff
@@ -0,0 +1,100 @@
...
+    },
+    "metrics log exists": {
+      topic: function (err, r) {
+        if (existsSync(process.env.METRICS_LOG_FILE)) {
+          this.callback();
+        } else {
+          fs.watchFile(process.env.METRICS_LOG_FILE, null, this.callback);
+        }
+      },
+      "metric fields are logged properly": function (event, filename) {
```
lloyd, line 72: this is a beautiful test to ensure we don't regress.

lloyd:
two nits, shouldn't block merge, this is great.  r+

note @vthunder - our metrics are fixed with this change which should go into production on august 1st.  Also, our metrics would have been under-reported.  this PR fixes that.  

Finally, note that we've change the name of the metrics file here, we should make sure to get that in the deployment ticket when we cut the train.

lib/static/views.js
zaach, line 94: Yep.

Issue #2086, opened 2012-07-18
siteName and siteLogo not scrubbed from id.get()


callahad:
User @oncletom reports that siteName and siteLogo are working for him using the old API. Should they be? We never tested those features with that API.

We should probably also check on returnTo.

thom4parisot:
Too bad, it was a cool stuff :-(

lloyd:
@oncletom why not use the .watch() api?

thom4parisot:
The short way: because it was really easy with `navigator.id.get` and there was no point for us to evolve towards `watch` API.

The detailed reason:
- we don't want/need to have Persona script on every page: we use it only for login;
- our app already handles the session, we don't need another tool doing the same thing
- if we don't provide the user email to the watch API, the behavior is a total nonsense – the `onlogin` is called for every page (yet we expected to have this called only after a login)
- I find it quite weird to provide the user email on every page, just for the sake of Persona
- and finally, it's a matter of taste, the watch API is weird. Why not sticking to usual libs? `navigator.id.watch('login', function(}{ … })` etc.

lloyd:
@oncletom Thanks for taking the time to lay out your concerns.  I really appreciate it.  Our thinking right now is that it's so important that we help users log out of a machine more easily, and so important that we streamline user experience for first time login, that we make some sacrifices - those that you describe.

As far as the resource hit (more resources on every page), the best thing I can think of doing is to focus on optimizing delivery of this content (size and latency).

As far as the api requirements (complex semantics, and needing to provide email address on every page), I'm trying to dream up how we might find a compromise and make this better.  The first part of the compromise I'm trying to find is in a new utility library that @callahad and I wrote:

https://github.com/lloyd/persona.js

I'd love your thoughts on this.  Does this _feel_ better?  Make more sense?  Fit your application better?

The second part of the compromise is challenging whether we really need email on every page.  I think it might be complex to pull off (and haven't finished thinking through fully how it would work), but what if we explore on the email list the possibility of relaxing this requirement?  What if instead of providing the email of who is logged in, you just tell us that _someone_ is logged in?

Anyhow, I take your objections seriously (I know other people will have them too), and am really curious how we can make this work better for you.

thom4parisot:
[...]

PR #2109, opened 2012-07-19
returnTo, siteName, and siteLogo only should work with the observer API....

lloyd:
...  issue #2086

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1194,17 +1194,17 @@
           },
           onlogout: function() {}
         });
-        options.oncancel = function() {
+        opts.oncancel = function() {
           if (callback) {
             callback(null);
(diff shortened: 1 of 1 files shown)


lloyd:
@callahad mind reviewing?

resources/static/include_js/include.js
```diff
@@ -1179,8 +1178,12 @@
         if (typeof callback === 'function') setTimeout(callback, 0);
       },
       // get an assertion
-      get: function(callback, options) {
-        options = options || {};
+      get: function(callback, passedOptions) {
+        var opts = {};
```
callahad, line 1182: Shouldn't `opts` be `options`? We reference the old `options` variable elsewhere in this function.

callahad:
r+

Issue #1973, opened 2012-07-10
cross-origin issues with fonts redux


jrgm:
I opened GH-1845, but closed it as stage was just not set up correctly. 

I happened to be on login.dev.anosrep.org and noticed this again (this is a FF only behaviour). Loading the main site page /, there are messages like the following. 

```
Timestamp: 7/9/12 5:35:02 PM
Error: downloadable font: download failed (font-family: "Open Sans" style:normal weight:300 stretch:normal src index:3): bad URI or cross-site access not allowed
source: https://static.dev.anosrep.org/v/c335fd209b/common/fonts/OpenSans-Light.woff
Source File: https://static.dev.anosrep.org/v/6ffc5e5417/production/browserid.css
Line: 0
Source Code:
@font-face {   font-family: "Open Sans";   font-style: normal;   font-weight: 300;   src: local("Open Sans Light"), local("OpenSans-Light"), url("https://static.dev.anosrep.org/v/1112689d33/common/fonts/OpenSans-Light.eot") format("embedded-opentype"), url("https://static.dev.anosrep.org/v/c335fd209b/common/fonts/OpenSans-Light.woff") format("woff"), url("https://static.dev.anosrep.org/v/1bf71be111/common/fonts/OpenSans-Light.ttf") format("truetype"); }
```

Presumably it is worried about the mismatch of login.dev.anosrep.org and static.dev.anosrep.org. We may need to send some access control headers with the content when this moves to stage and prod.

jrgm:
Hmm, I see that in stage now the content is served with 

access-control-allow-origin:    https://login.anosrep.org

and this error is not shown in the FF error console with train-2012.07.06

shane-tomlinson:
+1.  @lloyd - the fonts in Firefox are pretty janky on dev.

jrgm:
verified stage train-2012.07.20

PR #2110, opened 2012-07-19
add access-control-allow-origin to all static resources (excluding views...

lloyd:
...), to allow fonts to be requested cross domain.  fixes a regression introduced during the merge of router, for issue #1973


shane-tomlinson:
close #1973

Issue #1769, opened 2012-06-16
Windows install errors


seanmonstar:
Trying to install browserid locally onto a Windows machine causes some errors. These are what I've gathered so far, as I work around them, I'll add more that I find.
- ~~node-statsd specifically excludes Windows from the `os` property. - details in #1751~~
- ~~bcrypt fails to install (at version 0.4.1). There has been a lot of work on the repo to make it work in Windows, which seems to be getting close-ish: https://github.com/ncb000gt/node.bcrypt.js~~
- ~~jwcrypto fails - depedency on bigint, which uses old node-waf instead of node-gyp build system.~~
- ~~client-sessions fails - dependency on node-proxy. newest version of node-proxy would fix this.~~
- ~~postinstall script fails - if you get past the above issues, the browserid's postinstall script tries to execute `generate_ephemeral_keys.sh`, with shell not working on Windows, obviously.~~
- ~~process.on('SIGINT') throws~~
- ~~runDaemons overrides process.env.PATH~~
- ~~symlinks~~
- npm-lockdown doesn't work on Windows

shane-tomlinson:
@seanmonstar - how are we lookin?

seanmonstar:
I've been updating the list in the top description as things have been fixed. The only outstanding issue is that `lockdown` doesn't work right on Windows.

shane-tomlinson:
@seanmonstar - do you still have problems installing on Windows?

seanmonstar:
There's one last problem that: the current `scripts/lockdown.js` is an old version that doesn't work on Windows. There is an open pull request to fix the script in lockdown: https://github.com/mozilla/npm-lockdown/pull/24

Afterwards, the script would need to be copied to `scripts/lockdown.js`.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #2111, opened 2012-07-19
fixes to get it running on Windows

seanmonstar:
- Windows doesn't have signals, so process.on("SIGINT") throws
- changing process.env.PATH is a bad idea

covers parts of #1769

_note_: it won't actually run on Windows till all other PRs addressing #1769 are merged also.

### DIFF
--- a/lib/shutdown.js
+++ b/lib/shutdown.js
@@ -75,9 +75,7 @@ exports.handleTerminationSignals = function(app, callback) {
     };
   }
 
-  try {
+  if (!process.env.NO_SIGNALS) {
     process.on('SIGINT', endIt('INT')).on('SIGTERM', endIt('TERM')).on('SIGQUIT', endIt('QUIT'));
-  } catch (noSignals) {
-    // Window doesn't support signals
   }
 };
--- a/lib/shutdown.js
+++ b/lib/shutdown.js
@@ -75,5 +75,9 @@ exports.handleTerminationSignals = function(app, callback) {
     };
   }
 
-  process.on('SIGINT', endIt('INT')).on('SIGTERM', endIt('TERM')).on('SIGQUIT', endIt('QUIT'));
+  try {
+    process.on('SIGINT', endIt('INT')).on('SIGTERM', endIt('TERM')).on('SIGQUIT', endIt('QUIT'));
+  } catch (noSignals) {
+    // Window doesn't support signals
+  }
 };
--- a/scripts/run_locally.js
+++ b/scripts/run_locally.js
@@ -78,6 +78,21 @@ if (config.get('env').substr(0,5) === 'test_') {
   }
 }
 
+// Windows can't use signals, so lets figure out if we should use them
+// To force signals, NO_SIGNALS=true.
+// Otherwise, they will be feature-detected.
+var SIGNALS_PROP = 'NO_SIGNALS';
+if (!(SINGALS_PROP in process.env)) {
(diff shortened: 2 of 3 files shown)


lib/shutdown.js
```diff
@@ -75,5 +75,9 @@ exports.handleTerminationSignals = function(app, callback) {
     };
   }
 
-  process.on('SIGINT', endIt('INT')).on('SIGTERM', endIt('TERM')).on('SIGQUIT', endIt('QUIT'));
+  try {
+    process.on('SIGINT', endIt('INT')).on('SIGTERM', endIt('TERM')).on('SIGQUIT', endIt('QUIT'));
+  } catch (noSignals) {
```
shane-tomlinson: @seanmonstar - Is there any way of testing whether an environment supports signals before wrapping this with a try/catch?  I am worried that this may swallow valid exceptions thrown on *nix based systems when shutting down.

scripts/run_locally.js
```diff
@@ -133,7 +135,11 @@ daemonNames.forEach(function(dn) {
   });
 });
 
-process.on('SIGINT', function () {
-  console.log('\nSIGINT recieved! trying to shut down gracefully...');
-  Object.keys(daemons).forEach(function (k) { daemons[k].kill('SIGINT'); });
-});
+try {
```
shane-tomlinson: Same comment as above to detect whether the environment supports signals.  In Windows, do child processes correctly terminate whenever the parent process does?

lib/shutdown.js
seanmonstar: Well, since this is an async event callback, on *nix, when the signals are triggered, it will no longer be in this try/catch. The only sync thing here to catch is `process.on`.

There also isn't a decent way to see if the process supports signals. The error thrown isn't even useful ("Module not found", since it's trying to load an internal node signals binary). The only way to detect it earlier would be something like this:

``` js
var hasSignals = true;
try {
  function noop() {}
  process.on('SIGINT', noop);
  process.removeListener('SIGINT', noop);
} catch(err) {
  hasSignals = false;
}
```

scripts/run_locally.js
seanmonstar: Yea they do. This simply catchs the interrupt signal, which is usually Ctrl-C, which doesn't typically mean to shutdown. However, I guess we want Ctrl-C to shutdown everything. This would send SIGINT to the child processes, and then the above `lib/shutdown.js` would properly close DB connections and so forth.

seanmonstar:
Perhaps `shutdown` should be abstracted a little so it still shuts down gracefully on Windows?

lib/shutdown.js
shane-tomlinson: @seanmonstar - the hasSignals check above looks like a good way to go.

seanmonstar:
@shane-tomlinson the check sets `process.env.NO_SIGNALS`, and therefore will be available in the child processes since the env is copied over.

shane-tomlinson:
r+.  Manually merged into dev with commit 76e95224142868a464e6ef42e9e723ff08cda2ca.  

Instead of using NO_SIGNALS environment variable, changed it to SUPPORTS_SIGNALS.

Nice work @seanmonstar!

seanmonstar:
Well, I used a negative because I wasn't sure if in deployment the
run_locally script was used. If not, I wanted the processes to default to
using signals.

PR #2115, opened 2012-07-19
Ensure that when a user verifies in a different browser than what they r...

### LINKED ISSUE #2104: Verifying email reset in multiple browser exception in /wsapi/complete_reset
STR

1) Have an account with a secondary address
2) Sign in to BrowserID using same address in two browsers.
3) In Browser A, reset your password.
4) Open Verification Link in Browser B
5) Nothing happens in Browser B.

In Browser B, /wsapi/email_for_token returns "must_auth: false" because it already has a session to that user.  Because "must_auth" is set to "false", /wsapi/complete_reset is called without the user needing to enter a password.

/wsapi/complete_reset blows up saying "Authentication Required: Provide your password"

zaach:
...eset their password with, they must authenticate to complete the verification

Fixes #2104

### DIFF
--- a/lib/wsapi/email_for_token.js
+++ b/lib/wsapi/email_for_token.js
@@ -55,7 +55,8 @@ exports.process = function(req, res) {
                req.params.token === req.session.pendingCreation) {
         must_auth = false;
       }
-      else if (typeof req.session.pendingReverification === 'string') {
+      else if (typeof req.session.pendingReverification === 'string' ||
+               typeof req.session.pendingAddition === 'string') {
         must_auth = false;
       }
       // NOTE: for reverification, we require you're authenticated.  it's not enough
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@zaach - Thanks for the PR, it gets me further in my testing!  There is a change that needs made though, when verifying on the same browser where I already have a session, must_auth is returning true.

STR

1) Open dialog, if signed in sign out.
2) Enter new secondary email
3) In same browser, verify new secondary email.
4) Re-open dialog in same browser
5) add new secondary email
6) verify new secondary email in same browser but different tab.

/wsapi/email_for_token returns "must_auth: true"

lloyd:
closing this down and opening new PR with an additional patch on top of @zaach's work

Issue #2121, opened 2012-07-20
change-password dialog: wrong pw gives confusing error dialog


warner:
This is on both staging and dev.

STR:
- navigate to login.dev.anosrep.org or login.anosrep.org
- create an account with a secondary email address, with a password
- click the "Edit" button on the password line, to change the password
- type the wrong password into the "Old Password" box
- hit "done"
- The resulting error dialog says "We are very sorry. There has been an error! Please close this window and try again"
- the "See more info" link exposes text that says the error is in /wsapi/update_password and has a "reason" of "oldpass: Error: String is not in range"

Expected behavior: a nice dialog that says "Sorry, that's not your old password. Maybe you should try again?.

Also, I was surprised that there was only one New Password box. In most new-password situations (both initial account creation and change-password), I expect to see both a New Password and a Verify Password box, to guard against typos in the new password.

jrgm:
/cc @stomlinson, @lloyd 

Ah, I see. The key is that the incorrect old password is less than the required 8 characters. 

This is related to GH-2032 which was closed by GH-2053 about authenticate_user. 
But not for update_password called from the main site (I guess). 

I do _not_ see this on stage, so @bwarner could you check that again?

I'll mark this as 5-star for now, but I could punt if it's just the case where the "wrong" password is "too short".

lloyd:
fixed by PR 2126, confirmed on dev.

warner:
@jrgm you're right, it's not happening on stage. I must have gotten my tabs mixed up.

jrgm:
Thanks for checking @warner. Also your other comment about double-entry confirm of 
new password is GH-1702.

jrgm:
verified fixed with PR GH-2126 stage train-2012.07.20. A too short old password is disallowed by js in the manage page and the post is not sent on network.

PR #2126, opened 2012-07-20
Fix error on too short of old password when changing password.

shane-tomlinson:
- Generalize and cleanup unit tests relating to password changes.

issue #2121


lloyd:
r+

Issue #2118, opened 2012-07-20
main site - forgot password flow winds up showing /signin page


jrgm:
STR
1. Go to https://login.dev.anosrep.org and click Sign in.
2. Enter a known secondary email address and the 'Sign in' button to show the password field.
3. Click 'Forgot your password?'
4. Enter the password twice and click 'Reset Password'

Expected: page transitions to show text that an email link has been sent. Email is sent out.

Actual: page briefly shows that text but then redirects itself to /signin page. Email, though, is sent OK.

Reproduced in FF and Chrome on osx

jrgm:
Note: this does not happen on train-2012.07.06 or production. Gonna mark it 5-star.

lloyd:
r+

lloyd:
verified fixed on dev.

jrgm:
verified fixed on stage train-2012.07.20. Page transitions to show Forgot Password Confirm your email address... and the email is sent and received.

PR #2127, opened 2012-07-20
Fix forgot password redirection to /signin

shane-tomlinson:
- to page_helpers.emailSent, add a new input paramter - pollFuncName.  The poll functions are unique for each type of email sent.

issue #2118


lloyd:
r+

Issue #2088, opened 2012-07-18
error when verifying email addition in second browser


shane-tomlinson:
STR

1) In Browser A - Create an account with a primary
2) Open dialog to email picker, click "Add Email"
3) Add the first secondary address, set password, hit enter.
4) in Browser B - Open confirmation page, enter password.
5) Go back to Browser A.
6) Watch error screen appear.

![error message](https://img.skitch.com/20120718-ri9hu76c2hxqabr2s1ph6w3qty.png)

shane-tomlinson:
@lloyd - I brought this up while you were still here in London and remembered it again today as I was testing.  This scenario requires two browsers.  

When /wsapi/email_addition_status returns a status of "complete", Browser A's authentication level _should_ be set to "password"  Yet, when /wsapi/cert_key happens, it thinks Browser A is either no longer authenticated or authenticated to the wrong level.

lloyd:
confirmed that pull request #2135 fixes this issue.

jrgm:
For the initial STR in this issue, I went through this with desktop opera, safari, chrome, FF14, IE9/winxp, IE8/winxp, safari/ios5, stock/ics, chrome/ics, ff14/ics, stock-webkit/2.3.4, ff14/2.3.4 and they all have no error, the dialog dismisses and the user is logged in. Verified train-2012.07.30 and current dev for a few too. 

One oddity noted in GH-2242 (wtf? on that one).

jrgm:
I should note that for IE8 there's a bug in adding a primary atm GH-2239. But you can work around that to get the primary added by trying to do it a second time, which is how I confirmed the IE8 case.

PR #2129, opened 2012-07-20
Issues 2088 2104 address verification

shane-tomlinson:
This includes @zaach's PR for #2104 and mine for #2088

There could stand to be some polish.

This flow needs tested:

Create an account with a primary. 
Add a secondary, verify on alt browser.  
Log out.  
Sign in with primary.  
Add new secondary email.  Verify in alt browser.
At this point, the user should have to enter their password

When the user is forced to enter their password, there is a cancel button in the screen that should probably not be there.

### DIFF
--- a/resources/static/common/js/network.js
+++ b/resources/static/common/js/network.js
@@ -101,15 +101,8 @@ BrowserID.Network = (function() {
   }
 
   function handleAddressVerifyCheckResponse(onComplete, status, textStatus, jqXHR) {
-    if (status.status === 'complete') {
-      // The user at this point can ONLY be logged in with password
-      // authentication. Once the registration is complete, that means
-      // the server has updated the user's cookies and the user is
-      // officially authenticated.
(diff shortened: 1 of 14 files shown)


resources/static/common/js/network.js
```diff
@@ -101,15 +101,8 @@ BrowserID.Network = (function() {
   }
 
   function handleAddressVerifyCheckResponse(onComplete, status, textStatus, jqXHR) {
-    if (status.status === 'complete') {
```
shane-tomlinson, line 104: This was an incorrect assumption.  If a user is logged in to the assertion status and verifies their email in a second browser, they should be forced to enter their password before they are upgrade to the password status.

lloyd:
closing this down, let's move the party over to issue #2135 where we can review stomlinson's changes in isolation.  (@zaach's stuff is already reviewed and merged.

PR #2133, opened 2012-07-21
upgrade data formats

benadida:
so far, this is just upgrading the interface to jwcrypto that implements a slightly tweaked API to be ready for new data formats. This should NOT change any of the data formats we use yet.

### DIFF
--- a/lib/secrets.js
+++ b/lib/secrets.js
@@ -115,5 +115,7 @@ exports.publicKeyCreationDate = function(name, dir) {
 };
 
 exports.loadPublicKey = function(name, dir) {
-  return jwcrypto.loadPublicKey(JSON.stringify(readAndParseCert(name, dir)['public-key']));
+  var parsedCert = readAndParseCert(name, dir);
+  var pkString = parsedCert['public-key'] || parsedCert.publicKey;
+  return jwcrypto.loadPublicKey(JSON.stringify(pkString));
 };
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
         "express": "2.5.0",
         "gobbledygook": "0.0.3", 
         "mustache": "0.3.1-dev",
-        "jwcrypto": "0.3.2",
+        "jwcrypto": "https://github.com/mozilla/jwcrypto/tarball/2012-07-new-data-formats",
         "mysql": "0.9.5",
         "node-statsd": "https://github.com/downloads/lloyd/node-statsd/0509f85.tgz",
         "nodemailer": "0.1.24",
--- a/lib/keysigner/ca.js
+++ b/lib/keysigner/ca.js
@@ -29,7 +29,7 @@ function certify(hostname, email, publicKey, expiration, cb) {
   if (expiration == null)
     return cb("expiration cannot be null");
 
-  cert.sign(publicKey, {email: email},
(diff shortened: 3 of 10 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
         "express": "2.5.0",
         "gobbledygook": "0.0.3", 
         "mustache": "0.3.1-dev",
-        "jwcrypto": "0.3.2",
+        "jwcrypto": "https://github.com/mozilla/jwcrypto/tarball/2012-07-new-data-formats",
```
shane-tomlinson, line 21: @benadida - is desired to peg to a version of jwcrypto not published on npm?

benadida, line 21: @shane-tomlinson no this is a temporary thing while waiting for jwcrypto merge :)

lib/secrets.js
```diff
@@ -115,5 +115,7 @@ exports.publicKeyCreationDate = function(name, dir) {
 };
 
 exports.loadPublicKey = function(name, dir) {
-  return jwcrypto.loadPublicKey(JSON.stringify(readAndParseCert(name, dir)['public-key']));
+  var parsedCert = readAndParseCert(name, dir);
+  var pkString = parsedCert['public-key'] || parsedCert.publicKey;
```
shane-tomlinson, line 119: Is the parsedCert['public-key'] || ... to deal with both old and new public key formats?  Is there a sunset period on old certs and this code can be removed eventually?

If the code can be removed eventually, a // BEGIN TRANSITION CODE, // END TRANSITION CODE have been helpful to @lloyd and I to keep track of code that can be removed.

shane-tomlinson:
@benadida - the changes to the .sign calls are all consistent, but I wonder if a wrapper to jwcrypto.cert.sign could be implemented to present developers with a higher level interface.   

The following is format is repeated several times throughout the unit tests, it is easy to introduce errors when the structure of the objects to jwcrypto.cert.sign must be exactly as specified.

```
jwcrypto.cert.sign({publicKey: pubkey, principal: {email: user + "@" + domain}},
                  {issuer: domain, expiresAt: expiration, issuedAt: new Date()},
                  {}, _privKey, function(err, cert) {
...
```

To reduce the chances of programmer error, I would not change the signature to sign, but add a higher level helper:

```
signWithEmailPrincipal({
  publicKey: <key>,
  privKey: <key>
  user: <user>,
  issuer: <domain>,
  expiresAt: <date>
});
```

benadida:
@shane-tomlinson interesting ... On the flipside, this might increase errors, say if you misspell "issuer", it would end up in there misspelled as part of the extra payload, but not as the special field.

My preference would be to not add anything now, since jwcrypto isn't meant for most developers.

lib/secrets.js
shane-tomlinson, line 119: @benadida - after reading through the spec, `public-key` is the old field name whereas `publicKey` is the new, is there any danger of a cert containing both fields?  If so, would that be an error?  If not an error, should 'public-key' take precedence over 'publicKey' even though the latter is the newly specified field?

benadida, line 119: great question. I don't see any reason why this would cause issues, but the right way to do crypto code is to be _extremely_ strict. I will build this into jwcrypto:

https://github.com/mozilla/jwcrypto/issues/35

PR #2135, opened 2012-07-21
Issues 2088 and 2104

lloyd:
submitted on behalf of shane.  this is his four commits rebased onto current dev.  see issue #2129 for his comments.

### DIFF
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -193,7 +193,7 @@ $(document).ready(function() {
     });
   });
 
-  $(".specify button.logout").click(navigator.id.logout);
+  $(".specify button.logout").click(function() { navigator.id.logout() });
 
   $(".session button.update_session").click(function() {
     storage.loggedInUser = $.trim($('#loggedInUser').val());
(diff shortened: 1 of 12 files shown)


lloyd:
via local testing I have confirmed that this pile of patches fixes issue #2104

lloyd:
via local testing I have confirmed that this pile of patches fixes issue #2088

lloyd:
ok.  I've skimmed the code and it looks reasonable.  I've manually confirmed that this fixes issue #2088 and issue #2104.  All unit tests pass.

While this is a large patch, and it very well may introduce bugs I was not able to uncover in my manual testing.  it closes down two five star issues, so without it this train isn't going anywhere.

I will merge it, and will give QA guidance to focus attention on multi-browser cases that involve primary addresses.

NOTE: we should ask our selves as we uncover issues:
1. how prominent is this case (requires primary addresses, requires auth on separate browser, requires password reset?)
2. is there a user discoverable workaround.

If we find highly fringe issues for which there is a simple user discoverable workaround, considering we are not yet at beta, we may decide that they are not 5 star...  just sayin.

Issue #2136, opened 2012-07-23
Train-2012.07.20 l10n issues #2


mathjazz:
Some new l10n issues were reported by localizers. All of them are related to original strings. It is not neccessary to break the string freeze, but they should probably be fixed for the next train.
## Almost duplicates
- resources/views/signin.ejs:61
  - ``To verify that you own <strong %(primaryId)>address</strong>, you must sign in with your provider. A new window will be opened.`
- resources/views/signup.ejs:98
  - `To verify that you own <strong %(emailId)>address</strong>, you must sign in with your provider. A new window will be opened.`
## Strange structure

Why does it say "to verify that you own adress"? I think either "address" should be replaced by the actual e-mail address or there should be a "this" before "address".
- resources/views/signin.ejs:61
  - ``To verify that you own <strong %(primaryId)>address</strong>, you must sign in with your provider. A new window will be opened.`
- resources/views/signup.ejs:98
  - `To verify that you own <strong %(emailId)>address</strong>, you must sign in with your provider. A new window will be opened.`
## Another strange structure

IMHO should be: `Email <strong>%(emailId)</strong> is not registered.`
- resources/static/dialog/views/site/signin.ejs:
- resources/views/signin.ejs:
  - `<strong %(emailId)>Email</strong> is not registered. Would you like to <a %(signUpLink)>sign up</a> instead?`

PR #2141, opened 2012-07-23
Update the "To verify that you own" strings to unify copy and clean up bad grammar.

shane-tomlinson:
issue #2136

### DIFF
--- a/resources/views/signin.ejs
+++ b/resources/views/signin.ejs
@@ -58,7 +58,7 @@
             <ul class="notifications">
                 <li class="notification" id="verify_primary">
                   <p>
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@mathjazz - How are these changes looking?  

I made the first two requested changes, the `IMHO should be: Email <strong>%(emailId)</strong> is not registered.` change I want to talk through with @skinny97214 first.

mathjazz:
@shane-tomlinson That looks good, thanks!

But I've just found one tiny details that should be changed as well -- there's a double space between sentences, it should be single.

mathjazz:
I can't find these changes in train-2012.07.30. There's no hurry, we can wait, but I just wanted to check.

Issue #2130, opened 2012-07-20
[mainsite] a signedin primary user visits /forgot, ask for new passwords and email sent.


jrgm:
A very ridiculous scenario, but anyways...

STR
1. On the main site, signin as a primary email address (one that is not part of any account with other email addresses).
2. Manually go to the /forgot page

The user is show the normal UI with 'Email address' prefilled to the primary email address, and prompted for two passwords. Upon hitting reset password, the screen transitions to show 'We sent a confirmation email to foo@eyedee.me' and presumably that email is sent out. Not sure what would happen if that user received the email and clicked on the link. (I could set up an awsbox and hack it, but not a priority for me).

jrgm:
veriifed fixed

Issue #1679, opened 2012-06-04
/forgot page: email address field not disabled and has keyboard focus


jrgm:
Not sure this is really an issue, since it had this behaviour in train-2012.04.27 (and I think previous). But now that it contains the two additional password fields for entry, it seems odd.

STR:
1. For an existing secondary email address known to diresworb.org, click sign in and enter that email.
2. When prompted for the password, click forgot password.

Expected:
The email field is shown in a disabled state, and the keyboard is focused on the first password field.

Actual:
The email field has focus and can be edited.

Current production: train-2012.04.27:
![img](https://img.skitch.com/20120604-1u733d8qyjpkdkhh3rdey7ha3m.png)

Current stage: train-2012.05.25:
![img](https://img.skitch.com/20120604-1k33ekx6xeewee14rb83ew31cp.png)

jrgm:
One note, if the user edits the field and changes it to an unknown user, a tooltip is shown that it is an unknown user. 

If the user changes it to a known user, email is sent to that user. 

If changed to a primary email address, email is sent to the primary user, which seems wrong.

jrgm:
verified train-2012.08.17
- Ensure the reset password flow of the main dialog works as expected.
- Ensure the user cannot reset the password of a primary address.
- Ensure the /forgot page redirects to /signin if the email address is known.
- Ensure the /forgot page redirects to /signin if the email address is unknown.

PR #2148, opened 2012-07-24
Usability enhancements to the forgot password screen.

shane-tomlinson:
- Disable the email field.
- Focus the first password field instead of the email field.
- If no email is stored (user did not come from /signin), redirect the user to /signin.
- If email address that is stored is not known, redirect to /signin.
- If email address that is stored is known but a primary, redirect to /signin
- When calling User.addressInfo on a primary address, check whether the address is known or not.  Allows us to simplify the forgot page.

This addresses both #1679 and #2130.

Additional areas to test:
- Ensure the reset password flow of the main dialog works as expected.
- Ensure the user cannot reset the password of a primary address.
- Ensure the /forgot page redirects to /signin if the email address is known.
- Ensure the /forgot page redirects to /signin if the email address is unknown.


resources/static/pages/js/forgot.js
```diff
@@ -31,41 +30,86 @@ BrowserID.forgot = (function() {
...
+      return;
+    }
 
-    pageHelpers.setupEmail();
+    // We know an email address was stored, now check if it is registered.  If
+    // it is not registered, kick the user over to the signup page.  If it is
+    // registered, but a primary, kick them over to the signin page.
+    user.addressInfo(email, function(info) {
+      if (!info.known) {
+        doc.location.href = "/signup";
```
seanmonstar: the route `/signup` was removed with your other PR.

seanmonstar:
@shane-tomlinson only issue I can see is the redirect to `/signup`

seanmonstar:
I manually fixed that one thing. merging.

Issue #2120, opened 2012-07-20
Sign In with unknown email results in lame Sign In button


warner:
environment: both staging (login.anosrep.org) and dev (login.dev.anosrep.org). Both Chrome 21 and FF 17.0a1 (nightly).

STR:
- navigate to login.anosrep.org (or login.dev.anosrep.org)
- click "Sign In"
- enter a new non-primary email address
- get "ADDR is not registered. Would you like to `sign up` instead?" dialog. Note that the largest button on the screen is still the blue "Sign In" button on the bottom right.
- ignore suggestions and click "Sign In"
- nothing happens
- click again
- nothing happens

I'd prefer to see a new non-primary email address redirect me to the Sign Up page, so I can click on the largest button on the screen and get the most useful action. Failing that, I'd prefer that the non-functional Sign In button be disabled, leaving me to follow one of the two smaller links to the Sign Up page.

jrgm:
GH-1579 discusses the same STR and notes that the similar situation applies for the /signup page. I'm going to close GH-1579 (since finding anything old in GH is a hassle).

jrgm:
verified train-2012.08.17. The UI has been changed such that there is no separate signup and the button is always active (checking for inputs of course).

PR #2149, opened 2012-07-24
Usability updates to the signin and signup pages.

shane-tomlinson:
In the signin page:
- If the user arrives with or enters an unknown secondary email, redirect them to the /signup page.
- If the user arrives with or enters a known secondary email in storage, show the password field.

In the signup page:
- If the user arrives with or enters an unknown secondary email, show the password fields.
- If the user arrives with or enters a known secondary email, redirect them to the /signup page.

The page load could use some work, but I am going for an incremental improvement here.

issue #2120

### DIFF
--- a/resources/static/pages/css/style.css
+++ b/resources/static/pages/css/style.css
@@ -462,7 +462,7 @@ button.delete:active {
   margin-bottom: 10px;
 }
 
-.siteinfo, #congrats, .password_entry, #verify_password, .enter_password .hint, #unknown_secondary, .verify_primary .forminputs {
+.siteinfo, #congrats, .password_entry, #verify_password, .enter_password .hint, .verify_primary .forminputs {
   display: none;
 }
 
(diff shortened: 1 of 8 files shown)


seanmonstar:
Interesting. Technically, it works well. I just wonder if it's a little too magical.

What if I typo my email address? Or I use an email I haven't registered yet? Automatically redirecting me to Create Account is not what I was expecting. Would it be better to improve the "unknown secondary" styles we originally had?

If we did, I'd say the the "sign up" link in the message is hard to notice (blue versus a dark grey background), and the Sign In button stays enabled. Maybe the "sign up" link could look more like an inline button, and the Sign In button gets greyed out and disabled?

Though, I'm not UX. Maybe better to rope in @skinny97214.

/cc @warner

shane-tomlinson:
Closing this in favor of a new pull request #2161.  Closing.

Issue #2131, opened 2012-07-20
Perform CEF logging where indicated by security review


jedp:
Ensure that CEF/syslog logging occurs where indicated by the security review in https://wiki.mozilla.org/Security/Reviews/Identity/browserid

jedp:
I think this bug fell out of sight.  This has been in production for some months now.  Closing.

PR #2150, opened 2012-07-24
Added CEF logging at key points alongside our app logger

jedp:
Issue #2131

CEF (common event format) logging will enable dev and security ops to use tools like ArcSight to more closely track security and system events.  

To facilitate this, I have put an external CEF logging module together: https://github.com/jedp/node-cef.  It sends log messages via UDP to a syslog service.  The service is configured in our `lib/configuration.js`.

In this pull request, I have tried to make a more friendly API for the CEF logger, and implemented logging at key locations identified in this security review: https://wiki.mozilla.org/Security/Reviews/Identity/browserid

@ygjb @lloyd @ozten can you help me answer the following questions?
- Have I adhered properly to the CEF spec and practices?
- Is the CEF section in `lib/configuration.js` adequate (particularly for udp syslog)?
- Could the `cef_logger` API be made less obtrusive and more developer-friendly?

Many thanks!
j

### DIFF
--- a/lib/cef_logger.js
+++ b/lib/cef_logger.js
@@ -1,0 +1,169 @@
+/**
+ * A thin abstraction layer over the cef library
+ *
+ * This is deliberately not a part of the logging.js library because
+ * we want CEF logging calls to stand out separately in the code.
+ *
+ * See the cef module's README.md file for more information.
+ */
+
+const cef = require('cef');
+
+const STACK_FRAME_RE = new RegExp(/at ((\S+)\s)?\(?([^:]+):(\d+):(\d+)/);
+const THIS_FILE = __filename.split('/')[__filename.split('/').length - 1];
+
+var Logger = function(config) {
+  this.cef = cef.getInstance(config);
+
+  // used by _getCallerSignature to provide a unique signature for
+  // callers by module, function, and line number.
+  this._moduleIndex = {};
+  this._functionIndex = {};
+  return this;
+};
+
+/**
+ * The CEF logger supports eight logging severities.  In practice, we
+ * use only four:
+ *
+ *   emergency: Completely out of whack. Someone needs to look at
+ *              this. Harm to the application, user account, or system
+ *              security could have taken place.
+ *
+ *   alert:     Suspicious activity or application has non-validated
+ *              user input. Impact is not known.
+ *
+ *   warn:      Normal security application stuff, login failures,
+ *              password changes.
+ *
+ *   info:      Normal app activity. Logins and various kinds of
+ *              transactions.
+ */
+
+Logger.prototype = {
+
+  /**
+   * Return the function name, module, line, and column of the code
+   * that called into this logger.  Utility function for _emit().
+   */
+  _getCaller: function() {
+    var err = new Error();
+    Error.captureStackTrace(err);
+
+    // Throw away the first line of the trace
+    var frames = err.stack.split('\n').slice(1);
+
+    // Find the first line in the stack that doesn't name this module.
+    var callerInfo = null;
+    for (var i = 0; i < frames.length; i++) {
+      if (frames[i].indexOf(THIS_FILE) === -1) {
+        callerInfo = STACK_FRAME_RE.exec(frames[i]);
+        break;
+      }
+    }
+
+    if (callerInfo) {
+      return {
+        function: callerInfo[2] || null,
+        module: callerInfo[3] || null,
+        line: callerInfo[4] || null,
+        column: callerInfo[5] || null
+      };
+    }
+    return null;
+  },
+
+  /**
+   * _emit() - an private method to map the logging methods below
+   * to the cef logger methods.
+   */
+  _emit: function(severity, signature, name, extensions) {
+    extensions = extensions || {};
+    var options = {
+      signature: signature,
+      name: name,
+      extensions: extensions || {}
+    };
+    this.cef[severity](options);
+  },
+
+  /**
+   * The following logging functions take these arguments:
+   *
+   * @param signature
+   *        (integer)    A short string describing the part of the
+   *                     application that is being logged.  Used by
+   *                     ArcSight for correlation.  Yes, this is
+   *                     annoying and awkward to have to enter.
+   *
+   * @param name
+   *        (string)     A human-readable description.  This is your log
+   *                     message.
+   *
+   * @param extensions
+   *        (object)     A dictionary of special key/value pairs designated
+   *                     by the CEF standard or by ArcSight.  You're
+   *                     most likely to care about the ones provided by
+   *                     the cef helper function,
+   *                     extensionsFromHTTPRequest(req).
+   */
+
+  emergency: function(signature, name, extensions) {
+    this._emit('emergency', signature, name, extensions);
+  },
+
+  alert: function(signature, name, extensions) {
+    this._emit('alert', signature, name, extensions);
+  },
+
+  warn: function(signature, name, extensions) {
+    this._emit('warn', signature, name, extensions);
+  },
+
+  info: function(signature, name, extensions) {
+    this._emit('info', signature, name, extensions);
+  }
+};
+
+/**
+ * getInstance()     get a singleton instance of the logger with the
+ *                   config specified in our configuration.js file.
+ */
+var instances = {};
+var getInstance = module.exports.getInstance = function getInstance(config) {
+  // maybe override app config
+  config = config || require('./configuration').get('cef');
+
+  // make an instance key for this config
+  var configKey = [];
+  Object.keys(config).sort().forEach(function(key) {
+    configKey.push(key + '=' + config[key]);
+  });
+  // stringify the array
+  configKey = configKey.join(',');
+
+  if (!instances[configKey]) {
+    instances[configKey] = new Logger(config);
+  }
+
+  return instances[configKey];
+};
+
+/**
+ * combineHTTPEnvWithParams - utility to merge output of
+ * cef.extensionsFromHTTPRequest with additional params
+ *
+ * @param request
+ *        (object)    The http request object
+ *
+ * @param extensions
+ *        (object)    Optional extra cef extension dictionary
+ */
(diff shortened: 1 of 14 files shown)


lib/cef_logger.js
```diff
@@ -0,0 +1,169 @@
...
+ *              transactions.
+ */
+
+Logger.prototype = {
+
+  /**
+   * Return the function name, module, line, and column of the code
+   * that called into this logger.  Utility function for _emit().
+   */
+  _getCaller: function() {
```
jedp, line 80: In case you're wondering ... this function doesn't get called, but I'm leaving it in here for now in the hopes that I can make the cef `signature` automatically somehow and have one argument fewer in the log calls.  If not, I'll remove this dead code.

lib/wsapi/account_cancel.js
```diff
@@ -15,8 +17,12 @@ exports.i18n = false;
 exports.process = function(req, res) {
   db.cancelAccount(req.session.userid, function(error) {
     if (error) {
+      cef_logger.alert("ACCOUNT_CANCEL", "Error canceling account",
+                      mergeEnv(req, {duser: req.session.userid, msg: error}));
```
jedp: There are two things I hate about this:
1. The signature (`ACCOUNT_CANCEL`) has to be entered and be consistent across all other account cancel events.  Since our wsapi modules do one thing, I'm hoping that the module name can be used in some form as the signature.
2. The awkward merging of http request and additional params.  I'm thinking I want to hide all this in the logger function and let you pass in the http request object and additional params as separate and optional arguments.  That would remove the `mergeEnv` call.  On the downside, it would be potentially more confusing when looking at function calls and wondering why they had different argument types.

But if I made both those changes, this function call would instead be:

``` javascript
cef_logger.alert("Error canceling account", req, {duser: req.session.userid, msg: error});
```

This comment goes for every other cef log call.

package.json
```diff
@@ -5,6 +5,7 @@
     "dependencies": {
         "JSONSelect": "0.4.0",
         "bcrypt": "0.4.1",
+        "cef": "0.3.3",
```
jedp, line 8: That's published from this: https://github.com/jedp/node-cef

lib/wsapi/account_cancel.js
jaredhirsch: @jedp: if you can hack it together, (1) sounds great. And I totally agree with (2), it seems like you make that mergeEnv call every single time.

lib/cef_logger.js
```diff
@@ -0,0 +1,169 @@
...
+ * This is deliberately not a part of the logging.js library because
+ * we want CEF logging calls to stand out separately in the code.
+ *
+ * See the cef module's README.md file for more information.
+ */
+
+const cef = require('cef');
+
+const STACK_FRAME_RE = new RegExp(/at ((\S+)\s)?\(?([^:]+):(\d+):(\d+)/);
+const THIS_FILE = __filename.split('/')[__filename.split('/').length - 1];
```
zaach: Node's [path.basename](http://nodejs.org/api/path.html#path_path_basename_p_ext) might be useful here.

lib/wsapi/account_cancel.js
zaach: +1 for hiding the merge call. The first point would be nice to have, though it looks like some modules have multiple cef calls with distinct signatures?

lib/cef_logger.js
jedp: +1 thank you

lib/wsapi/account_cancel.js
jedp: Thanks for your feedback @6a68 and @zaach - It does seem likely that multiple cef calls from within the same module will have to have different signatures, yes.  But at least the merge call will be easy to take care of.

lib/cef_logger.js
```diff
@@ -0,0 +1,187 @@
...
+ *
+ * See the cef module's README.md file for more information.
+ */
+
+const path = require('path'),
+      cef = require('cef'),
+      http = require('http');
+
+const STACK_FRAME_RE = new RegExp(/at ((\S+)\s)?\(?([^:]+):(\d+):(\d+)/);
+const THIS_FILE = path.basename(__filename);
```
jedp, line 15: Simplified - thanks, @zaach

lib/wsapi/authenticate_user.js
```diff
@@ -53,13 +55,16 @@ exports.process = function(req, res) {
 
         if (err) {
           if (err.indexOf('exceeded') != -1) {
+            cef_logger.alert("LOAD_HIGH", "Load exceeded on auth request", req);
             logger.warn("max load hit, failing on auth request with 503: " + err);
             res.status(503);
             return fail("server is too busy");
           }
+          cef_logger.alert("BCRYPT_ERROR", "Internal error on bcrypt.compare", req, {msg: err});
```
jedp, line 63: Log methods can now take any number of object args that will all get mashed into the CEF "extensions" dictionary.  Here, it's `req` and the `{msg: err}` dictionary.  HTTP request objects are automatically converted to something CEF can digest.

tests/cef-logging.js
```diff
@@ -0,0 +1,139 @@
...
+  "HTTP request and other parameters": {
+    topic: function() {
+      var cb = this.callback;
+      var logger = cef_logger.getInstance();
+      var server = http.createServer(function(req, res) {
+        // combine http request and a couple other objects
+        cb(null, logger.info('PIE', 'I like pie',
+                            {msg: "I like pie"},
+                             req,
+                            {end: Date.now()}));
```
jedp, line 105: Added this test to make sure that http request objects are properly recognized and converted to CEF data, and also to ensure that it can be mixed in with other normal CEF key/value objects.

jedp:
Thanks, @6a68 and @zaach - I've updated the PR in response to your previous comments.

lib/wsapi/account_cancel.js
jedp: Ok changed; see further below.

lib/cef_logger.js
```diff
@@ -0,0 +1,187 @@
...
+
+Logger.prototype = {
+
+  /**
+   * Return the function name, module, line, and column of the code
+   * that called into this logger.  Utility function for _emit().
+   */
+  _getCaller: function() {
+    var err = new Error();
+    Error.captureStackTrace(err);
```
ozten, line 82: Cool, I didn't know about this function.

Playing around, I don't see what side-effect it has. err.stack will be populated after line 81.

No good docs on `captureStackTrace`, maybe it's needed if you use prototype inheritance with Error.

It doesn't hurt anything to call captureStackTrace, so this is a really minor knit pick and trolling for good links on it.

lib/wsapi/account_cancel.js
```diff
@@ -15,8 +16,12 @@ exports.i18n = false;
 exports.process = function(req, res) {
   db.cancelAccount(req.session.userid, function(error) {
     if (error) {
+      cef_logger.alert("ACCOUNT_CANCEL", "Error canceling account",
+                      req, {duser: req.session.userid, msg: error});
       wsapi.databaseDown(res, error);
     } else {
+      cef_logger.warn("ACCOUNT_CANCEL", "Canceled user account",
```
ozten, line 23: Should signatures be unique? Is it okay to use `ACCOUNT_CANCEL` for two different events?

ozten:
Great work!

lib/wsapi/account_cancel.js
jedp, line 23: Separate events can share the same signature.  So it's more of a gang tattoo than a signature proper.  That lets ArcSight correlate different events with a common signature.  It also unfortunately prevents me from auto-generating the signature and thus getting rid of this argument altogether!

I'm going to propose that ArcSight rename this field to GangTattoo.

lib/cef_logger.js
jedp, line 82: I somehow convinced myself that was necessary.  But Yes, when I'm already creating a new `Error`, there's no point.  Thanks for the catch!

jedp:
Thanks for your review, @ozten !

jaredhirsch:
I have a higher-level question here: is there any good reason we shouldn't just push all of our logging into syslog/udp in this way? The advantages I can think of are
- complex event processors, like all machine-learning thingies, do better with more data, and might even find unexpected correlations within that extra data
- we're now duplicating/fragmenting our logs--might be nice to keep it all in one place
- from an opsy standpoint, just dumping into syslog/UDP seems simpler in production situations: just one set of logs to capture, rotate, backup, and it's in the same standard place any process should log output.

ozten:
This isn't how [ArcSight](http://www.arcsight.com/) works. We use this across Mozilla web apps. This library is analogous to our python one.

jaredhirsch:
@ozten - cool, that makes sense. I'll save my questions on logging/monitoring for devops playtime :-)

lloyd:
"This pull request cannot be automatically merged."

seanmonstar:
So, does this just need to be rebased, and it's otherwise r+?

jedp:
Now that beta has happened (rock and roll!), I shall rebase and update ...

lib/wsapi/have_email.js
```diff
@@ -19,7 +20,11 @@ exports.args = {
 
 exports.process = function(req, res) {
   db.emailKnown(req.params.email, function(err, known) {
-    if (err) return wsapi.databaseDown(res, err);
+    if (err) {
+      return wsapi.databaseDown(res, err);
+      cef_logger.alert("DB_FAILURE", "emailKnown failed",
```
seanmonstar, line 25: It looks like this alert will never happen, since it's after the return.

seanmonstar:
@jedp I've rebased this myself, you can see it here: https://github.com/seanmonstar/browserid/compare/cef-logging

Besides [my most recent comment](https://github.com/mozilla/browserid/pull/2150/files#r1803020), it should be good to go, I think.

seanmonstar:
This ship is sinking, bail out!

Opening a new, rebased PR at #2615

PR #2161, opened 2012-07-26
Issue 2120 combine signin signup pages into /signin

shane-tomlinson:
This is aimed at improving the UX of the signin/signup flow on the main site.  To do so, the signin/signup pages have been combined into the /signin page and act very similarly to the dialog.

Combined many ideas that @seanmonstar had on how to improve this.


resources/static/common/js/lib/dom-jquery.js
```diff
@@ -328,6 +328,46 @@ BrowserID.DOM = ( function() {
...
+        },
+
+        /**
+         * Slide an element down
+         * @method slideDown
+         * @param {selector || element} elementToSlide
+         * @param {number} [animationTime]
+         */
+        slideDown: function( elementToSlide, animationTime ) {
+          return jQuery( elementToSlide ).slideDown( animationTime || 250 );
```
seanmonstar: Why is this the only animation to get a default 250ms? Should it have a default at all?

seanmonstar:
Would it be an improvement to change the "Sign In" header to "Sign Up", when 2 password fields show up?

Issue #1023, opened 2012-02-01
Enhancement request: Add a name to the BrowserID popup window


davehunt:
Using Selenium to test a site with BrowserID currently has to cycle through all windows to find one with the title 'Browser ID'. It would be possible to switch directly to the window if it had a name assigned to it.

Current:

```
for handle in selenium.window_handles:
    selenium.switch_to_window(handle)
        if selenium.title == 'Browser ID':
            break
```

Ideal:

```
selenium.switch_to_window('browserid')
```

tonychung:
improving support to our automation needs is always a good thing.   Help us, help you.

shane-tomlinson:
@lloyd - what do you think of adding this to WinChan?

shane-tomlinson:
Pull request to WinChan issued to allow window.name to be specified. https://github.com/lloyd/winchan/pull/26

@lloyd - can you review the WinChan change?

@benadida - in auth_with_idp/main.js, you set the window.name to auth_with_primary.  I spoke to @davehunt about opening the window with another name so they they can get a reference to the window, something like persona_dialog.  Once the Selenium has a handle to the window, the window name can be changed without ill effect.

jrgm:
verified train-2012.08.17

PR #2162, opened 2012-07-26
Give the Persona window the name __persona_dialog when opening. 

shane-tomlinson:
@benadida, @lloyd - Would you two give this a review?

Useful for Selenium tests which need an easy mechanism to get a reference to the window being tested.  Once Selenium has a reference to the window, it's name can be changed.

Include the new WinChan which allows window.name to be set.  Setting the name of the dialog to __persona_dialog on load.

This does not appear to interfere with primary signin where it changes the window.name to "auth_with_primary"

/cc @davehunt 

issue #1023

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -814,79 +814,11 @@
               }
             }
           };
-        },
-        onOpen: function(cb) {
-          var o = "*";
-          var msgTarget = isIE ? findRelay() : window.opener;
-          if (!msgTarget) throw "can't find relay frame";
-          function doPost(msg) {
-            msg = JSON.stringify(msg);
-            if (isIE) msgTarget.doPost(msg, o);
-            else msgTarget.postMessage(msg, o);
-          }
-
-          function onMessage(e) {
-            // only one message gets through
-            removeListener(window, 'message', onMessage);
-            var d;
-            o = e.origin;
-            try {
-              d = JSON.parse(e.data);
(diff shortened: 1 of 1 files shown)


resources/static/include_js/include.js
```diff
@@ -811,17 +814,89 @@
...
+            cb = undefined;
+            // explicitly close the window, in case the client is trying to reload or nav
+            try { window.close(); } catch (e) { }
+          };
+          addListener(window, 'unload', onUnload);
+          return {
+            detach: function() {
+              removeListener(window, 'unload', onUnload);
+            }
+          };
```
lloyd: this was manually removed from include.js, because it's the open-ee side of winchan, which is not needed in include.js

it is a hokey way to partially include a different project, but why add it back in this patch?

resources/static/include_js/include.js
```diff
@@ -811,17 +814,89 @@
...
         }
       };
     } else {
       return {
         open: function(url, winopts, arg, cb) {
           setTimeout(function() { cb("unsupported browser"); }, 0);
+        },
+        onOpen: function(cb) {
+          setTimeout(function() { cb("unsupported browser"); }, 0);
```
lloyd: ditto.

lloyd:
r+, suggest pruning the unused code, as mentioned above, tho.

jrgm:
verified train-2012.08.17

Issue #1682, opened 2012-06-04
after verifying email address,  /signup, /forgot redirects to /signin


jrgm:
This only happens for the XUL Fennec client, not for any other browser on Android (FF Beta, Aurora, Nightly, Stock, Chrome Beta). 

STR: 
1. Go to diresworb.org and click sign up; enter a secondary email address, two passwords and submit.
2. Get the email for verify_email_address and open in the same browser

Expected: The /signup page transitions to show the usual logged-in account management UI.

Actual: /signup transitions to /signin and asks for an email address. However the user is logged in at this point, since clicking on a linke to diresworb.org/ will show the account management UI.

adb logcat doesn't show any relevant error.

shane-tomlinson:
@jrgm - I did this a bit more generally, the email does not have to be verified in the same browser.  I was able to reproduce this verifying the email in a separate browser on a separate machine.  Digging in.

shane-tomlinson:
Link to #1555 - These may have very similar root causes.

shane-tomlinson:
I am now getting this for all browsers on the main site for either resetting a password or signing up.

jrgm:
Was this mistakenly closed in confusion over GH-2179? I don't see this in dev with FF, Chrome, but I definitely have this with IE8.

shane-tomlinson:
This is still reproducible on login.dev.anosrep.org and mistakenly closed as @jrgm says, re-opening.

shane-tomlinson:
Updating with modified STR:
1. Go to login.dev.anosrep.org and click sign up; enter a secondary email address, two passwords and submit.
2. Get the email for verify_email_address and open in a different browser.

Expected: The /signup page transitions to show the usual logged-in account management UI.

Actual: /signup transitions to /signin and asks for an email address.

PR #2163, opened 2012-07-26
Fix for users receiving a mustAuth response from the backend even though they set the password in this session.

shane-tomlinson:
If a user has set the password in this session (either dialog or main site) and they receive a mustAuth response from the verification poll, authenticate the user with the already entered email address and password.

This fixes users having to enter their password a second time on the /forgot and /signup pages after the verification is complete.

issue #1682


resources/static/common/js/user.js
```diff
@@ -103,17 +105,26 @@ BrowserID.User = (function() {
...
-    // RP and logged in.
-    var site = User.getReturnTo();
-    if (staged && site) storage.setReturnTo(site);
-
-    complete(onComplete, status);
+  function stageAddressVerification(email, password, stagingStrategy, onComplete, onFailure) {
+    // These are saved for the addressVerificationPoll.  If there is
+    // a stagedEmail or stagedPassword when the poll completes, try to
+    // authenticate the user.
+    stagedEmail = email;
```
shane-tomlinson, line 110: Every time an email/password is staged, save off the staged email and password to use on completion.

resources/static/common/js/user.js
```diff
@@ -148,6 +159,68 @@ BrowserID.User = (function() {
   }
 
   function addressVerificationPoll(checkFunc, email, onSuccess, onFailure) {
+    function userVerified(completionStatus) {
+      if (stagedEmail && stagedPassword) {
```
shane-tomlinson, line 161: If there is a stagedEmail and stagedPassword, re-authenticate the user using the stagedEmail and stagedPassword.

resources/static/common/js/user.js
```diff
@@ -148,6 +159,68 @@ BrowserID.User = (function() {
...
+          completeVerification(completionStatus);
+        }, onFailure);
+      }
+    }
+
+    function completeVerification(status) {
+      // As soon as the registration comes back as complete, we should
+      // ensure that the stagedOnBehalfOf is cleared so there is no stale
+      // data.
+      storage.setReturnTo("");
```
shane-tomlinson, line 195: This logic is taken verbatim from below.

shane-tomlinson:
Note that once this is merged, the old code in resources/static/dialog/js/modules/check_registration can be removed.

@lloyd - can you have a look at this?

seanmonstar:
@shane-tomlinson since the line comments seem needed to explain what's going on, should they be in the code instead?

shane-tomlinson:
@seanmonstar - outside of the last comment about "this logic is taken verbatim from below", the comments are already in the code as well, and slightly more verbose. I wanted to draw attention to them in the pull request.

seanmonstar:
This looks fine to me, the only thing stopping me from clicking merge is that it's 3 months old. I can click?

PR #2168, opened 2012-07-27
don't set process.env.SUPPORTS_SIGNALS = false

seanmonstar:
process.env converts all values set into strings,
so its better to just not set it.


shane-tomlinson:
don't set process.env.SUPPORTS_SIGNALS = false

Manually merged with checkin 9ca27e00279cfe37b87a3a84df7fc4ef2b1fe17c.

r+

Thanks @seanmonstar!

Issue #1639, opened 2012-05-24
Consider putting link to complete registration in email header


jedp:
When stage_user is called, browserid responds by mailing back a link for the user to click to finish his or her registration.

To make it easier to automate the registration process, it would be handy not to have to parse the link out of the body of an email, but instead simply to pull it from a header like X-BrowserID-Registration-Link.

nu?

fmarier:
Crazy idea: Thunderbird could parse that header and present something nice to the user :)

jedp:
fmarier++

lloyd:
let's do this when we redesign email!

shane-tomlinson:
PR #2174 manually merged with commit ce47dd5bf3e64d30e3b7ca5fdcb6afa894a8bd1a.  Closing.

jrgm:
verified train-2012.08.17

One quirk to note is that when doing e.g., signup on the main site X-Browserid-Relyingparty lists the hostname of the main site e.g., login.dev.anosrep.org, since there is no actual relying party in this case.

PR #2174, opened 2012-07-27
Add VerificationURL and RelyingParty to verification email headers

jedp:
While emails are on everybody's mind ...

Issue #1639

Adds X- headers containing verification url and RP in order to make automated handing easier and less susceptible to breakage if email templates or URL patterns change.

This would be used by https://github.com/lloyd/personatestuser.org

### DIFF
--- a/lib/email.js
+++ b/lib/email.js
@@ -30,24 +30,24 @@ if (smtp_params && smtp_params.host) {
 
 const TEMPLATE_PATH = path.join(__dirname, "static", "email_templates");
 
-// the underbar decorator to allow getext to extract strings 
+// the underbar decorator to allow getext to extract strings
 function _(str) { return str; }
 
 // a map of all the different emails we send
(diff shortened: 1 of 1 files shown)


jedp:
@lloyd would you mind giving this a thumbs up or down?

shane-tomlinson:
@jedp - could you update this for merging?

shane-tomlinson:
Manually merged into dev with checkin ce47dd5bf3e64d30e3b7ca5fdcb6afa894a8bd1a.

Created ephemeral deployment, checked that sent email contained two new headers.

r+

Nice work @jedp

jrgm:
verified train-2012.08.17

One quirk to note is that when doing e.g., signup on the main site X-Browserid-Relyingparty lists the hostname of the main site e.g., login.dev.anosrep.org, since there is no actual relying party in this case.

Issue #2176, opened 2012-07-28
sites_signed_in and sites_visited KPI renaming


shane-tomlinson:
PR #1807 introduced multiple new KPIs including sites_signed_in and sites_visited.

Neither of these names clearly convey their intent.

sites_signed_in should be number_sites_signed_in
sites_visited should be number_sites_remembered

jedp:
I particularly like the move to `number_sites_remembered`

@nmalkin , let's not forget to do some data migration in kpiggybank

nmalkin:
@jedp, is that how you think we should proceed? I was going to build in support for both names into the dashboard.

Now that I think about it, though, that would be _three_ different names, and the code would become irrelevant pretty soon. So, yeah, I'm liking the idea of data migration. I'll open a separate issue for that.

jedp:
Merged PR 2177

jrgm:
verified in local build train-2012.08.17.

PR #2177, opened 2012-07-28
KPI renaming

shane-tomlinson:
- sites_signed_in => number_sites_signed_in
- sites_visited => number_sites_remembered

issue #2176


jedp:
Thank you for making these nice!  r+

Issue #1345, opened 2012-03-28
/signin should not be accessible for authenticated users


cvan:
Steps to reproduce:
1. Log in @ https://browserid.org/signin
2. Load https://browserid.org/signin

Expected:
https://browserid.org/signin should redirect to https://browserid.org/ for authenticated users

Actual:
https://browserid.org/signin is still accessible even for authenticated users

lloyd:
we should probably address this when we rebrand browserid.org

shane-tomlinson:
Link to #1511

PR #2178, opened 2012-07-28
If the user hits the /signin or /forgot pages while signed in, redirect them to the "/" page.

shane-tomlinson:
issue #1345

Things to check with this pull request:
- Make sure an authenticated user can still access /about, /privacy, /tos, and /
- Make sure a user with cookies turned off can still access /about, /privacy, /tos, and /
- Make sure a user with javascript turned off can still view /about, /privacy, /tos, and /
- Make sure an authenticated user is redirected to / when going to the /signin or /forgot pages.

### DIFF
--- a/resources/static/pages/js/start.js
+++ b/resources/static/pages/js/start.js
@@ -18,7 +18,7 @@ $(function() {
       xhr = bid.XHR,
       network = bid.Network,
       token = pageHelpers.getParameterByName("token"),
-      path = document.location.pathname,
+      path = document.location.pathname || "/",
       moduleManager = bid.module,
       modules = bid.Modules,
       CookieCheck = modules.CookieCheck,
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
This pull request makes much more sense when looked at with the w=1 GET option:

https://github.com/mozilla/browserid/pull/2178/files?w=1

seanmonstar:
What do you think of a test case for this?

shane-tomlinson:
@seanmonstar - I want to add unit tests for this module, but without a refactor, there is no way to load the module multiple times.

seanmonstar:
Oh OK, fair enough. I hadn't noticed that the module was such an
anti-social singleton.

r+

jrgm:
verified train-2012.08.17

Issue #2179, opened 2012-07-28
going from /signin to /forgot via the "is this your address?" link redirects the user back to /signin


shane-tomlinson:
regressions are bad.

jrgm:
To flesh out STR (as I understand them)
1. Go to main site and click sign in. Enter the email of a known secondary email address and click sign in button.
2. Instead of entering a password, click on "forgot your password?"
3. Enter a new password twice and click 'reset password'
4. Go to email and click on the reset_password link to open in a new browser window (but same browser as the one you started in). 

Expected: window 2 shows "Your address has been verified". Window 1 should transition to show the management UI.

Actual: window 2 does as expected, but window 1 transitions to show the original /signin page with a prompt to enter email

Note: however with window 1 if you click on 'Mozilla Persona' you will see that you are in fact already signed-in.

shane-tomlinson:
@jrgm - you STRs are the STRs for #1682 and PR #2163.

The STRs for this particular bug are:

1) Go to main site /signin
2) Enter a known secondary
3) Click "Forgot your password?"
4) It redirects to the /forgot page, then redirects back to the /signin page.

jrgm:
@shane-tomlinson Ah sorry. Hmm. I cannot reproduce this with 07.20 or 07.30 or current dev with osx/chrome. What am I missing?

lloyd:
@shane-tomlinson

I cannot reproduce on ansorep.org (train-2012.07.20)
I cannot reproduce on login.dev.anosrep.org right now

what am I doing wrong?

lloyd:
cannot reproduce in our development environment anymore, nor can shane.  WFM, please re-open if you can figure out whether this is still relevant?

```
14:25 < stomlinson> lloyd: 2179 only affected dev (the current dev), it was never a problem in 7/20.
14:25 < stomlinson> and now I'm not getting it to reproduce.  grrr...
```

jrgm:
verified not a bug now

PR #2180, opened 2012-07-28
Fix the email not being saved when the user goes from /signin to /forgot via the "forgot your password?" link.

shane-tomlinson:
- Clean up behaviors when going between /signin and /forgot.
- Emails only need to be stored when toggling between /signin and /forgot. Instead of saving the email every time the user types a charcter, save the email when the user goes from /signin to /forgot via the "forgot your password?" link. Let the browser take care of auto-fill.
- pageHelpers.setupEmail now returns an email if it is available.
#2179

### DIFF
--- a/resources/static/pages/js/forgot.js
+++ b/resources/static/pages/js/forgot.js
@@ -53,11 +53,10 @@ BrowserID.forgot = (function() {
   }
 
   function redirectIfNeeded(doc, ready) {
-    // email addresses are stored if the user is coming from the signin or
-    // signup page.  If no email address is stored, the user browsed here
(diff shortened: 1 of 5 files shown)


lloyd:
r+ from code review, deploying and manually testing.

lloyd:
I cannot figure out the problem that this pull request solves.  @jrgm nor I can reproduce issue #2179 in any environment.

Testing out signin and forgot pages I see no regressions.  

Is this just code cleanup then?

shane-tomlinson:
@lloyd - This has become an irrelevant PR that fixes a problem that I never actually committed to dev.  The code cleanup is unnecessary.  Closing.

Issue #1984, opened 2012-07-11
Emphermeral deployments include a broken myfavoritebeer


ozten:
Ephemeral deployments still include $hostname.myfavoritebeer.org.

This example RP is broken, since it sources $hostname.hacksign.in/include.js

I spent a lot of time greping the VM image trying to understand what is going on, and then eventually wondered if we migrated to 123done...

We should fix myfavoritebeer or remove it completely form the deployment to avoid confusion.

stephendonner:
+1

jaredhirsch:
since myfavoritebeer is the reference site for the old navigator.id.get API, we should probably fix it, not abandon it.

@lloyd, I can steal this if you can provide support when I inevitably get stuck

jrgm:
@6a68 -> https://github.com/lloyd/myfavoritebeer.org/blob/master/server/main.js#L92

jaredhirsch:
@jrgm: thanks :-)

I managed to find it on my own: https://github.com/lloyd/myfavoritebeer.org/pull/57

jrgm:
Noted in the lloyd/myfavoritebeer.org/pull/57 that since we don't have stage and rely on an ephemeral we can't test with myfavoritebeer.org as an RP.

lloyd:
Will fix fri...

--lloyd

shane-tomlinson:
This was fixed with https://github.com/lloyd/myfavoritebeer.org/pull/57. Closing as fixed.

PR #2184, opened 2012-07-30
Pull in BrowserID-Tests

jaredhirsch:
# Overview

Import selenium tests (previously in mozilla/BrowserID-Tests), and a submodule test dependency (mozilla/bidpom), via a circuitous route. The route taken is irrelevant except in the probably rare case where we want to push or pull upstream code from mozilla/bidpom, which is a general-purpose library maintained by webqa.

This is existing code except for the new README inside the automation-tests directory and one commit that fixes some minor bugs stemming from upstream removal of selenium 1 (the RC API) code paths.

These tests can be run, today, against ephemeral instances via 123done. (And myfavoritebeers, once I fix a small, annoyingly high-impact, [bug](https://github.com/mozilla/browserid/issues/1984).)

This pull request description is pretty long because I had to jump through some annoying git hoops to come out with a configuration that met our needs.
# Gory details

mozilla/BrowserID-Tests has tests separated into 2 directories, 123done/ and myfavoritebeers/

These both source in mozilla/bidpom as git submodules under a directory called 'browserid', so you have a git submodule at /123done/browserid, and another (pointing at the same repo) at /myfavoritebeers/browserid. Git submodules are teh suck for branchy development, so I instead sourced bidpom using git-subtree.

In order to quickly make this git config change without write access to mozilla/BrowserID-Tests, I forked to 6a68/BrowserID-Tests, then git-subtree'd mozilla/bidpom into my fork. Then, I again used git-subtree to pull 6a68/BrowserID-Tests into 6a68/browserid. 

The net result of the git-subtreeing operations is that, from the perspective of browserid, the selenium tests inside BrowserID-Tests, and the stuff from the mozilla/bidpom repo, all just looks like normal files. However, we can still upstream to mozilla/bidpom.git, which might be useful in the future. We can probably do away with the BrowserID-Tests repo altogether, since those tests have been effectively pulled in.
### A bit about git-subtree vs git submodules

First off, warning: "git subtree" is different from "git-subtree". This ambiguity is unfortunate, as it makes googling a little tricky.

What we want is:
- to include test changes and production code changes in a single commit, for code reviewers and to keep history clean
- to be able to easily branch test code along with production code
- to be able to occasionally push or pull updates upstream (to mozilla/bidpom)

Git submodules don't track to branches; they track to commits. This makes it tricky and difficult to switch cleanly and quickly between experimental branches.

Git submodules are opaque to the super-project, so it's impossible to collect changes to super-project and submodule inside one commit, which, in our case, means code reviewers might have to traverse three repos (browserid, BrowserID-Tests, bidpom) to see related pieces of one commit. It's also a lot of annoying administrative work for the developer writing the code.

git-subtree solves both these problems. Apenwarr's [blog post](apenwarr.ca/log/?m=200904) and the git-subtree [help file](https://github.com/apenwarr/git-subtree/blob/master/git-subtree.txt) explain this really well; go have a look, if you're itnerested.

Note, git-subtree was added to mainline git in version 1.7.11 inside the contrib directory, meaning you have to install it manually. Its makefile doesn't quite work; see [here](https://github.com/mxcl/homebrew/issues/12897#issuecomment-6751404) for instructions. (Don't worry, it's straightforward.)

### DIFF
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -236,7 +236,10 @@
 
         // if window is unloaded and the client hasn't called cb, it's an error
         var onUnload = function() {
-          removeListener(isIE ? msgTarget : window, 'message', onDie);
+          try {
+            // IE8 doesn't like this...
+            removeListener(isIE ? msgTarget : window, 'message', onDie);
+          } catch (ohWell) { }
           if (cb) doPost({ a: 'error', d: 'client closed window' });
           cb = undefined;
           // explicitly close the window, in case the client is trying to reload or nav


automation-tests/123done/tests/test_add_another_email.py
```diff
@@ -0,0 +1,61 @@
...
+        Assert.contains("Thank you for signing up with Persona.", complete_registration.thank_you)
+
+        home_pg.wait_for_user_login()
+        Assert.equal(home_pg.logged_in_user_email, user['email'])
+
+        home_pg.click_logout()
+
+        second_user = MockUser()
+        bid_login = home_pg.click_sign_in(expect='returning')
+        bid_login.sign_in_add_another_email(second_user['email'])
```
zaach: This test fails here: `E       AttributeError: 'SignIn' object has no attribute 'sign_in_add_another_email'`

The definition seems to be missing from `sign_in.py`.

jaredhirsch: @zaach, thanks for pointing that out. I'm debugging that right now :-)

zaach:
The myfavoritebeer.org tests fail in both Firefox and Chrome with `E               Exception: Popup has not loaded
`. Besides that and the issue reported with the 123done.org sign in test, things run smoothly, though I notice tests tend to fail if the driver runs in the background without focus (maybe a general selenium problem).

automation-tests/myfavoritebeer/browserid/pages/sign_in.py
```diff
@@ -0,0 +1,224 @@
...
+        Base.__init__(self, selenium, timeout)
+
+        if self.selenium.title != self._page_title:
+            for handle in self.selenium.window_handles:
+                self.selenium.switch_to_window(handle)
+                WebDriverWait(self.selenium, self.timeout).until(lambda s: s.title)
+                if self.selenium.title == self._page_title:
+                    break
+            else:
+                raise Exception('Popup has not loaded')
```
zaach: myfavoritebeer.org tests are failing here.

jaredhirsch: I can't repro the test failures. What's the exact command-line incantation you're using?

automation-tests/123done/tests/test_add_another_email.py
jaredhirsch: @zaach, this is fixed and pushed into the pull request. Feel free to try again.

automation-tests/myfavoritebeer/browserid/pages/sign_in.py
jaredhirsch: @zaach: Oh, I bet I know what happened. Did you manually create a test account, and put its email and password inside credentials.yaml? If you don't, the dialog will time out. Give that a shot and let me know if it's still broken.

zaach:
```
py.test --destructive --credentials=credentials.yaml --baseurl=http://www.myfavoritebeer.org --driver=chrome --chromepath=/usr/local/bin/chromedriver -q myfavoritebeer
```

and

```
py.test --destructive --credentials=credentials.yaml --baseurl=http://www.myfavoritebeer.org --driver=firefox -q myfavoritebeer
```

The credentials should be okay (they worked for the 123done.org tests, at least).

Thanks!

automation-tests/123done/tests/test_add_another_email.py
zaach: @6a68 Awesome, works great.

automation-tests/myfavoritebeer/browserid/pages/sign_in.py
zaach: Doh, it should be dev.myfavoritebeer.org, not www.myfavoritebeer.org.

lloyd:
r-

First, let's not subtree in BrowserID-Tests - let's obsolete that repo and move tests in tree.

Second, let's create one script that does everything for you - pulls down jar files and kicks off the tests in one fell swoop - this directly increases chances that we'll run these things.  Suggestion: should be easy to point at an ephemeral instance.

warner:
FYI, I read over the git-subtree docs, and you're right, it looks ideal for pulling in a copy of upstream `bidpom` without imposing a burden on regular users. Just be sure to add a note somewhere with the commands needed to update the local copy or push to upstream (and a pointer to the git-subtree docs, since they aren't part of a normal git install).

+1 to @lloyd's suggestion to copy the BrowserID-Tests code into the browserid repo and then kill off BrowserID-Tests.

PR #2189, opened 2012-07-31
make placeholder text in inputs lighter grey

### LINKED ISSUE #2187: Persona sign-up page's placeholder input texts same font as regular input
This screenshot demonstrates how the placeholder texts appear just like regular input texts. The placeholder texts should be faint degree of grey or something. At the moment I'm just not sure if it's my text or the default text in there. 
http://cl.ly/IQ7o

seanmonstar:
only Firefox seems to make the placeholder text
be the same color as the normal text, so this will
set the placeholder to a lighter grey

fixes #2187

### DIFF
--- a/resources/static/common/css/style.css
+++ b/resources/static/common/css/style.css
@@ -141,6 +141,12 @@ input[type=password]:disabled {
     border-color: #b2b2b2;
 }
 
+input[type=text]:-moz-placeholder,
+input[type=email]:-moz-placeholder,
+input[type=password]:-moz-placeholder {
+   color:#aaa;
+}
+
 label {
   display: block;
 }


shane-tomlinson:
@seanmonstar - I like this change.  I think we should also not rely on webkit styling to #aaa by default and should set it for both webkit and IE10.

http://www.kirupa.com/html5/styling_placeholder_text_css.htm

IE10 has the :-ms-input-placeholder CSS selector, webkit uses :-webkit-input-placeholder

With that change, this gets an r+

shane-tomlinson:
Manually merged after adding the -ms-input-placeholder and -webkit-input-placeholder.  ffc5c4ea305bce6c92d7b8abbc135acd132ca741

r+, thanks @seanmonstar!

PR #2190, opened 2012-07-31
catch winchan removeListener error in IE8

### LINKED ISSUE #2183: IE doesn't remember you (was: [ie8] train-2012.07.20 js exception when signing in)
STR
1. On stage login.anosrep.org and beta.123done.org /sign_in with a clean profile and a known secondary email address
2. Enter the password and hit next

Expected: dialog dismisses and you are logged in

Actual: This JS exception is thrown. You are logged in but 1) if you reload beta.123done.org you're logged in state is not retained, and 2) clicking logout does nothing.

```
Line: 3
Character: 28936
Code: 0
Error Message: Permission denied
URL: https://static.login.anosrep.org/v/1ea909f058/production/en_US/dialog.js
```

In my local dev with train-2012-07.20 the exception in unminified code is:

```
Line: 13
Character: 5
Code: 0
Error Message: Permission denied
URL: http://10.250.2.124:10002/common/js/lib/winchan.js
```

seanmonstar:
removeListener throws in IE8

pending on lloyd/winchan#27 getting merged in, this contains those changes.

would fix #2183

### DIFF
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -236,7 +236,10 @@
 
         // if window is unloaded and the client hasn't called cb, it's an error
         var onUnload = function() {
-          removeListener(isIE ? msgTarget : window, 'message', onDie);
+          try {
+            // IE8 doesn't like this...
+            removeListener(isIE ? msgTarget : window, 'message', onDie);
+          } catch (ohWell) { }
           if (cb) doPost({ a: 'error', d: 'client closed window' });
           cb = undefined;
           // explicitly close the window, in case the client is trying to reload or nav


shane-tomlinson:
@seanmonstar - There is the external WinChan which is used for the popup, the version inside of dialog.js should be updated for this as well.  (another instance of dynamic creation of include.js would be handy)

seanmonstar:
The include.js specifically removed the onOpen methods, perhaps to shave
bytes since they aren't used.
On Jul 31, 2012 3:24 AM, "Shane Tomlinson" <
reply@reply.github.com>
wrote:

lloyd:
r+

jrgm:
Hmm, so this isn't on train-2012.07.30. I still get the exception as originally reported in GH-2183 (but the logged in state does hold upon reload or restart of IE8 now with the other fixes added to train-2012.07.30).

jrgm:
Hey @lloyd, we need to pick this one up for train-2012.07.30 I still have this (non-fatal) js error when using IE8

jrgm:
verified train-2012.08.17

Issue #1859, opened 2012-06-28
Rpm version string improvement


david-caro:
Hi,
I'm using the browserid server at work, and the version string does not adapt to the requirements of the company. I'd like to change the version string to something a little more descriptive, here's my proposal (as I explained to fetep):

_Major.branch[-num-commits_hash-of-last]_

**Version string (just modified a little):** _MAJOR.tagname/branchname_
This is the the closest tag name (or the branch name, if there are no tags before the branch) prepended with the _MAJOR_ (a custom number to reflect a major release, mandatory if the changes made result on a code that is not backwards compatible).

**Release string:** _Num-of-commits_sha_
This will be the number of commits since the tag/branch and the short sha of the last commit, to reflect that the code is different in the case of packages from different repos or a rebase (bad rebase!). This can be easily retrieved with the describe command.

Also adding the source repository on the description, to keep track of the source of the code. And the info about the translations.

A sample version string will be this:

_0.train_2012.07.15-12_fa3613f_

That means the 12th commit after the tag _train_2012.07.15_. If you commit again, the tag will be something like:

_0.train_2012.07.15-13_11fa5d_

Even if the hash is not incremental, the version string is incremental (because the number of commits is placed before the hash). And, if you use a different repo with the same tag/branch, and the same number of commits, you see that the hash has changed (that doesn't necessarily mean a bump on the version).

What do you think?

lloyd:
@fetep what do you think about an update to rpm scripts to append sha to version?

david-caro:
Seeing that no one complains about this, I've changed the build script and the spec file to use this new version format. Opened the Pull request https://github.com/mozilla/browserid/pull/2192

lloyd:
@gene1wood how does this look to you guys?

ozten:
+1

FWIW

gene1wood:
Is the "get_ci_count" function still being used now that you're determining the number of commits with the "git describe" queries?

david-caro:
No, it's not used here, but I've found some problems when calculating the commits when there's a branch and no tags newer, and I'm preparing a new commit (that uses this function again).
I hope to upload it today (I'm just testing it, but the new version of browserid requires node >=0.8 and I'm just installing it to check if everything is ok.

gene1wood:
Cool

lloyd:
I closed down the pull request referencing this because it was stale.  Do we still have people that care deeply about this issue?  @gene1wood or @6a68 - any thoughts?

[...]

PR #2192, opened 2012-07-31
Fixes: mozilla/browserid#1859

david-caro:
Using the new version string format (and some minor improvements)

### DIFF
--- a/scripts/rpmbuild.sh
+++ b/scripts/rpmbuild.sh
@@ -3,14 +3,90 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-set -e
+usage()
+{
+    cat <<EOU
+    Usage: $0 [-hv] [major]
+
+    Creates an rpm package of the current version
+
+    Options:
+        -h
+            Show this help
+        -v
+            Be verbose
+
+    Parameters:
+        major
+            Major version to use, defaults to '0'
+
+
+EOU
+    exit ${1:-0}
+}
+
+get_locale()
+{
+    [[ -d locale/.svn ]] \
+    && svn up locale \
+    || {
+        [[ -d locale ]] \
+        || svn co  http://svn.mozilla.org/projects/l10n-misc/trunk/browserid/locale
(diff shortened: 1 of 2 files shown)


scripts/rpmbuild.sh
```diff
@@ -3,14 +3,90 @@
...
+    || {
+        [[ -d locale ]] \
+        || svn co  http://svn.mozilla.org/projects/l10n-misc/trunk/browserid/locale
+    }   
+    LOCALE_REV="$(svn info locale/ | sed -n -e "s,^Rev.*n: ,,p")"
+}
+
+get_ver_string()
+{
+    branch="$(git branch | grep '*')"
```
gene1wood: Maybe something like this instead? : 

branch="$(git symbolic-ref HEAD 2>/dev/null)" || branch="(unnamed branch)"
branch=${branch##refs/heads/}

scripts/rpmbuild.sh
```diff
@@ -3,14 +3,90 @@
...
+    branch="${branch:2}"
```
gene1wood: And remove this line

scripts/rpmbuild.sh
david-caro: That's better thanks

david-caro:
I'm having problems testing this new code, 'cause when building it complains about a missing file named  resources/static/common/js/lib/bidbundle.js, that existed before (in the tain-2012.07.30 branch at least).
When executing the env CONFIG_FILES=config/l10n-all.json scripts/compress part.

Somebody can help me get past this so I can fully test it?
Thaks!

david-caro:
BTW. the file: https://github.com/mozilla/browserid/blob/train-2012.07.30/resources/static/common/js/lib/bidbundle.js

seanmonstar:
@david-caro That file used to be a symlink, but that was removed to place nice on Windows. `npm install` will now make that file directly.

You can fix it doing:

```
rm -rf node_modules/
npm install
```

lloyd:
this is out of date, we will be more responsive moving forward.  But given no action in 6 months and that it doesn't merge cleanly, I'm going to close it down.

PR #2202, opened 2012-08-01
Add a console message if only one of privacyPolicy or termsOfService is defined.

shane-tomlinson:
issue #1438

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1076,9 +1076,9 @@
       }
     }
 
-    function log(message) {
+    function warn(message) {
       try {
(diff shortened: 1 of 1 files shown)


seanmonstar:
just a general question regarding logging deprecating warnings: is it better to console.warn() deprecation _warnings_?

seanmonstar:
otherwise, looks good from here

shane-tomlinson:
@seanmonstar - excellent suggestion.  According to MDN, IE down to 8 supports console.warn - https://developer.mozilla.org/en-US/docs/DOM/console.warn

Will update.

PR #2208, opened 2012-08-02
Add an example IdP for dynamic identifiers

ozten:
You can play with this by going to the test RP on :10001

Enter alice@dynamicidexample.domain

You will be authenticated as 123456@dynamicidexample.domain

See comments at the top of scripts/serve_example_dynamic_id_primary.js

Note: **dynamic-identifiers** is the long lived branch to support BlueVia IdP work. This pull request is against that empty branch, recently taken from the **dev** branch.


ozten:
@lloyd I pushed two new commit to this. Now manual testing ends with:

'uncaught exception: invalid email' state.js line 275

@benadida - The cert that this example IdP issues, doesn't contain protocol: null. Does jwcrypto need to be updated to use the protocol property?
https://github.com/mozilla/browserid/blob/dynamic-identifiers-example-idp/scripts/serve_example_dynamic_id_primary.js#L122

ozten:
Hmmm, I'm noticing a difference in how jwcrypto.cert.sign is called in our example RPs versus the [Certifier](https://github.com/mozilla/browserid-certifier/blob/master/lib/keysigner-compute.js#L15)

ozten:
FYI https://github.com/mozilla/jwcrypto/pull/44

shane-tomlinson:
@ozten - are you keeping the dynamic-identifiers branch up to date with the latest dev versions?

ozten:
Sadly, no.

ozten:
Won't fix

Issue #2215, opened 2012-08-03
call to navigator.id.get with no options or navigator.id.getVerifiedEmail throws exception


shane-tomlinson:
```
TypeError: passedOptions is undefined
[Break On This Error]   

opts.privacyPolicy =  passedOptions.privacyPolicy || undefined;
```

This affects the current staging - 07.20.2012

jrgm:
verified train-2012.08.17

PR #2216, opened 2012-08-03
Make sure navigator.id.get can be called with no options.

shane-tomlinson:
This fixes both navigator.id.get called with no options and navigator.id.getVerifiedEmail, which is still officially supported.

issue #2215

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1188,6 +1188,7 @@
       // get an assertion
       get: function(callback, passedOptions) {
         var opts = {};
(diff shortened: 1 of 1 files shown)


resources/static/include_js/include.js
```diff
@@ -1188,6 +1188,7 @@
       // get an assertion
       get: function(callback, passedOptions) {
         var opts = {};
+        passedOptions = passedOptions || {};
```
lloyd, line 1191: nice.

Issue #2218, opened 2012-08-03
IE will not keep you logged in (observer API, current production)


lloyd:
STR:
1. log into 123done.org on IE (8, 9, 10)
2. restart your IE browser
3. return to 123done.org

expected: you're still logged in
actual: you ain't

lloyd:
fixed in dev and train-2012.07.30

seanmonstar:
Was this what latchkey was running into? It was in prod?

jrgm:
Verified. With [ ie8, ie9, ie10 ] X [ train0730.123done.org, dev.123done.org ] after signin, then quit IE, then return to 123done in IE, I recover my signed-in state correctly.

PR #2220, opened 2012-08-03
Fix "IE doesn't keep you logged in between restarts (or page reloads)" p...

lloyd:
...roblems by disabling cookie check in iframe.  issue #2218 & issue #2183

fix for dev branch

### DIFF
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -16,6 +16,15 @@
 
   network.init();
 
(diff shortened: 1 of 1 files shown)


lloyd:
@shane-tomlinson gave me a verbal r+.  woohoo.

jrgm:
Verified. With [ ie8, ie9, ie10 ] X [ train0730.123done.org, dev.123done.org ] after signin, then quit IE, then return to 123done in IE, I recover my signed-in state correctly.

Issue #1659, opened 2012-05-29
if rp specifies all of privacyPolicy,termsOfService and the old params, the old params trump the new


jrgm:
In GH-1643 the params privacyURL and tosURL were changed to privacyPolicy and termsOfService. The old parameters can still be used and you can even mix and match. 

But if someone sets all four, the deprecated privacyURL and tosURL values trump the new params' values. 

Very minor point.

jrgm:
verfied train-2012.08.17

PR #2228, opened 2012-08-06
Sanity checks for the deprecated tosURL and privacyURL

shane-tomlinson:
- Use termsOfService over tosURL
- Ensure only one of termsOfService or tosURL is defined
- Use privacyPolicy over privacyURL
- Ensure only one of privacyPolicy or privacyURL is defined

@lloyd - I feel like the exceptions that are thrown when both termsOfService and tosURL/privacyPolicy and privacyURL should be done in dialog.js.  This would ensure that native callers of the dialog code are acting as we expect as well.

I placed these in include.js for now because watch has a similar check for loggedInEmail

issue #1659


jrgm:
verified train-2012.08.17

Issue #2227, opened 2012-08-06
example RP - loggedInEmail still used in navigator.id.watch


shane-tomlinson:

jrgm:
verified train-2012.08.17

PR #2229, opened 2012-08-06
Change loggedInEmail to loggedInUser in the example RP

shane-tomlinson:
To test:
- Make sure the example RP still works.
- Make sure the console message that says "loggedInEmail has been deprectated" is no longer seen.

issue #2227


zaach:
Confirmed deprecation warning is gone and RP works. r+

PR #2232, opened 2012-08-06
update bcrypt to 0.7.1

seanmonstar:
this adds Windows support to bcrypt, as long as you're
using nodejs >= 0.8.5

fixes deprecation warning since newer bcrypt has new API

one of the last pieces for #1769

### DIFF
--- a/lib/bcrypt-compute.js
+++ b/lib/bcrypt-compute.js
@@ -6,10 +6,10 @@ const bcrypt = require('bcrypt');
 
 process.on('message', function(m) {
   if (m.op === 'encrypt') {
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
Same comment as https://github.com/mozilla/browserid/pull/2238#issuecomment-7554042

I'm nervous about setting dev to force node v0.8.5 if our servers/ephemeral instances are using node v0.6.15

seanmonstar:
This doesn't force 0.8.5. Bcrypt still works with 0.6. It just needs 0.8.5
to successfully compile on Windows.

shane-tomlinson:
Closing this, re-opened as #2245 which has an additional commit.

PR #2233, opened 2012-08-06
Selenium MKII

jaredhirsch:
Second pass at integrating selenium tests into browserid.

For reference, here's the first pull request: https://github.com/mozilla/browserid/pull/2184

Changes from the first version:
- Added a super-simple test runner, `run.py`, that gets you going in no time flat.
  - If you're using an old python, it'll tell you.
  - If you don't have virtualenv or pip installed, it'll detect this and tell you what to do.
  - If you haven't installed the test dependencies, it'll install them automatically.
    - You can double-check everything's installed via `run.py --install`.
  - By default, `run.py` runs one test that should always pass: it creates a dev.123done user, logs in, and logs out.
  - You can point that one test at ephemeral instance "foo" via `run.py --target=foo`.
  - You can run all the tests via `run.py --all`.
  - `run.py` _should_ (not verified) work on Windows as well as posix platforms.
  - The automation-tests README has been updated to mention `run.py` in a brutally short tl;dr at the top.
- Weird git voodoo got slightly simpler:
  - BrowserID-Tests was copied straight in, not pulled in as a git-subtree.
    - We've abandoned its history but made our history look simpler.
  - Only one copy of the upstream bidpom repo was pulled in as a git-subtree.
    - Because of python module loading rules, you now have to do `python -m py.test` instead of just `py.test` if you manually run tests, but that's no big deal. Updated README to reflect this.


jaredhirsch:
merging from browserid/import-selenium-tests branch instead of from 6a68/dev

PR #2235, opened 2012-08-06
Import selenium tests

jaredhirsch:
Second pass at integrating selenium tests into browserid.

For reference, here's the first pull request: https://github.com/mozilla/browserid/pull/2184

Changes from the first version:
- Added a super-simple test runner, `run.py`, that gets you going in no time flat.
  - If you're using an old python, it'll tell you.
  - If you don't have virtualenv or pip installed, it'll detect this and tell you what to do.
  - If you haven't installed the test dependencies, it'll install them automatically.
    - If you don't trust it, you can manually trigger install of dependencies via `run.py --install`.
  - By default, `run.py` runs one test that should always pass: it creates a dev.123done user, logs in, and logs out.
  - You can point that one test at ephemeral instance "foo" via `run.py --target=foo`.
  - You can run all the tests via `run.py --all`.
  - `run.py` _should_ (not verified) work on Windows as well as posix platforms.
  - The automation-tests README has been updated to mention `run.py` in a brutally short tl;dr at the top.
- Weird git voodoo got slightly simpler:
  - BrowserID-Tests was copied straight in, not pulled in as a git-subtree.
    - We've abandoned its history but made our history look simpler.
  - Only one copy of the upstream bidpom repo was pulled in as a git-subtree.
    - Because of python module loading rules, you now have to do `python -m py.test` instead of just `py.test` if you manually run tests, but that's no big deal. Updated README to reflect this.


shane-tomlinson:
@6a68 - This is even more awesome than the first round, the run.py script looks like it makes the setup process _much_ easier.  Since I already have all of the deps installed, I have not tested this functionality yet.

Some suggestions:
- Rename credentials.yaml to credentials.yaml.example and have people to copy it to credentials.yaml.  Add credentials.yaml to the ".gitignore" file.  This would prevent credentials.yaml from always showing up in a developer's modified file list.
- Expand run.py so that it is possible to specify which driver to use and have it automatically set the correct command line parameters.
- The section in the README about writing automation tests is useful, but it leaves me scratching my head as to what I'd have to do to have a new test committed to both our mozilla/browserid repo and the upstream bidpom.

SO nice!  I'm excited to have these.

jaredhirsch:
@shane-tomlinson thanks for taking a look! devopsing right now, but will make the suggested changes and update the pull request friday at the latest

Issue #2226, opened 2012-08-06
jwcrypto@0.4.0 depends on concat-stream@0.0.8 which requires node >= 0.8.0


shane-tomlinson:
It is impossible to update ephemeral instances right now:
## Remote install

```
remote: npm http 200 https://registry.npmjs.org/mkdirp/-/mkdirp-0.3.3.tgz
remote: npm http 200 https://registry.npmjs.org/concat-stream/-/concat-stream-0.0.8.tgz
remote: 
remote: npm ERR! Unsupported
remote: npm ERR! Not compatible with your version of node/npm: concat-stream@0.0.8
remote: npm ERR! Required: ["node >= 0.8.0"]
remote: npm ERR! Actual:   {"npm":"1.1.16","node":"0.6.15"}
remote: npm ERR! 
remote: npm ERR! System Linux 3.2.18-1.26.6.amzn1.x86_64
remote: npm ERR! command "node" "/usr/local/bin/npm" "install" "--production"
remote: npm ERR! cwd /tmp/deploy11271-4312-6azzre
remote: npm ERR! node -v v0.6.15
remote: npm ERR! npm -v 1.1.16
remote: npm ERR! code ENOTSUP
remote: npm ERR! message Unsupported
remote: npm ERR! errno {}
remote: npm http GET https://registry.npmjs.org/esprima
remote: npm http 200 https://registry.npmjs.org/esprima
remote: npm http GET https://registry.npmjs.org/esprima/-/esprima-0.9.9.tgz
remote: npm http 200 https://registry.npmjs.org/esprima/-/esprima-0.9.9.tgz
remote: npm ERR! 
remote: npm ERR! Additional logging details can be found in:
remote: npm ERR!     /tmp/deploy11271-4312-6azzre/npm-debug.log
remote: npm not ok
remote: 
To app@23.22.161.152:git
 * [new branch]      HEAD -> master
{ 'ie-cookies': { created: 1344246182306 } }
[10:43][~] $ 
```
## Local install

```
npm http 200 https://registry.npmjs.org/mkdirp/-/mkdirp-0.3.3.tgz
npm http 200 https://registry.npmjs.org/concat-stream/-/concat-stream-0.0.8.tgz
npm ERR! error installing http-browserify@0.1.2
npm ERR! error installing browserify@1.13.5
npm ERR! error installing jwcrypto@0.4.0
npm ERR! Unsupported
npm ERR! Not compatible with your version of node/npm: concat-stream@0.0.8
[...]

PR #2236, opened 2012-08-07
upgrade to jwcrypto 0.4.2 which fixes windows and properly installs bigi...

lloyd:
...nt into envs where it is supported

issue #2226


warner:
r+!

PR #2238, opened 2012-08-07
add nodejs 0.8 to travis-ci

seanmonstar:


seanmonstar:
travis took it on like a champ. http://travis-ci.org/mozilla/browserid/builds/2054278

shane-tomlinson:
@seanmonstar - Are there plans to upgrade everybody/servers to node 0.8.*? 

I'm nervous about setting Travis to use 0.8.0 if our ephemeral instances and servers are using v0.6.15.

lloyd:
This tests both 0.6 and 0.8, it doesn't remove v0.6.

--lloyd

lloyd:
:+1:

ozten:
This is awesome!

Where in the schedule would this fall, roughly?

I don't think we want to upgrade node (if it isn't a hard dependency), until we've got confidence in our deployment systems.

seanmonstar:
We want to upgrade eventually, since 0.8 is faster, has more secure TLS,
and many dependencies will begin to require it. However, right now we don't
need to.

This will just show that when we want to upgrade, it will be OK, because
tests pass.

jrgm:
verified

PR #2243, opened 2012-08-08
a command-line tool for looking at browserid stuff in localStorage sqlite databases on Firefox, Chrome and Safari

jrgm:
I wrote this a while ago, and updated to work with chrome and safari when I realized that was trivial.
- useful for a quick and dirty command-line way to check things cert expiry times, kpi interaction_data pending, confirmed/denied state, and other stuff
- this requires node-sqlite3 which is a binary install so I've made this tool just whine if that has not been manually installed instead of adding yet another dependency in package.json. On linux and osx just do `npm install sqlite3` with node 0.6.17 (You may incur some time to bring in node-gyp first-time). On win32, I'm not sure about how.
- I've used this on osx and ubuntu for firefox, chrome, safari.

### DIFF
--- a/scripts/inspect_localstorage.js
+++ b/scripts/inspect_localstorage.js
@@ -1,0 +1,249 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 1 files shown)


jrgm:
verified train-2012.08.17

PR #2249, opened 2012-08-08
exclude .svn in addition to .git when building RPM

floatingatoll:
We noticed that while inspecting an RPM that there were /opt/browserid/locale/*/.svn/ directories. Exclude those from the RPM.


jrgm:
verified train-2012.08.17 in code

Issue #1532, opened 2012-05-02
`navigator.id.get(f)` will call f twice in some circumstances


washort:
I ran across this while attempting to implement persistent login. If you log in via clicking the button, then reload the page, 'callback invoked' gets logged twice.

``` javascript
<html>
<body>
click it: <button>clicky</button>
</body>
<script src="https://browserid.org/include.js"></script>
<script>
function doBrowserId(opts) {
navigator.id.get(function(assertion) {
console.log("callback invoked");
}, opts);
}
document.querySelector('button').onclick = doBrowserId;
doBrowserId({"silent": true});
</script>
</html>
```

chrismendis:
+1

shane-tomlinson:
@washort, @chrismendis - the "silent: true" option and event navigator.id.get have been deprecated in favor of the watch api.  More information can be found at: https://developer.mozilla.org/en/DOM/navigator.id

@lloyd, have you ever seen this happen with the new API?

chrismendis:
I upgraded to the new Watch API, and it's working as expected. Thanks, Mozilla devs!

I noticed that if I set the `loggedInEmail` option in my `navigator.id.watch()` call, then the `login` callback is never fired. I'm definitely setting the `loggedInEmail` option to the email that was used to log in during the previous session. (I just use a cookie to keep track of the email.)

EDIT: I just realize that the observed behaviour is the expected behaviour, per the Observer API documentation. Please ignore the previous paragraph.

shane-tomlinson:
@lloyd - have you seen this behavior?

shane-tomlinson:
@lloyd - I can verify this behavior on prod through dev. When calling navigator.id.get with the silent: true option, the callback is called twice.  The callback is first called with null, the second time with an assertion.

STR (very round about)

0) Make sure your version of Persona has https://github.com/mozilla/browserid/pull/2216 applied.
1) Load up my modified MFB at https://github.com/shane-tomlinson/myfavoritebeer.org/tree/configurable_verifier
2) start MFB using 

```
BROWSERID_URL=http://10.246.29.67:10002 VERIFIER_URL=http://10.246.29.67:10000 node server/main.js 
```

but replacing the IP addresses with valid values
3) In a browser, load up your instance of MFB
4) Sign in
5) Click Sign out
5) In the MFB repo, edit line 98 of static/js/main.js to:

```
navigator.id.get(gotVerifiedEmail, { silent: true });
```

[...]

PR #2255, opened 2012-08-09
Deprecate silent option + ensure the callback is only called once for navigator.id.get with the silent: true option.

shane-tomlinson:
- Show a console.warn message that silent has been deprecated.
- Bring over some helper functions from PR #2228 to check for deprecated options, print console messages.

issue #1532


lloyd:
r+ - tests pass - IE8 works - code looks right.

PR #2256, opened 2012-08-09
Make the dialog window minimizable in Firefox.

shane-tomlinson:
issue #1594


seanmonstar:
The [docs](https://developer.mozilla.org/en-US/docs/DOM/window.open#Window_functionality_features) say this will add the minimize button to the window, but only if `dialog=1`. If dialog is not 1, then it will have the normal buttons by default.

Is there reason it should be `dialog=1` in the first place? Perhaps to remove the maximize button?

Issue #2258, opened 2012-08-09
dialog becomes unresponsive in set password screen for new account if api call is throttled


shane-tomlinson:
STR

1) Open dialog, if logged in, sign out.
2) Enter new secondary address
3) Enter password
4) Close dialog
5) Re-open dialog
6) Enter same secondary address
7) Enter password
8) throttle tooltip appears
9) Try again, no tooltip, no response.

This bug affects 2012.07.20, 2012.07.30, and current dev.

lloyd:
I don't think this issue is a 5 star and warrants merge to 07.30.

The user sees the tooltip in the dialog.  they must wait, and then click the button again.  I think we should merge the fix, but I don't think that it's critical we merge it into 07.30.

testing fix, if it works merging into dev, and closing down the 07.30 targeted PR.

lloyd:
fixed in dev!

shane-tomlinson:
@lloyd - I am re-opening this after giving it more thought.  The user does _not_ have to try to enter their email address 3 times within one minute, only twice.  The first time the email is throttled, the user can no longer proceed.

If this still does not warrant a fix, that is fine, but the clarification needs to be made.

lloyd:
Ok, let's confirm understanding.  Here's what the user has to do:
1. click sign in on a website
2. type their email address into the dialog
3. pick a password.
4. click next
5. close the dialog
6. click sign in on a website
7. type the same email address as in 2
8. pick a password
9. click next
10. wait for the tooltip (that says "we just sent an email to this address, try again later")
11. once the tooltip disappears, click next again

Expected: tooltip is redisplayed if less than 60s have elapsed between #4 and #11
Actual: no tooltip is shown, button is non-operational until 60s have elapsed between #4 and #11

@stomlinson - have I got this right?

@jrgm - want this merged into 07.30?  is this hotfix worthy?

shane-tomlinson:
@lloyd - you got it right until the "until 60 seconds has elapsed between #4 and #11".

Once the tooltip is shown, the button is dead forever, the user cannot continue unless they re-open the dialog.

seanmonstar:
We've been hoping that 07.30 is the beta1 train.

lloyd:
Given schedule, I wonder if we should aim for current dev to become beta w focused DDT starting Monday...

--lloyd

jrgm:
verified train-2012.08.17

PR #2259, opened 2012-08-09
Fix the unresponsive set password screen if the staging wsapi returned a throttled response.

shane-tomlinson:
issue #2258

To properly test:
- Make sure the user can still hit "enter" after a throttled response for staging a new account, forgot password, and add the first secondary email to a primary only account.

Conflicts:
    resources/static/test/cases/dialog/js/misc/state.js


seanmonstar:
not 5 star, so not landing in stage

PR #2261, opened 2012-08-09
use node 0.6.17 in dev and ephemeral deployments

lloyd:


seanmonstar:
I love how post_deploy.js has a .js extension, but is a bash script.

PR #2264, opened 2012-08-10
remove symlinks, added postinstall script

### LINKED ISSUE #2263: symlinks don't resolve in Windows
there's a couple symlinks in the repository, that when checked out on Windows, don't resolve. So, some resources don't load.
- bidbundle.js
- dialog/views/site

adding to #1769, even though they're not technically install errors; they still prevent it from running properly.

seanmonstar:
- unix symlinks are incompatible with Windows
- postinstall now generates keys, and sets up files

fixes #2263


.gitignore
```diff
@@ -9,6 +9,8 @@
 /resources/static/build
 /resources/static/production
 /resources/static/i18n
+/resources/static/common/js/lib/bidbundle.js
+/resoueces/static/dialog/views/site
```
shane-tomlinson: @seanmonstar - a slight typo here, this should be resources instead of resoueces

shane-tomlinson:
I like this approach and overall this works fine for local dev, but it breaks when trying to run unit tests on ephemeral instances - the symlinks are not properly set up.

I tried modifying scripts/awsbox/post_deploy.sh to call scripts/postinstall.js instead of scripts/generate_ephemeral_keys.js, but all code is first staged into a temporary directory and then moved to its new directory.  Node seems to create absolute path symlinks when linking directories.  Since code is staged into a temp directory and then moved, the links are broken.

Can we somehow call the linker on awsbox instances once the code is moved to its end location?

Do we care?  Does anybody run the unit tests on ephemeral instances?

seanmonstar:
Oh good point. It's likely because of the path.join() making it absolute.
I'll look again Monday.

seanmonstar:
OK, so I can make the links relative, which should fix moving it around on awsboxes.

Node on Windows actually errored with relative symlinks, as a bug, but it just got fixed: https://github.com/joyent/node/commit/9918e5d4f6fbc8fe63d0395170a2daef2484d17a

So, I can make the script set up relative paths, and it will work fine for *nix, and for Windows if they have nodejs 0.8.8, most likely.

shane-tomlinson:
@seanmonstar - Before doing that, let's find out from @jrgm and @jbonacci whether they actually run unit tests on AWS instances.  If not, it makes more sense to get Windows working for you using the absolute path symlinks until we migrate to node 0.8.8.

@jbonacci, @jrgm - do you guys run front end unit tests on AWS instances?

seanmonstar:
How does it only affect units tests? If it moves the directory afterwards to then run the server, it will be missing the bidbundle.js file.

jrgm:
> do you guys run front end unit tests on AWS instances?

I do not run the unit tests from the command line on awsboxes. (I do though run the front-end tests in browser pulling from awsbox.)

But I was wondering about the same thing as @seanmonstar.

shane-tomlinson:
@seanmonstar, @jrgm - regarding why the dialog still works - in https://github.com/mozilla/browserid/pull/2264/files#L4R24, you are copying the bidbundle.js file instead of creating a symlink to it.

shane-tomlinson:
@seanmonstar, @jrgm - My last comment needs updating.  In https://github.com/mozilla/browserid/pull/2264#issuecomment-7638849 I point out that I modified "post_deploy" to call scripts/postinstall.js instead of just generate_ephemeral_keys.js. That calls the code that copies bidbundle.js

seanmonstar:
@shane-tomlinson i've fixed the other comments (like typo in ,gitignore), and made the symlink use a relative path. This should work just fine on *nix in the 0.6.x branch of nodejs.

It correctly makes the symlink and everything on Windows with the newest nodejs 0.8.8.

scripts/postinstall.js
```diff
@@ -0,0 +1,42 @@
...
+  src = path.join(__dirname, src);
+  dest = path.join(__dirname, dest);
+  var destParent = path.dirname(dest);
+  var cwd = process.cwd();
+  process.chdir(destParent);
+
+  if (existsSync(dest)) {
+    fs.unlinkSync(dest);
+  }
+  var relSrc = path.relative(destParent, src);
```
shane-tomlinson, line 17: Awesome. I didn't know this existed.

shane-tomlinson:
@seanmonstar - These changes work on OSX in a fresh directory after performing npm install. Front end unit tests pass
and account creation works on ephemeral instances.

Should postinstall.js be called on postupdate as well? If bidbundle is updated using the npm update command, the new bidbundle.js file will not be copied over to the client. See https://npmjs.org/doc/scripts.html

Other than that, this gets an r+.

Issue #1990, opened 2012-07-11
deprecate getVerifiedEmail


benadida:
We should begin the deprecation process for it with the beta launch.

warner:
ben's comment suggests this is beta1-req, not sure if we still feel that way, tagging to solicit discussion

callahad:
I like beginning the deprecation process for `getVerifiedEmail()` with beta1 since it affords us the opportunity to walk through deprecation, monitor for usage, add console alerts, etc. in anticipation of real, future deprecations.

The only place I'm aware our thinking has changed is with regard to `get()`, which we'll be keeping around for a little while longer.

jrgm:
deprecation warning added in PR-2267. Verified in train-2012.08.17

PR #2267, opened 2012-08-10
Adding a deprecation warning for navigator.id.getVerifiedEmail

shane-tomlinson:
issue #1990


zaach:
works as intended, r+

Issue #2231, opened 2012-08-06
In dialog, changing the email chosen then 'Forgot password' causes email to be sent to initial email not the corrected one.


gion-andri:
I found a bug on todays version of translate.personatest.org. 
1. Sign in (e.g. at http://translate.123done.org) with a wrong email adress (for example mistyped). 
2. Click cancel
3. Enter your correct email (from an existing account)
4. Click 'Forgot your password?'
5. Enter a new password and send the form

I get an email on the wrong (mistyped) adress with a link to activate a new account. The password for the existing account isn't changed.

adamkalis:
I can also reproduce this bug. The mail was sent and received to the first wrong email address (the one at the first step).

jrgm:
Yep, indeed.

jrgm:
Update summary

jrgm:
verified train-2012.07.30 for the original STR.

PR #2273, opened 2012-08-13
Fix users who enter wrong email address, cancel, enter correct email address then forgot password.

shane-tomlinson:
- Add the notion of momentos to the state machine so that back properly works.

issue #2231

### DIFF
--- a/resources/static/dialog/js/misc/state.js
+++ b/resources/static/dialog/js/misc/state.js
@@ -36,7 +36,7 @@ BrowserID.State = (function() {
             // Save a snapshot of the current state off to the momentos. If
             // a state is ever cancelled, this momento will be used as the
             // new state.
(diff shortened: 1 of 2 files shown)


resources/static/dialog/js/misc/state.js
```diff
@@ -17,15 +17,37 @@ BrowserID.State = (function() {
 
   function startStateMachine() {
     /*jshint validthis: true*/
-    var self = this,
+    // Self has been changed from a reference to this to a reference to the
```
seanmonstar, line 20: I personally find this naming confusing. `self` is usually always `this`. If it's not going to be, than a better name is probably warranted: `state`? `currentState`?

seanmonstar:
Besides the naming, this has an r+ from me.

PR #2276, opened 2012-08-13
Add clear localstorage step to change password test to ensure specific w...

zacc:
This step is to clear the browserid artefacts from the localstorage and ensure that for the second login step we will not hit the 'is this your computer' workflow.

This will stabilise slow IE tests that were falling into the 'is this your computer' workflow and failing.

### DIFF
--- a/automation-tests/123done/tests/test_change_password.py
+++ b/automation-tests/123done/tests/test_change_password.py
@@ -44,6 +44,7 @@ class TestChangePassword:
         account_manager.click_password_done()
 
         account_manager.click_sign_out()
+        mozwebqa.selenium.execute_script('localStorage.clear()')
 
         home_pg.go_to_home_page()
 


klrmn:
so, i think this PR is still useful and should be merged.

i did, however, get a javascript error from a line like this running in IE8 on sauce yesterday. i didn't keep it around to see if it was transient or repeatable...but there exists the possibility that IE8 just doesn't support HTML5 well enough to do this.

shane-tomlinson:
@klrmn, @zacc - IE8 has serious localStorage issues, but localStorage.clear() should not be one of them. I just tried it repeatedly in a test environment to make sure.

@klrmn - if you see problems with localStorage.clear, an easy work around is:

```
for(var key in localStorage) localStorage.removeItem(key);
```

jaredhirsch:
looks good to me. merging in.

Issue #2265, opened 2012-08-10
IE8: After verifying second secondary email address, page is not redirected to RP.


shane-tomlinson:
This is intermittent and I am trying to figure out exact STRs

Initial STRs when I first saw the problem but have not since been able to.

1) In IE8, open the dialog using an already existing secondary address.
2) Add a new secondary address to the account
3) Open the verification page in a new IE8 tab
4) Page says "Your email has been verified" but never redirects.

Reported by @zacc

zacc:
Saucelabs job video showing the failure.
http://saucelabs.com/jobs/3dd96e0838b24262acc962ab118190a1

zacc:
Actually that video is IE9
the equivalent test running on XP/IE8 at Saucelabs passes.

zacc:
The test that found this is test_add_another_email_address.py

When this is resolved revert this commit to restore coverage to IE9
https://github.com/mozilla/BrowserID-Tests/commit/0e3b7624b17e39a3669cf288d56b7f626b2d3a81

jaredhirsch:
@zacc: Rather than pull that commit into browserid, I'm kind of inclined to just let the test keep on failing, so that @shane-tomlinson can easily verify he's fixed it (by running the affected test in his dev environment). Thoughts?

zacc:
The test will show up with a small x while it is failing and a capital X in the test results when it is passing again. Or if you're looking in the HTML report it will say XPass. (meaning eXpected fail but got Pass).

Actually can you use a slightly different (preferred) code solution when you move this into the main repo? Just looks neater.

``` python
@pytest.mark.xfail("config.getvalue('browser_name') == 'internet explorer' and \
                   int(config.getvalue('browser_version')) == 9", 
                   reason='IE Local Storage https://github.com/mozilla/browserid/issues/2265')
def test_can_add_another_email(self, mozwebqa):
```

jaredhirsch:
ok, since we're close to cutting the beta train and this might not get fixed, I'll pull over the fix.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

[...]

PR #2278, opened 2012-08-13
Mark add-email test as expected to fail on IE9. Refs #2265, fixes #2269.

### LINKED ISSUE #2269: Import changes from upstream BrowserID-Tests
Two commits landed in mozilla/BrowserID-Tests this week. Since we don't treat that repo as a first-class upstream citizen, copy the diffs over into our automation-tests directory.

This is going to keep happening until we've completed migration to a Jenkins that Identity devs control, and which points at the tests inside browserid, not the tests inside BrowserID-Tests.

jaredhirsch:
This change imported from defunct "upstream" repo mozilla/BrowserID-Tests 0e3b76

Once this change is merged over, we'll be up to date with BrowserID-Tests, so it can be obsoleted (and I'll add a note to its README to this effect); next, the tests will be migrated to ci.mozilla.org, and BrowserID-Tests will be deleted.

The essence of the change is that, for IE9, while #2265 is open, we expect a certain test (add secondary email) to fail. We express this by decorating the test with a pytest.xfail decorator.

This diff doesn't map directly over from the BrowserID-Tests repo because the affected test was moved into bidpom without the corresponding 123done test being deleted. I previously deleted that duplicate test (123done/test_add_another_email), so I added the annotation to the surviving version of the test (browserid/check_add_email).

### DIFF
--- a/automation-tests/browserid/tests/check_add_email.py
+++ b/automation-tests/browserid/tests/check_add_email.py
@@ -17,6 +17,9 @@ import restmail
 class TestSignIn(BaseTest):
 
     @pytest.mark.travis
+    @pytest.mark.xfail("config.getvalue('browser_name') == 'internet explorer' \
+                       and int(config.getvalue('browser_version')) == 9", 
+                       reason='IE Local Storage https://github.com/mozilla/browserid/issues/2265')
     def test_add_email(self, mozwebqa):
         user = self.create_verified_user(mozwebqa.selenium, mozwebqa.timeout)
         user.additional_emails.append('%s_1@restmail.net' % user.id)


zacc:
Hey 6a68, The tests in the browserid(bidpom)/tests/ are not the ones you want to be running. They are not the tests the Web QA team have been running via Jenkins. They are the tests designed to test bidpom.

The tests in 123done/tests/ are the ones I have written specifically to test browserid integration with 123done.org. The test framework I have set up inside 123done/tests/ is more suitable for expanding upon, built with that in mind.

Although the two tests are both very similar. I don't think we should have two folders with tests in different places (3 when you take into account mfb tests too).

Chat later.

jaredhirsch:
we're just going to delete tests that are going to fail. so I'll do that.

zacc:
Why would you do that?
Are we going to add it again later?

PR #2284, opened 2012-08-14
update version.js to run asynchronously and eliminate race conditions in code

lloyd:
The way that `version.js`, a tiny library to determine the currently running code version, worked made it impossible to reliably determine the software version at startup.  Because it would return a bogus (randomly generated) version until the `git` program could return the actual version, client code would have to wait.  

This improvement makes it possible to write a command line program (or server) that gets the version without any setTimeouts as soon as it is available.

This is motivated by the automation testing work that's ongoing.

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -32,11 +32,10 @@ function setSHA(s) {
 if (config.get('env') === 'production') {
   try {
     var contents = fs.readFileSync(path.join(__dirname, '..', 'resources', 'static', 'ver.txt'));
-    sha = contents.toString().split(' ')[0];
-    if (sha.length !== 7) throw "bad sha in ver.txt";
-    setSHA(sha);
+    var reportedSHA = contents.toString().split(' ')[0];
+    if (reportedSHA.length !== 7) throw "bad sha in ver.txt";
+    setSHA(reportedSHA);
   } catch(e) {
-    sha = undefined;
     logger.debug('cannot read code version from ver.txt: ' + e);
   }
 }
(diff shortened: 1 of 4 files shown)


lloyd:
travis failure is red herring: http://travis-ci.org/#!/mozilla/browserid/jobs/2122302

lib/version.js
```diff
@@ -24,6 +34,7 @@ if (config.get('env') === 'production') {
     var contents = fs.readFileSync(path.join(__dirname, '..', 'resources', 'static', 'ver.txt'));
     sha = contents.toString().split(' ')[0];
```
shane-tomlinson: @lloyd - this re-uses the global sha variable.  If the sha's length is not 7, the error will be correctly thrown, but any functions that call module.exports after the incorrect sha is set will have the incorrect value returned.

lib/wsapi/session_context.js
```diff
@@ -50,7 +50,7 @@ exports.process = function(req, res) {
       data_sample_rate: config.get('kpi_backend_sample_rate')
     };
     if (config.get('enable_code_version')) {
```
shane-tomlinson, line 69: Are we still actually using this?  The code_version check on the front end was disabled in favor of using ETAGs and SHAs in the URL.

seanmonstar:
So, this needs some cleanup?

shane-tomlinson:
@seanmonstar - it does, I pinged @lloyd about it earlier today.

shane-tomlinson:
@lloyd - ping - this thing is suffering from bit rot.

lloyd:
r+ - we can merge this monday.

PR #2285, opened 2012-08-14
Issue 2254 runpy enhancements

jaredhirsch:
In response to some of @shane-tomlinson and @lloyd suggestions:
- .gitignore config files and bid_selenium and results directories
- move example config files to *.yaml.example
- add --everywhere command to run.py to run all our browsers (alas, except android) against sauce labs
  - lloyd has already added our sauce sekrets to the persona dev bundle and this script looks for a sauce.yaml file, if not found, it'll pull the creds out of the env vars and write to the file for you. aw yeah.

After thinking it over and some discussion with @lloyd, I really don't think I want to add additional granularity to run.py--if you want more than it offers, you should look at the README cookbook options and drop down a level to py.test. Otherwise, I'm going to have to abstract over a huge, huge API that py.test and mozwebqa (a pytest plugin installed as part of pip install) expose. Ugh.


automation-tests/credentials.yaml.example
```diff
@@ -0,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# insert your sauce labs credentials here and save as sauce.yaml
+username: <value>
+password: <value>
+api-key: <value>
```
lloyd: was this file overwritten by accident?

jaredhirsch: yep, fixed :)

PR #2289, opened 2012-08-14
Persona org server smoke tests

klrmn:
https://moztrap.allizom.org/manage/cases/?filter-productversion=2&filter-suite=31


jaredhirsch:
Hey @klrmn - thanks so much for contributing these tests. There are some things we'll probably tweak, but I want to land all this code without further delay. I did a little git twiddling and it looks like these can be merged straight in; see pull request #2315.

PR #2315, opened 2012-08-15
Personaorg tests

jaredhirsch:
Resubmission of @klrmn's pull request 2289


jaredhirsch:
After thinking it over (and discussion with @ozten), I'm going to pull these in rather than wait. See related pull request #2289 for related discussion.

PR #2316, opened 2012-08-15
Removing DEPLOYMENT.md per lloyd

ozten:
Yep...


lloyd:
r+ - this file is so old...  it causes more confusion than provides value.  good riddance!

PR #2318, opened 2012-08-15
Add email type (primary or secondary) to DOM

graingert:
Add email type (primary or secondary) to DOM

This allows my user style to embolden my primary email

### DIFF
--- a/resources/static/pages/js/manage_account.js
+++ b/resources/static/pages/js/manage_account.js
@@ -85,7 +85,8 @@ BrowserID.manageAccount = (function() {
 
     _(emails).each(function(item) {
       var e = item.address,
-          identity = _.template(template, { email: e });
+          t = item.info.type,
+          identity = _.template(template, { email: e, type: t });
 
       var idEl = $(identity).appendTo(list);
       idEl.find(".delete").click(cancelEvent(removeEmail.bind(null, e)));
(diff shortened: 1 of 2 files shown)


graingert:
"Executing your before_install (sudo apt-get install libgmp3-dev) took longer than 8 minutes and was terminated."

shane-tomlinson:
@graingert - Thanks for the PR!  Do you want these classes to only be assigned in the manage page or do you want these to be applied in the dialog?  The PR as written applies classes to the manage page.

shane-tomlinson:
@graingert - Can you define the concrete use case of this PR?  

Once BigTent comes online, all Yahoo, GMail and Hotmail addresses are going to be marked "primary" as well. Is that going to affect what you are trying to do?

graingert:
This is only for my personal userstyle that changes the style of .secondary and .primary

shane-tomlinson:
@graingert - If your styling should be applied to the dialog, can you add the appropriate logic to browserid/resources/static/dialog/views/pick_email.ejs as well?

shane-tomlinson:
@graingert - yo, ping. This is suffering from bit rot. If you want this merged in, can you rebase and add the appropriate logic to resources/static/dialog/views/pick_email.ejs as well?

lloyd:
closing after 2 months of inactivity.  please re-open with a new commit if still desired.

In general, us doing this would imply that our DOM classes are an external contract that people can rely on.  That's a bit of commitment that scares me.  

Further, Given the way that the dialog works now, I'm not sure this is possible (we don't _know_ the email type in the front end until it's actually selected)

PR #2321, opened 2012-08-15
avoid running tests that require --email

klrmn:
avoiding them in two ways,
- -m travis
- skip if mozwebqa.email == None

also, longer timeouts


klrmn:
i don't think this constitutes a failure on the PR's part:

"I'm sorry but the VM stalled during your build and was not recoverable.
We are continuosly working on test run stability, please email support@travis-ci.org if this issue persists."

Issue #1500, opened 2012-04-25
Previously selected email address not remembered if user's session expires.


shane-tomlinson:
This may be a dup of another, but I cannot find it.

If a user's session expires and the user must re-authenticate, BrowserID forgets which address a user last used.

STR
1) Open a new browser profile that can sit dormant for 2 weeks.
2) Sign in to myfavoritebeer.org using an account with multiple addresses that you do not need to use again.  Sign in using email A.
3) Sign in to myfavoritebooze.org using email B.
4) 2 weeks + 1 day later, sign in to myfavoritebeer.org again using email A.
5) Open the dialog for myfavoritebooze.org

Expected:
Email B is highlighted.

Actual:
No email is highlighted.

lloyd:
+1

seanmonstar:
haha, "STR: open a browser and come back in 2 weeks"

anyways, @shane-tomlinson did that commit clear this up?

shane-tomlinson:
@seanmonstar - The commit is the start, but I want to make sure I have all the logic right and get some feedback before creating a PR.

jaredhirsch:
Bumping to Beta 3; not going to get fixed prior to Beta 2 announcement.

jaredhirsch:
Taking. Some hints from @shane-tomlinson [in ircloggr](http://irclog.gr/#show/irc.mozilla.org/identity/296343)

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #2335, opened 2012-08-16
Retain site<->email associations until a user either explicitly logs out or a new user logs in.

shane-tomlinson:
- On user.logoutUser, remove all emails, loggedIn emails, and site<->email associations.
- When a user's session expires, invalidate all certs (remove keys) but do not remove the email address.
- Depend on checkAuthenticationAndSync to remove any old email addresses and site associations for users who log out.

issue #1500


shane-tomlinson:
Closing this until some refinements are made.

PR #2336, opened 2012-08-16
tell selenium not to use proxy for firefox and chrome

klrmn:
also, rename misnamed env


lloyd:
Why wouldn't we want to disable proxying globally?

klrmn:
@lloyd , you're welcome to disable proxying globally, but this'd be the script i can't work on because my sauce is empty.

klrmn:
@lloyd  implemented this in a different pull.

Issue #2277, opened 2012-08-13
Make main site /signup into a 301 Moved Permanently to /signin


jrgm:
Hey, love the new UI on dev for omni-signin.

But (there's always a but :-) ) the former URL /signup is now 404. Seems to me that should be a '301 Moved Permanently' to /signin.

jrgm:
verified in train-2012.08-17

PR #2337, opened 2012-08-16
Issue #2277 signup redirect

zaach:
Add a permanent redirect from /signup to /signin. #2277


tests/page-requests-test.js
```diff
@@ -64,14 +65,11 @@ suite.addBatch({
   'GET /add_email_address':      respondsWith(200),
   'GET /confirm':                respondsWith(200),
   'GET /reset_password':         respondsWith(200),
-  'GET /confirm':                respondsWith(200),
```
shane-tomlinson, line 67: Nice cleanup to remove the dups

PR #2338, opened 2012-08-16
many updates to run.py

klrmn:
- tell selenium not to use proxy
- reduce restriction to python 2.6
- write out a credentials.yaml file if it doesn't exist
- do not use sauce
- save results to different folders so you end up with all of them rather than just the most recently finished
- make -e run both FF and Chrome
- make 'which' work on windows

Tested on OSX and WinXP


klrmn:
i'm about to test run.py on windows, but i have to create a whole environment there first...

jaredhirsch:
Rebased this and merged in as pull request 2366. Closing this one.

Issue #1039, opened 2012-02-02
ensure that when a primary comes up, existing certs by secondary are invalidated in the dialog


benadida:
otherwise it's a fail for the user.

zaach:
@benadida this will matter for big-tent, yes?

benadida:
@zaach yes, given the way we're now doing bigtent with non-chained public keys, we should indeed fix this.

shane-tomlinson:
@benadida - where would the failure be in this case, at the verifier?

Do we have to worry about the reverse situation as well, if a primary functionality for a domain goes down and Persona thinks a domain is a secondary (from the conversation happening in the mailing list)?

callahad:
The reverse seems like a reasonable concern. Maybe the broader principle is that the UA shouldn't send an assertion unless the backing IdP is the same one that would be resolved as per normal?

We'd lose the cool story about how eyedee.me could do down, yet users could still log in to sites that had eyedee.me's pubkey cached, but that kind of seems like a mess anyway...

lloyd:
this is not a beta1 req. moving.  (bigtent will launch post-beta1 and we have other idp related issues)

lloyd:
This has been taken to the mailing list to determine a more holistic solution to handling IdP's coming online and going offline - https://groups.google.com/forum/?fromgroups=#!topic/mozilla.dev.identity/ArXmGF3OnW4

This issue discusses a single component of the problem and is not useful.  Closing down and will open specific issues to get this done and make the world safe for bigtent.

PR #2341, opened 2012-08-17
Convert secondary addresses to primaries when a primary is stood up.

shane-tomlinson:
- Clear any existing secondary certs for the domain.
- Convert "secondary" to "primary" for addresses in the domain.

issue #1039

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -375,40 +375,6 @@ BrowserID.User = (function() {
     },
 
     /**
-     * Create a primary user.
-     * @method createPrimaryUser
-     * @param {object} info
-     * @param {function} onComplete - function to call on complettion.  Called
-     * with two parameters - status and info.
-     * Status can be:
-     *  primary.already_added
-     *  primary.verified
-     *  primary.verify
-     *  primary.could_not_add
-     *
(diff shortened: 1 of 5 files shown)


resources/static/common/js/user.js
```diff
@@ -375,40 +375,6 @@ BrowserID.User = (function() {
     },
 
     /**
-     * Create a primary user.
```
shane-tomlinson, line 378: This was never used, so I took it and its tests out.

resources/static/test/cases/common/js/user.js
```diff
@@ -180,40 +184,6 @@
   });
 
 
-  asyncTest("createPrimaryUser with primary, user verified with primary - expect 'primary.verified'", function() {
```
shane-tomlinson, line 183: createPrimaryUser was never used, so I removed the declaration from user.js as well as its tests.

shane-tomlinson:
Note, this PR is not yet ready.

If the user is already authenticated to Persona and is displayed the email picker, the email type is never re-checked. This means existing addresses will not be correctly converted from primary to secondary.

shane-tomlinson:
Closing for a fuller solution.

Issue #2287, opened 2012-08-14
[ie9] Failure/Error when signing-up or cancelling-out of primary user dialog with eyedee.me 


jrgm:
STR

1) Go to https://login.dev.anosrep.org/signin
2) Enter a primary email address @eyedee.me
3) Dialog button transitions from Next to Verify button after looking up address_info
4) Click Verify to bring up the /sign_in dialog
5) In the dialog, click Cancel 

On IE9 (but not on desktop Chrome/Opera/Safari/FF) this error is shown:

![img](https://img.skitch.com/20120815-kexmsyj3c6tb5y12misyp99ktp.png)

(Upon closing that dialog, the main site page also shows an error screen that the dialog had failed.)

jrgm:
Actually, the signup path fails the same way. Related to the other open issue about IE8 failiing? 
WIth IE9 you can complete the signup of a primary with eyedee.me via the dialog when launced from an RP.

shane-tomlinson:
@jrgm - When I try using eyedee.me from dev.123done.org, I get redirected to "https://eyedee.me/browserid/" and am shown a 404 error in IE8.  Not opportune.

callahad:
When we determine the root cause of this, please ping me and me know. If this is completely on the IdP side, I should probably document it as a caveat for IdP developers.

shane-tomlinson:
@jrgm, @jbonacci, @lloyd - This bug affects users whether they cancel from the primary or not.  Even if a user successfully signs in with their IdP, they are redirected from the IdP to the dialog instead of the special "authenticate_with_primary" page.

A fix (PR #2342) has already been issued.

PR #2342, opened 2012-08-17
Fix IE failing when authenticating with IdP from main site's /signin

shane-tomlinson:
- window.name is used to keep track of where to redirect the user once authentication with the IdP is complete. IE was losing window.name when redirected to the IdP because it was set from within the dialog.
- Set the window.name to auth_with_primary when window.open is called.
- Update WinChan to use the version that allows the setting of window.name

issue #2287


seanmonstar:
Yup, it works!

Issue #2346, opened 2012-08-17
manage_page.js should be converted to a PageModule.


shane-tomlinson:

shane-tomlinson:
Closing with merge of #2347

PR #2347, opened 2012-08-17
Convert the manage page to a PageModule.

shane-tomlinson:
This is a no-user facing change PR. I converted the managePage to be a PageModule so that it is consistent with other modules.

issue #2346

### DIFF
--- a/resources/static/pages/js/manage_account.js
+++ b/resources/static/pages/js/manage_account.js
@@ -17,79 +17,83 @@ BrowserID.manageAccount = (function() {
       pageHelpers = bid.PageHelpers,
       cancelEvent = pageHelpers.cancelEvent,
       confirmAction = confirm,
+      complete = helpers.complete,
       doc = document,
       tooltip = bid.Tooltip,
       authLevel;
 
   function syncAndDisplayEmails(oncomplete) {
+    var self=this;
     user.syncEmails(function() {
-      displayStoredEmails(oncomplete);
+      displayStoredEmails.call(self, oncomplete);
     }, pageHelpers.getFailure(errors.syncEmails, oncomplete));
   }
 
   function displayStoredEmails(oncomplete) {
     var emails = user.getSortedEmailKeypairs();
     if (_.isEmpty(emails)) {
-      $("#content").hide();
+      dom.hide("#content");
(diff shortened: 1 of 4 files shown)


resources/static/pages/js/manage_account.js
```diff
@@ -202,48 +203,46 @@ BrowserID.manageAccount = (function() {
...
-    var template = new EJS({ text: $("#templateManage").html() });
-    var manage = template.render({});
-    $("#hAlign").after(manage);
+      var self=this,
+          oncomplete = options.ready,
+          template = new EJS({ text: $("#templateManage").html() }),
+          manage = template.render({});
 
-    dom.bindEvent("#cancelAccount", "click", cancelEvent(cancelAccount));
+      $("#hAlign").after(manage);
```
seanmonstar: jQuery $ slipped in, instead of bid.DOM?

resources/static/pages/js/manage_account.js
```diff
@@ -202,48 +203,46 @@ BrowserID.manageAccount = (function() {
...
-    dom.bindEvent("button.edit", "click", startEdit);
-    dom.bindEvent("button.done", "click", cancelEdit);
-    dom.bindEvent("#edit_password_form", "submit", cancelEvent(changePassword));
+      self.click("#cancelAccount", cancelAccount);
 
-    user.checkAuthentication(function(auth_level) {
-      authLevel = auth_level;
+      self.bind("button.edit", "click", startEdit);
+      self.bind("button.done", "click", cancelEdit);
+      self.bind("#edit_password_form", "submit", cancelEvent(changePassword));
```
seanmonstar: should this be `Module.submit` instead?

seanmonstar:
looks good besides my comments

resources/static/pages/js/manage_account.js
shane-tomlinson: I changed off of the $'s to dom.\* calls. Thanks for pointing this out.

resources/static/pages/js/manage_account.js
shane-tomlinson: Great catch. Changed it.

Issue #2345, opened 2012-08-17
development IdP is not keyboard accessible.


shane-tomlinson:
Relate to #869

PR #2348, opened 2012-08-17
Make the example IdP more keyboard accessible.

shane-tomlinson:
- Add autofocus to the appropriate button or input fields.
- Enable the form to be submitted using the keyboard in "sign_in.html"

issue #2345


zaach:
r+, waiting for word from @lloyd before merging

Issue #2349, opened 2012-08-17
"sign in" should come before "add another email address" and "this is not me" in email picker tab ordering


shane-tomlinson:
Specific issue related to #869

PR #2350, opened 2012-08-17
Change the default tab ordering so the submit buttons come before the "this is not me" and "Add another email address" buttons.

shane-tomlinson:
issue #2349


shane-tomlinson:

Issue #2352, opened 2012-08-17
"forgot your password" should come after "sign in" in tab ordering on /signin page


shane-tomlinson:
link to #869

PR #2354, opened 2012-08-17
Make "forgot your password" come after "sign in" in the tab ordering.

shane-tomlinson:
issue #2352


shane-tomlinson:

Issue #2351, opened 2012-08-17
"Sign Up" should be first element tabbed to on the / page.


shane-tomlinson:
Debatable.

link to #869

PR #2355, opened 2012-08-17
Set the "signup" button as the first item in the tab ordering.

shane-tomlinson:
issue #2351


shane-tomlinson:

PR #2358, opened 2012-08-17
remove platform dependant "which" code

graingert:
remove platform dependant "which" code, Now use try/catch import to check

### DIFF
--- a/automation-tests/run.py
+++ b/automation-tests/run.py
@@ -5,6 +5,7 @@ import os
 import platform
 import subprocess
 import sys
+import pkg_resources
 
 
 def main():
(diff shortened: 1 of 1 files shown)


klrmn:
i've both fixed which() in https://github.com/mozilla/browserid/pull/2338, and used the which() function to determine if tests can be run on chrome...since the chromedriver isn't a python package, i'm not sure if pkg_resources will be able to do the same job.

jaredhirsch:
@graingert: first off, thanks so much for contributing this pull request! I think I've seen you in #identity, but we haven't gotten a chance to work together before.

I'm inclined to pull in the which-related changes in @klrmn's linked pull request instead, but please don't let that discourage you from participating in the future :-)

graingert:
@6a86: `which`ever is the best fix! I don't want this to turn into a `which` hunt.

It might be best to keep "which" for `chromedriver` and use `pkg_resources` for `pip` etc

klrmn:
i would certainly be cool with using the right tool for the right job.

graingert:
merged changes from master into new pull: https://github.com/mozilla/browserid/pull/2375

jaredhirsch:
@graingert: OMG it's full on! Double `which` pun all the way across the sky! OMG. OMG. What does this mean?

I'll get to that other pull request in a bit. Jenkins wrestling all week long.

PR #2362, opened 2012-08-20
Split up storage into distinct models for cohesion.

shane-tomlinson:
This is the first of a several part refactor to start to clean up storage.js

Storage.js lost its cohesion long ago.  This is an effort to split the distinct parts into their own files under the "model" subdirectory.  Each model is then mixed in to storage.js, giving BrowserID.Storage the exact same API that it had before. To show that this is indeed the case, 0 unit tests have been modified.

The next parts to this cleanup will be:
- Split off unit tests into their own distinct files.
- Reduce much of the boilerplate code to do JSON.stringify and JSON.parse.
- Default localStorage values should be defined with each model so when setDefaultValues is called, it sets the correct value automatically.

### DIFF
--- a/lib/static_resources.js
+++ b/lib/static_resources.js
@@ -152,6 +152,7 @@ exports.resources = {
     '/common/js/models/users_computer.js',
     '/common/js/models/logged_in.js',
     '/common/js/models/return_to.js',
+    '/common/js/models/main_site.js',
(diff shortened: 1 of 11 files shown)


shane-tomlinson:
Closing until refinements are made.

Issue #2282, opened 2012-08-14
[opera] front end tests - flakiness if opera is not top-most window 


jrgm:
STR

1) start opera and load https://login.dev.anosrep.org/test/
2) do absolutely nothing on the same desktop
3) tests pass
3) start another application window; load https://login.dev.anosrep.org/test/ in opera and move the other application's window above opera
4) tests fail

When the tests fail, there are two errors:

```
Tests completed in 47107 milliseconds.
1327 tests of 1329 passed, 2 failed.

dialog/js/modules/pick_email: email associated with site - check correct email (1, 2, 3)Rerun
checked element is focused

dialog/js/modules/pick_email: single email, no email associated with site - check first radio button (1, 2, 3)Rerun
the lone email address is still focused for keyboard navigation
```

I don't see the same with other browsers. Mostly I'm just documenting this desktop opera quirk here.

This is most likely part of f73f0ad7. Phantomjs on osx front-end tests report two instances of 'currently unable to focus elements'.

jrgm:
fixed and verified in PR-2363 but this issue wasn't closed. Closing.

PR #2363, opened 2012-08-20
Fix the Opera unit test flakiness when checking for element focus.

shane-tomlinson:
issue #2282


zaach:
Confirmed this fixes the issue on Opera. r+

PR #2368, opened 2012-08-20
Issue 2364 upstream selenium changes

jaredhirsch:
pulling in some upstream bidpom changes.

@klrmn, these changes didn't apply totally cleanly. It looks like some changes were made to the bidpom check_sign_in test as part of pull request 2366, to rename mozwebqa.email to credentials['email'] and a couple similar changes, which aren't present upstream. 

If you think I've merged in the conflict correctly, and I think I have, then I guess I'll issue a pull request back to bidpom to catch the updates. Otherwise, we'll be ever-so-slightly diverged from bidpom permanently, which will absolutely bite us in the long run.


jaredhirsch:
doh! Just read shane's docs on pull request conventions, realized IM DOIN IT WRONG.

jaredhirsch:
reissued as https://github.com/mozilla/browserid/pull/2370

PR #2369, opened 2012-08-20
fix calling id.watch() from the head

### LINKED ISSUE #2252: Calling navigator.id.watch() from the document <head> breaks onlogin/onlogout
If you take the updated [Python cookbook example](https://github.com/mozilla/browserid-cookbook/blob/master/python/python.cgi) and apply the following patch (i.e. moving the `navigator.id.watch()` call from the `body` to the `head` of the page):

```
diff --git a/python/python.cgi b/python/python.cgi
index dba2f07..a559c32 100755
--- a/python/python.cgi
+++ b/python/python.cgi
@@ -10,8 +10,6 @@ def print_header(email = 'null'):

     print """<!DOCTYPE html><html><head><meta charset="utf-8">
-</head>
-<body>
 <script>
 navigator.id.watch({
     loggedInEmail: %s,
@@ -26,6 +24,8 @@ navigator.id.watch({
     },
 });
 </script>
+</head>
+<body>
 <form id="login-form" method="POST">
 <input id="assertion-field" type="hidden" name="assertion" value="">
 </form>""" % email
```

This problem occurs because:
1. `navigator.id.logout()` won't do anything if [`loggedInUser` isn't set properly](https://github.com/mozilla/browserid/blob/26a7264a31be3c210d7257bd0ab785ba5c040d3f/resources/static/communication_iframe/start.js#L96)
2. `loggedInUser` isn't set in the `navigator.id.watch()` call if [the communication channel isn't open](https://github.com/mozilla/browserid/blob/26a7264a31be3c210d7257bd0ab785ba5c040d3f/resources/static/include_js/include.js#L1071)
3. the communication channel is not open because while we're in the `head` of the document, [`dom.body` is null](https://github.com/mozilla/browserid/blob/26a7264a31be3c210d7257bd0ab785ba5c040d3f/resources/static/include_js/include.js#L998)

seanmonstar:
stores loggedInUser into a local variable, and notifies
the Channel whenever it finally builds

fixes #2252

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -983,7 +983,28 @@
     }
 
     var commChan,
+        waitingForDOM = false,
         browserSupported = BrowserSupport.isSupported();
+
+    function domReady(callback) {
+      if (document.addEventListener) {
+        document.addEventListener('DOMContentLoaded', function contentLoaded() {
+          document.removeEventListener('DOMContentLoaded', contentLoaded);
+          callback();
+        }, false);
+      } else if (document.attachEvent && document.readyState) {
+        document.attachEvent('onreadystatechange', function ready() {
+          var state = document.readyState;
+          // 'interactive' is the same as DOMContentLoaded,
+          // but not all browsers use it, sadly.
+          if (state === 'loaded' || state === 'complete' || state === 'interactive') {
+            document.detachEvent('onreadystatechange', ready);'
+            callback();
+          }
+        });
+      }
+    }
+
 
     // this is for calls that are non-interactive
     function _open_hidden_iframe() {
@@ -991,10 +1012,19 @@
       // IFRAME as doing so will cause an exception to be thrown in IE6 and IE7
       // from within the communication_iframe.
       if(!browserSupported) return;
+      var doc = window.document;
+
+      // can't attach iframe and make commChan without the body
+      if (!doc.body) {
+        if (!waitingForDOM) {
+          domReady(_open_hidden_iframe);
+          waitingForDOM = true;
+        }
+        return;
+      }
 
       try {
         if (!commChan) {
-          var doc = window.document;
           var iframe = doc.createElement("iframe");
           iframe.style.display = "none";
           doc.body.appendChild(iframe);
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -966,6 +966,8 @@
       ready: null
     };
 
+    var loggedInUser;
+
     var compatMode = undefined;
     function checkCompat(requiredMode) {
       if (requiredMode === true) {
@@ -1022,6 +1024,13 @@
           commChan.bind('login', function(trans, params) {
             if (observers.login) observers.login(params);
           });
+
+          if (loggedInUser) {
+            commChan.notify({
+              method: 'loggedInUser',
+              params: loggedInUser
+            });
+          }
         }
       } catch(e) {
         // channel building failed!  let's ignore the error and allow higher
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@seanmonstar - I ran the PR against a modified example RP, see the attached diff.  @fmarier's observation that document.body is null is still a problem, this prevents the communication iframe from being created if navigator.id.watch is called within the head of the document.

This is not a problem in the initial RP's tab because the assertion is generated in the dialog and passed back through WinChan, making everything appear to operate correctly in the initial tab. There _is_ a problem if the user creates a new account or adds an email to an existing account and they must verify their email. When the email verification page is redirected back to the RP, navigator.id.watch tries to create the communication iframe to generate the assertion. Since document.body does not yet exist, the iframe is not created and no assertion is ever generated.

Perhaps doing a poll for the existence of document.body? Waiting for DOMContentLoaded would work in browsers that support it, for IE8 we would have to use onload. The danger of waiting for onload is that it waits for all document resources to finish, if for some reason an image takes a long time to download or times out, the user could be waiting for an extremely long time.

diff:
http://pastebin.mozilla.org/1768337

seanmonstar:
Oh I see what you mean now. I had thought similarly, but when I tested, I was looking at the console output. The callbacks all being triggered, since the observers are held onto regardless of the iframe. And I was clicking logout first, which tries to add the iframe again.

I'll listen for `DOMContentLoaded` and `readyStateChange` for IE8.

resources/static/include_js/include.js
```diff
@@ -981,18 +983,48 @@
...
+        document.addEventListener('DOMContentLoaded', function contentLoaded() {
+          document.removeEventListener('DOMContentLoaded', contentLoaded);
+          callback();
+        }, false);
+      } else if (document.attachEvent && document.readyState) {
+        document.attachEvent('onreadystatechange', function ready() {
+          var state = document.readyState;
+          // 'interactive' is the same as DOMContentLoaded,
+          // but not all browsers use it, sadly.
+          if (['loaded', 'complete', 'interactive'].indexOf(state) > -1) {
```
shane-tomlinson: IE8 unfortunately doesn't support Array.prototype.indexOf.

shane-tomlinson: @seanmonstar - if I apply the following diff, IE8 acts as expected:

```
diff --git a/resources/static/include_js/include.js b/resources/static/include_js/include.js
index 979bca2..013f05f 100644
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -997,9 +997,12 @@
           var state = document.readyState;
           // 'interactive' is the same as DOMContentLoaded,
           // but not all browsers use it, sadly.
-          if (['loaded', 'complete', 'interactive'].indexOf(state) > -1) {
-            document.detachEvent('onreadystatechange', ready);
-            callback();
+          var events = ['loaded', 'complete', 'interactive'];
+          for(var i = 0, eventName; eventName = events[i]; ++i) {
+            if(eventName === state) {
+              document.detachEvent('onreadystatechange', ready);
+              callback();
+            }
           }
         });
       }
```

seanmonstar: I forget that so often...

I'm inclined to switch it to `if (state === 'loaded' || state === 'complete' || state === 'interactive')`

seanmonstar:
@shane-tomlinson changed away from indexOf

resources/static/include_js/include.js
```diff
@@ -981,18 +983,48 @@
...
+          document.removeEventListener('DOMContentLoaded', contentLoaded);
+          callback();
+        }, false);
+      } else if (document.attachEvent && document.readyState) {
+        document.attachEvent('onreadystatechange', function ready() {
+          var state = document.readyState;
+          // 'interactive' is the same as DOMContentLoaded,
+          // but not all browsers use it, sadly.
+          if (state === 'loaded' || state === 'complete' || state === 'interactive') {
+            document.detachEvent('onreadystatechange', ready);'
```
shane-tomlinson, line 1001: There is an extra ' on the end of this line.

resources/static/include_js/include.js
```diff
@@ -981,18 +983,48 @@
...
+        document.addEventListener('DOMContentLoaded', function contentLoaded() {
+          document.removeEventListener('DOMContentLoaded', contentLoaded);
+          callback();
+        }, false);
+      } else if (document.attachEvent && document.readyState) {
+        document.attachEvent('onreadystatechange', function ready() {
+          var state = document.readyState;
+          // 'interactive' is the same as DOMContentLoaded,
+          // but not all browsers use it, sadly.
+          if (state === 'loaded' || state === 'complete' || state === 'interactive') {
```
shane-tomlinson, line 1000: Much simpler than my solution.

shane-tomlinson:
@seanmonstar - I removed the extra ' and manually merged this in with commit 18c7a2f7c486a9f722fedf1f8d40b9cdef589cf5.  

close #2252

Issue #2281, opened 2012-08-14
Distinct KPI for password reset screen


nmalkin:
Though the password reset screen is distinct from the regular set-password screen, the same event is emitted when it is displayed: _screen.set_password_. 

I propose having an event _screen.reset_password_ for this situation.

This is valuable, because, right now, there is no good way of identifying a user who is trying to reset his or her password through the KPI event stream. (The current best solution is to look for _screen.check_registration_ but **not** _user.email_staged_ or _user.user_staged_.)

(Similar to #1825).

nmalkin:
> The current best solution is to look for _screen.check_registration_ but **not** _user.email_staged_ or _user.user_staged_.

Even that's not enough, as it turns out.

@shane-tomlinson, do you have any idea what subset of events could uniquely identify a user as being in the password reset flow? (Until we get this implemented.)

shane-tomlinson:
@nmalkin - you wanna review the above to see if it gives you what you need?

PR #2373, opened 2012-08-21
KPI updates for reset password flow

shane-tomlinson:
Add the following new events to the KPI event stream:
- screen.reset_password
- user.reset_password_staged
- user.reset_password_confirmed
- user.reverify_email_staged
- user.reverify_email_confirmed

There was also a spelling error in user.user_confirmed, it was previously user.user_confrimed.

issue #2281


nmalkin:
The changes look good to me, and I've confirmed that both sets of new events (reset_password and reverify_email) are emitted during the respective flows.

jrgm:
verified train-2012.10.12

PR #2378, opened 2012-08-21
add docs about snippets to docs/changes/ (which also keeps the directory alive)

warner:
This sets up the changelog-generation process that @shane-tomlinson and I were talking about this morning.

### DIFF
--- a/docs/changes/Changelog-Maintenance.md
+++ b/docs/changes/Changelog-Maintenance.md
@@ -1,0 +1,35 @@
+
+# Changelog Snippets
+
+This directory contains snippets of text that should be added to the Changelogs for various audiences. For details, please see:
+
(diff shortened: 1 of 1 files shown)


lloyd:
I don't know that I'm completely bought into snippets.  It may be too heavyweight for what we're doing.  See https://github.com/mozilla/browserid/wiki/Reviewing-Pull-Requests

warner:
Cool by me. Feel free to pull out docs/changes/ . My only thought on having devs modifing `ChangeLog` directly is that it's not always apparent which train a given pull-request will eventually land in, so maybe they should add their entries to a non-specific section at the beginning of the file, instead of to the Train-XYZ section. But, we could also just double-check those lists manually when the train departs.

Issue #1863, opened 2012-06-28
alias en_UK et al to en_US


seanmonstar:

lloyd:
talk to @lloyd for context!

shane-tomlinson:
@lloyd, @seanmonstar, @ozten, @mathjazz Instead of aliasing en_UK to en_US, would it be better to rename our en_US locale just plain en and we let this be the fallback?

It seems kind of silly that if a user specifies en_UK as their default language and then es as their second, the user sees es - at least this was the case last time I checked.

Would this approach work for other languages as well, it looks like we have zn-CN and zn-TW,and nb-NO with no fallbacks to plain zn or nb?

ozten:
This has been discussed previously (and i'll defer to l10n-drivers if I mispeak or the policy has changed but...)

1) I'm okay with changing en_US to en, because I don't think there are any en_\* locales which would be offended.

Aside: What % of our users have en_UK,es ?

2) We can get into real political trouble by doing this for other locales. I think it's even documented in the unit tests. 

**Please don't do this without the help of l10n-drivers.**

shane-tomlinson:
@ozten, @mathjazz - who are the appropriate people to get in contact with on l10n-drivers?  My contact so far has been limited to @mathjazz.

mathjazz:
Pascal (our web l10n driver): "on mozilla.org en-\* are aliased to en-US, if we need a specific page that really targets England (to announce an event for example), we can bypass the alias"

milossh:
Aside from that, there's no and there should be no zh, as per community.

seanmonstar:
That's fine with me. As long as someone who has `en_UK, cn` doesn't get
Chinese when we can clearly serve him English.

jrgm:
verified in stage with the current set of localizations on train-2012.10.12

jrgm:
Oh crap. Spoke too soon. If the browser does not send an Accept-Language header (e.g., in FF remove all the languages in the preferences UI, or in about:config set 'intl.accept_languages' to be '') then /sign_in loads content that has the HTML lang attribute set to 'en-US' and tries to load https://login.anosrep.org/production/en_US/dialog.js which is 404. The dialog is broken in this case.

Admittedly, not sending any Accept-Language header is unusual but not impossible. I'll open a five-star issue about this specific case.

jrgm:
Okay, GH-2645 was a configuration issue in stage and fixed now.

PR #2384, opened 2012-08-23
Change the default language from en-US to en

shane-tomlinson:
This involves no changes for the localizers. 

The default language was set to "en" in all of the configuration files. All references to en-US were changed to "en" in supported languages.  en-\* languages will fall back to "en" if an exact match is not made. Unsupported languages fall back to "en" because it is the default language.

issue #1863

### DIFF
--- a/config/aws.json
+++ b/config/aws.json
@@ -7,10 +7,10 @@
   "kpi_backend_sample_rate": 1.0,
   // for amazon deployments, enable it-CH which is the trigger language
   // for localization tests db-LB (which is a testing language where chars
-  // are inverted and reversed), and en-US.
+  // are inverted and reversed), and en.
   // This set can be overridden by adding more to config.json on the VM.
   "supported_languages": [
-    "en-US", "it-CH"
+    "en", "it-CH"
   ],
   "debug_lang": "it-CH",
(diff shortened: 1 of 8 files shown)


shane-tomlinson:
@mathjazz, @lloyd - Since @ozten is out, can I get a review on this?

@mathjazz - This PR involves 0 changes to the SVN repo or the way the localizers do their work.

mathjazz:
Shane, this looks good to me.

jrgm:
verified in stage with the current set of localizations on train-2012.10.12

jrgm:
Oh crap. Spoke too soon. If the browser does not send an Accept-Language header (e.g., in FF remove all the languages in the preferences UI, or in about:config set 'intl.accept_languages' to be '') then /sign_in loads content that has the HTML lang attribute set to 'en-US' and tries to load https://login.anosrep.org/production/en_US/dialog.js which is 404. The dialog is broken in this case.

Admittedly, not sending any Accept-Language header is unusual but not impossible. I'll open a five-star issue about this specific case.

jrgm:
Okay, GH-2645 was a configuration issue in stage and fixed now.

PR #2392, opened 2012-08-24
Complete registration test fix

lloyd:
This should fix the most prominent failure in our unit tests which arises on many of our VMS, usually the slower ones.  

The problem was that we were trying to test the registration landing page when verifying an email that was staged via 123done.  Because the landing page redirects after ~5 seconds, we were doing a WebDriverWait with a non-automic test case.  That test case would often (reproducibly > 50% of the time on osx/ffx saucelabs vm) fetch an element before page redirect, and try to extract the textual contents of that element after page redirect.

Now we wait for the redirect to occur which alleviates the timing issues.  

In order to fix this I selectively merged some of @klrmn's fixes to the complete_registration bit of bidpom.  

While some tests might be broken by this bidpom change, all we care about right now is 123done tests.

### DIFF
--- a/automation-tests/123done/tests/test_change_password.py
+++ b/automation-tests/123done/tests/test_change_password.py
@@ -27,9 +27,8 @@ class TestChangePassword:
         email = inbox.find_by_index(0)
 
         # Load the BrowserID link from the email in the browser
-        mozwebqa.selenium.get(email.verify_user_link)
         from browserid.pages.complete_registration import CompleteRegistration
-        CompleteRegistration(mozwebqa.selenium, mozwebqa.timeout)
+        CompleteRegistration(mozwebqa.selenium, email.verify_user_link)
 
         mozwebqa.selenium.get(mozwebqa.server_base_url)
         from browserid.pages.account_manager import AccountManager
--- a/automation-tests/123done/tests/test_change_password.py
+++ b/automation-tests/123done/tests/test_change_password.py
@@ -28,7 +28,7 @@ class TestChangePassword:
 
         # Load the BrowserID link from the email in the browser
         from browserid.pages.complete_registration import CompleteRegistration
-        CompleteRegistration(mozwebqa.selenium, email.verify_user_link)
+        CompleteRegistration(mozwebqa, email.verify_user_link)
 
         mozwebqa.selenium.get(mozwebqa.server_base_url)
         from browserid.pages.account_manager import AccountManager
--- a/automation-tests/123done/tests/test_new_user.py
+++ b/automation-tests/123done/tests/test_new_user.py
@@ -28,7 +28,7 @@ class TestNewAccount:
 
         # Load the BrowserID link from the email in the browser
         from browserid.pages.complete_registration import CompleteRegistration
-        complete_registration = CompleteRegistration(mozwebqa.selenium, email.verify_user_link)
+        complete_registration = CompleteRegistration(mozwebqa, email.verify_user_link)
 
         home_pg.go_to_home_page()
         Assert.equal(home_pg.logged_in_user_email, user['email'])
--- a/automation-tests/123done/tests/test_new_user.py
+++ b/automation-tests/123done/tests/test_new_user.py
@@ -27,13 +27,8 @@ class TestNewAccount:
         email = inbox.find_by_index(0)
 
         # Load the BrowserID link from the email in the browser
-        mozwebqa.selenium.get(email.verify_user_link)
         from browserid.pages.complete_registration import CompleteRegistration
-        complete_registration = CompleteRegistration(mozwebqa.selenium, mozwebqa.timeout)
-
-        # Check the message on the registration page reflects a successful registration!
-        Assert.contains("Thank you for signing up with Persona.", complete_registration.thank_you)
+        complete_registration = CompleteRegistration(mozwebqa.selenium, email.verify_user_link)
 
         home_pg.go_to_home_page()
-
         Assert.equal(home_pg.logged_in_user_email, user['email'])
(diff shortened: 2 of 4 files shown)


automation-tests/browserid/pages/complete_registration.py
```diff
@@ -12,20 +12,39 @@
 
 class CompleteRegistration(Base):
 
+    _page_title = 'Mozilla Persona: Complete Registration'
     _email_locator = (By.ID, 'email')
     _password_locator = (By.ID, 'password')
     _finish_locator = (By.CSS_SELECTOR, 'div.submit > button')
     _thank_you_locator = (By.ID, 'congrats')
 
-    def __init__(self, selenium, timeout, expect='success'):
```
lloyd, line 20: note, @klrmn changed the signature of CompleteRegistration().  First, Timeouts are not specified everywhere in high level code.  Second, the caller provides the URL to redirect to, rather than the Callee redirecting before invoking CompleteRegistration.  Third, the values of 'expect' are widened to handle the various states of the landing page.

I agree with all of these changes.

automation-tests/run_saucelabs
```diff
@@ -42,6 +42,7 @@ var globalPythonArgs = {
   "--saucelabs": sauceYAMLPath,
   "--webqatimeout": 90,
   "--destructive": null,
+  "-n": 5,
```
lloyd, line 45: an orthogonal change to the PR which makes `./run_saucelabs` execute individual tests in parallel.  This imposes greater load on sauce but get's er done faster (the slowest test is now total test time which is about 123s for me.  awesomesauce.)

automation-tests/123done/tests/test_change_password.py
```diff
@@ -27,9 +27,8 @@ def test_can_change_user_password(self, mozwebqa):
         email = inbox.find_by_index(0)
 
         # Load the BrowserID link from the email in the browser
-        mozwebqa.selenium.get(email.verify_user_link)
         from browserid.pages.complete_registration import CompleteRegistration
-        CompleteRegistration(mozwebqa.selenium, mozwebqa.timeout)
+        CompleteRegistration(mozwebqa.selenium, email.verify_user_link)
```
lloyd: updates to client code for new signature of CompleteRegistration()

automation-tests/123done/tests/test_new_user.py
```diff
@@ -27,13 +27,8 @@ def test_can_create_new_user_account(self, mozwebqa):
         email = inbox.find_by_index(0)
 
         # Load the BrowserID link from the email in the browser
-        mozwebqa.selenium.get(email.verify_user_link)
         from browserid.pages.complete_registration import CompleteRegistration
-        complete_registration = CompleteRegistration(mozwebqa.selenium, mozwebqa.timeout)
-
-        # Check the message on the registration page reflects a successful registration!
-        Assert.contains("Thank you for signing up with Persona.", complete_registration.thank_you)
```
lloyd, line 35: This assertion is meaningless.  It cannot be safely checked, because the target landing page redirects.  Now instead of testing for this language inside the page, we test for successful redirection inside CompleteRegistration().  This is sufficient and more robust for the purpose of this test.  (if the page successfully redirects, then and the email is displayed on 123done, then all the mechanism must be working).

Useful to note that we hope to completely eliminate user visible elements of the landing page that performs the redirect, so this change future proofs us a bit too.

automation-tests/123done/tests/test_new_user.py
```diff
@@ -27,13 +27,8 @@ def test_can_create_new_user_account(self, mozwebqa):
...
+        complete_registration = CompleteRegistration(mozwebqa.selenium, email.verify_user_link)
 
         home_pg.go_to_home_page()
```
klrmn, line 33: this line might be superfluous, since the page redirects to 123done's homepage without help from the test.

automation-tests/123done/tests/test_change_password.py
klrmn: check.

automation-tests/browserid/pages/complete_registration.py
```diff
@@ -12,20 +12,39 @@
...
-        Base.__init__(self, selenium, timeout)
+    def __init__(self, mozwebqa, url, expect='redirect'):
+        """
+        class init method
+        :Args:
+        - url - the confirmation url from the email
+        - expect - redirect/success/reset/verify (default redirect)
+        """
+        Base.__init__(self, mozwebqa)
```
klrmn: does the Base referred to in this file expect (self, mozwebqa) or (self, selenium, timeout)? there was no change to the import, nor a change to base.py in this directory.

https://github.com/mozilla/browserid/blob/dev/automation-tests/browserid/pages/base.py#L12 says it takes (self, selenium, timeout=60), so you need to send (self, self.selenium, self.timeout)

automation-tests/123done/tests/test_new_user.py
lloyd, line 33: I actually tried to remove it and got an exception when trying to access .logged_in_user_email.  I concluded it was necessary, but I agree it sucks to add technically unnecessary work (page load).  Ideas to allow home_pg to extract logged_in_user_email without a navigation?

automation-tests/browserid/pages/complete_registration.py
lloyd: I'll add a fix.  Why does the code work at all?

lloyd: ok, it was working because I changed client code, so while CompleteRegistration's **init** called the argument `mozwebqa`, it was actually a selenium object.  The fix is to adjust both client code and this call to `Base.__init__()` - running tests now and will add commit.

klrmn:
r+

Issue #2017, opened 2012-07-12
add different callback and discourage onready()


benadida:
new callback should be something like onstatusquo(), which gets called if neither onlogin nor onlogout are about to be called. We should then phase out onready().

callahad:
How certain are we of this change? Should I discourage use of onready in MDN now*, while we work on finalizing its deprecation plan and the exact details of onstatusquo?

\* Where now means as soon as MDN comes back up for edits.

callahad:
I believe the decision is to go with `onvalidsession`, but awaiting further discussion from @lloyd and co.

warner:
this sounds like it needs to be nailed down soon (to allow onready() to be phased out in a reasonable timeframe), but not necessarily before beta1

callahad:
As per discussion with @lloyd, we're confident that we want to do this, but the details are still up in the air. Since many sites are using the Observer API without using onready, I'm going ahead and pulling this from the docs but without formally deprecating it.

**We have not yet announced this deprecation!** Do NOT modify / remove `onready` without discussing a deprecation plan with @fmarier or me.

callahad:
To be implemented in #2648 as `onmatch`.

PR #2401, opened 2012-08-27
Add note-to-ourselves re: onready's deprecation

callahad:
onready is in a weird state of limbo: we know we're going to deprecate it, we
know what we're going to replace it with, and we know we want to discourage its
use for the moment, but we haven't actually planned, implemented, or announced
its replacement.

We haven't announced its deprecation, either.

Adding two small comments in-code to ensure that we don't inadvertantly remove
this without giving RPs proper notice. After all, if onready stops firing, we
seriously break sites that rely on it.

See GH-2017 for more info.

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1040,6 +1040,7 @@
               commChan.call({
                 method: 'loaded',
                 success: function(){
+                  // NOTE: Do not modify without reading GH-2017
                   if (observers.ready) observers.ready();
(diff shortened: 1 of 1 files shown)


callahad:
This seems silly, but seriously, we don't want to remove this before giving RPs enough of a heads up.

Because we've already pulled it from our docs, it'd be super easy to say "Oh, it was experimental and it hasn't been documented in months. Nuke it."

And that would break RPs.

PR #2410, opened 2012-08-29
Add new "Sign in" PNGs, compress existing ones

callahad:


lloyd:
skinny wants us to provide click states as well, but let's merge this now as its an incremental improvement.

PR #2414, opened 2012-08-31
BrowserID.persona_test_user() with unit tests

klrmn:
this has not been plumbed into any actual tests yet

### DIFF
--- a/automation-tests/browserid/browser_id.py
+++ b/automation-tests/browserid/browser_id.py
@@ -4,6 +4,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import urllib2
+import json
+
 import selenium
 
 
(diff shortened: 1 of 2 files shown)


klrmn:
@jedp - in terms of what I am aware the bidpom BrowserID class needs, personatestuser.org is working. In terms of writing unit tests, we should contemplate how many layers of testing test tools makes sense or whether the tests making use of the tools are test enough.

@davehunt, @zac, @6a68 - 
This method should just Do The Right Thing for actual client projects like marketplace-tests. 

Since tests run on 123done.org and such need to know which environment they are in (dev/stage/prod) to use this, another layer is needed to detect the env from the --baseurl. I'm still contemplating the possible ways of doing this.

klrmn:
oh, and plumbing this into the various test sets should speed them up considerably.

jaredhirsch:
I guess it's not clear to me what the intent of this persona_test_user is, exactly. Can you provide some explanation?

Is this a way to create test users outside of the regular flow?

klrmn:
@6a68 - many test cases have 'registered user exists' as a precondition. this method gives them exactly that, without using selenium, so it runs faster.

the test(s) that test creation of users will go thru the regular steps using selenium, this is just for use in preconditions.

jaredhirsch:
ah, ok. just looked at personatestuser.org and it made sense. Definitely good to have out-of-band ways to create users, rather than depending on the creation flow we're supposed to be testing.

PR #2370, opened 2012-08-21
Issue 2364 upstream selenium changes

jaredhirsch:
The PR formerly known as https://github.com/mozilla/browserid/pull/2368, this time with more following-documented-procedures cowbell.

My comments from the original PR:

> pulling in some upstream bidpom changes.
> 
> @klrmn, these changes didn't apply totally cleanly. It looks like some changes were made to the bidpom check_sign_in test as part of pull request 2366, to rename mozwebqa.email to credentials['email'] and a couple similar changes, which aren't present upstream.
> 
> If you think I've merged in the conflict correctly, and I think I have, then I guess I'll issue a pull request back to bidpom to catch the updates. Otherwise, we'll be ever-so-slightly diverged from bidpom permanently, which will absolutely bite us in the long run.

PR #2415, opened 2012-08-31
Fix complete registration

klrmn:
so, the problem was that in https://github.com/mozilla/browserid/commit/1adbe2f70b4727f2f4bb1b68706173e1e9ef7baa and https://github.com/mozilla/browserid/commit/9780d7933ff0d921b9cdf53e6534230008c9c54e, the tests in the browserid directory were not updated.

i could have fixed it by just fixing the bidpom tests, but bidpom is meant to be able to be used by people who are not using the pytest-mozwebqa plugin, so i fixed it to re-enable that.

### DIFF
--- a/automation-tests/123done/pages/home.py
+++ b/automation-tests/123done/pages/home.py
@@ -61,6 +61,10 @@ class HomePage(Page):
 
     @property
     def logged_in_user_email(self):
+        WebDriverWait(self.selenium, self.timeout).until(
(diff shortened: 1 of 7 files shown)


klrmn:
this is needed before https://github.com/mozilla/browserid/pull/2370 can be merged in.

jaredhirsch:
R+ looks swell :-) thanks for cleaning this up

PR #2416, opened 2012-08-31
Train 2012.08.17 automated test fixes

lloyd:
a squash of all automation test fixes made in dev.  This change contains nothing outside of automation-tests/, and should get the tests riding with train-2012.08.17 as stable as dev tests.


jaredhirsch:
ok, tests passing for me too.

Issue #2382, opened 2012-08-22
Re-enable siteName and siteLogo for .get()


callahad:
Last month we decided that we should support `siteName` and `siteLogo` with the `.get()` API after noticing that they were inadvertently available (#2086, #2109). This reverses a decision made in #2070.

We just need to update https://github.com/mozilla/browserid/blob/d6cf9538185623d53c87d2a0889868ec857823e8/resources/static/include_js/include.js#L1243-L1246 and add tests.

shane-tomlinson:
Closing with the merge of #2421

jrgm:
verified train-2012.10.12 (via beta.myfavoritebeer.org)

PR #2421, opened 2012-09-04
Allow siteName and siteLogo when using navigator.id.get

shane-tomlinson:
issue #2382


callahad:
r+ Tests pass and it looks good to me. Travis is complaining about the backend tests on node 0.8. Not sure what's going on there.

Should we add unit tests to ensure that these options continue working when passed to `.get()`?

shane-tomlinson:
@callahad - there are no real unit tests for include.js. It's a sad state of affairs, I am hoping to test this using the Selenium tests.

seanmonstar:
It was just a timeout on the tests. In she goes.

Issue #2388, opened 2012-08-24
return to management UI when verify_email_address in a second browser 


jrgm:
1) In browser 1 sign up as a new secondary email address from 123done
2) Load the verify_email_address link in browser 2
3) Enter the password
4) browser 1 transitions to show logged in to 123done
5) browser 2 transitions to show 'Your address has been verified!', but stays on that page.
6) Reloading the page in browser 2 shows the error screen, but clicking on the Persona logo shows the account management UI

It would be nice if the UI did a redirect to '/' after a short interval.

jrgm:
verified on train1012.personatest.org for the current state of the dev branch (what will be train-2012.10.12).

jrgm:
verified above, but this issue wasn't closed. Closing.

PR #2424, opened 2012-09-04
If there is no saved site info in localStorage, redirect to the management page after verification is complete.

shane-tomlinson:
This makes it so the user is always redirected _somewhere_ after verification.
- If the user is verifying in the same browser as they started and started at an RP, the user will be redirected to the RP.
- If the user is verifying in the same browser as they started and started at Persona, the user will be redirected to Persona's management page.
- If the user is verifying in a different browser than they started at, the user will be redirected to Persona's management page.

issue #2388


klrmn:
thank you for specifying all of the use cases.

jrgm:
Verified handled correctly in current dev (train-2012.10.12) for the three cases noted above.

PR #2428, opened 2012-09-05
change loggedInUser to == instead of ===

seanmonstar:
loggedInUser starts as undefined, and later can become a real value, or be set explicitly to null. So, when checking if there is a loggedInUser, we must allow either undefined or null to match.

This would trigger an initial onlogout event, immediately, since it's first set to undefined, and then checked for null.

Unless this as by design? If so, we should add to the comment that this has been considered.

### DIFF
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -58,7 +58,7 @@
         // the resources or cost required to generate an assertion
         if (assertion) chan.notify({ method: 'login', params: assertion });
         loggedInUser = email;
-      } else if (loggedInUser !== null) {
+      } else if (loggedInUser != null) {
         // only send logout events when loggedInUser is not null, which is an
         // indicator that the site thinks the user is logged out
         chan.notify({ method: 'logout' });


seanmonstar:
Some people may end up doing something like this:

``` js
var currentUser = '{{ session.email }}';
navigator.id.watch({
  loggedInUser: currentUser
});
```

If the user isn't logged in, it could be setting loggedInUser to an empty string. With the current check, Persona would think that the RP is reporting someone is logged in.

What if I updated this pull request to check for truthiness, instead?

``` js
} else if (!loggedInUser) {
```

/cc @lloyd

callahad:
Requiring `null` to explicitly signal "no user is logged in" and `undefined` for "I do not know if a user is logged in" was by design. I'm wary about changing this right after Beta, especially since it would alter existing behavior.

@lloyd is the master of this API, though.

lloyd:
yeah, having null and undefined imply different things is part of the api - we need a distinction between "I don't know who is logged in" and "noone is logged in" to preserve existing semantics.

callahad:
@lloyd Before closing this entirely, what would you think about having `null` and `""` (emptystring) be equivalent?

I'm still wary about Beta / API stability, but that one in particular seems like a common and easy mistake with most templating languages.

I'm completely on the fence. If you're not strongly swayed, close it down :)

lloyd:
null and empty string being equivalent seems fine to me.  slight preference for requiring the developer to specify `null` lest we surpress logout events and surprise the developer.

callahad:
Making that change would let me change

```
{% if session.email %}
  var currentUser = '{{ session.email }}';
{% else %}
  var currentUser = null;
{% endif %}
```

to

```
var currentUser = '{{ session.email if session.email }}';
```

Still, I'm not sure it's worth the cost in terms of breaking backwards compatibility and opening the slippery slope of "why not all falsey values other than undefined?"

callahad:
Not being able to promise API stability for beta 1 would suck, and this works for now. Siding with lloyd and closing.

PR #2435, opened 2012-09-06
Update bidpom readme

klrmn:
this should have been a part of https://github.com/mozilla/browserid/pull/2365


jaredhirsch:
@klrmn - I'm pretty confused by the git log here. Is it possible to just include the last commit in this PR?

PR #2442, opened 2012-09-06
update README

klrmn:
this should have been a part of https://github.com/mozilla/browserid/pull/2365

now with better branch tracking


jaredhirsch:
hey! that was quick. thanks for cleaning that up.

PR #2443, opened 2012-09-06
make the url parameter optional

klrmn:
tests that open the url themselves should pass

now with better branch tracking

### DIFF
--- a/automation-tests/123done/tests/test_new_user.py
+++ b/automation-tests/123done/tests/test_new_user.py
@@ -28,6 +28,7 @@ class TestNewAccount:
 
         # Load the BrowserID link from the email in the browser
         from browserid.pages.complete_registration import CompleteRegistration
-        complete_registration = CompleteRegistration(mozwebqa.selenium, mozwebqa.timeout, email.verify_user_link)
+        mozwebqa.selenium.get(email.verify_user_link)
+        complete_registration = CompleteRegistration(mozwebqa.selenium, mozwebqa.timeout)
 
         Assert.equal(home_pg.logged_in_user_email, user['email'])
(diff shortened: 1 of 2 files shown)


klrmn:
@6a68 this needs to be a part of what gets pushed to bidpom/prod

jaredhirsch:
Hey @klrmn - I find this diff, in isolation, pretty confusing. Submitting PRs works better when they are a little less granular, as there's not much context here to explain what this fixes, or why it's useful.

As a general guideline, I'd suggest keeping series of related commits together inside one PR against one repo, instead of submitting a PR for every commit. If the commit message itself doesn't provide context, then the PR itself needs to explain what the code does--as in this case, where I'm looking at the code a couple weeks after the PR was submitted.

Just for reference, here are some good guidelines on git commit messages: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

I also would like to not get into minor details on what does and doesn't get upstreamed to bidpom; it's too much work, and not IMO worth it. I'll periodically bundle up all our diffs and submit upstream, and will generally leave it to upstream folks to figure out what diffs they do or don't want to accept.

klrmn:
Hi @6a68,

I submitted this PR because I realized after the fact that
https://github.com/mozilla/browserid/pull/2392 broke the bidpom tests, and
if it were pushed to the bidpom repo, it would change the api, and would
therefore break the tests of client users like marketplace-tests.

I apologize for not including enough context.

jaredhirsch:
hey @klrmn - no worries! Just trying to provide some constructive feedback :-)

jaredhirsch:
we're moving from python selenium to node selenium; closing accordingly.

Issue #2444, opened 2012-09-07
fix CREATE_TEST_USERS mysql insert statement


jrgm:
It's missing a placeholder for the hashed password.

PR #2445, opened 2012-09-07
add missing placeholder on INSERT statement for adding test users

jrgm:
Fixes GH-2444.  @warner can you look at this?

### DIFF
--- a/lib/db/mysql.js
+++ b/lib/db/mysql.js
@@ -591,7 +591,7 @@ exports.cancelAccount = function(uid, cb) {
 
 exports.addTestUser = function(email, hash, cb) {
   client.query(
-    "INSERT INTO user(passwd, lastPasswordReset) VALUES(FROM_UNIXTIME(?))",
+    "INSERT INTO user(passwd, lastPasswordReset) VALUES(?, FROM_UNIXTIME(?))",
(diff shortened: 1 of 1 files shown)


jrgm:
(Although, drive by comment while I'm looking at this, why not let the database pick NOW(). e.g.,:

```
 exports.addTestUser = function(email, hash, cb) {
   client.query(
-    "INSERT INTO user(passwd, lastPasswordReset) VALUES(?, FROM_UNIXTIME(?))",
-    [ hash, now() ],
+    "INSERT INTO user(passwd, lastPasswordReset) VALUES(?, NOW())",
+    [ hash ],
     function(err, info) {
       if (err) return cb(err);
```

But that's not part of this pull request.

warner:
Oh, hey, that's an important fix.. r=warner.

What calls `addTestUser()`? Probably nothing in the test suite (since they pass).

(switching to a DB-provided NOW() is probably correct.. I'd have to research how mysql does timestamps to be sure it'd behave the same way. The original idea was to allow `now()` to be dependency-injected so that e.g. unit tests could artifically advance the clock, to exercise "time passing" without actually having to stall the tests by a couple of seconds.)

PR #2449, opened 2012-09-07
statsd module should log uncaught exceptions

ozten:
Fixes Issue#2440

Adds a 'base' module which should provide lowest level features
while server is starting up.

Uses winston in addition to statsd for capturing exceptions.


seanmonstar:
Only comment is that "base" seems a little too generic. It'd be tempting to make this a dumping ground of all sorts of unrelated utility functions. Maybe "basic-exceptions"?

Otherwise looks fine.

ozten:
Agreed. I was imagining that there might be other small things that need to exist before `require`s. Maybe over engineering.

lloyd:
Not sure I would have been an r+ for this one...  It appears exceptions are logged to console with `UNCLE:` prepended.  My questions are:
1. should we actually take the process down with a non-zero exit code?
2. should these go in the error log?
3. Is `UNCLE:` a useful prefix?  The first time I saw it I was confused.  (it is cute, our processes "cry uncle", but..)
4. What happens to stderr on stage and in production?  Is this output discoverable?
5. would a statsd counter that we could monitor on and fires everytime uncaught exceptions are raised be useful?

lib/baseExceptions.js
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * The most base level features that everything else depends on while
+ * booting up the service
+ */
+
+var handleUncaughtException = function (err) {
+  console.error('UNCLE:', err.stack || err);
```
ozten, line 11: Agreed, UNCLE is silly.

lloyd:
I care much less about `UNCLE:` than the other stuff - making sure we're aware of uncaught exceptions, their count and nature, in production and stage.  so that's 1, 2, 4, and 5.

PR #2451, opened 2012-09-07
Use personatestuser in persona_server

klrmn:
persona_server/tests/test_manage_account.py .....
persona_server/tests/test_sign_in_unit.py ..


jaredhirsch:
we're moving from python selenium to node selenium; closing accordingly.

PR #2452, opened 2012-09-08
attempt to fix window switch / close errors

klrmn:
this has been tested against WinXP IE8, but will have to be merged to find out whether it reduces the number of failures on ci

3 occasions of 2 sporadic failures in ci logs as of 5pm friday

seen at
https://ci.mozilla.org/view/Persona/job/identity.browserid.dev.windows.xp.ie.8/1315/testReport/junit/123done.tests.test_new_user/TestNewAccount/test_can_create_new_user_account

self = <browserid.pages.sign_in.SignIn object at 0x188cf10>

```
def close_window(self):
```

> ```
>   self.selenium.close()
> ```

browserid/pages/sign_in.py:53:

E       WebDriverException: Message: u'ERROR Internal Server Error'

---

seen at
https://ci.mozilla.org/view/Persona/job/identity.browserid.dev.windows.7.ie.9/1324/testReport/junit/123done.tests.test_sign_in/TestSignIn/test_that_user_can_sign_in/

https://ci.mozilla.org/view/Persona/job/identity.browserid.dev.windows.7.ie.9/1313/testReport/junit/123done.tests.test_change_password/TestChangePassword/test_can_change_user_password/

self = <browserid.pages.sign_in.SignIn object at 0x2a3ff10>
selenium = <selenium.webdriver.remote.webdriver.WebDriver object at
0x29b9bd0>
timeout = 90, expect = 'new'

```
def __init__(self, selenium, timeout, expect='new'):
    Base.__init__(self, selenium, timeout)
```

> ```
>   selenium.switch_to_window('__persona_dialog')
> ```

browserid/pages/sign_in.py:38:

E       NoSuchWindowException: Message: u"No window found

### DIFF
--- a/automation-tests/browserid/pages/sign_in.py
+++ b/automation-tests/browserid/pages/sign_in.py
@@ -35,6 +35,9 @@ class SignIn(Base):
     def __init__(self, selenium, timeout, expect='new'):
         Base.__init__(self, selenium, timeout)
 
+        # wait for second window to open
+        WebDriverWait(self.selenium, self.timeout).until(
+            lambda s: s.window_handles > 1)
         selenium.switch_to_window('__persona_dialog')
 
         if expect == 'new':
@@ -50,7 +53,11 @@ class SignIn(Base):
             raise Exception('Unknown expect value: %s' % expect)
 
     def close_window(self):
-        self.selenium.close()
+        try:  # race condition with window closing by itself
+            self.selenium.close()
+        except:
+            # if it fails, the framework will capture a screenshot
+            pass
 
     @property
     def signed_in_email(self):


automation-tests/browserid/pages/sign_in.py
```diff
@@ -50,7 +53,11 @@ def __init__(self, selenium, timeout, expect='new'):
             raise Exception('Unknown expect value: %s' % expect)
 
     def close_window(self):
-        self.selenium.close()
+        try:  # race condition with window closing by itself
+            self.selenium.close()
+        except:
+            # if it fails, the framework will capture a screenshot
+            pass
```
jrgm, line 61: So is the test explicitly closing the popup dialog which will dismiss itself? Why would it do that?

klrmn, line 61: the code doesn't differentiate between the pop-up that doesn't close itself on sign-up from the pop-up that does close itself on successful sign-in. the code attempts to close both types, when it's done with them.

jrgm, line 61: Ah, ok. So the popup on signup does close itself on verifying the email if you leave it open. Both paths are valid things to test, yes. Although it should differentiate the cases between signin and signup since prematurely closing the signin popup is a race of its own.

jaredhirsch:
@klrmn - This is great! Way to pitch in on the stabilization stuff :-)

I will merge this branch locally today and test against IE8 a few times to see if I can, first, repro the race condition, and second, repro the fix fixing the race condition.

automation-tests/browserid/pages/sign_in.py
```diff
@@ -35,6 +35,9 @@ class SignIn(Base):
     def __init__(self, selenium, timeout, expect='new'):
         Base.__init__(self, selenium, timeout)
 
+        # wait for second window to open
+        WebDriverWait(self.selenium, self.timeout).until(
+            lambda s: s.window_handles > 1)
```
lloyd, line 40: I agree this should fix ie9, based on my understanding of the problem.

lloyd:
r+ !  merging, once this runs 10x on dev we can merge into stage.  I think this will address the IE9 issue where window is not spawned at the time we attempt to switch to it, which is one of the most frequent failures in the dev env.

awesome work @klrmn

PR #2454, opened 2012-09-10
removing gobbledygook symlink

seanmonstar:
symlink no worky on Windows, so make a copy
at postinstall instead


seanmonstar:
pick me! :D

PR #2456, opened 2012-09-11
Make sure method name in comment matches the current function name

fmarier:
It looks like some of these functions got renamed but the comments were not updated.


fmarier:
The commit diff would have been way more useful if it had an 2-3 extra lines of context!

shane-tomlinson:
Thanks for the update @fmarier.

r+  I will merge whenever I find out status of the dev branch.

PR #2468, opened 2012-09-12
Fix change password stage

klrmn:
@6a68, @lloyd: this should fix the persistent failure of test_can_change_user_password on stage but not dev.

the fix has been tested for both dev and stage.


jaredhirsch:
klrmn - thanks for submitting the fix! This is huge.

Two things: first, please rebase your branch to squash those two commits together, then push -f to your remote branch. This will destructively rewrite history, which is ok in the case of a short-lived branch like this, and it means the second spurious commit (and commit log noise) won't be merged in with the fix.

Second thing: I'd prefer to keep config details like URLs out of the source code and in config files--does mozwebcfg (or whatever it's called) permit multiple stanzas for different configurations? I've got a train to catch so can't look right this second, but if this seems hard to do, I'm cool to merge this as submitted. If it's easy to do, let's do it :-)

klrmn:
the second thing is kinda out of scope. it sounds like this would be a part of #identity needing to get together with @davehunt to figure out how to use pytest-mozwebqa without using --baseurl the way it's intended.

jaredhirsch:
@klrmn - looks great! nice squashing. thanks for the input, merging this in.

davehunt:
@klrmn @6a68 feel free to enhance pytest-mozwebqa to take an environment argument, which can be matched to a section in the config file. This is something I've considered before, but we should continue to support the existing behaviour.

The config file might look like:

```
[DEFAULT]
baseurl: 'http://www.example.com'
api: 'webdriver'
tags: 'tag1, tag2'

[staging]
baseurl: 'http://www.elpmaxe.com'
```

klrmn:
@davehunt - at the moment, my brain is breaking at trying to figure out how
to ignore --baseurl and run specified tests on a selection of [123done,
persona server, myfavoritebeer ]

davehunt:
Ignore --baseurl? Sounds like you need a different plugin, that's kinda fundamental. :)

klrmn:
that would be why it's breaking my brains.
we could simply put 'http://google.com/' in mozwebqa.cfg to make
pytest-mozwebqa go, but then each test will have to be parameterized with
the sites valid for that test, doctored for the right environment, open the
site inside the test.

at least that's what i've come up with so far.

davehunt:
I wouldn't put anything in baseurl that you're not intending on testing. I guess I'm not understanding the problem... If you need to test multiple sites, just run py.test multiple times.

PR #2475, opened 2012-09-13
Update about page screenshot. Fixes #2469

### LINKED ISSUE #2469: How does it work: screenshot has old browserid.org domain in the URL bar
Dialog screenshot

callahad:


shane-tomlinson:
r+

@jrgm, @lloyd - does this change warrant a merge before this train is cut?  If not, I will merge on Monday.

jrgm:
Yeah, this is so simple, we should just do it.

PR #2480, opened 2012-09-14
fixupAbsolutePath: filter out scheme-relative URLs

fmarier:
The [current check](https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/dialog.js#L108) to ensure that only absolute paths are accepted fails to take into account scheme-relative URLs like `//foo.com`.

These URLs end up in `fixupURL` and [get the origin prepended to them](https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/dialog.js#L86) to something like `https://origin.example.com//foo.com`, which is invalid but still follows our same-origin restrictions.

So the solution is to require that the character after the leading slash be anything but a slash.

### DIFF
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -83,7 +83,7 @@ BrowserID.Modules.Dialog = (function() {
     if (typeof(url) !== "string")
       throw "urls must be strings: (" + url + ")";
     if (/^http(s)?:\/\//.test(url)) u = URLParse(url);
-    else if (/^\//.test(url)) u = URLParse(origin + url);
+    else if (/^\/[^\/]/.test(url)) u = URLParse(origin + url);
     else throw "relative urls not allowed: (" + url + ")";
     // encodeURI limits our return value to [a-z0-9:/?%], excluding <script>
     var encodedURI = encodeURI(u.validate().normalize().toString());
@@ -105,7 +105,8 @@ BrowserID.Modules.Dialog = (function() {
   }
 
   function fixupAbsolutePath(origin_url, path) {
-    if (/^\//.test(path))  return fixupURL(origin_url, path);
+    // Ensure URL is an absolute path (not a relative path or a scheme-relative URL)
+    if (/^\/[^\/]/.test(path))  return fixupURL(origin_url, path);
 
     throw "must be an absolute path: (" + path + ")";
   }
--- a/resources/static/test/cases/dialog/js/modules/dialog.js
+++ b/resources/static/test/cases/dialog/js/modules/dialog.js
@@ -614,6 +614,23 @@
     });
   });
 
+  asyncTest("get with a scheme-relative siteLogo URL - not allowed", function() {
+    createController({
+      ready: function() {
+        mediator.subscribe("start", function(msg, info) {
+          ok(false, "start should not have been called");
+        });
+
+        var retval = controller.get(HTTPS_TEST_DOMAIN, {
+          siteLogo: "//example.com/image.png"
+        });
+
+        equal(retval, "must be an absolute path: (//example.com/image.png)", "expected error");
+        testErrorVisible();
+        start();
+      }
+    });
+  });
 
   asyncTest("get with returnTo with https - not allowed", function() {
     createController({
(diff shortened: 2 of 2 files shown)


seanmonstar:
@fmarier add a test case and I'd say we're good to go!

fmarier:
I've rebased that commit onto the latest dev (so that the front-end tests actually pass) and I've added two test cases to make sure we don't reintroduce this small bug.

lloyd:
this fix broke marketplace testing.  specifically, now we do not support wierd ass urls like '//developers/'.  If the servers automatically dedup '//' rather than redirecting, and the website's client code doesn't care, should we?

What was the actual problem that this commit intended to fix, in my current caffeinated state I cannot understand?  This change did change api semantics and it did cost some folks some time: https://bugzil.la/808024

seanmonstar:
Specify `//foo.com` lets you declare an absolute url, without the scheme. So it can just adopt the current scheme (http or https)

lloyd:
if a site passes us `returnTo: "//foo.com"` and we normalize that to "http://example.com//foo.com", how will  that site be able to undermine our same-origin policy?

I think I'm just wondering if we need to harden our normalization code and not explicitly block '//' ?

seanmonstar:
It looks like the code doesn't normalize, but throws an error if someone tries to use `//foo.com`. I think that's better than (sometimes wrongly) guessing it's a typo and was supposed to be only one slash.

fmarier:
@lloyd my reasoning for fixing this was that the initial check was not checking for absolute paths properly and that it was letting URLs on different domains through. While the other function happened to fix it, it could change in the future and allow code to escape the same-origin requirements if it relies on the first function to ensure that it's an absolute path.

If we want to allow these URLs, what I would suggest is duplicating the check to make it clear that we are thinking about schema-relative URLs and breaking them on purpose when we encounter a double-slash at the beginning.

So perhaps something like:

```
function fixupURL(origin, url) {
...
  if (/^http(s)?:\/\//.test(url)) u = URLParse(url);
  else if (/^\/\//.test(url)) u = URLParse(origin + url); // scheme-relative URLs will get turned into same-host absolute paths
  else if (/^\//.test(url)) u = URLParse(origin + url);
  else throw new Error("relative urls not allowed: (" + url + ")");
```

The other function would revert to /^\// like it was prior to this change.

I'm not a fan of code duplication, but I can't think of another way to make it clear that changes to that else clause impact a totally different type of URLs. Any other ideas?

lloyd:
@fmarier ok, I understand now.  

Your proposed fix seems a fine way to balance regression risk but still not needlessly disallow silly paths.  Another way to manage regression risk is to add create a selenium test that excercises the snot out of this code?

My main point here was that because `//developers/` is a valid (but silly) url path and not a scheme-relative url, it seems like we should allow it - especially because we have a case where our over-aggressive checks broke others tests.

@seanmonstar I wasn't saying we should assume `//developers/` was meant to be `/developers/` and change it, I'm just saying if it isn't an absolute url, but it is a syntactically correct absolute path, lets pass it through appended to origin.

shane-tomlinson:
@lloyd - Your proposal seems to have a very small benefit whereas I worry that it will introduce confusion. If we make a change to allow //<resource>, is this going to confuse an separate small subset of people who will then think scheme relative URLs are allowed? 

I propose that we stay consistent with the browsers. If somebody enters in a //<resource> for an image src or an anchor href, how does a browser handle this?

lloyd:
I understand the tension, we are choosing between two fringe cases that are mutually exclusive.  But because it affects our API, it's worth the attention.

Our API contract currently implicitly is:

"siteLogo is an absolute path, however neither leading double slashes nor a path of '/' are allowed"

"returnTo is a absolute path, however it may not begin with leading double slashes"

I think these conditions are not useful.  When we say its a path, then we've completely and precisely specified it and we get to write less prose in our spec.  (And have a simpler contract that helps native implementors).

As far as what would browsers do, if you type a path with leading double slashes, it would send GET '//foo/' to the server, it would not block you from navigating.

IMG SRC is different because it may be a path or an absolute URL, so there is ambiguity where syntax overlaps. But now I'm curious, what _does_ IE8 do with:



I say @callahad makes the call and we update code, tests, and specs.

-- lloyd (thumb-typing)

ozten:
Have we considered requiring a fully qualified url **and we** render a protocol relative url?

lloyd:
The idea behind making these arguments paths rather than urls was to make it obvious that urls with distinct from origins are not allowed.

lloyd:
And keep in mind, the problem that marketplace hit is just by calling .request() from a page with a couple slashes, your popup displays an error to users.  (Because in this case _we_ supply the returnTo parameter based on the document location).

The main issue here is we changed our API after beta without warning and broke people.  The secondary issue is how things should ideally work for returnTo and siteLogo.

callahad:
So... I feel like `////foo` is a valid path, just an unconventionally formatted one. I think we should support that and make it more evident in the docs that `siteLogo` and `returnTo` takes _paths_, not _URLs_.

Issue #1566, opened 2012-05-09
[ie8] red line at top of screen for /signin


jrgm:
Don't know how I didn't notice this little line before:

![img](https://img.skitch.com/20120510-8nqg9hqcwba4sek2x5gwn5wj48.png)

That is with IE8 in standards mode on a freshly created user account on winxp. Minor cosmetic thing.

jrgm:
@lloyd I'm giving this to @shane-tomlinson  to consider fixing for final polish.

seanmonstar:
This looks like it was with previous design. I don't see it in IE9 Compat Mode...

jrgm:
This is still there with a real IE8 browser looking at login.dev.anosrep.org/signin. 

It's the &lt;noscript&gt; element styled here https://github.com/mozilla/browserid/blob/dev/resources/static/pages/css/style.css#L6

jrgm:
Hmm, IE3. It did have the first CSS support ;-)

But no XP/IE8. I doubt it's VM v. PC issue. Could be subtle DLL differences, but the as I read the computed style with ie8 devtools, that red line should be there. 

Anyways, this was always a minor issue. 

(p.s., to be clear, you have to look at /signin, not just /).

shane-tomlinson:
@jrgm, @jbonacci - Have I mentioned that you two have some awesome bug finding and debugging prowess?  I'll have a look as soon as I have access to my laptop again (the HD has taken a turn for the worse).

shane-tomlinson:
Closing as fixed.

PR #2486, opened 2012-09-14
Fix IE8 showing the red line on the signin page.

shane-tomlinson:
IE8 has a rendering bug with the noscript tag, see http://stackoverflow.com/questions/2912175/noscript-tag-appears-even-if-javascript-is-turned-on-in-ie8

Placing the noscript contents inside of another tag that can be styled correctly in IE8

issue #1566

### DIFF
--- a/resources/static/pages/css/style.css
+++ b/resources/static/pages/css/style.css
@@ -3,7 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 
-noscript {
+#noscript-warning {
+  position: absolute;
   position: fixed;
   display: block;
   background-color: #ef1010;
--- a/resources/views/partial/noscript-warning.ejs
+++ b/resources/views/partial/noscript-warning.ejs
@@ -1,0 +1,9 @@
+<noscript>
+  <!-- IE8 has a rendering bug with noscript tags, content must be placed
+       inside of another element to be styled -->
+  <p id="noscript-warning">
+    <%- gettext('We\'re sorry, Persona requires that Javascript is enabled.') %>
+  </p>
+</noscript>
+
+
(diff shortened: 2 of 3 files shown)


resources/static/test/testHelpers/helpers.js
```diff
@@ -56,6 +56,12 @@ BrowserID.TestHelpers = (function() {
   var TestHelpers = {
     XHR_TIME_UNTIL_DELAY: 100,
     setup: function() {
+      // partials are not supported by the client side EJS. Create
+      // a standin if it is not available that does nothing.
```
shane-tomlinson: so dirty it makes my skin crawl.

seanmonstar: Page templates are rendered on the server though, and I don't see this partial used by a dialog template...

Plus, there won't even **be** a client-side EJS once #2455 merges.

shane-tomlinson: _normally_ the page templates are rendered on the server. For client side QUnit tests, main site templates are also rendered in the tests, causing a failure anywhere that "partial" is used. Got any ideas on how we can make this less crappy, or do you think once #2455 merges, we'd be able to take this statement out?

seanmonstar: Oh I see, even once #2455 is merged, the compiled function will still contain a call to `partial`.

shane-tomlinson: @seanmonstar - I have added a comment[https://github.com/mozilla/browserid/pull/2455#issuecomment-8609628] to #2455 to address partials in client side templates. What'cha think?

seanmonstar:
I manually rebased out the `window.partial` stuff, and merged in.

Issue #2309, opened 2012-08-15
several pages do not show javascript disabled warning


shane-tomlinson:
STR

1) Open your favorite browser, disable javascript
2) visit https://login.dev.anosrep.org/forgot
3) Where's the "Persona requires Javascript to be enabled" warning?

also affects:
- add_email_address
- verify_email_address
- reset_password
- confirm

seanmonstar:
that PR handled all those cases, right?

PR #2487, opened 2012-09-14
Issue 2309 forgot js disabled

shane-tomlinson:
Depends on PR #2486 being merged first

issue #2309


seanmonstar:
`partial` support is in #2455, so depends on that.

seanmonstar:
I manually rebased out the `window.partial` stuff, and merged in.

PR #2488, opened 2012-09-14
Restart processes after rpm installation

ozten:
Bump version from 0 to 1
Use daemontools svc -t to restart all browserid
processes deployed on this machine.

@gene1wood @jrgm - How does this look? Seems to work in the Vagrant intcluster.

### DIFF
--- a/scripts/browserid.spec
+++ b/scripts/browserid.spec
@@ -1,7 +1,7 @@
 %define _rootdir /opt/browserid
 
 Name:          browserid-server
-Version:       0.2012.08.17
+Version:       1.2012.08.17
 Release:       1%{?dist}_%{svnrev}
 Summary:       BrowserID server
 Packager:      Gene Wood <gene@mozilla.com>
@@ -39,6 +39,10 @@ done
 mkdir -p %{buildroot}%{_rootdir}/config
 cp -p config/l10n-all.json %{buildroot}%{_rootdir}/config
 cp -p config/l10n-prod.json %{buildroot}%{_rootdir}/config
+
+%post
+echo "Restarting daemons under /var/service/browserid-*"
+/command/svc -t /var/service/browserid-*
 
 %clean
 rm -rf %{buildroot}
--- a/scripts/browserid.spec
+++ b/scripts/browserid.spec
@@ -42,7 +42,7 @@ cp -p config/l10n-prod.json %{buildroot}%{_rootdir}/config
 
 %post
 echo "Restarting daemons under /var/service/browserid-*"
-/command/svc -t /var/service/browserid-*
+/usr/local/bin/svc -t /service/browserid-*
 
 %clean
 rm -rf %{buildroot}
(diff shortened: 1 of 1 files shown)


gene1wood:
This looks good to me. simple change

ozten:
Why do we need this?

The currently planned deployment mechanism will only allow a restart of the browserid processes **after** the drain traffic / restore traffic. So production traffic would be sent to old processes running in a newly deployed env until the restart is executed.

This adds a step during rpm installation to restart all browserid processes. [This happens after the file installation step](http://www.rpm.org/max-rpm/s1-rpm-inside-scripts.html) and before the removal of files from the previous version of the rpm.

gene1wood:
@jrgm Yes, our puppet scripts localize and for systems in stage they "svc -d" the webhead service in favor of the webhead-loadtest services.

This change (adding a "svc -t") should play nice with this.

scripts/browserid.spec
```diff
@@ -40,6 +40,10 @@ mkdir -p %{buildroot}%{_rootdir}/config
 cp -p config/l10n-all.json %{buildroot}%{_rootdir}/config
 cp -p config/l10n-prod.json %{buildroot}%{_rootdir}/config
 
+%post
+echo "Restarting daemons under /var/service/browserid-*"
+/usr/local/bin/svc -t /service/browserid-*
```
floatingatoll, line 45: Does this rpm install /service/browserid-*?

Will this restart any unrelated services that aren't part of this RPM?

Are there already %requires entries for whatever rpm installs /usr/local/bin/svc?

Is the order of restarts important, such as static and browserid before router?

scripts/browserid.spec
```diff
@@ -1,7 +1,7 @@
 %define _rootdir /opt/browserid
 
 Name:          browserid-server
-Version:       0.2012.08.17
+Version:       2.2012.08.17
 Release:       1%{?dist}_%{svnrev}
```
floatingatoll, line 5: Should we be increment Release: 1 to Release: 2, rather than the leftmost major version number, or is this normal for minor changes in BrowserID versioning?

ozten, line 5: Yes, I think your right. What is that 0 for in front of Version?

I'll update Version to `0.2012.08.17` and Release: to `2%{?dist}_%{svnrev}`

scripts/browserid.spec
ozten, line 45: > Does this rpm install /service/browserid-*?

Yes

> Does this rpm install /service/browserid-*?

I don't believe so, we discussed this point in IRC.

> Are there already %requires entries for whatever rpm installs /usr/local/bin/svc?

No. We discussed this. Using `%post` is a temporary solution. Please feel free to fix this, I don't know enough to make that change.

Gene was okay w/o that dependency and moving this solution out into puppet once a long term solution is settled on.

> Is the order of restarts important, such as static and browserid before router?

I don't think so. @gene1wood @jrgm ?

PR #2495, opened 2012-09-17
Issue 2484 api usage kpi

shane-tomlinson:
For issue #2484

### DIFF
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -110,6 +110,20 @@ BrowserID.Modules.Dialog = (function() {
     throw "must be an absolute path: (" + path + ")";
   }
 
+  function validateRPAPI(rpAPI) {
+    var VALID_RP_API_VALUES = [
+      "watch_without_onready",
+      "watch_with_onready",
+      "get",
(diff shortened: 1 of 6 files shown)


resources/static/dialog/js/modules/dialog.js
```diff
@@ -110,6 +110,20 @@ BrowserID.Modules.Dialog = (function() {
...
+    var VALID_RP_API_VALUES = [
+      "watch_without_onready",
+      "watch_with_onready",
+      "get",
+      "getVerifiedEmail",
+      "internal"
+    ];
+
+    if (_.indexOf(VALID_RP_API_VALUES, rpAPI) === -1) {
+      throw "invalid value for rp_api: " + rpAPI;
```
jedp, line 123: I like.  Ensures that we keep this current with future api changes.

jrgm:
Verified train-2012.10.12. For my own notes, the keys are now:

```
email_type
event_stream
lang
new_account
number_emails
number_sites_remembered
number_sites_signed_in
orphaned
rp_api
sample_rate
screen_size
timestamp
```

where rp_api is 'get' for train1012.myfavoritebeer.org, and 'watch_with_onready' for train1012.123done.org.

Issue #2483, opened 2012-09-14
flexible-graphic.png's dimensions are not defined in /about


shane-tomlinson:
This slows down rendering of the /about page.

shane-tomlinson:
Because of image scaling issues on mobile, closing as WONTFIX.

PR #2496, opened 2012-09-17
Set the image sizes

shane-tomlinson:
issue #2483

### DIFF
--- a/resources/views/about.ejs
+++ b/resources/views/about.ejs
@@ -19,13 +19,13 @@
                 </div>
 
                 <div class="graphic">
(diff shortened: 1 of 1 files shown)


callahad:
How does this handle the scaling / responsive design aspects of the page? (I'll pull and investigate myself in a second)

shane-tomlinson:
@callahad - great question, I meant to check before I issued the PR but got side tracked.

callahad:
Answer: It doesn't scale proportionally :)

Resized browser showing squished images

shane-tomlinson:
Closing and the original issue as WONTFIX.

PR #2505, opened 2012-09-20
adding jshintrc files

### LINKED ISSUE #2478: add a .jshintrc file
Our guidelines demand we run jshint before push code (on save is even better), but jshint allows us to declare different rules on each computer.

We have a few jshint settings hiding in the comments at the top of files, so why not just include a [root level `.jshintrc`](https://github.com/jshint/node-jshint/#within-your-projects-directory-tree).

seanmonstar:
3 jshintrc files, because of different environments:
1. node
2. browser
3. browser tests

fixes #2478

### DIFF
--- a/resources/static/.jshintrc
+++ b/resources/static/.jshintrc
@@ -1,0 +1,23 @@
+{
+  "passfail": false,
+  "maxerr": 100,
+  "browser": true,
+  "forin": false,
+  "boss": true,
+  "noarg": true,
+  "undef": true,
+  "browser": true,
+  "laxbreak": true,
+  "laxcomma": true,
+  "eqeqeq": true,
+  "eqnull": true,
+  "expr": true,
+  "predef": [
+    "BrowserID",
+    "_",
+    "$",
+    "jQuery",
+  ],
+  "shadow": false,
+  "supernew": false
+}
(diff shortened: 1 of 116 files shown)


resources/static/.jshintrc
```diff
@@ -0,0 +1,23 @@
+{
+  "passfail": false,
+  "maxerr": 100,
+  "browser": true,
+  "forin": false,
+  "boss": true,
+  "noarg": true,
+  "undef": true,
+  "browser": true,
```
shane-tomlinson: there are two browser declarations (I just realized this is an error in my original)

resources/static/test/.jshintrc
```diff
@@ -0,0 +1,33 @@
...
+    "jQuery",
+    "ok",
+    "equal",
+    "strictEqual",
+    "notEqual",
+    "start",
+    "stop",
+    "asyncTest",
+    "test",
+    "raises"
```
shane-tomlinson: since this is for QUnit tests, should add "module" to this list.

shane-tomlinson:
@seanmonstar - I added a PR to your branch with some cleanup - https://github.com/seanmonstar/browserid/pull/6

seanmonstar:
@shane-tomlinson I pulled that in, and cleaned out a couple globals not needed, like `EJS` and `alert`.

Issue #2485, opened 2012-09-14
is diajog/js/modules/dialog.js->getVerifiedEmail needed?


shane-tomlinson:
This was initially used as an access point for Pancake and Soup and is probably no longer needed.

shane-tomlinson:
PR has been merged, closing.

PR #2511, opened 2012-09-21
remove dialog.js->getVerifiedEmail, it is no longer used.

shane-tomlinson:
remove dialog.js->getVerifiedEmail.

issue #2485


seanmonstar:
r+, will merge Monday

PR #2512, opened 2012-09-21
pre-process cachify in templates

### LINKED ISSUE #2502: Allow cachify in dialog views
`cachify` is needed to show some inline images, and it outputs a unique URL so we can cache bust later. However, it only exists on the server. So if a dialog view has inline images and needs cachify, the function doesn't exist.

seanmonstar:
`cachify` in templates should be pre-processed before they get compiled from EJS. This allows us to use `cachify` in dialog templates, to show inline images and such.

fixes #2502


lib/templates.js
```diff
@@ -7,10 +7,32 @@ const
...
+  // cachify depends on an md5 hash of the contents of the file, so we
+  // can't determine that once we're on the client. This replaces
+  // instances of `cachify('some-url')` to the complete URL.
+  //
+  // Any instances of variables being used are part of the expression
+  // will FAIL.
+  //
+  // Bad: cachify( base + '/my-image.png')
+  return String(text).replace(cachifyRE, function(m, url) {
+    return "'" + cachify.cachify(url) + "'";
```
ozten, line 32: Nice way to do it without evaluating the template and keeping the code DRY.

lib/templates.js
```diff
@@ -7,10 +7,32 @@ const
...
 
 var bundles = {};
 
+
+const commentsRE = /<!--[\s\S]*?-->/g;
+function stripComments(text) {
+  return String(text).replace(commentsRE, '');
+}
+
+const cachifyRE = /cachify\(\s*["']([^"']*)["']*\s*\)/g; // :-(
```
shane-tomlinson, line 21: Turn that frown upside down!

shane-tomlinson:
@seanmonstar - I am trying to figure out a good way of testing this using compressed resources. Unfortunately at the moment the front end unit tests when run on ephemeral instances or with the "use_minified_resources" config item set to false still use uncompressed resources, meaning what you are hoping to test with your cachify check is never tested.

Running frontend unit tests on ephemeral instances is also blocked on #2517

seanmonstar:
It's tested. In the tests, it should replace the cachify call to the string "/dialog/css/style.css", and the test correctly passes.

It's true that the test suite is not able to test compression is working correctly. For that, I just ran `./scripts/compress`, started the server, searched the compressed JS file for the resouce, and saw it properly had its md5 hash prefixed on it.

resources/static/test/cases/common/js/renderer.js
```diff
@@ -39,6 +39,14 @@
     ok($("#focusButton").length, "template loaded with partial");
   });
 
+  test("render dialog template with cachify", function() {
+    renderer.render("#formWrap .contents", "test_template_cachify");
+
+    var expected = /\/dialog\/css\/style\.css$/;
```
shane-tomlinson, line 45: I like that once our front end tests work with production resources on ephemeral instances, this will just work.

resources/static/test/mocks/cachify.js
```diff
@@ -1,14 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-window.cachify = (function() {
```
shane-tomlinson, line 4: less code is a good thing.

Issue #2515, opened 2012-09-24
error when returnTo is empty or set to "/"


shane-tomlinson:
This breaks all dev sites because fixupAbsolutePath marks `returnTo: "/"` as invalid.

PR #2516, opened 2012-09-24
Issue 2515 fixup absolute path

shane-tomlinson:
This is a two commit PR.

1) The first commit addresses issue #2515 - `error when returnTo is empty or set to "/"`
2) The second commit is a huge cleanup of the dialog.js tests.

issue #2515


fmarier:
That fix looks good to me. r+

For the record, this bug was introduced (by me) in #2480 / 60d61f59dd3c17a4f65e90fa694e9fb4269591e6.

Issue #2175, opened 2012-07-27
Enable node-debug mode


jaredhirsch:
Wouldn't it be nice to drop into a REPL to introspect the stack at some interesting or busted point in execution?

@jedp pointed out that we'd probably have to start all the child processes using the --debug option, but maybe that's the only change needed?

@jedp also smartly suggested we could set an env var to toggle debug mode. yum, repl for breakfast :-)

benadida:
yes!!

PR #2522, opened 2012-09-25
debug mode

zaach:
Addresses issue #2175. 

I've noticed that debugger statements aren't caught during the initial loading of the app. There's an optional to have the debugger break at the beginning so it has time to catch those statements, but you'd have to manually tell each process to continue. That would be a pain, so I've neglected to use that option.


jaredhirsch:
lgtm, though I'm pretty unfamiliar with all the things :-)

Issue #2517, opened 2012-09-24
cannot run frontend tests on ephemeral instances


shane-tomlinson:
The routes for `mocks/templates.js` and `mocks/site-templates.js` are set up to only serve up if the resources are not minified.

Front end unit tests are still allowed to be run on ephemeral instances. For the unit tests, ephemeral instances still serve up individual resources, so these two resources must still be included in the routes list.

While this is not user facing, this is important because @jrgm on occasion runs frontend tests on ephemeral instances.

jrgm:
belated but yes front end tests run on ephemeral instances on train-2012.10.12

PR #2524, opened 2012-09-25
Allow site-templates.js and templates.js to be requested on ephemeral instances but not on stage or prod.

shane-tomlinson:
issue #2517


seanmonstar:
Oh hey, I didn't even notice that `if` statement. This is definitely the better place for that.

seanmonstar:
merged since it prevented testing of a feature that had already landed

Issue #2525, opened 2012-09-25
When Persona takes longer than it should, error message is incorrect


nth10sd:
The message "We are sorry, this is taking a longer than it should." has an excess "a".

shane-tomlinson:
Thanks for filing this @nth10sd, I have issued a pull request for review.

warner:
Originally in bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=794306 , please close after the fix lands.

PR #2528, opened 2012-09-26
Remove the extra "a" from "We are sorry, this is taking a longer than it should."

shane-tomlinson:
issue #2525


zaach:
Will this affect i18n? Is it okay to merge?

Issue #2521, opened 2012-09-24
Persona should stop trying to log into sites if it gets into an infinite loop of rapid failures


callahad:
Related to #2520 and https://github.com/mozilla/django-browserid/issues/82. There are a bunch of scenarios in which users can get into infinite loops with failed logins.

If Persona automatically fires the `onlogin` action several times in succession for the same account/domain, it should detect that the login is failing and back off.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

Issue #2520, opened 2012-09-24
Need better docs for what to do when login fails


callahad:
If login fails, it's easy to get stuck in an infinite loop where the error page calls `.watch()` and gets another `onlogin` action, ad infinitum. I need to improve docs to discuss handling this.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #2529, opened 2012-09-26
Allow navigator.id.logout() before session loads

yourcelf:
Prior to this patch, `navigator.id.logout` no-ops if it is called
before a session has loaded (e.g., before the query to login.persona.org
returns, and before "onlogin" has fired).

This patch changes the behavior of `logout` by having it set a flag
which prevents "onlogin" from firing again until the next time
`navigator.id.request` is called.

The assumption is that the intention behind a call to
`navigator.id.logout` is to clear any session, whether it's locally
cached or not; and that this intention should be honored until a new
session is requested through `navigator.id.request`.

Discussion of the rationale is present in issues #2521, #2520, and the
following mailing list thread:

https://groups.google.com/forum/?fromgroups=#!topic/mozilla.dev.identity/mDj8oLKn8aI


shane-tomlinson:
@yourcelf - Thanks for the PR and pardon the delay in review and merge. I agree with your argument that if logout is called before .watch or .request, .login should not be called.

After looking over the code, I couldn't help but think there was a simpler way to do this. I felt like we were masking the problem instead of fixing it.

I am attaching a diff that _should_ result in the same behavior. Could you test to make sure this is the case?

```
diff --git a/resources/static/communication_iframe/start.js b/resources/static/communication_iframe/start.js
index 5bee8ee..b447aaa 100644
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -92,8 +92,18 @@
   });

   chan.bind("logout", function(trans, params) {
-    if (loggedInUser != null) {
+    // The assumption is that the intention behind a call to
+    // navigator.id.logout is to clear any session, whether it's locally
+    // cached or not; and that this intention should be honored until a new
+    // session is requested through navigator.id.request.
+    // See https://github.com/mozilla/browserid/pull/2529
+    setRemoteOrigin(trans.origin);
+    // loggedInUser will be undefined if none of loaded, loggedInUser nor
+    // logout have been called before. This allows the user to be force logged
+    // out.
+    if (loggedInUser !== null) {
       storage.setLoggedIn(remoteOrigin, false);
+      loggedInUser = null;
       chan.notify({ method: 'logout' });
     }
   });
diff --git a/resources/static/include_js/include.js b/resources/static/include_js/include.js
index df83776..6edccf6 100644
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1254,7 +1254,7 @@
         // allocate iframe if it is not allocated
         _open_hidden_iframe();
         // send logout message if the commChan exists
-        if (commChan) commChan.notify({ method: 'logout' });
+        if (commChan) commChan.call({ method: 'logout', success: function() {} });
         if (typeof callback === 'function') {
           warn('navigator.id.logout callback argument has been deprecated.');
           setTimeout(callback, 0);
```

A complete PR with this diff can be found in #2916. Could you review that to make sure the behavior is consistent with your PR?

shane-tomlinson:
Closing in favor of #2916.

PR #2531, opened 2012-09-27
Use connect-fonts to serve up browser/locale specific fonts to reduce overall download size.

shane-tomlinson:
READY FOR INITIAL REVIEW & IDEA GENERATION

Work that still needs done:
- I was very confused by what static-resource-test.js was trying to do where I marked XXX. Help needed here.
- When I enable locales in config/aws.json and push to an ephemeral instance, a lot of errors are displayed that look like the svn repo is never fetched.


jedp:
It sounds like reducing the size of the font download will give us a huge win in speed.  I'm particularly interested in seeing this happen, because most b2g users will be using Persona across 2G data networks, where you can practically hear the individual bits as they sail lazily through the air.

lib/static_resources.js
```diff
@@ -113,16 +113,14 @@ var dialog_js = und.flatten([
   ]]);
 
 exports.resources = {
-  '/production/dialog.css': [
-    '/common/fonts/fonts_common.css',
-    '/common/fonts/fonts_dialog.css',
+  '/production/:locale/dialog.css': [
+    '/:locale/opensans-regular,opensans-light,opensans-bold/fonts.css',
```
ozten, line 117: This file currently has one asset per line. Please put a couple line returns in here at the commas.

lib/font_config.js
```diff
@@ -0,0 +1,99 @@
...
+ * https://github.com/shane-tomlinson/connect-fonts/blob/master/README.md
+ *
+ * Two fields are exported, fonts and locale_to_subdirs.
+ * fonts is the font configuration expected by connect-fonts,
+ * locale_to_subdirs returns a mapping of locale->directory where the
+ * fonts for a locale are located.
+ *
+ */
+
+const
```
ozten, line 20: This file is an adapter between **connect-fonts** and our app. This could just be part of `lib/configuration.js`.

Providing config/fonts.json is nice, but it can be picked up with the normal `CONFIG_FILES` mechanism, etc. We might want this so that you can have config/fonts-prod.json which matches l10n-prod.json.

Unless the design is for this font configuration to always have config for all locales and then the i18n sub-system enabels/disables locales. This is actually probably better, since there is one place to enable and disable locales. We just need to make adding fonts and font config part of the checklist for adding a new locale.

CC @mathjazz so he knows about this work.

mathjazz, line 20: ACK

shane-tomlinson, line 20: @mathjazz - It looks scary, but it's really not.  The default is to be very cautious on what gets served to users, if a font for a locale is not found or enabled, the full font set (what is currently served to all users) will be used.

@ozten - I see what you are saying about enabling/disabling locales in one place, but locales currently do not map 1:1 to directories. 2 locales could map to the same directory. For example, both hungarian and russian could use the cyrillic font set. We could force locales to be 1:1 with a directory and use symlinks from locale->font directory, but we have to worry about Windows developers. 

Additionally, there are no adverse effects if a font for a locale is enabled, but the locale itself is disabled for Persona. No extra bytes are served to any users. The locale mapping here is locale->directory.

There is the additional issue that it is fine for a locale's font to be enabled in the font server, but the locale itself be disabled in the Persona configuration. Additional configured font locales will have no adverse affect.

Maybe like you say, all this should be abstracted into a connect-fonts-opensans npm module and the configuration is not needed here. I am heavily leaning this way. Would you be willing to lend me an ear and some of your ideas to get that moving?

shane-tomlinson:
Due to @ozten's awesome comments, I am taking this down and re-thinking connect-fonts and how to make it better.

PR #2537, opened 2012-09-27
Avoid using GNUisms and BASHisms

grimreaper:
This should be a portable project and should run without relying on bash.

/bin/bash is always wrong
/usr/bin/env bash is sometimes correct
/bin/sh is always correct

### DIFF
--- a/docs/SETUP_FREEBSD.md
+++ b/docs/SETUP_FREEBSD.md
@@ -1,0 +1,14 @@
+Installing Dependencies on FreeBSD
+---------------------------------
+
+Run the following to install necessary dependencies:
+
+    cd /usr/ports
+    make -C www/node install clean
+    make -C math/gmp install clean
+    make -C www/npm install clean
+    make -C devel/git install clean
+
+or  with pkgng
+
+    pkg add  www/node math/gmp www/npm devel/git
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -31,7 +31,7 @@ Running tests with `npm test` will use a json database by default.  To
 test using MySQL, you will need to grant `browserid` privileges to
 create tables.  You can then run the mysql suite with, e.g.,
 
-```bash
+```sh
 NODE_ENV=test_mysql MYSQL_USER=browserid MYSQL_PASSWORD=browserid npm test
 ```
 
@@ -58,7 +58,7 @@ There are two test suites:
 By default the test runner will run them all. You can limit it to one
 suite by setting `WHAT_TESTS` in your environment.  For example:
 
-```bash
+```sh
 WHAT_TESTS=front npm test
 ```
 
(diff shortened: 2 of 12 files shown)


callahad:
I haven't reviewed, but I'm definitely +1 for this sort of change.

matthewp:
If the goal is portability we should also rewrite the Makefile in something non-posix.  Perhaps ShellJs (created by a Mozillian I believe): https://github.com/arturadib/shelljs

h4ck3rm1k3:
Bash and gnu are very portable, where dont they run?

grimreaper:
bash is not installed by default on FreeBSD. In addition it should never be installed in /bin on FreeBSD. I believe the same is true for { net, open, dragonfly } bsd.  Many operating systems prefer to not use GNU software and specifically avoid GNU extensions.

Maybe people prefer using zsh instead of bash, and may even deinstall bash.

bashisms are not portable. gnuisms are not portable. POSIX is generally portable (except if you care about windows).  please don't give into the embrace, extend, and extinguish model. ;)

h4ck3rm1k3:
umm, prefer is one thing. I am talking about portability not politics. Show me where gnu bash does not run or cannot be made to run.

seanmonstar:
@matthewp that is awesome! Being the only Windows dev on browserid, that's pretty important to me. I'll get on it so we use shelljs.

grimreaper:
h4ck3rm1k3: this has nothing to do with "can be made to run" - this about using _standards compliant_ shell scripting technique. Many operating systems do not install bash by default and users should not be required to do in order to run a simple shell script.

Show me one platform Internet Explorer can not be made to run with wine. Should we use IE specific code because "IE can be made to run" ?

Remember we are talking about portability, not politics: use standards compliant code, not bash extensions. (p.s. the use of ";)" indicates a joke)

h4ck3rm1k3:
Ok @grimreaper sounds fair to me. 
If you want to make an build system/configure that runs out of the box and does not need gnu, have you looked at cmake? 
we are talking about build scripts here? I am just learning about the js world, and come from the c world. 
Normally projects use automake/autoconf or cmake or scons to create the build system and it does this in a portable way. 
I have not reviewed all bash scripts but sure, writing bash scripts might not be a good idea if you have better tools for it, porting them to a javascript shell sounds funky but i dont know if it the right direction. I will need some time to get up to speed here. Just trying to help.
mike

h4ck3rm1k3:
Oh, now i see the commits, yeah that looks fine to me! I did not see that this was a pull but just jumped into the discussion about bash. 
mike

callahad:
If it ain't broke... ShellJS looks cool, and probably makes sense for some of our internal, administrative scripts, but I'm not sure if we'll be able to use it everywhere. I'd be in favor of merging these changes (pending review) and resolving this bug independent of a move to ShellJS.

seanmonstar:
@callahad Maybe not in every single script (though, eventually, why not?), but some of the scripts are indeed broken on Windows (test_backend, for example).

callahad:
@lloyd will review and merge this once we double-check that this totally works on our end.

Thanks for the pull request, @grimreaper !

lloyd:
code looks great, if anyone wants to verify the scripts execute fine on linux and macos we can hit the green button, otherwise I'll try to make time to manually test shortly.

lloyd:
tested, manually merging.  r+

PR #2543, opened 2012-09-28
Corrects the source order on the label/input password fields

icaaq:
The order of labels and input password fields are not optimal right now. If using a screenreader the output becomes

`label old password`
`label new password`
`input old password`
`input new password` 

which is confusing.

This pullrequest fixes so that the labels are more connected to the corresponding input field.

### DIFF
--- a/resources/static/pages/css/style.css
+++ b/resources/static/pages/css/style.css
@@ -304,10 +304,6 @@ button.delete:active {
   display: inline-block;
 }
 
-#edit_password input[type=password] {
-  width: 40%;
-}
-
 .showedit {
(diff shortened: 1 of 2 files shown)


seanmonstar:
Awesome, thanks. I'll take a look.

callahad:
As an aside, thank you @icaaq for filing these fantastic accessibility pull requests! Your contributions are extremely valuable.

icaaq:
@callahad Thanks!

seanmonstar:
yup, still looks correct. merged in. thanks again!

Issue #314, opened 2011-09-27
Saved passwords are not automatically filled in when logging in with BrowserID on Firefox 6


ghost:
OS: Windows 7 32-bit
Browser: Firefox 6.0 (Mozilla/5.0 (Windows NT 6.1; rv:6.0) Gecko/20100101 Firefox/6.0)

STR:
1. Open http://beta.myfavoritebeer.org/ in Fx6.
2. Click on the Sign in button.
3. Enter your credentials and click on the Sign In button.
4. Click on the "Remember password" button (displayed in a pop-up).
5. Select an email address and finish the login process.
6. Log out.
7. Click on the Sign in button from MyFavoriteBeer again, then click on the "This is not me" link.
8. Type the same email address in the corresponding textbox, then click on the Tab key (or click inside the password textbox).
   1. Click on the Cancel button.
   2. Click on the Sign in button from MyFavoriteBeer again.
   3. Type the same email address in the corresponding textbox, then press the Tab key (or click inside the password textbox).

Expected Results:
 The password is automatically filled in the corresponding field both at step 8 and steps 11.

Actual Results:
 The password is not automatically filled in no matter if I manually type in the email, select it from a drop-down, click inside the password field or press Tab.

ghost:
I should also mention I verified the Fx Password Manager and the password was actually saved.

ghost:
This issue reproduces on Firefox (6.0, 7.0.1) on multiple OSs (Win XP, Win 7, Mac 10.6, Android). It works fine on IE9.

shane-tomlinson:
@jbonacci -

https://bugzilla.mozilla.org/show_bug.cgi?id=355063 outlines a Firefox bug that shows that authentication content added to the DOM after DOMContentLoaded does not receive any love from the password manager.

This leads me to think I should pre-fill the form with the authentication content so that we do not have this problem.

shane-tomlinson:
There are several bugs in Bugzilla related to this:

https://bugzilla.mozilla.org/show_bug.cgi?id=583578 - Prevents the password field from being updated with required email.

https://bugzilla.mozilla.org/show_bug.cgi?id=670137 - Closes the "remember password" dialog when a window is opened (or closed as is our case)

https://bugzilla.mozilla.org/show_bug.cgi?id=631800 - Prevents passwords from being saved if originally entered in a popup window.

shane-tomlinson:
[...]

PR #2556, opened 2012-10-05
Password manager auto-fill works in Firefox!

shane-tomlinson:
- Move the email/password form to always be in the DOM. This is a requirement for the Firefox password manager to auto-fill passwords.
- Update all the unit tests.
- Remove the authenticate template, move the contents into dialog.ejs

Open issues:
- There are multiple HTML elements with the "email" and "password" IDs. This will require these elements to be renamed in either the authentication screen or in the other screens. Changes to these IDs will require changes to bidpom.

issue #314


shane-tomlinson:
@6a68, @davehunt, @zacc - Can you weigh in on the process necessary to change element IDs in bidpom?

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -174,12 +182,13 @@ BrowserID.Modules.Authenticate = (function() {
       lastEmail = options.email || "";
 
       var self=this;
-      self.renderDialog("authenticate", {
-        siteName: options.siteName,
-        email: lastEmail
-      });
+
+      dom.addClass(BODY_SELECTOR, AUTHENTICATION_CLASS);
+      dom.setInner(RP_NAME_SELECTOR, options.siteName);
```
seanmonstar, line 187: The `siteName` is escaped before somewhere?

shane-tomlinson, line 187: @seanmonstar - https://github.com/mozilla/browserid/blob/71341882496bf4573335a0f96693b40e15bc2ee9/resources/static/dialog/js/modules/dialog.js#L228

seanmonstar, line 187: I see it, its escaped in `dialog.get()` at the beginning of the request.

Issue #2561, opened 2012-10-08
internal api needs logout and logoutEverywhere functions.


shane-tomlinson:
B2G and other internal/native implementations that make use of the API need a way of logging the user out of a specific site as well as a general logout function.

shane-tomlinson:
link to #1416

lloyd:
Lets include the motivating use case in issues, it helps people understand and prioritize.

-- lloyd (thumb-typing)

seanmonstar:
Motivating case is that it seems the B2G dialog will use the internal API, and there's currently no internal logout. So, priority is like 9000.

lloyd:
Cool, Ty @seanmonstar!

So this is for the dialog to sign in at time of first app purchase? Or for dialog when signing in to websites? Or both?

What are desired semantics of logout?  Logs you out locally?  Form the current site?  

How about logoutEverywhere?  Does that log you out on other devices?  All sites on current device?  Destroy your session with persona services?

Meta question is how do functions designed for internal use or use by sites on the web map into b2g...

-- lloyd (thumb-typing)

shane-tomlinson:
@lloyd - you are not supposed to be working today.

lloyd:
Yeah man.  But my baby is sleeping on me.  Also, I think it is critically important we focus on describing high level product requirements in bug reports, rather than implementation details.  So I guess I'm on a crusade here :)

-- lloyd (thumb-typing)

shane-tomlinson:
@lloyd - 

These are both thin wrappers around their user.js equivalents. For now, these will act exactly like they do in the normal web flow.

Calling BrowserID.internal.logout will log the user out of the specified origin on the current device. When you log out of a site on one device, it has no effect on another device. This has never been synced across devices.

Calling BrowserID.internal.logoutEverywhere clears your session and marks all sites as logged out on the current device - this too has never been synced across devices. A better name for this may be logoutAllSites.

The meta question will be hammered out further tomorrow with @benadida. The internal API is going to be updated, driven by high level requirements (some of which are outlined in the etherpad that you and @anantn wrote 6 months ago) In theory, I would like the internal API to provide an interface similar to navigator.id so there is minimal friction for developers embedding the dialog in web views.

seanmonstar:
This was fixed!

PR #2562, opened 2012-10-08
Issue 2561 internal api logout

shane-tomlinson:
This is a two commit PR - when reviewing the code it will be less scary to look at these in their separate parts.

1) test cleanup.
2) add the BrowserID.internal.logout and BrowserID.internal.logoutEverywhere.

issue #2561


benadida:
this looks awesome. r+. Shane, I will let you merge as I'm staying away from production code from now on :)

PR #2564, opened 2012-10-09
Instrument the dialog to measure startup time

shane-tomlinson:
A series of commits to start to instrument the dialog startup to find inefficiencies.


resources/static/common/js/browserid.js
```diff
@@ -7,8 +7,7 @@
 
   window.BrowserID = window.BrowserID || {};
 
-  // Define some constants.
-  _.extend(window.BrowserID, {
+  var bid = BrowserID, constants = {
```
seanmonstar: I don't see why the change is needed. Still, even if kept, the `bid` var is never referenced again.

shane-tomlinson: I have moved this to be the very first script of the set, before any libraries. This is so I can get the exact time before any other scripts are parsed. Upon completion of script parsing, start.js is run. I can take the diff between parse end and parse begin to find the total parse time. If we start shaving script bytes and realize gains in parsing, I would like to be able to quantify how much time we are saving just in parsing.

seanmonstar: I see, you're removing the need for underscore. However, when this is compressed into a single script, won't it have to parse the whole thing before it executes any of it?

shane-tomlinson: I am unsure of how the browsers handle running code while parsing. javascript_parse may in fact not be the parse time, but the amount of time it takes to run all the statements in between the top of the script and the bottom. I need a better understanding of how browsers handle this.

At first glance, the NavigationTiming (https://dvcs.w3.org/hg/webperf/raw-file/tip/specs/NavigationTiming/Overview.html) API does not seem to offer any solutions for this either.

seanmonstar: No, It doesn't. I assume it has to parse before executing, since it must do function hoisting.

A for-sure time would be a new `script` tag, above the main one, that only has `JS_START_TIME = new Date()` in it, and never gets combined into the rest.

seanmonstar: However, that might be tainted by the download time of the main script.

fmarier:
I left two small comments in the commits, but overall it looks good to me.

I guess the front-end tests that failed on Travis need to be fixed before merge.

shane-tomlinson:
@fmarier - Added the note that enabling measuring the dom loading time breaks CSP.

The concern about the script in the HEAD not being wrapped I believe is because you were looking at an old version of the code.

And finally, fixed the unit test. This error was due to us changing all the exceptions to be "throw new Error" based instead of "throw <string>" based on dev, when I rebased I did not update this exception. This is now updated.

seanmonstar:
CSP doesn't allow inline scripts?

fmarier:
@seanmonstar If you want inline scripts or styles in CSP, you have to allow "unsafe-inline" which kinda defeats the purpose of CSP (protecting against an attacker's inline scripts).

fmarier:
Not sure why Travis is still marked as failed (maybe it hasn't run since Shane's last update?), but r+ from me.

PR #2572, opened 2012-10-10
upgrade lockdown to 0.0.3

seanmonstar:
last issue for #1769

@zaach At first I replaced the `preinstall` to just be the command `lockdown`, but if you start from a fresh install, `lockdown` doesn't exist. So, you almost need to include the script locally, but that seems ugly. Possible alternative was `npm install lockdown`, and then `npm install`...

### DIFF
--- a/lockdown.json
+++ b/lockdown.json
@@ -7,6 +7,9 @@
   },
   "active-x-obfuscator": {
     "0.0.1": "089b89b37145ff1d9ec74af6530be5526cae1f1a"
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@seanmonstar - can you rebase this?

seanmonstar:
rebased!

zaach:
r+. We may change the way lockdown bootstraps in a later version.

PR #2573, opened 2012-10-10
change base module to baseExceptions

seanmonstar:
@ozten this is the change to `baseExceptions`. If that sounds like a good name, we can merge into your branch, and then merge yours in dev.


seanmonstar:
merge target is #2449

bin/browserid
```diff
@@ -4,7 +4,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-require('../lib/base.js').addExceptionHandler();
+require('../lib/baseExceptions.js').addExceptionHandler();
```
ozten, line 7: Cool!

PR #2576, opened 2012-10-10
fixes for testsetup, etc.

jrgm:
- fix the backward order of arguments to wd.prototype.newSession. (We weren't
  actually running any browser but the default FF/Vista one as a result. Now
  that we are trying to run Windows, we don't seem to be able to. But that's
  next to fix.)
- fix some accidental globals and shadows
- use underscore.extend to reduce boilerplate
- remove unused testSetup.mergeOpts method
- node-inspector doesn't like one-liner functions


jrgm:
Oops. Sorry, I shouldn't have branched again for the second set of fixes.

PR #2579, opened 2012-10-10
we throw Errors here, nothing else

seanmonstar:
give us stack traces!

_Note_: This doesn't include library code, like WinChan or bidbundle. That will come later.


ozten:
+1000 thanks for doing this.

zaach:
r+, wooo!

PR #2584, opened 2012-10-11
Fix issues with internal.get API

### LINKED ISSUE #2580: internal.get fails if modules not registered yet
We noticed this when we tested on siginifically slower awsboxes, that waiting for the window.load event isn't enough before you call `internal.get`.

Instead of making everyone else aware of this, `internal.get` should wait till the `dialog` module is ready before trying to use it.

seanmonstar:
1. On a slower domain, like an awsbox, there was a timing issue. They didn't know when it was safe to call `internal.get`, since the `dialog` was not ready until after the `session_context` returned from the cookie check.
2. B2G was passing in a `WrappedJSObect` for options, which we then tried to append some properties to as defaults. This caused permission errors, so now we create a local copy using `_.extend()`.

@shane-tomlinson Feel free to tell me that the polling is dirty (it feels a bit dirty), but these commits are currently on https://b2g.personatest.org and working for them.

fixes #2580

### DIFF
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -67,7 +67,10 @@
       callback && callback(assertion || null);
     }
 
-    options = options || {};
+    // Make our own copy, since we assign properties to it.
+    // There was an error doing that in B2G, since we got a
+    // WrappsJSObject.
+    options = _.extend({}, options);
 
     var silent = !!options.silent;
     if(silent) {
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -85,16 +85,24 @@
     else {
       // Use the standard dialog facilities to get the assertion, pass the
       // options block directly to the dialog.
-      var controller = moduleManager.getRunningModule("dialog");
-      if(controller) {
-        options.rp_api = "internal";
-        controller.get(origin, options, complete, complete);
-      }
-      else {
-        complete();
-      }
+      options.rp_api = "internal";
+      get(origin, options, complete);
     }
   };
+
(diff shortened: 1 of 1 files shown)


zaach:
r+. On production there were also timing issues when trying to log in a second time.

shane-tomlinson:
@seanmonstar - I agree that the polling feels dirty, mostly because it feels like it ignores the root problem.

The issue we have is timing:

1) When the dialog javascript is loaded, the cookie check is done.
2) The cookie check calls session_context.
3) While session_context is out, window.onload is triggered.
4) B2G calls BrowserID.internal.get
5) Since session_context is still out, the dialog module has not yet been registered.
6) BrowserID.internal.get fails because the dialog module is not available.

Digging into network.js->cookiesEnabled, the entire cookie check is surrounded by a "withContext({..});" Above the withContext call, there is a comment "Make sure we get context first or else we needlessly send a cookie to the server" which is from checkin cc39d218.

That comment is important because withContext added with the original cookie check when we set the max-age of the cookie and let it expire. Calling session_context first was an optimization to avoid sending the "test" cookie to the server.

We are now immediately expiring the cookie so it is never sent to the server. withContext's original purpose is taken care of by other means.

On the surface, we could remove the call to session_context which makes the cookieCheck synchronous, but the function signature is still asynchronous. Removing the call to session_context would cause cookieCheck to terminate immediately and in theory start the remainder of the dialog initialization faster than currently happens. 

Since the initial dialog load and cookie check are asynchronous, there is no simple way of knowing when the dialog is ready. We do not have a "ondialogready" event or anything like that.

A poll here is suitable. Adding a comment for the reasons in a separate commit.

Also issuing a PR to simplify the cookie check based on the above comments.

PR #2588, opened 2012-10-12
Fix ephemeral instances serving of static resources

### LINKED ISSUE #2587: ephemeral deployments not serving static resources
With change 1173c2d7c173e9deabe9209b4136006935f1bfaa we broke serving of static resources on ephemeral instances.  Because ephemeral aws deployments use `production.json` then overlay `aws.json`, the addition of `public_static_url` to the former caused ephemeral instances to serve static resources from `https://static.login.persona.org`.

lloyd:
This change adds an overlay to `aws.json` to _undefine_ `public_static_url` which causes the fallback logic to work as it should, and static resources to be served from the dynamic name of the ephemeral instance.

BONUS: update to awsbox 0.2.18 to get latest security patches built in and speed up VM creation by ~200s.  

closes #2587


lloyd:
@seanmonstar you want this one?  I think you have context here...

PR #2621, opened 2012-10-18
Add a timeout when trying to fetch a well-known. Fixes #2603.

### LINKED ISSUE #2603: "We are sorry, this request is taking to long" is displayed every time for '@yahoo.co.uk' email addresses
Browser: FF 16.0.1, Chromium 20.0.1132.47
OS: Ubuntu 12.04
1. Go to www.beta.123done.org or www.beta.myfavoritebeer.org and press on Sign in with Persona
2. In the sign in dialog enter any '@yahoo.co.uk' email address and press on next. 

Expected: 
"Confirm your email address" message  pops up

Actual:
"We are sorry, this request is taking to long" message is displayed every time for a few seconds before you are ask to "Confirm your email address"

I only saw this for 'yahoo.co.uk' email addresses

jaredhirsch:
See discussion in #2603 of how SSL connections to some IdPs, e.g. yahoo domains of the form https://yahoo.co.uk, hang forever.

There's a 15 second timeout somewhere in squid or node, but the front-end dialog code looks like it shows the "this is taking a while" error message after 10 seconds.

So, to strike a balance between letting slow IdPs negotiate SSL connection and serve the well-known in time, but also forcing fallback to happen in a timely manner, I pegged the timeout at 8 sec.

I'm not sure how to easily test this, but I could be convinced it's worth spending time figuring out how to do it. My immediate thoughts are: hack the request object to always fire a timeout, monkey-patch the clock to make it look like the timeout has elapsed, or use a wrapper around node to really hold the socket open long enough to cause a timeout (don't like this idea, seems slow and hacky for a unit test).

### DIFF
--- a/lib/primary.js
+++ b/lib/primary.js
@@ -13,6 +13,7 @@ logger = require('./logging.js').logger,
 urlparse = require('urlparse'),
 jwcrypto = require("jwcrypto"),
 config = require("./configuration.js"),
+primaryTimeout = config.get('declaration_of_support_timeout_ms'), 
 secrets = require("./secrets.js");
 
 // alg
@@ -206,7 +207,7 @@ var getWellKnown = function (domain, delegates, cb) {
     req.abort();
     logger.debug('timeout trying to load well-known for ' + domain + ': ' + e.toString());
     handleProxyIDP();
-  }, config.get('declaration_of_support_timeout_ms'));
+  }, primaryTimeout);
   req.on('response', function() { clearTimeout(reqTimeout) });
 
   req.on('error', function(e) {
--- a/lib/primary.js
+++ b/lib/primary.js
@@ -206,7 +206,7 @@ var getWellKnown = function (domain, delegates, cb) {
     req.abort();
     logger.debug('timeout trying to load well-known for ' + domain + ': ' + e.toString());
     handleProxyIDP();
-  }, 8000);
+  }, config.get('declaration_of_support_timeout_ms'));
   req.on('response', function() { clearTimeout(reqTimeout) });
 
   req.on('error', function(e) {
--- a/lib/primary.js
+++ b/lib/primary.js
@@ -201,7 +201,16 @@ var getWellKnown = function (domain, delegates, cb) {
     }, handleResponse);
   }
 
+  // front-end shows xhr delay message after 10 sec; timeout sooner to avoid this
+  var reqTimeout = setTimeout(function() {
+    req.abort();
+    logger.debug('timeout trying to load well-known for ' + domain + ': ' + e.toString());
+    handleProxyIDP();
+  }, 8000);
+  req.on('response', function() { clearTimeout(reqTimeout) });
+
   req.on('error', function(e) {
+    if (reqTimeout) { clearTimeout(reqTimeout); }
     logger.debug(domain + ' is not a browserid primary: ' + e.toString());
     handleProxyIDP();
   });
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -225,8 +225,9 @@ var conf = module.exports = convict({
   process_type: 'string',
   email_to_console: 'boolean = false',
   declaration_of_support_timeout_ms: {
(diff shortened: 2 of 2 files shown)


lloyd:
is it better to address this in the application (this pull request) or in squid? /cc @gene1wood

jaredhirsch:
Good question. I was wondering if this was too shallow a fix; sounds like I
should keep digging.

gene1wood:
@lloyd In the app. Conceptually this should be defined by the app since this is driven by desired user experience, instead of by the proxy which should have it's logic driven by desires for security. Also, squid can't do complex business logic related to timeouts, where you might want to get fancy with timeout behaviour down the road based on the target domain or something.

gene1wood:
Another approach could be to begin recording/keeping track of domains and their last known well-known value so we could make more intelligent decisions about things based on the age of our local record about a domain and the value. We could use this to prevent potential flapping of a primary if it sits on the timeout threshold. We could also set internal caching behaviour based off of previous responses from domains for their well-known file (e.g. A domain that previously didn't have a well-known file, maybe we only check once a day whereas an active primary we check more often to make sure they haven't changed their mind). We could also extend the timeout for domains which have ever been a primary since they probably have more intention about serving the well-known file than some other domain that's never before served a well-known file.

lloyd:
I don't believe req.setTimeout will do what you want @6a68.  This does not give you an end-to-end timeout for the request, but rather "Sets the socket to timeout after timeout milliseconds of inactivity on the socket."

There are many cases where this timeout can be _much_ longer than 8 seconds.  Suggest an application level timeout as was done in `http_forward.js` for very similar reasons.

jaredhirsch:
thanks @lloyd for the feedback, will fix. Shall I close and resubmit, or just update this PR?

lloyd:
@6a68 fielders choice

jaredhirsch:
updated PR per @lloyd's suggestions, waiting for further input on whether/what to unit test. 

@shane-tomlinson, if you have ideas around testing a request timeout, I'd love to hear 'em :-)

lib/primary.js
```diff
@@ -201,7 +201,16 @@ var getWellKnown = function (domain, delegates, cb) {
     }, handleResponse);
   }
 
+  // front-end shows xhr delay message after 10 sec; timeout sooner to avoid this
+  var reqTimeout = setTimeout(function() {
+    req.abort();
+    logger.debug('timeout trying to load well-known for ' + domain + ': ' + e.toString());
+    handleProxyIDP();
+  }, 8000);
+  req.on('response', function() { clearTimeout(reqTimeout) });
```
jrgm, line 211: Magic number! We should make that configurable.

lloyd, line 211: +1 to making that configurable

jaredhirsch:
Sounds good. Not totally sure where to put the parameter--maybe config/{local, production, ephemeral}.json?

seanmonstar:
Config goes in `lib/configuration.js`. Look at some of the others to give it a reasonable default, and an ENV var to overwrite.

jaredhirsch:
Thanks, @seanmonstar. Looking at lib/configuration.js, I noticed that there's an existing config parameter, declaration_of_support_timeout_ms, that is used by our dev proxy, and looks like it does the same thing--but I'm not completely sure.

jaredhirsch:
bump! @seanmonstar, @lloyd, mind taking another look at this five-star?

lib/primary.js
```diff
@@ -201,7 +201,16 @@ var getWellKnown = function (domain, delegates, cb) {
     }, handleResponse);
   }
 
+  // front-end shows xhr delay message after 10 sec; timeout sooner to avoid this
+  var reqTimeout = setTimeout(function() {
+    req.abort();
+    logger.debug('timeout trying to load well-known for ' + domain + ': ' + e.toString());
+    handleProxyIDP();
+  }, config.get('declaration_of_support_timeout_ms'));
```
seanmonstar: nit: lift this into a `const` at the top, so we only ask config once.

PR #2623, opened 2012-10-19
2565 dynamic wellknown

warner:
This adds the domain= queryarg. The code still needs tests before being ready to merge.
/cc @ozten for help.


lib/primary.js
```diff
@@ -188,26 +189,26 @@ var getWellKnown = function (domain, delegates, cb) {
     req = http.get({
       host: httpProxy.host,
       port: httpProxy.port,
-      path: 'https://' + domain + WELL_KNOWN_URL,
+      path: 'https://' + fromDomain + WELL_KNOWN_URL + "?domain=" + principalDomain,
```
ozten, line 194: It would be okay if we didn't have the domain query string parameter when principalDomain and fromDomain have the same value.

I don't think it hurts to be consistent and always send the domain query string parameter.

lib/primary.js
```diff
@@ -65,16 +65,16 @@ function parseWellKnownBody(body, domain, delegates, cb) {
...
         return cb("Too many hops while delegating authority " + JSON.stringify(dels));
       }
-      logger.debug(domain + ' is delegating to ' + v[k]);
+      logger.debug(fromDomain + ' is delegating to ' + v[k]);
       // recurse into low level get /.well-known/browserid and parse again?
       // If everything goes well, finally call our original callback
-      delegates[domain] = dels.length;
-      getWellKnown(v[k], delegates, function (err, nbody, ndomain, ndelegates) {
+      delegates[fromDomain] = dels.length;
+      getWellKnown(v[k], principalDomain, delegates, function (err, nbody, newDomain, ndelegates) {
```
ozten, line 74: Nice, seperation of fromDomain, newDomain, and principalDomain makes this clearer.

ozten:
I don't see why the httpProxy isn't happy, but in this branch

```
curl http://192.168.186.138:10002/wsapi/address_info?email=eozten%40delegat.es
{"type":"secondary","known":false}
```

Which isn't right. On dev is resolves to 

```
{"auth":"https://mockmyid.com/browserid/sign_in.html","prov":"https://mockmyid.com/browserid/provision.html","type":"primary"}
```

warner:
@lloyd or @zaach , does that bin/proxy change seem reasonable?

lloyd:
the bin/proxy change looks reasonable to me.  We should audit squid's configuration in stage and see if there is anything to be done there.  I believe you can check to see if squid is behaving properly by jumping into the staging env and using curl.

ozten:
Okay, I've added to tests/auth_with_assertion.js. You can run this test with 

```
./node_modules/.bin/vows tests/auth-with-assertion-test.js
```

It currently fails with

```
·{ success: false, reason: 'issuer 'bigtent.domain' may not speak for emails from 'yahoo.com'" }
```

I've also merged dev in to keep this up to date. Hopefully I've got @6a68 timeout patch in there correctly as there was a conflict.

When you run the test, you'll also want to tail the logs under var/logs. It's weird that yahoo.com is reported as having primary support... I think that is reported in terms of 127.0.0.1:10002 being a secondary (?).

ozten:
I don't see the code in lib/static/views.js which renders the well-known file being run during unit tests, nor manual testing. I guess this is expected, since the shim and verifier short circuit and using config to hand off to bigtent.

lib/primary.js
```diff
@@ -244,19 +247,19 @@ exports.checkSupport = function(domain, cb, delegates) {
       }
 
       try {
-        parseWellKnownBody(body, domain, cbdelegates, function (err, r) {
+        var r = parseWellKnownBody(body, newDomain, principalDomain, cbdelegates, function (err, r) {
```
warner: where does this 'r' get used? Seems confusing to have it overlap with the inner function argument.

seanmonstar:
Just wanted to point out dev now includes tests against jshint. You may want to rebase against dev.

ozten:
Awesome, good point. Thanks!

ozten:
r+ me and warner

rebase merged in [commit aaaa](https://github.com/mozilla/browserid/commit/aaaaf6c4d5089e5a2a6fae6c49ad12c932b861e9)

Issue #2628, opened 2012-10-23
upgrade to node v0.8.12 in dev


lloyd:
motivated and discussed on mailing list:

https://groups.google.com/forum/?fromgroups=#!topic/mozilla.dev.identity/st_eL7kUpUw

and

https://groups.google.com/forum/?fromgroups=#!topic/mozilla.dev.identity/GtvnAk4cjHs

PR #2630, opened 2012-10-23
Upgrade to node v0.8.12

lloyd:
This updates node in dev to v0.8.12 - issue #2628

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     },
     "devDependencies": {
         "vows": "0.5.13",
-        "awsbox": "0.2.18",
+        "awsbox": "0.3.1",
         "irc": "0.3.3",
         "jshint": "0.7.1",
         "minimatch": "0.2.6",
--- a/package.json
+++ b/package.json
@@ -54,6 +54,6 @@
         "start": "node ./scripts/run_locally.js"
     },
     "engines": {
-        "node": ">= 0.8.12"
+        "node": ">= 0.8.11"
     }
 }
--- a/package.json
+++ b/package.json
@@ -54,6 +54,6 @@
         "start": "node ./scripts/run_locally.js"
     },
     "engines": {
-        "node": ">= 0.6.17"
+        "node": ">= 0.8.12"
     }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ before_install:
  - "mysql -e 'create database browserid;'"
 
 node_js:
- - 0.6
  - 0.8
 
 notifications:
--- a/lockdown.json
+++ b/lockdown.json
@@ -7,7 +7,7 @@
   },
   "active-x-obfuscator": {
     "0.0.1": "089b89b37145ff1d9ec74af6530be5526cae1f1a"
-	},
+  },
   "addressparser": {
     "0.1.3": "9e9ab43d257e1ae784e1df5f580c9f5240f58874"
   },
@@ -21,7 +21,7 @@
     "0.0.5": "971b8995078d83c80f2372f134c496e71b293a46"
   },
   "awsbox": {
-    "0.2.18": "7cfebaa0c21de2e054d02a8ced97d021124fe962"
+    "0.3.1": "bbd433489d7d73c8b38c4f60f39ab6b034e73584"
   },
   "bcrypt": {
     "0.7.1": "923e2623331211adcab6ac84ec4fcd41713e1e58"
@@ -331,8 +331,8 @@
   },
   "xoauth2": {
     "0.1.3": "4448cb94fb1032c02cb2c81e95008674dcd2d1c6"
-  },	
+  },
   "zeparser": {
     "0.0.5": "03726561bc268f2e5444f54c665b7fd4a8c029e2"
-	}
+  }
 }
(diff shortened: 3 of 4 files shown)


gene1wood:
Services Ops yum repositories don't currently have nodejs-0.8.12 in them. The method that we use to build nodejs into an rpm as detailed here ( https://intranet.mozilla.org/Services/Ops/NodeJS ) is to use this spec file : https://github.com/kazuhisya/nodejs-rpm/blob/master/nodejs.spec

Would you use this to do your testing (a nodejs 0.8.12 install built using that RPM spec)?

I'm not sure how this should go from a workflow and responsibility standpoint (service ops build the rpm and send it to dev, dev build and test and send to ops, etc)

Also, this reveals a potential risk in the current package that if we were to add a new version of nodejs into mrepo, the next time a browserid machine is built it would pull the newest nodejs version (since there's no explicit version indicated in the browserid-server spec. It's probably good that we define an explicit version as you've done @lloyd

lloyd:
Hmm.  Until we have the integration cluster we probably will use completely distinct means of installing 0.8.12.  We should add a task to that crew to figure out how to use your bits.

For this time around are you willing to build the rpm, @gene1wood ?  Is there any risk of us affecting prod by putting this into mrepo?  Shall we hotpatch explicit versioning into prod before that?

gene1wood:
Searching through puppet it appears that the only hosts that would be impacted by putting a newer nodejs in mrepo would be our client[0-9][0-9] servers. They have some qa_loadtesting setup related to marteau ( https://github.com/mozilla-services/marteau ) that installs nodejs and doesn't constrain the version.

@jrgm might be able to speak to the impact of upgrading nodejs on these systems

@lloyd I'm fine with building the new nodejs rpm and adding it to mrepo once we get a go ahead from @jrgm on these "clientXX" machines.

jrgm:
Really? Marteau uses nodejs? Looks pure python to me, but I'm not really familiar with it.

Otherwise, I use node for browserid/bin/load_gen load tests, but upgrading to 0.8 is what I'd want for that.

package.json
```diff
@@ -54,6 +54,6 @@
         "start": "node ./scripts/run_locally.js"
     },
     "engines": {
-        "node": ">= 0.6.17"
+        "node": ">= 0.8.11"
```
shane-tomlinson, line 57: The PR is for node v0.8.12, but this is saying v0.8.11 is OK.

shane-tomlinson:
@lloyd - When upgrading from v0.6.17 to v0.8.11, running npm install, and then running the backend tests, the tests segfaulted on the bcrypt-compatibility-test. I _believe_ this is due to compiling node v0.8.11 with a different version of XCode than what was originally used to compile the bcrypt or jwcrypto native extensions. After I removed the npms and forced a recompilation of native extensions, the tests passed.

This points out that some developers who have recently upgraded their x-code may have to do this as well.

```
[12:44][browserid-remotes(dev)] $ ps-test-backend 
~/development/browserid-remotes ~/development/browserid-remotes
tests/add-email-with-assertion-test.js
·················· 
✓ OK » 18 honored (4.623s)
tests/auth-with-assertion-test.js
·········· 
✓ OK » 10 honored (4.220s)
tests/bcrypt-compatibility-test.js
node(39918,0x7fff76d55180) malloc: *** error for object 0x100a274c0: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
./scripts/test_backend: line 22: 39918 Abort trap: 6           vows $file
~/development/browserid-remotes
[12:45][browserid-remotes(dev)] $ 
```

package.json
lloyd, line 57: I lowered the range to allow us to continue to work on travis-ci, which is still at 0.8.11.

My view is that package.json expresses the node.js version you must have to develop on the software, while browserid.spec defines the precise version we'll use in production.  It does mean we're expressing ranges in two places, but this isn't us repeating ourselves because the ranges _mean_ different things.

lloyd:
@shane-tomlinson My understanding is that your node_modules must be installed and built using the same version of node that they will run under.  This is even true for minor version differences.

I believe you saw a segfault because you npm installed using `0.6.17` (linking binary node modules against `0.6` headers), then ran underneath `0.8.11`

I understand you ran npm install _after_ installing `0.8.11`, but that doesn't rebuild binary modules.  `npm rebuild` would, in fact, rebuild binary modules and should have addressed your problem.

summary: I think this is a known limitation with node (no binary compatibility in modules), I do not believe this is related to different x-code versions, and I think there's nothing to be done here.

sane analysis?

package.json
gene1wood, line 57: Good catch @shane-tomlinson I kept reading 0.8.12 in this ticket but swearing I'd seen 0.8.11 somewhere.

gene1wood:
@jrgm sounds ok with this so I'm happy.

I'll build and inject the new node version and then merge this

seanmonstar:
Doesn't our README say to first `rm -rf node_modules` before running `npm
install` after a pull?

gene1wood:
I've uploaded the nodejs 0.8.12 to mrepos : https://bugzilla.mozilla.org/show_bug.cgi?id=805090

gene1wood:
r+

Issue #2629, opened 2012-10-23
[dev] dialog stuck at "Connecting to Persona..." in chrome


lloyd:
STR:
1. visit dev.123done.org
2. click the button
3. notice the dialog is stuck

cannot reproduce on firefox, but occurs on recent chrome both in OSX and linux

lloyd:
I bisected this issue and found a the offending commit: 

```
3e49e6f06060e7ff03db1316f49644808cde88e7 is the first bad commit
commit 3e49e6f06060e7ff03db1316f49644808cde88e7
Author: Shane Tomlinson <stomlinson@mozilla.com>
Date:   Tue Oct 16 13:30:06 2012 +0100

    Remove withContext from the cookie check to speed up dialog startup.

    withContext will be still called when an XHR request needs the context.

    Updated the unit tests to make use of network.init({ cookiesEnabledOverride: xxx }); instead of     transport.setContextInfo("cookies_enabled", xxx); The context info's cookies_enabled field was never actually checked anywhere.

:040000 040000 9294495a9c46977250bd848d82a7717f28c96ebf 0fe60b9e41f3437058baa485ef00c1a5d77c106e M  resources
```

lloyd:
in chrome, there is a console error: "Uncaught TypeError: Cannot read property 'd' of undefined"

lloyd:
I've dug down into this issue.  It seems that the change @shane-tomlinson made change our load timing (of course).  It further seems like winchan.js, subsequent to invocation of 'onOpen' expects a single message to be posted to it.  Further, it seems like browserify uses postMessage to implement a browser happy version of `.nextTick()` so nodejsy code can run in your browser.

So this change causes a browserify message to be posted, that gets caught by winchan after invocation of onOpen, that causes winchan to puke, because the message is not what it is expecting.  And this only manifests on chrome with minified resources.

lloyd:
fixed in dev.

benadida:
oh man, I can see why @lloyd vomited.

@lloyd is there a way to look at the postMessage event source to make this judgment call rather than the slightly less robust check you're doing? Not a blocker, just a suggestion.

lloyd:
On Oct 24, 2012, at 4:59 PM, Ben Adida notifications@github.com wrote:

> oh man, I can see why @lloyd vomited.
> 
> @lloyd is there a way to look at the postMessage event source to make this judgment call rather than the slightly less robust check you're doing? Not a blocker, just a suggestion.
> 
[...]

PR #2631, opened 2012-10-23
issue #2629 - fix winchan to allow others to use postMessage without bre...

lloyd:
...aking (specifically, drop messages which are not properly formed 'initial request' messages from our spawner)

### DIFF
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -196,13 +196,17 @@
         }
 
         function onMessage(e) {
-          // only one message gets through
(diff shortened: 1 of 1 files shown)


lloyd:
got @shane-tomlinson r+ in IRC.

seanmonstar:
Maybe we could work in checking `e.source` vs the window we expect to be talking to?  Perhaps comparing it against `window.opener`?

lloyd:
This would be better.  Careful to test on ie.  also, we can make "request" more unique.  

-- lloyd (thumb-typing)

Issue #2624, opened 2012-10-21
[dev] signin with primary email address not working in dev


jrgm:
STR
1. Load http://dev.123done.org and click signin.
2. In the dialog, enter an @eyedee.me email address.
3. When prompted, click 'sign in with eyedee.me'.
4. On eyedee.me, enter a password and click 'Sign in'.

The dialog returns to /sign_in#AUTH_RETURN, but doesn't complete signin.
Eventually, the error screen is shown with 'Action: Checking Email Address'.

This is working fine in stage and prod.

ozten:
I can reproduce, following the STR.

https://eyedee.me/api/whoami

```
{"user":"ozten"}
```

I don't see the next expected call, a POST to eyedee.me/api/cert_key

Interestingly, **no repro** if I log into eyedee.me directly first.

Unable to troubleshoot with Chrome/Safari Dev Tools, getting "Uncaught TypeError: Cannot read property 'd' of undefined " dialog.js line 3.

ozten:
Eyedee.me always uses the SHIM from https://browserid.org. Is this a valid test case? Maybe this isn't failing as soon as it did before?

lloyd:
taking

lloyd:
NOTE: @ozten 's comment about "cannot read property 'd' of undefined" was a separate five star, issue #2629, now fixed in dev.

Thinking the easiest way to move forward on this one is to bisect and isolate the commit where the breakage was introduced.

lloyd:
from deep in the bowels... 

```
error: "runtime_error"
id: 264228
message: "{"message":"generator isn't seeded"}"
```

This may mean that the codepath used in #AUTH_RETURN is not seeding the random number generator.  This may be related to load time optimizations, as I believe that the session_context call seeds said generator...

PR #2633, opened 2012-10-23
Fix primary support on dev

lloyd:
A fix for issue #2624.

This PR should fix primary support on dev.  Specifically, our support was broken during key generation and certification post in-dialog authentication.  This likely introduced with 3e49e6f06060e7ff03db1316f49644808cde88e7 where we removed a call to session_context at dialog startup.  

A side effect of calling session_context is that it gets server supplied entropy to seed jwcrypto's random number generator.  With shane's optimization, apparently session_context was getting called before generateKeypair for all codepaths except `#AUTH_RETURN` - the path where we return from the IdP after an in-dialog redirect.

This PR fixes the problem by calling `session_context` during provisioning if it has not yet been called, to ensure the PRNG is seeded.

### DIFF
--- a/resources/static/common/js/provisioning.js
+++ b/resources/static/common/js/provisioning.js
@@ -4,7 +4,9 @@
 BrowserID.Provisioning = (function() {
   "use strict";
 
-  var jwcrypto = require("./lib/jwcrypto");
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thank you for the excellent description of the problem. Looks good here. r+

PR #2634, opened 2012-10-23
Fix siteName in dialog regression

### LINKED ISSUE #2632: [dev] siteName not being used in '%s uses Persona instead of usernames....' in dialog.js
Oops.

img

seanmonstar:
Since we needed to access this element by selector now, it was given a reasonable id of `rp_name`. However, `rp_name` is already the id of another element. So, then selecting by id would return the previous element. This PR just picks a new, unique id for our fancy little element.

fixes #2632


shane-tomlinson:
I am comfortable with this fix as a fix to the problem as described.

The larger problem of elements with duplicate IDs is not yet fully handled. The IDs of the tooltips relating to the email address and password are duplicated in add_email.ejs and set_password.ejs. I tested across several browsers (Firefox OSX, Chrome OSX, Fennec, Android Native browser, IE8 on WinXP) to see if this would cause a problem, so far all the tooltips show when expected.

This fix gets an r+, opening a new bug for the other duplicate IDs.

Nice work @seanmonstar.

PR #2640, opened 2012-10-24
remove undefined variable: e.toString()

### LINKED ISSUE #2636: undefined variable used to call toString on error path from GH-2621
At line 208 here https://github.com/mozilla/browserid/blob/380f33a6fabc239971b97ba51cd0b0c9f73e7a15/lib/primary.js#L205 the code calls e.toString(), but there is no `e` defined.

/me bows to jshint@0.9.1 (not the version jshint-tests uses). I'll file a bug to update that version.

Marking five-star to simply remove the `+ e.toString()` on that line for the next train.

seanmonstar:
fixes #2636

Also, while I was at it, I fixed all the other jshint warnings, since this warning was hard to notice amid the sea of warnings.
- jshint option white was flooding us with "wah wah wah you need an space between function and (). wah"

### DIFF
--- a/lib/primary.js
+++ b/lib/primary.js
@@ -39,11 +39,12 @@ try {
 
 // This becomes async
 function parseWellKnownBody(body, domain, delegates, cb) {
+  var v;
   try {
-    var v = JSON.parse(body);
+    v = JSON.parse(body);
   } catch(e) {
     return process.nextTick(function() {
-      if (cb) cb("malformed declaration of support for '" + domain + "': " + e.toString());
+      if (cb) cb("malformed declaration of support for '" + domain + "': " + String(e));
     });
   }
   const want = [ 'public-key', 'authentication', 'provisioning' ];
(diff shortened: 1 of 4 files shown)


lib/primary.js
```diff
@@ -39,11 +39,12 @@ try {
...
 // This becomes async
 function parseWellKnownBody(body, domain, delegates, cb) {
+  var v;
   try {
-    var v = JSON.parse(body);
+    v = JSON.parse(body);
   } catch(e) {
     return process.nextTick(function() {
-      if (cb) cb("malformed declaration of support for '" + domain + "': " + e.toString());
+      if (cb) cb("malformed declaration of support for '" + domain + "': " + String(e));
```
shane-tomlinson, line 47: @seanmonstar - What is the difference here? I tried to find some info online but am coming up short.

shane-tomlinson:
Tests pass, I am still able to log in using an @example.domain email address, looks good. 

I am just wondering why the change from e.toString to String(e)

seanmonstar:
toString can fail if the value is a primitive, null, or undefined. Casting
to String is a little safer, and uses toString if it exists. It's also 3
characters shorter :-)

PR #2658, opened 2012-10-29
pass event through when using DOMModule event binding

seanmonstar:
such as this.click(selector, function(e) {});

this lets us access the target, since we can be doing delegation. I recently needed this for adding a new click handler in b2g stuff, and felt it was worth being in dev.

@shane-tomlinson r?

### DIFF
--- a/resources/static/common/js/helpers.js
+++ b/resources/static/common/js/helpers.js
@@ -53,7 +53,7 @@
   function cancelEvent(callback) {
     return function(event) {
       event && event.preventDefault();
-      callback.call(this);
+      callback.call(this, event);
     };
   }
 


shane-tomlinson:
@seanmonstar - this is going to break things on the front end - travis is showing some failures as well.  To do what you want to do, I have traditionally used the normal "this.bind" route instead of the shortcut.

```
this.bind(<selector>, 'click', cb)
```

seanmonstar:
What does it break? Looking in travis, the only broken test is actually a test making sure the event isn't passed. I haven't really noticed anything broken in our b2g dialog.

shane-tomlinson:
@seanmonstar - 

dialog->js->modules->required_email.js::signIn 
dialog->js->modules->verify_primary_user.js::cancel, 
common->js->modules->extended-info.js::open 

These three are hooked up to the click handler and take a callback as their first parameter. 

The callbacks are used for unit testing only, but there are at least these 3 locations where an update to the function signature and associated unit tests will have to be made.

seanmonstar:
Very well, I'll back this out and use `.bind()`.

PR #2663, opened 2012-10-30
fix API regression where website would get onlogout invoked even when pa...

lloyd:
...ssing 'null' for loggedInUser


lloyd:
ditching this PR in favor of @seanmonstar's patch

PR #2664, opened 2012-10-30
fix undefined check of loggedInUser

### LINKED ISSUE #2662: API semantics broken: onlogout fired even if loggedInUser: null
the API is not supposed to invoke the onlogout callback if a website supplies `loggedInUser: null`.  This behavior was broken in commit f94a3b40364f3db78568de7630944d4262eb8699

Sites that rely on this API contract and redirect upon onlogout events can get into an infinite reload loop

seanmonstar:
fixes #2662

@lloyd r?

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1056,7 +1056,7 @@
             if (observers.login) observers.login(params);
           });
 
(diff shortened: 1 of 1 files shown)


lloyd:
r+

PR #2666, opened 2012-10-31
update to awsbox 0.3.3

lloyd:
includes:
- latest security patches
- support for `local_hooks` and renaming of `hooks` to `remote_hooks`
- support for `update` verb that git pushes and run local_hooks if any

r?


seanmonstar:
Perhaps I should have mentioned it here: https://github.com/lloyd/awsbox/commit/302637dc24d21cd7cbf7e9f437b41961553ae292#commitcomment-2083051

PR #2676, opened 2012-11-05
Issue 2674 add issuer locally

ozten:
Updates `localStorage.emails` with `issuer` at the same level as `cert`.


ozten:
sigh, wrong branch.

Issue #2674, opened 2012-11-04
Clear certificates from storage when new issuer is detected


ozten:
Currently under `localStorage.emails` we store:

```
{
  "ozten@mockmyid.com":{
     "created":"2012-11-03T02:35:58.934Z",
     "type":"primary",
     "updated":"2012-11-03T02:35:58.934Z",
     "pub":{
        "algorithm":"DS",
        "y":"391eb7...367",
        "p":"ff60...17",
        "q":"e21e04f911d1ed7991008ecaab3bf775984309c3",
        "g":"c52a4a0...233397a"
     },
     "priv":{
        "algorithm":"DS",
        "x":"1c296f623ebb86c9c3be48c4815915d2336b352c",
        "p":"ff60048...e17",
        "q":"e21e04f911d1ed7991008ecaab3bf775984309c3",
        "g":"c52a4a0ff3...896233397a"
     },
     "cert":"eyJhbGciO...bgdkA",
     "verified":true
  }
}
```

We'll parse an existing certificate and make sure the issuer matches address info's issuer property.

jrgm:
The issuer is already a field available in the cert, no? Or is this a different value of issuer?

ozten:
True, the issuer is in the cert, but we'd have to base64 decode it and parse it out. I think we want it in a handy place.

lloyd:
+1 to @jrgm's idea, let's not duplication information.  will present concrete suggestion in review.  thanks @jrgm

ozten:
Okay, I see how I can use `require` to get jwcrypto and then `jwcrypto.extractComponents(cert).payload.iss` to get the issuer.

OT: I was surprised to find we have a `require` implementation, seems like it would be useful throughout our frontend code.

ozten:
There are some tricky timing issues (if done another way), but it looks like adding this to pick_email and authenticate is the most promising...

https://github.com/ozten/browserid/compare/mozilla:primary-transition-ux...ozten:issue-2674-invalidate-via-addr-info

I'll try to manually test with hardcoded data in a bit, before issuing a pull request.

ozten:
The behavior is identical to the `notMe` state. The user has to re-enter their email address.

This is not ideal, but changing issuers will happen so infrequently... maybe it is okay for a first cut?

We can also probably do better by using this codepath in the transition detected codepaths, where we know we can clear a cert w/o checking address info.

ozten:
https://github.com/mozilla/browserid/pull/2775

[...]

PR #2677, opened 2012-11-05
Added issuer to localStorage. Fixes issue #2674

ozten:
Updates `localStorage.emails` with `issuer` at the same level as `cert`.


ozten:
Refreshed this pull request.

ozten:
We went another way with this.

PR #2679, opened 2012-11-05
Adding everywhere parameter to navigator.id.logout

ozten:
`navigator.id.logout()` remains the same.

`navigator.id.logout({everywhere: true})` clears the server side session as well as the `localStorage.loggedIn` status.


resources/static/communication_iframe/start.js
```diff
@@ -92,6 +92,8 @@
   });
 
   chan.bind("logout", function(trans, params) {
+    if (params && params.everywhere) user.logoutUser();
```
shane-tomlinson, line 95: @ozten - is this something that should be exposed to _all_ RPs or only the marketplace? Do we want to allow any RP to specify that you should be logged out on all sites? That seems like it leaves a lot of room open for user confusion.

ozten, line 95: My understanding is that it is for all RPs based on @benadida's requirements.

callahad, line 95: As the designated end-user curmudgeon, this fells bad / strange / confusing to me. Has Crystal reviewed this?

Edit: Wait, unless I'm misunderstanding what this does...

Edit: Also, this is against the **b2g branch**. So all of my knowledge / objections / etc go out the window :)

ozten:
Here is the [Marketplace bug#808831](https://bugzilla.mozilla.org/show_bug.cgi?id=808831)

shane-tomlinson:
@ozten - this needs to be updated before merge. And like @callahad - this is against the B2G branch so all of my objections are nullified.

ozten:
I think [Bug#808831 Comment#3](https://bugzilla.mozilla.org/show_bug.cgi?id=808831#c3) makes this PR obsolete.

Issue #2682, opened 2012-11-06
apply the browserid.spec fix for CONTRIBUTORS and en/en-us to dev branch


jrgm:
Ugghh. This commit on train-2012.10.12 did not make it into the dev branch.

https://github.com/mozilla/browserid/commit/49cffe72afdd93772d42bf42fa474be1194b60c4#scripts/browserid.spec

I can't find the bug or issue that specifically documents this, but without that, the static webhead in stage is crashing on startup trying to load the CONTRIBUTORS file.

See also https://bugzilla.mozilla.org/show_bug.cgi?id=806475

jrgm:
For clarity, PR-2683 is for dev, PR-2685 is for train-2012.10.26, and PR-2684 was me hitting the wrong button.

jrgm:
Fixes were committed to dev and train-2012.10.26. Marking this one fixed.

jrgm:
(Ignore. Leaving a breadcrumb: 
https://github.com/mozilla/browserid/issues/2645, 
https://github.com/mozilla/browserid/issues/1863, 
https://bugzilla.mozilla.org/show_bug.cgi?id=805482)

PR #2685, opened 2012-11-06
fix missing CONTRIBUTORS and renaming of en-US and update Changelog in train-2012.10.26

jrgm:
Fixes GH-2682 for train-2012.10.26


jrgm:
The Travis build failed on 'Executing your before_install (sudo apt-get install libgmp3-dev) took longer than 8 minutes and was terminated.' for the frontend test, but both backends passed ok. This is bogo.

lloyd:
r+

Issue #2566, opened 2012-10-09
Can't confirm an email address using mail client's embedded IE browser


mamacdon:
My e-mail client is a corporate deployment of Lotus Notes 8.5, which uses (on Windows platforms) an embedded Internet Explorer 8 as its built-in web browser. When I try to confirm a Persona email address from Lotus Notes, the embedded browser fails to load the confirmation page and my email address is not confirmed.

Steps are:
1. Add a new email address to my Persona profile.
2. In Lotus Notes, open the confirmation email.
3. Click the "Click to confirm this email address" link.
4. I'm taken to a page and see a script error:

```
An error has occurred in the script on this page.
Line: 8
Char: 14823
Error: 'JSON' is undefined
Code: 0
URL: https://static.login.persona.org/v/53e7265f01/production/en_US/browserid.js
```

This probably happens because the embedded browser is running in compatibility mode. Setting an 'X-UA-Compatible' <meta> tag in the confirmation page should fix it.

jrgm:
Thanks for the bug report. A couple of questions:
1. When you say "Add a new email address.." ... Did you do this in some other browser than the Lotus Notes embedded one? 
2. I downloaded Note 8.5.3 on Windows/XP and the embedded IE says it is "Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; Trident/4.0; .NET CLR 1.1.4322; .NET CLR 2.0.50727; .NET CLR 3.0.04506.30; .NET CLR 3.0.04506.648; .NET CLR 3.0.4506.2152; .NET CLR 3.5.30729; .NET4.0C; .NET4.0E)". Are you sure that your embedded IE is IE8?

That's IE7, and we do not support that (except to nicely inform the user of that fact). The verify_email_address and reset_password pages though are not very nice in IE7.

callahad:
We should totally be using the `<meta http-equiv="X-UA-Compatible" content="IE=Edge"/>` meta tag everywhere.

mamacdon:
1. Yes -- I used Google Chrome to add the email address to my Persona account. Lotus Notes only came into play when I checked my email, and tried to click the confirmation link. (The problematic part of the above steps is from Step 2 onward.)
2. No, I was wrong to assume the embedded browser was IE8. My user-agent string is:

```
[...]

PR #2687, opened 2012-11-07
add a meta tag to all documents to fix environments with an embedded IE ...

lloyd:
...forced into compatibility mode.

This makes it possible to verify your email account with a program like lotus notes that displays the verification page in an embedded browser rather than in your default/desktop browser.

closes issue #2566


lloyd:
@jrgm says r+

PR #2692, opened 2012-11-08
Adding forceAuthentication parameter to request

ozten:
Adding forceAuthentication parameter to `navigator.id.request`.

If `true`, force the user to re-enter their credentials. This meets the b2g requirements for wrong PIN entry lockout. See [Bug #808831](https://bugzilla.mozilla.org/show_bug.cgi?id=808831)

**Note:** The user has to type their email address in. If we want to try to do something better, please file a new GH issue with those requirements.

**Note:** This PR makes [PR 2679](https://github.com/mozilla/browserid/pull/2679) obsolete.

### DIFF
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -68,15 +68,15 @@ pre {
     </li><li>
       <input type="checkbox" id="siteName">
       <label for="siteName">Supply Site Name</label><br />
-    </li>
     </li><li>
       <input type="checkbox" id="siteLogo">
       <label for="siteLogo">Supply Site Logo</label><br />
-    </li>
     </li><li>
       <input type="checkbox" id="returnTo">
       <label for="returnTo">Supply returnTo</label><br />
-    </li>
+    </li><li>
+      <input type="checkbox" id="forceAuthentication">
+      <label for="forceAuthentication">Force Authentication</label><br />
     </li>
   </ul>
     <button class="assertion">Get an assertion</button>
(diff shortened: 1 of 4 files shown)


seanmonstar:
Are we concerned at all that other RPs can access the API if the user is a FxOS device? @benadida ?

benadida:
fair question @seanmonstar . I don't think it's a problem, but worth filing a followup bug to track this issue in case we want to change our minds. Can you do that?

ozten:
Good call. I think this would be addressed in the native layer. I haven't chatted with Jed yet about how that will work.

callahad:
I'm a big +1 on finding ways to avoid leaking the B2G/Marketplace stuff into other RPs, even if it's undocumented.

benadida:
@callahad we've structured this API call so that it still makes sense beyond B2G/Marketplace, so I don't think we need to worry about complete isolation on this particular issue. Forced re-authentication is a problem that we've heard from a number of Relying Parties.

Now, I'm not saying this will be supported in our shim just yet, but that if it leaks to other B2G Apps, it's probably not a big deal (I wouldn't want to spend considerable resources preventing that.)

callahad:
@benadida I'm more worried about leaking to non-B2G apps -- "hey guys, use this other shim over here and you get these cool features!"

benadida:
@callahad not part of any shim, only part of the B2G native implementation (and alternate server) for now.

callahad:
@benadida Oh, cool! Carry on, then :)

PR #2694, opened 2012-11-09
transition_to_primary shown once

### LINKED ISSUE #2404: Make secondary to primary transition screen show one time only
When we notice that a user's has a yahoo.com email address, we display a "Yahoo.com makes this easy" screen.

This is good because it informs the user that something is different and that they don't type in their Persona password anymore.

But we shouldn't this same screen every time.

Can we just display this screen once and skip this step for the second and subsequent flows?

This came up as a BigTent UX bug, but we should think about the general case.

seanmonstar:
This includes changes to address_info request, and adds in complete_transition request. Won't actually work without the backend making the expected changes.

@lloyd halp!

merge target is primary-transition-ux. fixes #2404.


resources/static/dialog/js/modules/verify_primary_user.js
```diff
@@ -32,10 +34,23 @@ BrowserID.Modules.VerifyPrimaryUser = (function() {
     complete(callback);
   }
 
+  function transitionComplete(callback) {
+    /*jshint validthis: true*/
+    var self = this;
+    user.completeTransition(email, function onSuccess(status) {
+      // do we care about status?
```
shane-tomlinson: @seanmonstar - this is one thing I was wondering here too, would completeTransition ever call the callback with a non-successful status or would it call the error callback directly?

resources/static/common/js/network.js
```diff
@@ -749,6 +749,22 @@ BrowserID.Network = (function() {
...
+    },
+
+    /**
+     * Mark the transition of this email as having been completed.
+     * @method completeTransition
+     * @param {string} [email] - The email that transitioned.
+     * @param {function} [onComplete] - Called whenever complete.
+     * @param {function} [onFailure] - Called on XHR failure.
+     */
+    completeTransition: function(email, onComplete, onFailure) {
```
seanmonstar, line 761: The Etherpad suggests this must authed. Does that mean anything for the front-end call, or can I just assume that the server will return a failure if the user isn't authenticated?

https://id.etherpad.mozilla.org/primary-transition-ux-swarm

resources/static/dialog/js/modules/verify_primary_user.js
```diff
@@ -48,20 +63,30 @@ BrowserID.Modules.VerifyPrimaryUser = (function() {
...
-        siteName: data.siteName,
-        idpName: data.idpName
-      });
-
-      if (data.siteTOSPP) {
-        dialogHelpers.showRPTosPP.call(self);
-      }
+      user.addressInfo(email, function onSuccess(info) {
+        if (showsPrimaryTransition(info.state)) {
+          self.submit = transitionComplete;
```
shane-tomlinson, line 67: One thing you could do here for clarity is just set self.submit to transitionComplete instead of verify on Module declaration. From my brief look at the code, it does not look like self.submit is ever called when submit is set to transitionComplete. I was initially confused where transitionComplete was getting called.

resources/static/test/cases/dialog/js/modules/verify_primary_user.js
```diff
@@ -33,112 +34,193 @@
     }
   });
 
-  test("personaTOSPP true, requiredEmail: true - show TOS/PP", function() {
+  asyncTest("personaTOSPP true, requiredEmail: true - show TOS/PP", function() {
```
shane-tomlinson, line 38: A remnant of requiredEmail that is never actually used anywhere. Issue #2705 is now open to remove this.

resources/static/common/js/network.js
shane-tomlinson, line 761: @seanmonstar - you can call user.checkAuthentication. If the status is already known (and it should be at this point), no new network request will be made.

resources/static/dialog/js/modules/verify_primary_user.js
seanmonstar, line 67: You mean reverse what is currently happening? Make transitionComplete the default, and change submit to verify if it is not needed?

resources/static/dialog/js/modules/verify_primary_user.js
seanmonstar: Our spec suggests complete_transition will return with `{ success: boolean }`. Perhaps that should be checked in `network.js`, and trigger the error callback if success is false?

Issue #2678, opened 2012-11-05
Add wsapi complete_transition


ozten:
Add a new wsapi `complete_transition` -  indicates that a user has been displayed the "transition to primary" screen.

```
AUTHENTICATED (assertion?) POST request

arguments: "email"

Side affect - will update the ? with ?

returns JSON object { success: boolean }
```

ozten:
From lloyd's [comments on a commit](https://github.com/ozten/browserid/commit/a4ba788e057ceca3ef5003fb4559cc4b1eb1a745#commitcomment-2141598):

I think you should:
1. verify the the currently authenticate user owns the email address in question
2. verify the email address specified is a primary email address
3. update the database!  set email.lastUsedAs to `primary`

Then I think in cert_email we need to update database and set email.lastUsedAs to `secondary`

lloyd:
landed in 781184b5f2b6372c0ce44fff046f1719a78553f9

PR #2709, opened 2012-11-13
First stab at updating cert_key and adding complete_transition wsapis

ozten:
Fixes Issue #2678

To avoid duplicating a regex, I've added a new wrapper function `checkEmailSupport` which delegates to the existing `checkSupport`. A change in behavior is that during errors we log the full email address instead of the domain, not sure if that is kosher.


lib/db.js
```diff
@@ -86,6 +86,7 @@ exports.onReady = function(f) {
   'isStaged',
   'lastStaged',
   'listEmails',
+  'emailLastUsedAs', // TEMPORARY
```
lloyd: what does "TEMPORARY" mean?

lib/db/json.js
```diff
@@ -29,18 +29,25 @@ const ESC = JSON.stringify;
...
- *  ]
+ * {
+ *   "users":[
+ *     {
+ *       "id": <numerical user id>,
+ *       "password": <string password and salt info>,
+ *       "lastPasswordReset": <seconds-since-epoch, integer>,
+ *       "emails":{
+ *         "syncer@somehost.com":{
+ *           "type": <string secondary|primary>,
```
lloyd, line 40: we need to go through and obliterate type completely.  The time to do this is probably after we land the last API changes in a separate pass?

lib/db/json.js
```diff
@@ -472,6 +479,32 @@ exports.listEmails = function(uid, cb) {
   });
 };
 
+/* TODO: This is only temporary for a unit test, would be listEmails */
```
lloyd: oh, I see, actually not.  Given that now we're going to _not_ include type in list_emails response, this function probably will hang around.

lib/db/mysql.js
```diff
@@ -591,6 +591,25 @@ exports.listEmails = function(uid, cb) {
...
+      cb('emailLastUsedAs Expected 1 row, got ' + rows.length + ' for ' + email);
+    } else {
+      cb(null, rows[0].lastUsedAs);
+    }
+  });
+};
+
+exports.updateEmailLastUsedAs = function(email, type, cb) {
+  var typeEnum = ['primary', 'secondary'];
+  if (typeEnum.indexOf(type) === -1) return cb ('Invalid type for updating email.lastUsedAs');
```
lloyd: process.nextTick here?

lib/primary.js
```diff
@@ -260,6 +260,12 @@ exports.checkSupport = function(principalDomain, cb) {
   });
 };
 
+const emailRegex = /\@(.*)$/;
+exports.checkEmailSupport = function(email, cb) {
```
lloyd: would it be too magic to have `.checkSupport` just deal with email addresses when passed a string with `@`?

lib/wsapi/cert_key.js
```diff
@@ -46,13 +46,15 @@ exports.process = function(req, res) {
...
       // to cause them to be forwarded.
       req.body = req.params;
-
       forward(keysigner, req, res, function(err) {
         if (err) {
           logger.error("error forwarding request to keysigner: " + err);
           httputils.serverError(res, "can't contact keysigner");
           return;
         }
+	db.updateEmailLastUsedAs(req.params.email, 'secondary', function (err) {
```
seanmonstar: indentation is off

lib/db/mysql.js
```diff
@@ -591,6 +591,25 @@ exports.listEmails = function(uid, cb) {
...
+    } else if (rows.length !== 1) {
+      cb('emailLastUsedAs Expected 1 row, got ' + rows.length + ' for ' + email);
+    } else {
+      cb(null, rows[0].lastUsedAs);
+    }
+  });
+};
+
+exports.updateEmailLastUsedAs = function(email, type, cb) {
+  var typeEnum = ['primary', 'secondary'];
```
seanmonstar: elevate this to a `const` outside the function. only need to construct the array once.

lib/wsapi/address_info.js
```diff
@@ -26,15 +26,10 @@ exports.args = {
...
-
 exports.process = function(req, res) {
-  // parse out the domain from the email
-  var m = emailRegex.exec(req.params.email);
-
-  primary.checkSupport(m[1], function(err, r) {
+  primary.checkEmailSupport(req.params.email, function(err, r) {
     if (err) {
-      logger.info('"' + m[1] + '" primary support is misconfigured, falling back to secondary: ' + err);
+      logger.info('"' + req.params.email + '" primary support is misconfigured, falling back to secondary: ' + err);
```
lloyd: privacy point: logging email addresses when not strictly necessary increases our chances of capturing where people are going on the web (because we can correlate to verifier logs).  it's a nit...

lib/wsapi/cert_key.js
```diff
@@ -46,13 +46,15 @@ exports.process = function(req, res) {
...
+	  if (err) logger.error("cert_key unable to update email.lastUsedAs in the database");
+	});
```
lloyd: this spacing looks funny.  (nit)

lib/wsapi/complete_transition.js
```diff
@@ -0,0 +1,52 @@
...
+  'email': 'email'
+};
+exports.i18n = false;
+
+exports.process = function (req, res) {
+  var email = req.params.email;
+  db.userOwnsEmail(req.session.userid, email, function(err, owned) {
+    if (err) return wsapi.databaseDown(res, err);
+    // not same account? big fat error
+    if (!owned) return httputils.badRequest(res, "that email does not belong to you");
```
lloyd, line 33: is a 403 better?  `.forbidden()` ?

lloyd, line 33: "In servers where authentication is required, this commonly means that the provided credentials were successfully authenticated but that the credentials still do not grant the client permission to access the resource (e.g. a recognized user attempting to access restricted content)."

lib/db/mysql.js
ozten: Good call.

lib/wsapi/complete_transition.js
```diff
@@ -0,0 +1,52 @@
...
+    primary.checkEmailSupport(email, function (err, r) {
+      var notPrimary = false;
+      if (err) {
+        logger.error('"' + email + '" primary support is misconfigured, falling back to secondary: ' + err);
+        notPrimary = true;
+      } else if (r && r.urls) {
+        //3) update the database! set email.lastUsedAs to primary
+        db.updateEmailLastUsedAs(email, 'primary', function (err) {
+          if (err) return httputils.serverError(res, "Unable to update database");
+          return res.json({success: true});
```
seanmonstar: I've wondered this as I make the completeTransition function in the dialog: do we even care about the `success` param? Will it ever be returned false? It looks like if it is false, a badRequest is returned instead.

tests/cert-emails-test.js
```diff
@@ -139,6 +140,23 @@ suite.addBatch({
           assert.equal(certs_and_assertion.certificates[0].split(".").length, 3);
           assert.equal(certs_and_assertion.assertion.split(".").length, 3);
         },
+      },
+      "after a short wait": {
+        // In practise, db.emailLastUsedAs is sometimes called before
+        // db.updateEmailLastUsedAs has been called by cert_key wsapi...
+        topic: function (err, r) {
+          setTimeout(this.callback, 500);
+	},
```
seanmonstar: indentation

lib/wsapi/complete_transition.js
```diff
@@ -0,0 +1,52 @@
...
+    // not same account? big fat error
+    if (!owned) return httputils.badRequest(res, "that email does not belong to you");
+
+    primary.checkEmailSupport(email, function (err, r) {
+      var notPrimary = false;
+      if (err) {
+        logger.error('"' + email + '" primary support is misconfigured, falling back to secondary: ' + err);
+        notPrimary = true;
+      } else if (r && r.urls) {
+        //3) update the database! set email.lastUsedAs to primary
```
lloyd: wonky comment

lib/wsapi/cert_key.js
ozten: Thanks! Will fix.

lib/wsapi/complete_transition.js
```diff
@@ -0,0 +1,52 @@
...
+        db.updateEmailLastUsedAs(email, 'primary', function (err) {
+          if (err) return httputils.serverError(res, "Unable to update database");
+          return res.json({success: true});
+        });
+      } else {
+        notPrimary = true;
+      }
+      if (notPrimary) return httputils.badRequest(res, "email is not a primary");
```
lloyd, line 49: would this be more readable?

```
if (r && r.urls) {
  // udpate db etc
} else {
  if (err) logger.error(err);
  httputils.badRequest()
} 
```

??

tests/cert-emails-test.js
```diff
@@ -139,6 +140,23 @@ suite.addBatch({
           assert.equal(certs_and_assertion.certificates[0].split(".").length, 3);
           assert.equal(certs_and_assertion.assertion.split(".").length, 3);
         },
+      },
+      "after a short wait": {
+        // In practise, db.emailLastUsedAs is sometimes called before
+        // db.updateEmailLastUsedAs has been called by cert_key wsapi...
+        topic: function (err, r) {
+          setTimeout(this.callback, 500);
```
lloyd, line 148: ewww.  sleeping in tests!  :P

tests/cert-emails-test.js
```diff
@@ -139,6 +140,23 @@ suite.addBatch({
...
+	},
+        "email table lastUsedAs updated": {
+          topic: function(err, certs_and_assertion) {
+            // TODO used listEmails here, once that work is done
```
lloyd: eliminate todo.  likey that work won't be done!

lib/wsapi/address_info.js
seanmonstar: This actually changes the log to include more identifiable data. Before, `m[1]` would have just been the primary domain. perhaps more than a nit?

ozten: Agreed, as called out in the PR summary.

I considered refactoring to

```
primary.checkSupport(primary.domainFrom(email), ...)
```

or 

```
primary.checkSupport(emailutils.domainFrom(email), ...)
```

lib/wsapi/complete_transition.js
ozten, line 33: Good call. Should l update https://github.com/ozten/browserid/blob/dev/lib/wsapi/cert_key.js#L29 also?

lib/wsapi/complete_transition.js
ozten, line 49: Wouldn't that have to be this?

```
if (!err && r && r.urls) {
  // udpate db etc
} else {
  if (err) logger.error(err);
  httputils.badRequest()
} 
```

lib/wsapi/complete_transition.js
ozten: I don't document my code often, but when I do... it's with wonky comments.

tests/cert-emails-test.js
ozten, line 148: Tests work hard, they need sleep too!

tests/cert-emails-test.js
ozten: Okay, wanted to communicate during the review that

1) I think we can do more tests once all the backend is updated
2) I have a temporary function in db.js that is only used in unit tests and will be removed once listEmails includes lastUsedAs.

Will remove.

lib/db/json.js
ozten: Hmm, I don't see enough of the new design to understand, but will take out the TODOs.

lib/primary.js
```diff
@@ -260,6 +260,7 @@ exports.checkSupport = function(principalDomain, cb) {
   });
 };
 
+exports.emailRegex = /\@(.*)$/;
```
ozten, line 263: @lloyd @seanmonstar this gets ride of logging email addresses and keeps regex reusable.

Issue #2711, opened 2012-11-14
[train-2012.11.09] The 'input#authentication_email' in the dialog does not have focus for unauthed user


jrgm:
STR
1. Use a browser where the user is not known (or click 'this is not me' and open the dialog again)
2. use either 123done or mfb and bring up the /sign_in dialog

Expected: /sign_in dialog presents text input box with 'enter email address' and has focus.

Actual: focus is not set to the text input box.

My vmware fusion is updating atm, but I have this on all the osx browsers, and I expected it's not different on windows and linux.

This is a regression from current production train-2012.10.12

lloyd:
merged into train-2012.11.09 and dev.  wewt.

jrgm:
This was fixed in train-2012.11.09 (currently on stage) and rechecked with train-2012.11.23 on ephemeral instance, but I hadn't noted it verified here. Now noting.

PR #2723, opened 2012-11-15
Focus the #authentication_email field when first loading the dialog and the user is not authenticated.

shane-tomlinson:
issue #2711


lloyd:
r+, tested, @jrgm on board.  merging to dev and train-2012.11.09

PR #2727, opened 2012-11-16
Send internal parameter {silent:true} for silent assertions

jedp:
This PR adds an internal parameter to the internal api that enables RPs to know whether an assertion was acquired without user interaction (`{silent: true}`).

The parameter will not be visible to normal RPs.

This is a prerequisite for https://bugzilla.mozilla.org/show_bug.cgi?id=790141, "Native implementation of get() and getVerifiedEmail()"

### DIFF
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -178,7 +178,6 @@
       log('checking and emitting');
       // this will re-certify the user if neccesary
       user.getSilentAssertion(loggedInUser, function(email, assertion) {
-        log('silent return: email, assertion ', email, assertion);
         if (email) {
           // only send login events when the assertion is defined - when
           // the 'loggedInUser' is already logged in, it's false - that is
@@ -210,8 +209,10 @@
     }
 
     function doLogin (params) {
-      log('doLogin', params);
-      callback({ method: 'login', assertion: params });
+      log('doLogin (with silent assertion)');
+      // Through the _internalParams, we signify to any RP callers that are 
+      // interested that this assertion was acquired without user interaction.
+      callback({ method: 'login', assertion: params, _internalParams: {silent: true} });
     }
 
     function doLogout () {


jedp:
@seanmonstar do you mind if I nominate you to review this?

jedp:
I've pushed this code to http://notoriousb2g.personatest.org if anyone wants to play with it live

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -210,8 +209,10 @@
     }
 
     function doLogin (params) {
-      log('doLogin', params);
-      callback({ method: 'login', assertion: params });
+      log('doLogin (with silent assertion)');
+      // Through the _internalParams, we signify to any RP callers that are 
+      // interested that this assertion was acquired without user interaction.
+      callback({ method: 'login', assertion: params, _internalParams: {silent: true} });
```
ozten, line 215: Is `doLogin` only called in code paths that guarantee there was no user interaction?

resources/static/include_js/include.js
```diff
@@ -1280,9 +1280,13 @@
 
         checkCompat(true);
         internalWatch({
-          onlogin: function(assertion) {
+          onlogin: function(assertion, internalParams) {
             if (callback) {
-              callback(assertion);
+              if (internalParams) {
+                callback(assertion, internalParams);
```
ozten: Consider documenting the shape of this callback, where it is originally declared, since it has an optional parameter.

Great stuff!

seanmonstar:
Why is the include.js changed? Doesn't B2G provide it's own navigator.id code?

resources/static/dialog/js/misc/internal_api.js
jedp, line 215: That's the intent.  This happens via the internal api when getting a silent assertion only.  The only way into that code should be through paths that don't involve the user.

resources/static/include_js/include.js
jedp: Yes, definitely.  Thanks for your comments, @ozten!

jedp:
@seanmonstar b2g only implements the `navigator.id.*` functions.  It relies on the shim and the internal api to a large extent.  (Specifically, it is included by gaia here: https://github.com/mozilla-b2g/gaia/blob/master/apps/system/js/identity.js (line 43 is where it is displayed).

Without changing `include.js`, there's no way to get an extra parameter from internal api back to the caller, is there?

seanmonstar:
@jedp The include.js is only needed to implement `nagivator.id.*`. The dialog and communication_iframe both don't use the include. The shim doesn't fully execute, since it detects that `navigator.id` already exists thanks for the native implementation.

jedp:
ah so i only need the change to internal_api?  I could swear I tried that and it didn't work.  But that would be great!  I'll try and update the PR

jedp:
@seanmonstar I've removed the unnecessary changes to include.js.  Thanks for your vigilance!

seanmonstar:
@jedp Looks correct. I notice this only affects `watch`. Does it need to there if `get` is called also?

jedp:
@seanmonstar good catch! but this is actually the intent.  in b2g, `get` is implemented internally using the observer api.  It does a `watch` followed by a `request`.  If the `watch` returns a silent assertion, it is simply ignored; there is always a `request` to force the popup to open.  This `get` call is the whole reason for the `{silent: true}` parameter.  Without it, the caller has no way to tell whether an assertion given to the `onlogin` handler is from `watch` or `request`.

PR #2734, opened 2012-11-19
Remove eyedee.me related wclickIfExists invocations

shane-tomlinson:
- When redirecting to eyedee.me, add a small delay before continuing to ensure time for the click handler to be hooked up.
- Renenable the returning-user, remove-email, and add-primary-to-secondary tests.

@6a68 - mind reviewing this?


jaredhirsch:
@shane-tomlinson This is the sorta situation where abstracting out something like POMs, but flow-oriented, would be nice (instead of having to duplicate the fix across the codebase). However, I don't want us to rush to clunky abstraction, so r+ from me.

Issue #2722, opened 2012-11-15
return "Server Too Busy" when we are.


lloyd:
currently all of our processes that handle requests keep going when they are maxed out.  This causes them to not respond to health checks in time and drop out of the load balancer.

When this happens, all connected clients are affected and get error codes.  

Rather than hitting this point, processes should pre-emptively return 503 error codes (as soon as possible when over 90% usage), and keep running and responding to health checks in a timely fashion.

We should spin _more_ processes and run them all under 90% even at max load.

shane-tomlinson:
closing since @lloyd already merged the PR.

PR #2743, opened 2012-11-19
refuse requests when the server is too busy

lloyd:
issue #2722

### DIFF
--- a/lib/load_gen/activities/add_email.js
+++ b/lib/load_gen/activities/add_email.js
@@ -56,37 +56,27 @@ exports.startFunc = function(cfg, cb) {
       email: email,
       site: userdb.any(user.sites)
     }, function (err, r) {
+      err = common.checkResponse(err, r);
       if (err) return cb(err);
-      if (!r || r.code !== 200) {
-        var msg = 'failed to add email: ' + email + ' to existing user ' +
-          user.emails[0];
-        winston.error(msg);
-        return cb(msg);
-      }
+
       // now get the verification secret
       wcli.get(cfg, '/wsapi/fake_verification', context, {
         email: email
       }, function (err, r) {
+        err = common.checkResponse(err, r);
         if (err) return cb(err);
-        if (!r || r.code !== 200) {
-          err ='failed to fetch verification token for email: ' + email;
-          winston.error(err);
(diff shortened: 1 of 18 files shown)


shane-tomlinson:
@lloyd - It'd be awesome to have 2 fewer bugs listed but this needs to be rebased and @jrgm is worried that 20ms is a too low of a default.

lloyd:
absolutely, I'm on the hook to rebase and address @jrgm's concerns.  I'll do this today.

lib/load_gen/activities/add_email.js
```diff
@@ -56,37 +56,27 @@ exports.startFunc = function(cfg, cb) {
...
-        return cb(msg);
-      }
+
       // now get the verification secret
       wcli.get(cfg, '/wsapi/fake_verification', context, {
         email: email
       }, function (err, r) {
+        err = common.checkResponse(err, r);
         if (err) return cb(err);
-        if (!r || r.code !== 200) {
```
jrgm, line 71: Needs to check for err.code !== 'nonJSON' here.

lib/load_gen/activities/reset_pass.js
```diff
@@ -65,31 +65,24 @@ exports.startFunc = function(cfg, cb) {
     pass: user.password,
     site: userdb.any(user.sites)
   }, function (err, r) {
+    err = common.checkResponse(err, r);
     if (err) return cb(err);
```
jrgm, line 69: Needs to check for err.code !== 'nonJSON' here, I seem to recall.

jrgm:
Hmm, github is having issues atm, and won't let me add another line note. But about line 72 is activities/signup, that also needs to check for nonJSON in response from fake_verification (or that's what I recall; sorry, thought I had added a comment here about these places to fix, but apparently did not).

lloyd:
at this point I believe all open issues are addressed.  I did my best to load test this on AWS - https://gist.github.com/4489337

But load testing on AWS cannot quite excercise these failures.  I feel comfortable enough with this to merge and vet it in our staging environment, because the actual code changes are not that extensive.

I leave it up to @jrgm whether this is really merge worthy at this point.

Issue #2747, opened 2012-11-20
testHelpers.setup done callbacks do not check for the existence of the fixtures.


shane-tomlinson:
This causes problems in tests that make use of personatestuser when run against ephemeral instances.

Link to #2736

PR #2748, opened 2012-11-20
Make sure the testSetup.setup fixtures exist before dereferencing them.

shane-tomlinson:
In all of the tests, make sure the testSetup.setup fixtures exist before dereferencing them. Make sure any errors in testSetup.setup are propagated to the test's done function so that vows catches them.

issue #2747


lloyd:
r+

PR #2761, opened 2012-11-22
B2g forceIssuer support

ozten:
What is it?

if you

1) go to http://192.168.186.138:10001/
2) check Force Issuer to issuer.domain and click "Get an Assertion"
3) enter a primary email like 'test@example.domain' and click Next
4) Create a password
5) Get VERIFICATION URL out of logs and go there

Note: You're logged in and the assertion was backed by 'issuer.domain'.

Note: The assertion is verified by the verifier.

This is because config/local.json adds this setting.

Caveats:

This patch is rushed and the frontend code is fairly horrible. Lots of duplicate code, I don't like the use of the `User.forceIssuer` as a global variable, etc.

How did storage get so ugly?
Why forceIssuerEmails?  
localStorage is a flat key value store, so we couldn't do:
- storage.default.emails and storage.fxos.emails
- dynamic keys (such as using fxos.login.persona.org) don't work well with removeItem and other APIs here.
- We can refactor all writes to emails to use a default or issuer based bucket

Why keep storage.forceIssuerEmails up to date for all addresses?
`user.checkAuthenticationAndSync` is called without us knowing which specific 
email the user is trying to forceIssuer on.


bin/keysigner
```diff
@@ -79,11 +80,18 @@ try {
 
 
 // and our single function
-app.post('/wsapi/cert_key', validate({ 'email': 'email', 'pubkey': 'pubkey', 'ephemeral': 'boolean' }), function(req, resp) {
+app.post('/wsapi/cert_key', validate(certKey.args), function(req, resp) {
   var startTime = new Date();
+  var forceIssuer = req.params.forceIssuer;
+  if ('default' !== forceIssuer &&
```
seanmonstar: is `forceIssuer` required? It doesn't look like it in `wsapi/cert_key`.

If it's not passed, then the `undefined` value would trigger an error. Is this correct behavior?

ozten: Yes, it's required.

AFAICT any `args` property on a `/wsapi/*.js` module is required.

One of the _few_ improvements of this patch is to remove the duplicate declaration of cert_key arguments.

lib/db/json.js
```diff
@@ -300,66 +316,84 @@ function getAndDeleteRowForSecret(secret, cb) {
...
+          addIt();
+        }
+      });
     });
   });
 };
 
+function withType(email, cb) {
+  var domain = email.replace(/[^@]*@/, '');
+  primary.checkSupport(domain, function (err, urls) {
```
seanmonstar: This creates an outgoing request, right? Slow servers could make this much slower. Not sure if there's an answer to this, though.

lib/db/mysql.js
```diff
@@ -363,20 +383,35 @@ function getAndDeleteRowForSecret(secret, cb) {
     });
 }
 
+function withType(email, cb) {
```
seanmonstar: This function exists also in `db/json.js`.

bin/keysigner
ozten: Thanks for showing me how wsapi calls can have optional arguments. I'll remove `default` as a value and make this optional throughout the code.

lib/db/mysql.js
ozten: I'll extract into a new file dbutils.js.

ozten:
Closing - re-worked in PR#2854

Issue #2762, opened 2012-11-22
Allow tooltip.showTooltip to specify either a reference DOM element or text and an anchor element.


shane-tomlinson:
Groundwork for #2641, which is 4*

PR #2764, opened 2012-11-23
Allow tooltip.showTooltip to either be called with a reference DOM element or text and an anchor.

shane-tomlinson:
issue #2762

### DIFF
--- a/resources/static/common/js/lib/dom-jquery.js
+++ b/resources/static/common/js/lib/dom-jquery.js
@@ -199,7 +199,13 @@ BrowserID.DOM = ( function() {
         * @return {boolean} true if the element has the attribute, false otw.
         */
         hasAttr: function( element, attrName ) {
-            var val = jQuery( element )[ 0 ].getAttribute( attrName );
+            var el = jQuery( element )[ 0 ],
+                val = null;
+
+            if (el) {
+              val = el.getAttribute( attrName );
+            }
+
             return val !== null;
         },
 
--- a/resources/static/common/js/tooltip.js
+++ b/resources/static/common/js/tooltip.js
@@ -9,13 +9,16 @@ BrowserID.Tooltip = (function() {
       TOOLTIP_MIN_DISPLAY = 2000,
       READ_WPM = 200,
       bid = BrowserID,
+      dom = bid.DOM,
       renderer = bid.Renderer,
       hideTimer,
       tooltip;
 
-  function createTooltip(el) {
+  function createTooltip(tooltipText) {
+    // There is only one global tooltip, update its reference to the new
+    // tooltip. All other tooltips should have been removed by this point.
     tooltip = renderer.append("body", "tooltip", {
-      contents: el.html()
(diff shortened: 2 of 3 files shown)


resources/static/common/js/lib/dom-jquery.js
```diff
@@ -199,7 +199,13 @@ BrowserID.DOM = ( function() {
...
         */
         hasAttr: function( element, attrName ) {
-            var val = jQuery( element )[ 0 ].getAttribute( attrName );
+            var el = jQuery( element )[ 0 ],
+                val = null;
+
+            if (el) {
+              val = el.getAttribute( attrName );
+            }
+
```
seanmonstar, line 208: `return jQuery(element).attr(attrName) !== undefined;` is safe, regardless if the element exists or not.

Not necessarily better, just point it out.

resources/static/common/js/tooltip.js
```diff
@@ -43,46 +46,57 @@ BrowserID.Tooltip = (function() {
...
-        tooltipAnchor = tooltipEl.attr("for");
-
-    if (tooltipAnchor) {
-      // The tooltip should be anchored to another element.  Place the tooltip
-      // directly above the element and remove it when it is no longer needed.
-      tooltipEl = createTooltip(tooltipEl);
-      anchorTooltip("#" + tooltipAnchor);
-    }
+  function showTooltipString(tooltipText, tooltipAnchor, complete) {
+    createTooltip(tooltipText);
```
seanmonstar, line 63: It looks like if another tooltip is shown before the previous one finishes animating, the reference will be lost because `createTooltip` clobbers it. Then, the callback for `animateTooltip` calls `removeTooltips`, and has the wrong reference. No?

resources/static/common/js/lib/dom-jquery.js
shane-tomlinson, line 208: Nice!

resources/static/common/js/tooltip.js
shane-tomlinson, line 63: Well, I'll be... You are correct. Fixing.

shane-tomlinson, line 63: On further inspection, the only way for showTooltipString to be called is via showTooltip. On line 91, any existing tooltips (even ones that are mid-animation) are removed before a new one is created. Only one tooltip at a time. Phew.

The "only one tooltip shown at a time" test should cover this case.

seanmonstar, line 63: That's true, but the animation callback would still be called after the timeout, right? And then it calls `removeTooltips`, which now has the new tooltip?

shane-tomlinson, line 63: I see it now! You are right. Opening a new PR 'cause I'm a complete and utter goon.

shane-tomlinson, line 63: https://github.com/mozilla/browserid/pull/2930 for you @seanmonstar

Issue #2740, opened 2012-11-19
wd-extension: wait until buttons are enabled before attempting to click


jaredhirsch:
We sometimes try to click buttons before their event handlers are wired up. Add a method that polls on the server until a specified CSS selector or attribute is present/absent from an element. We'll use this util to wait for the .disabled class/disabled attr to be gone before attempting to click.

If we want to poll on the wd side, it looks like the JSON wire protocol supports fetching computed CSS for an element, so that's one way to dig for the presence of a class, and we can get the value of arbitrary attributes (you can also check if an el is enabled).

If we want to poll on the browser, maybe we can use wd's extension method waitForConditionInBrowser.

Related to #2739 (use uniform set of class/attribute transitions to indicate buttons are enabled/disabled).

PR #2765, opened 2012-11-26
Issue 2740 submit in tests

shane-tomlinson:
beginning of updates to better handle timing problems related to disabled submit buttons in both the dialog and on eyedee.me

### DIFF
--- a/automation-tests/lib/wd-extensions.js
+++ b/automation-tests/lib/wd-extensions.js
@@ -1,7 +1,7 @@
 // add helper routines onto wd that make common operations easy to do
 // correctly
 
-const wd        = require('wd/lib/webdriver')
+const wd        = require('wd/lib/webdriver'),
       utils     = require('./utils.js'),
       timeouts  = require('./timeouts.js');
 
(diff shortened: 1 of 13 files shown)


shane-tomlinson:
Link to issue #2740

jaredhirsch:
as discussed in triage: seems simpler for users to just always use wclick instead of having to know to use wclick sometimes, and wsubmit other times.

automation-tests/lib/wd-extensions.js
```diff
@@ -257,3 +270,26 @@ wd.prototype.wclear = function(opts, cb) {
     self.clear(el, cb);
   });
 };
+
+// click on a submit button when form submission is enabled.
+wd.prototype.wsubmit = function(opts, cb) {
```
shane-tomlinson, line 275: So, @6a68 and I decided that combining these make more sense because it reduces the cognitive load on the developers. Going to modify this in the morning.

shane-tomlinson:
Closing this down for a new PR with a combined wclick and wsubmit.

PR #2771, opened 2012-11-27
Update wclick to handle disabled buttons.

### LINKED ISSUE #2740: wd-extension: wait until buttons are enabled before attempting to click

shane-tomlinson:
@6a68 or @jrgm, mind having a look at this?
- Check for the disabled attribute on the button.
- Check for the "submit_disabled" class on the body.

fixes #2740.

### DIFF
--- a/automation-tests/lib/wd-extensions.js
+++ b/automation-tests/lib/wd-extensions.js
@@ -1,7 +1,7 @@
 // add helper routines onto wd that make common operations easy to do
 // correctly
 
-const wd        = require('wd/lib/webdriver')
+const wd        = require('wd/lib/webdriver'),
       utils     = require('./utils.js'),
       timeouts  = require('./timeouts.js');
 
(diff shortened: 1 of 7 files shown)


jaredhirsch:
Looks great! The wclick changes make it super readable.

PR #2772, opened 2012-11-27
Update wclickIfExists to handle closed windows.

### LINKED ISSUE #2741: this-is-my-computer & selenium bug workaround
We could hack the time to ensure this never/always appears.

Or, add a wd-extension method: set implicit wait down to zero, poll on the remote browser at 1 sec intervals, handle the window-gone exception gracefully

shane-tomlinson:
Set the implict wait timeout to 0 and check for the window gone exception. This is to get around a bug in Webdriver where a request never returns when trying to query for an element on a window that has been closed.

fixes #2741

### DIFF
--- a/automation-tests/lib/wd-extensions.js
+++ b/automation-tests/lib/wd-extensions.js
@@ -1,7 +1,7 @@
 // add helper routines onto wd that make common operations easy to do
 // correctly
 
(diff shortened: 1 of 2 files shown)


jaredhirsch:
looks good. fingers crossed that this finally fixes the window gone hanging errors.

PR #2775, opened 2012-11-27
Issue 2674 invalidate via address info

ozten:
This is a catch all for invalid certs.

With the various transition patches, we'll also check and clear them. (Those patches haven't dropped yet.)

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -1022,7 +1022,39 @@ BrowserID.User = (function() {
         }, onFailure);
       }
     },
-
+    /**
+     * Checks for outdated certificates and clears them from storage.
+     * Returns original info or null if email info is now invalid.
+     * @param {string} email - Email address to check.
+     * @param {object} info - Output from addressInfo callback
+     * @return {object} or null
+     */
+    checkEmailIssuer: function(email, info) {
+      function clearCert(email, idInfo) {
+        delete idInfo.cert;
+        delete primaryAuthCache[email];
+        storage.addEmail(email, idInfo);
+      }
+      prepareDeps();
+      var identity = User.getStoredEmailKeypair(email);
+      if (identity && identity.cert && info && info.issuer) {
+        var prevIssuer;
+        try {
+          prevIssuer = jwcrypto.extractComponents(identity.cert).payload.iss;
+        } catch (e) {
+          // error parsing the certificate!  Maybe it's of an old/different
+          // format?  just delete it.
+          helpers.log("Looking for issuer, error parsing cert for"+ email +":" + e);
+          clearCert(email, identity);
+          return null;
+        }
+        if (prevIssuer && info.issuer !== prevIssuer) {
+          clearCert(email, identity);
+          return null;
+        }
+      }
+      return info;
+    },
     /**
      * Add an email address to an already created account.  Sends address and
      * keypair to the server, user then needs to verify account ownership. This
--- a/resources/static/dialog/js/modules/authenticate.js
+++ b/resources/static/dialog/js/modules/authenticate.js
@@ -68,6 +68,9 @@ BrowserID.Modules.Authenticate = (function() {
       addressInfo = info;
       dom.removeAttr(EMAIL_SELECTOR, 'disabled');
 
+      // Used for side effect of clearing certificate in local storage
+      user.checkEmailIssuer(email, info);
+
       if ("offline" === info.state) {
         self.close("primary_offline", info, info);
       }
--- a/resources/static/dialog/js/modules/pick_email.js
+++ b/resources/static/dialog/js/modules/pick_email.js
@@ -61,8 +61,13 @@ BrowserID.Modules.PickEmail = (function() {
     var valid = checkEmail.call(self, email);
     if (valid) {
       user.addressInfo(email, function (info) {
-        self.close("email_chosen", _.extend({ email: email }, info));
-      }, function() {
+        info = user.checkEmailIssuer(email, info);
+        valid = !!info;
+        if (valid)
+          self.close("email_chosen", _.extend({ email: email }, info));
+        else
+          self.notMe();
+      }, function(xhr) {
         //TODO: Handle xhr failure
       });
     }
--- a/resources/static/test/cases/common/js/user.js
+++ b/resources/static/test/cases/common/js/user.js
@@ -1322,6 +1322,47 @@
     );
   });
 
+  // JWCrypto relies on there being a random seed.  The random seed is
+  // gotten whenever network.withContext is called.  Since this is
+  // supposed to mock the cert clearing step...
+  // add a random seed to ensure that we can get our keypair.
+  jwcrypto.addEntropy("H+ZgKuhjVckv/H4i0Qvj/JGJEGDVOXSIS5RCOjY9/Bo=");
+  function makeCert(email, issuer, cb) {
+    var opts = {algorithm: "DS", keysize: 256};
+    jwcrypto.generateKeypair(opts, function (err, keyPair) {
+      ok(!err, "We can generate a keypair");
+      // We don't care about idPSecretKey... reuse user's keyPair here
+      var idpSecretKey = keyPair.secretKey;
+
+      jwcrypto.cert.sign({
+        publicKey: keyPair.publicKey,
+        principal: {email: email}
+      }, {
+        issuer: issuer, expiresAt: new Date() + 10000, issuedAt: new Date()}, null, idpSecretKey, function (err, cert) {
+          ok(!err, "We can create a cert");
+          cb(cert);
+        });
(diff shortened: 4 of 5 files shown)


resources/static/dialog/js/modules/pick_email.js
```diff
@@ -5,6 +5,7 @@ BrowserID.Modules.PickEmail = (function() {
   "use strict";
 
   var bid = BrowserID,
+      jwcrypto,
```
shane-tomlinson: Is jwcrypto used?

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -61,8 +62,13 @@ BrowserID.Modules.PickEmail = (function() {
...
     if (valid) {
       user.addressInfo(email, function (info) {
-        self.close("email_chosen", _.extend({ email: email }, info));
-      }, function() {
+	info = user.checkEmailIssuer(email, info);
+        valid = !!info;
+        if (valid)
+	  self.close("email_chosen", _.extend({ email: email }, info));
+	else
+	  self.notMe();
```
shane-tomlinson: Is the idea to log the user out if checkEmailIssuer returns null?

ozten: I chose to log them out, the ideal thing would be to re-provision them.

Problem - we've just removed their cert but they are far along in the state machine.

Solution - ?

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -68,6 +68,9 @@ BrowserID.Modules.Authenticate = (function() {
       addressInfo = info;
       dom.removeAttr(EMAIL_SELECTOR, 'disabled');
 
+      // Used for side effect of clearing certificate in local storage
+      user.checkEmailIssuer(email, info);
```
shane-tomlinson, line 72: Would it make more sense to push this call down into user.addressInfo?

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -61,8 +62,13 @@ BrowserID.Modules.PickEmail = (function() {
     var valid = checkEmail.call(self, email);
     if (valid) {
       user.addressInfo(email, function (info) {
-        self.close("email_chosen", _.extend({ email: email }, info));
-      }, function() {
+	info = user.checkEmailIssuer(email, info);
```
shane-tomlinson: This is why I am wondering if it makes sense to push the user.checkEmailIssuer call into user.addressInfo, it is called twice directly upon returning from user.addressInfo.

resources/static/common/js/user.js
```diff
@@ -1022,7 +1022,39 @@ BrowserID.User = (function() {
...
+     * Checks for outdated certificates and clears them from storage.
+     * Returns original info or null if email info is now invalid.
+     * @param {string} email - Email address to check.
+     * @param {object} info - Output from addressInfo callback
+     * @return {object} or null
+     */
+    checkEmailIssuer: function(email, info) {
+      function clearCert(email, idInfo) {
+        delete idInfo.cert;
+        delete primaryAuthCache[email];
```
shane-tomlinson, line 1035: Nice catch on deleting from the primaryAuthCache.

resources/static/dialog/js/modules/authenticate.js
ozten, line 72: Maybe. It would make addressInfo depend on the state of storage, not sure if that is a good thing.

resources/static/common/js/user.js
```diff
@@ -1022,7 +1022,39 @@ BrowserID.User = (function() {
...
       }
     },
+    /**
+     * Checks for outdated certificates and clears them from storage.
+     * Returns original info or null if email info is now invalid.
+     * @param {string} email - Email address to check.
+     * @param {object} info - Output from addressInfo callback
+     * @return {object} or null
+     */
+    checkEmailIssuer: function(email, info) {
```
shane-tomlinson, line 1032: This function could use a test to make sure it is operating how you expect it to.

resources/static/test/mocks/xhr.js
```diff
@@ -170,6 +170,8 @@ BrowserID.Mocks.xhr = (function() {
       "get /wsapi/address_info?email=testuser%40testuser.com valid": { type: "secondary", state: "known" },
       "get /wsapi/address_info?email=testuser%40testuser.com known_secondary": { type: "secondary", state: "known" },
       "get /wsapi/address_info?email=testuser%40testuser.com unknown_secondary": { type: "secondary", state: "unknown" },
+      // Not sure why this passes on dev, but I had to add testuser2 here...
```
shane-tomlinson, line 173: This is worrisome. Can you open an issue about this and assign it to me to investigate?

ozten, line 173: Sure https://github.com/mozilla/browserid/issues/2776

resources/static/dialog/js/modules/pick_email.js
shane-tomlinson: I think you should be able to do something like:

```
self.close("primary_user", infoBeforeItWasModifiedByCheckEmailIssuer);
```

This is only a theory.

resources/static/dialog/js/modules/pick_email.js
ozten: Nope, good catch!

resources/static/common/js/user.js
ozten, line 1032: Added a test case. It found that I should call prepareDeps... :)

resources/static/test/cases/common/js/user.js
```diff
@@ -1322,6 +1322,48 @@
...
+  jwcrypto.addEntropy("H+ZgKuhjVckv/H4i0Qvj/JGJEGDVOXSIS5RCOjY9/Bo=");
+  function makeCert(email, issuer, cb) {
+    var opts = {algorithm: "DS", keysize: 256};
+    jwcrypto.generateKeypair(opts, function (err, keyPair) {
+      ok(!err, "We can generate a keypair");
+      // We don't care about idPSecretKey... reuse user's keyPair here
+      var idpSecretKey = keyPair.secretKey;
+
+      jwcrypto.cert.sign({
+	publicKey: keyPair.publicKey,
```
seanmonstar: indentation

seanmonstar:
could you rebase out my commit? I'm afraid it's going to delete some stuff that I later decided should stay.

ozten:
I'll give it a shot. git hasn't been my friend lately...

seanmonstar:
```
git fetch mozilla
git rebase -i mozilla/primary-transition-ux
```

And delete my commit from the list.

ozten:
Conflict, will rebase again and resolve conflicts.

resources/static/test/mocks/xhr.js
seanmonstar, line 173: The same line exists also at 177?

resources/static/dialog/js/modules/pick_email.js
seanmonstar: @shane-tomlinson do you think this should be done? or is it something to look into in the future?

ozten:
Fixed whitespace issues that caused Travis failure.

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -61,8 +61,13 @@ BrowserID.Modules.PickEmail = (function() {
...
     if (valid) {
       user.addressInfo(email, function (info) {
-        self.close("email_chosen", _.extend({ email: email }, info));
-      }, function() {
+        info = user.checkEmailIssuer(email, info);
+        valid = !!info;
+        if (valid)
+          self.close("email_chosen", _.extend({ email: email }, info));
+        else
+          self.notMe();
```
shane-tomlinson, line 69: @seanmonstar - the rebase seems to have blown away our conversaion that was here.

General flow was:
- I asked if @ozten wanted to log the user out. 
- @ozten stated that a more desirable experience would be to re-provision the user.
- I suggested using 'self.close("primary_user")'
- @seanmonstar responded with - `@shane-tomlinson do you think this should be done? or is it something to look into in the future?`

self.close("primary_user", infoFromAddressInfo) _should_ cause the address to be re-provisioned if needed. Since checkEmailUser will remove the address' cert, the address will need to be re-provisioned. After the re-provisioning happens, "email_chosen" will be called again.

The hardest part is going to be making sure cancelling auth from the IdP still works correctly.

resources/static/test/mocks/xhr.js
shane-tomlinson, line 173: If that is on both lines, the first one should be able to be removed.

ozten:
Travis is green now.

https://travis-ci.org/mozilla/browserid/builds/3405108

PR #2777, opened 2012-11-28
Prefetch dialog components from the communication_iframe

fmarier:
When the browser is idle, it will start fetching and caching the dialog's Javascript, CSS and images. This works on IE9 and Firefox
(it's currently disabled in Chromium).

The trick is to make sure that the communication iframe is open while the user is busy reading the RP's site. When using the
Observer API, this happens automatically at the end of the navigator.id.watch() call.

More background on this feature here: http://feeding.cloud.geek.nz/2012/11/prefetching-resources-to-prime-browser.html

### DIFF
--- a/resources/views/communication_iframe.ejs
+++ b/resources/views/communication_iframe.ejs
@@ -4,6 +4,15 @@
 <head><title>non-interactive iframe</title>
   <meta http-equiv="X-UA-Compatible" content="IE=Edge"/>
   <meta charset="utf-8">
+  <%- cachify(util.format('/production/%s/dialog.js', locale), {tag_format: '<link rel="prefetch" href="%s">'}) %>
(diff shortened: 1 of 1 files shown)


fmarier:
I also thought it might be a good idea to [add a new helper function to connect-cachify](https://github.com/mozilla/connect-cachify/pull/23). In which case we could make this patch a bit more readable: https://github.com/fmarier/browserid/commit/e78a8a69307a6eb7bd159bbc0d9a8a80742860b5

jaredhirsch:
do we want to pull in this optimization? @shane-tomlinson, @fmarier, thoughts?

let's close this if we don't want to pull this code in right now, so that we can get to PR inbox zero (theoretically at least).

shane-tomlinson:
@6a68 - not yet, not until we slim down the overall bundle.

fmarier:
I'll submit a new pull request once we're ready to merge. I have to simplify this one anyways since Austin merged my connect-cachify PR.

PR #2778, opened 2012-11-28
Issue 2717 auth primary transition

ozten:
Detects transition_to_primary and shows dialog.

### DIFF
--- a/resources/static/dialog/js/misc/state.js
+++ b/resources/static/dialog/js/misc/state.js
@@ -215,6 +215,13 @@ BrowserID.State = (function() {
 
     handleState("user_confirmed", handleEmailConfirmed);
 
+    handleState("upgraded_primary_user", function (msg, info) {
+      user.completeTransition(info.email, function () {
+        info.state = 'known';
+        redirectToState("email_chosen", info);
+      }, complete(info.complete));
+    });
+
     handleState("primary_user", function(msg, info) {
       self.addPrimaryUser = !!info.add;
       var email = self.email = info.email,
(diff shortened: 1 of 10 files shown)


resources/static/dialog/js/misc/state.js
```diff
@@ -215,6 +215,13 @@ BrowserID.State = (function() {
 
     handleState("user_confirmed", handleEmailConfirmed);
 
+    handleState("upgraded_primary_user", function (msg, info) {
+      user.completeTransition(info.email, function () {
+        info.state = 'known';
+        redirectToState("email_chosen", info);
+      }, complete(info.complete));
```
seanmonstar: doesn't this execute info.complete immediately, instead of passing the complete function?

resources/static/dialog/js/misc/state.js
```diff
@@ -313,14 +320,21 @@ BrowserID.State = (function() {
...
           // Email is a primary and the cert is available - the user can log
           // in without authenticating with the IdP. All invalid/expired
           // certs are assumed to have been checked and removed by this
           // point.
           redirectToState("email_valid_and_ready", info);
-        }
+        } else if ("transition_to_primary" === info.state) {
+          startAction("doUpgradeToPrimaryUser", info);
+          complete(info.complete);
+	}
```
seanmonstar: indent

seanmonstar:
I know they're less fun to write, but we need tests for this new case.

resources/static/dialog/js/misc/state.js
ozten: Wow, great catch. Thanks.

ozten:
Updated based on feedback.

Added one test, but later in the release we can continue to add more vows and qunit tests.

PR #2780, opened 2012-11-28
not authed transition to secondary

### LINKED ISSUE #2770: Not authed / secondary / transition_to_secondary
If the user isn't logged in, and types an email that is known to us that used to be a primary, but is now a secondary, they need to be shown a transition explaining they need to use the password they created previously on Persona.

Not having a password previously is a different issue.

seanmonstar:
fixes #2770

### DIFF
--- a/resources/static/common/js/helpers.js
+++ b/resources/static/common/js/helpers.js
@@ -73,6 +73,11 @@
     }
   }
 
+  var emailRe = /.*@/;
+  function getDomainFromEmail(email) {
+    return String(email).replace(emailRe, "");
+  }
+
   _.extend(helpers, {
     /**
      * Get an email from a DOM element and validate it.
@@ -127,7 +132,9 @@
      * @method log
      * @param {string} msg
      */
-    log: log
+    log: log,
+
+    getDomainFromEmail: getDomainFromEmail
   });
 
 
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -264,7 +264,7 @@ BrowserID.User = (function() {
   }
 
   function getIdPName(addressInfo) {
-    return addressInfo.email.replace(/.*@/, "");
+    return helpers.getDomainFromEmail(addressInfo.email);
   }
 
   /**
--- a/resources/static/dialog/js/modules/authenticate.js
+++ b/resources/static/dialog/js/modules/authenticate.js
@@ -26,10 +26,20 @@ BrowserID.Modules.Authenticate = (function() {
       BODY_SELECTOR = "body",
       AUTHENTICATION_CLASS = "authentication",
       FORM_CLASS = "form",
+      AUTHENTICATION_LABEL = "#authentication_form label[for=authentication_email]",
+      EMAIL_LABEL = "#authentication_form .label.email_state",
+      TRANSITION_TO_SECONDARY_LABEL = "#authentication_form .label.transition_to_secondary",
+      PASSWORD_LABEL = "#authentication_form .label.password_state",
+      IDP_SELECTOR = "#authentication_form .authentication_idp_name",
       currentHint;
 
   function getEmail() {
     return helpers.getAndValidateEmail(EMAIL_SELECTOR);
+  }
+
+  function hasPassword(info) {
+    return (info && info.email && info.type === "secondary" && 
+      (info.state === "known" || info.state === "transition_to_secondary" ));
   }
 
   function initialState(info) {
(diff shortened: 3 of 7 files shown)


resources/static/common/js/helpers.js
```diff
@@ -127,7 +132,9 @@
      * @method log
      * @param {string} msg
      */
-    log: log
+    log: log,
+
+    getDomainFromEmail: getDomainFromEmail
   });
```
ozten, line 138: Thanks for making this reusable.

resources/static/common/js/user.js
```diff
@@ -264,7 +264,7 @@ BrowserID.User = (function() {
   }
 
   function getIdPName(addressInfo) {
-    return addressInfo.email.replace(/.*@/, "");
+    return helpers.getDomainFromEmail(addressInfo.email);
```
ozten, line 267: Good cleanup. [primary_offline.js](https://github.com/mozilla/browserid/blob/primary-transition-ux/resources/static/dialog/js/modules/primary_offline.js#L23) could use that too.

No worries about adding that to this patch.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -26,18 +26,29 @@ BrowserID.Modules.Authenticate = (function() {
...
+      (info.state === "known" || info.state === "transition_to_secondary" ));
+  }
+
   function initialState(info) {
     /*jshint validthis: true*/
     var self=this;
 
     self.submit = checkEmail;
-    if(info && info.email && info.type === "secondary" && info.state === "known") {
+    if(hasPassword(info)) {
```
ozten, line 50: Excellent.

resources/static/test/cases/dialog/js/modules/authenticate.js
```diff
@@ -25,6 +25,11 @@
       PASSWORD_SELECTOR = "#authentication_password",
       FORGOT_PASSWORD_SELECTOR = "#forgotPassword",
       BODY_SELECTOR = "body",
+      AUTHENTICATION_LABEL = "#authentication_form label[for=authentication_email]",
+      EMAIL_LABEL = "#authentication_form .label.email_state",
+      TRANSITION_TO_SECONDARY_LABEL = "#authentication_form .label.transition_to_secondary",
+      PASSWORD_LABEL = "#authentication_form .label.password_state",
+      IDP_SELECTOR = "#authentication_form .authentication_idp_name",
```
ozten, line 33: It's good that this is consistent with how the existing code is written, but it's a DRY violation to have these constants in two files.

Consider exporting these constants off of `BrowserID.Modules.Authenticate`.

seanmonstar, line 33: I agree. This made me twitch while I was pasting it.

resources/views/dialog.ejs
```diff
@@ -27,7 +27,10 @@
                     <ul class="inputs">
 
                         <li>
-                            <label for="authentication_email"><%= gettext('To sign in with Persona, please enter your email address.') %></label>
+                            <label for="authentication_email"></label>
+                            <span class="label email_state"><%= gettext('To sign in with Persona, please enter your email address.') %></span>
+                            <span class="label transition_to_secondary"><%- format(gettext('%(idp) no longer allows you to log into Persona with your %(idp) password. You must now use your existing Persona password to log in with this email address.'), { idp: '<strong class="authentication_idp_name"></strong>'}) %></span>
```
ozten, line 32: Shouldn't the IdP be interpolated into the span before format puts it into the string... or do we do this via a DOM manipulation?

It's nice that you've kept HTML out of the string... but it also increase the risk of a XSS. We'll need to be careful about validation with email addresses before we parse out a domain name.

resources/views/test.ejs
```diff
@@ -34,6 +34,10 @@
         <div id="authentication_form">
           <input type="text" id="authentication_email" />
           <input type="password" id="authentication_password" class="returning" />
+          <label for="authentication_email"></label>
```
ozten, line 38: Same DRY comments, clearly this would be very hard to fix. Duplicating HTML here seems like bad idea bears, but we won't change during this sprint.

resources/views/dialog.ejs
seanmonstar, line 32: Yea, when this view is rendered before the dialog shows, so we won't have the IdP name yet.

The email that the domain is received from had to first get through `validation.email`

ozten:
Ran tests locally, I get an error too.

seanmonstar:
I just ran them locally, and got all green. I rebased and force pushed, and now travis is green.

PR #2782, opened 2012-11-28
Make sure emails are trimmed of whitespace when a user signs up.

### LINKED ISSUE #2781: PhantomJS: /signin page. New email addresses are not trimmed before submission
This appears to be a PhantomJS only problem, other browsers are automatically trimming the address.

shane-tomlinson:
fixes #2781


jaredhirsch:
r+

PR #2787, opened 2012-11-28
Handle the different "Window gone" exceptions thrown by various browsers...

### LINKED ISSUE #2784: Selenium: Handle various browser exceptions for "Window gone"
Chrome returns:
"NoSuchWindow: findElement execution failed; The window could not be found"

Firefox returns:
"Exception: Window not found. The browser window may have been closed."

IE returns:
"Unable to find element with css selector" and a status of 13.

Opera always returns the exact same thing. No way to differentiate if the window is gone yet.

shane-tomlinson:
....

fixes #2784


jaredhirsch:
now is the winter of test discontent
made glorious summer by this pull request

PR #2789, opened 2012-11-28
transition_no_password flow

### LINKED ISSUE #2779: Not authed / secondary / transition_no_password
If the user isn't logged in, and types an email that is known to us that used to be a primary, but is now a secondary, and have never used a password with this email, they need to be shown a transition explaining they need to create a password for use with Persona.

seanmonstar:
fixes #2779

This shows a new message, but otherwise follows the reset password flow. @lloyd tells me that everything else should just work. Like magic.

### DIFF
--- a/resources/static/dialog/views/set_password.ejs
+++ b/resources/static/dialog/views/set_password.ejs
@@ -10,7 +10,13 @@
       <ul class="inputs">
           <li>
             <% if (!password_reset) { %>
-              <%= gettext("Your email address is new to us. Please create a password to use with Persona.") %>
+              <% if (transition_no_password) { %>
+                <%- format(gettext("%(idp) no longer allows you to log into Persona with your %(idp) password. Please create a new password to use with your Persona account."), {
+                  idp: "<strong>" + escape(domain) + "</strong>"
+                }) %>
+              <% } else { %>
+                <%= gettext("Your email address is new to us. Please create a password to use with Persona.") %>
+              <% } %>
             <% } %>
           </li>
 
--- a/resources/static/test/cases/dialog/js/modules/authenticate.js
+++ b/resources/static/test/cases/dialog/js/modules/authenticate.js
@@ -140,6 +140,17 @@
     testUserUnregistered();
   });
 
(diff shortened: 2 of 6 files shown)


seanmonstar:
Should have pointed out, this depends on https://github.com/seanmonstar/browserid/commit/0103c44901a951b18adb2eb30b5687d8bd2249c6#L0R137

seanmonstar:
Which probably means I'll need to rebase this once the other PR lands, cause this will likely make travis fall over (I forgot I had copied the function over while working on this PR)

resources/static/dialog/views/set_password.ejs
```diff
@@ -10,7 +10,13 @@
       <ul class="inputs">
           <li>
             <% if (!password_reset) { %>
-              <%= gettext("Your email address is new to us. Please create a password to use with Persona.") %>
```
ozten, line 13: Nice re-use of templates.

resources/static/test/cases/dialog/js/modules/authenticate.js
```diff
@@ -140,6 +140,17 @@
     testUserUnregistered();
   });
 
+  asyncTest("checkEmail with transition_no_password, transition message", function() {
+    $(EMAIL_SELECTOR).val("registered@testuser.com");
+    xhr.useResult("secondaryTransitionPassword");
```
ozten, line 145: I don't see a commit for `secondaryTransitionPassword` in the Mock xhr.

resources/static/test/cases/dialog/js/modules/set_password.js
```diff
@@ -58,6 +57,17 @@
...
+    controller.destroy();
+    createController({
+      email: "transition@password.no",
+      transition_no_password: true
+    });
+    var msg = $("#set_password .inputs li").text();
+    ok(msg.indexOf("no longer allows"), "transition message shown");
+    ok(msg.indexOf("password.no"), "message shows IdP domain");
+  });
+
```
ozten, line 70: Very cool.

ozten:
You rule the school! Travis is green, merging.

PR #2796, opened 2012-11-29
show different copy when transition_to_primary

### LINKED ISSUE #2794: Show new copy when transition_to_primary
blocks #2404

seanmonstar:
fixes #2794

### DIFF
--- a/resources/static/dialog/views/verify_primary_user.ejs
+++ b/resources/static/dialog/views/verify_primary_user.ejs
@@ -7,9 +7,16 @@
     <%= format(gettext("%s makes this easy"), [idpName]) %>
   </h3>
 
-  <p>
-    <%= format(gettext("Persona lets you use your %(emailProvider) account to sign into sites like %(aWebsite)."),
+  <p class="description">
+    <% if (transition_to_primary) { %>
+      <%=format(gettext("Your %(emailProvider) account has been upgraded to work even better with Persona. We will redirect you to %(emailProvider). Once you verify your account there, you will be signed into %(aWebsite)."), {
+        emailProvider: idpName,
+        aWebsite: siteName
+      })%>
+    <% } else { %>
+      <%= format(gettext("Persona lets you use your %(emailProvider) account to sign into sites like %(aWebsite)."),
                { emailProvider: idpName, aWebsite: siteName }) %>
+    <% } %>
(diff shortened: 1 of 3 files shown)


resources/static/dialog/views/verify_primary_user.ejs
```diff
@@ -8,8 +8,15 @@
   </h3>
 
   <p>
-    <%= format(gettext("Persona lets you use your %(emailProvider) account to sign into sites like %(aWebsite)."),
+    <% if (transition_to_primary) { %>
+      <%=format(gettext("Your %(emailProvider) account has been upgraded to work even better with Persona. We will redirect you to %(emailProvider). Once you verify your account there, you will be signed into %(aWebsite)."), {
```
ozten: NIce!

resources/static/test/cases/dialog/js/modules/verify_primary_user.js
```diff
@@ -187,6 +189,8 @@
       auth_url: "http://testuser.com/sign_in",
       ready: function r() {
         testElementExists("#verifyWithPrimary");
+        var text = $(".form_section h3 + p").text();
+        ok(text.indexOf("has been upgraded") !== -1, "shows upgraded transition message");
```
ozten: Cool. Patch looks good. I'll merge when the branch is green.

seanmonstar:
Rebased in the fix from lloyd, so that should fix up travis.

resources/static/test/cases/dialog/js/modules/verify_primary_user.js
shane-tomlinson: This is a pretty brittle selector, what about slapping ID or class on the p element in case it ever moves in the DOM?

ozten:
Frontend tests failed with [multiple calls to password_set](https://travis-ci.org/mozilla/browserid/jobs/3410840)

resources/static/test/cases/dialog/js/modules/verify_primary_user.js
seanmonstar: Yea, I should. Doing...

seanmonstar:
Rebased in latest again, and added a class like @shane-tomlinson suggested.

Issue #1501, opened 2012-04-25
A more secure verifier?


callahad:
Verifiers (ours included) should probably not accept an assertion signed by a fallback if the domain in question is a native Identity Provider.

This mitigates the security risk of browserid.org being compromised, as fraudulent assertions from browserid.org for users with a native Identity Provider would get rejected.

This question has come up two or three times on the mailing list.

benadida:
Yes, absolutely. Not probably. Absolutely :)

boinggg:
like

warner:
we aren't doing this already? I'll study the code to see where the pubkey lookup is happening and what we do when it succeeds.

One deep question: how "sticky" should it be? Specifically, how long should the verifier remember that it saw an IdP pubkey? If an IdP publishes their key, then gives up and removes it, at what point should the verifier start accepting certs from the fallback? (similarly, how long until the browser's implementation stops trying to talk to the primary and switches to the fallback?)

I've suggested before that the keys published in `/.well-known/browserid` would benefit from expiration dates.. I think they would help here too, and would behave just like DNS TTLs or HTTP cache headers. If an IdP ever publishes a pubkey, even for just a moment, they should expect browsers and verifiers to pay attention to it until the latest expiration date passes.

We should probably also study the handoff behavior. If at T=0 the browser checks the IdP, sees no `.well-known/browserid`, uses the fallback, and generates an assertion, then at T=1 the IdP starts publishing their pubkey, then at T=2 the verifier checks that assertion, should it fail? And is there a feedback mechanism to cause the browser to try again, starting with the IdP? For assertions, the window of possible breakage is pretty small (things might be broken for the first 5 minutes after the IdP starts offering browserid service, but after that they'll be ok). For browser certificates, it might be more like a day. Worth studying, at any rate.

benadida:
@warner we're not doing this yet because we haven't figured out the upgrade behavior completely... but it's worth a careful consideration if you have time!

seanmonstar:
@warner is this important enough for beta1? And can it be done quickly enough? If not, we should tag it beta2

warner:
I don't think it can be done quickly enough: I suspect that getting the handoff right will involve revisiting a lot of code (in particular to have the browser implementation respond to expired pubkeys by re-fetching a new key, and maybe copy some of the verification checks into the browser so it can predict what the verifier will do and fix any problems before delivering the assertion).

[...]

PR #2797, opened 2012-11-29
Implement a Great User Experience as email providers implement persona

lloyd:
_NOTE:_  This pull request is not ready for merge, but is hungry for review.

In order to turn on bigtent and to prepare for popular email providers to implement persona support, these changes attempt to make UX great for users with email addresses that gain (or loose) "primary" persona support.  Language and screens are added, and more.
## 10k Foot Overview

This change-set:
- Introduces new screens and new language for all transition cases (still in progress) to help users understand what's going on.
- Adds mechanisms to allow for zero-downtime switch flipping for primaries
- Removes much state from the frontend and letting the server manage it
## API Changes (@francois and @zaach, please review?)
- list_emails now returns a list of emails without any additional information
- address_info now returns a "state" for emails, that tells the frontend what flow it should display to the user
- address_info now returns the current issuer for the domain
- update_password is now callable by a `assertion` authenticated user (does not require pw auth)
- used_address_as_primary is added, which is a means for the frontend to tell the backend that the user has used an address as a primary (causing a transition UX to not be displayed subsequently)
- session_context now has a `has_password` key which the frontend can use to know definitively whether the user has set a password
## Front End Work: (@shane-tomlinson, please review?)
- several new screens added
- we no longer store email `type` nor `verified` in local storage, everywhere where we were using these from local storage, we are now getting them from address_info
- address_info is called Every Time the user picks an email address, which is the point where we determine what state to transition to
- everywhere where we determined if the user has a password by looking for a local secondary address, we now defer to `session_context`'s `has_password` property
## Database Changes (@warner and @gene1wood, please review?)
- `lastUsedAs` field is added, `type` field is removed.  they both contain either `primary` or `secondary`.  the difference is for `type` it was not expected to change, while `lastUsedAs` is the last form the email address was in when the user used it.  `lastUsedAs` may be seeded from `type` for the purposes of database migration
- `idp` table added.  This holds recently seen idps and a timestamp of when they were last seen.  This allows us to consider an idp broken when they stop serving `.well-known` without an explicit signal.
## Security Questions (@warner and @francois please review?)
- new information leaks in `address_info`
- loosing of require auth level in `update_password`
## Spec changes (@callahad and @warner and @benadida, please review?)
- new parsing of .well-known documents
- addition of `disabled` property in these documents, which is a signal to the _implementation_ that it should fall back to secondary and the primary wishes to opt out
- new SHOULD behavior of verifier: it should verify assertions whenever `public-key` is defined, but ignore `disabled`


jaredhirsch:
@lloyd @jrgm: suppose we wanted to add some browser automation to test this new behavior.
1. Following our pattern of just verifying the happy paths, what might be the most useful tests to add?
2. Given that we're testing client behavior across changes in IdP state, what would we need to do to create an IdP useful for testing? Could the new test features be hacked into eyedee.me, or do we need a new service for just these tests?

lloyd:
@6a68 the 6 user stories here: https://id.etherpad.mozilla.org/primary-transition-ux-swarm

they are _not_ in priority order.

I hope to start implementing these tomorrow, and I believe we'll have to make a new server to be able to test them. I'd love to vet the approach with you guys. /cc @jrgm

bin/browserid
```diff
@@ -126,6 +128,17 @@ db.open(config.get('database'), function (error) {
     return setTimeout(function() { process.exit(1); }, 0);
   }
 
+  // As we detect primaries are online in the browserid process,
+  // we must record this.  This allows us to handle cases where
+  // IdP's transition online or off.
+  primary.events.on('idp_seen', function(domain) {
+    wsapi.requestToDBWriter({
+      path: '/wsapi/saw_idp?domain=' + domain
```
benadida: suggest URI-encoding the parameter, just to be safe. Would be better to use a URI encoding library in general.

lib/well-known-parser.js
```diff
@@ -0,0 +1,78 @@
...
+  // or supported: true
+
+  // is this a delegation document?
+  if (doc.authority && typeof doc.authority === 'string') {
+    return {
+      type: "delegation",
+      authority: doc.authority
+    };
+  }
+  // if a non-string authority is specified, we ignore it.
```
benadida: should we be stricter here, just to prevent weird code paths? If there's an authority field and it's not a string, maybe that's a malformed well-known?

lib/well-known-parser.js
```diff
@@ -0,0 +1,78 @@
...
+  // there are three main types of support documents
+  // 1. "supported" - declares the domain is a browserid authority,
+  //    contains public-key, authentication, and provisioning
+  // 2. "delegation" - declares the domain allows a different domain
+  //    to be authoritative for it.
+  // 3. "disable" - domain declares explicitly that it wants a secondary
+  //    to be authoritative.
+
+  // is this a "disable" document?
+  if (doc.supported === 'false' || doc.supported === false) {
```
benadida: I thought I saw discussion that this field should be called "disabled: true". Now, I prefer staying away from double negatives and doing what this code does, I just want to make sure we're in agreement on what that should be.

lib/wsapi/cert_key.js
```diff
@@ -41,18 +41,20 @@ exports.process = function(req, res) {
...
       // to cause them to be forwarded.
       req.body = req.params;
-
       forward(keysigner, req, res, function(err) {
         if (err) {
           logger.error("error forwarding request to keysigner: " + err);
           httputils.serverError(res, "can't contact keysigner");
           return;
         }
+        db.updateEmailLastUsedAs(req.params.email, 'secondary', function (err) {
```
benadida: this introduces a DB write on every cert_key, which will likely change the performance profile of our app, since that used to be read-only. Is there an easy way to determine those rare cases when that update is needed? Stashing something in the req data structure from when that user row was read?

lib/wsapi/address_info.js
```diff
@@ -8,15 +8,48 @@ primary = require('../primary.js'),
...
+    }
+  },
+  false: {
+    "primary": {
+      "primary": "known",
+      "secondary": "transition_no_password"
+    },
+    "secondary": {
+      "primary": "transition_to_primary",
+      "secondary": "tranistion_no_password"
```
fmarier: typo: tranIStion => tranSItion

fmarier:
I'm not all that confident about my review because I wasn't sure how to review my part. I initially looked through the commits, but found that some of them were undoing/modifying others and so I gave up on that and instead
tried to find the files I cared about in the diff and examined how they changed.

So I stared at the changes for a while and then went back to the specific changes that @lloyd highlighted. I came up with a list of things I wanted to verify and then dived back into the diff:
1. Examine the `update_password` code to make sure that the old password check is working fine.
2. Make sure that `session_context` doesn't expose the hashed password through its new has_password.
3. Review the new `used_address_as_primary` API call.
4. Look at the new info returned by `address_info` to make sure it's not a problematic leak.

It all seemed fine to me. So I would give a small r+ for the parts I looked at. But hopefully others will look at it too.

shane-tomlinson:
Front end tests need to pass:

```
➜  ~PERSONA_ROOT git:(primary-transition-ux) ✗ ps-test-frontend 
~/development/browserid ~/development/browserid
· · · · router (48168): info: forwarding request: /common/js/error-messages.js -> http://127.0.0.1:10010/common/js/error-messages.js
loading
Test multiple calls to password_set for stage_email, simulate throttling - call doAddEmail with correct email for each died, exception and test follows
TypeError: Result of expression 'actions.info[actionName]' [undefined] is not an object.
function () {
    xhr.setContextInfo("auth_level", "password");
    testStagingThrottledRetry("stage_email", "doStageEmail");
  }
currently unable to focus elements, focus check skipped - try focusing the unit test page
currently unable to focus elements, focus check skipped - try focusing the unit test page
currently unable to focus elements, focus check skipped - try focusing the unit test page
currently unable to focus elements, focus check skipped - try focusing the unit test page
'waitFor()' finished in 41351ms.
dialog/js/misc/state: multiple calls to password_set for stage_email, simulate throttling - call doAddEmail with correct email for each (2, 0, 2)
   - doStageEmail called
   - Died on test #2: Result of expression 'actions.info[actionName]' [undefined] is not an object. - { "message": "Result of expression 'actions.info[actionName]' [undefined] is not an object.", "line": 46, "sourceId": 218710184, "sourceURL": "http://127.0.0.1:10002/test/cases/dialog/js/misc/state.js", "expressionBeginOffset": 1380, "expressionCaretOffset": 1404, "expressionEndOffset": 1409, "name": "TypeError" }
Tests completed in 41164 milliseconds.
1413 tests of 1415 passed, 2 failed.
✗ · · 

  frontend unit tests
    ✗ pass!
      » expected 0,
    got  1 (===) // vows.js:91
✗ Broken » 6 honored ∙ 1 broken (45.696s)
~/development/browserid
➜  ~PERSONA_ROOT git:(primary-transition-ux) ✗ 
```

There look to be two problems. 

1) A tested action looks like it is no longer defined.
2) password_set is being triggered twice when doStageEmail completes.

resources/static/common/js/helpers.js
```diff
@@ -127,7 +132,9 @@
      * @method log
      * @param {string} msg
      */
-    log: log
+    log: log,
+
+    getDomainFromEmail: getDomainFromEmail
```
shane-tomlinson, line 137: I can't find a test for this.

resources/static/common/js/network.js
```diff
@@ -749,6 +752,28 @@ BrowserID.Network = (function() {
...
+    },
+
+    /**
+     * Mark the transition of this email as having been completed.
+     * @method usedAddressAsPrimary
+     * @param {string} [email] - The email that transitioned.
+     * @param {function} [onComplete] - Called whenever complete.
+     * @param {function} [onFailure] - Called on XHR failure.
+     */
+    usedAddressAsPrimary: function(email, onComplete, onFailure) {
```
shane-tomlinson, line 743: This is missing a test.

resources/static/common/js/network.js
```diff
@@ -455,6 +455,9 @@ BrowserID.Network = (function() {
           newpass: newPassword
         },
         success: function(status) {
+          // successful change of password will upgrade a session to password
+          // level auth
+          if (status) auth_status = "password";
```
shane-tomlinson, line 439: This should have some tests in resources/static/test/cases/common/js/network.js between lines 601 and 617 to make sure this is operating as expected and somebody else doesn't break those expectations in the future.

resources/static/common/js/network.js
```diff
@@ -619,10 +622,10 @@ BrowserID.Network = (function() {
           // storage.
           // update our local storage map of email addresses to user ids
           if (userid) {
-            storage.updateEmailToUserIDMapping(userid, _.keys(emails));
+            storage.updateEmailToUserIDMapping(userid, emails.emails);
```
shane-tomlinson, line 604: Rename the "emails" parameter to the "success" callback to "info" or "resp" - something that better conveys what the object is. It is no longer emails only.

resources/static/test/cases/common/js/user.js
```diff
@@ -463,7 +462,9 @@
   });
 
   asyncTest("canSetPassword with only primary addresses - expect false", function() {
```
shane-tomlinson, line 464: This test needs remained to say something about the user not having a password.

resources/static/test/cases/common/js/user.js
```diff
@@ -472,7 +473,9 @@
   });
 
   asyncTest("canSetPassword with secondary addresses - expect true", function() {
```
shane-tomlinson, line 475: This test needs renamed to say something about the user having a password.

resources/static/test/cases/common/js/user.js
```diff
@@ -572,23 +574,8 @@
...
-      testObjectValuesEqual(status, {
-        success: false,
-        reason: "verified_email"
-      });
-
-      start();
-    }, testHelpers.unexpectedXHRFailure);
-  });
-
   asyncTest("requestEmailReverify with owned unverified email - false status", function() {
```
shane-tomlinson, line 556: Rename this test to remove the "unverified"

resources/static/test/cases/common/js/user.js
```diff
@@ -572,23 +574,8 @@
     );
   });
 
-  asyncTest("requestEmailReverify with owned verified email - false status", function() {
```
shane-tomlinson, line 575: Has the concept of verified vs unverified emails gone away completely, even from the backend? What happens if the user requests an email re-verification but they have not verified the email in the first place?

If this concept still exists, the xhr mock should be updated and both cases handled in tests.

resources/static/common/js/user.js
```diff
@@ -830,30 +800,10 @@ BrowserID.User = (function() {
...
-
-            persistEmail({
-              email: email,
-              type: emailInfo.type || "secondary",
-              verified: emailInfo.verified
-            });
-          });
-
-          // update the type and verified status of stored emails
-          _.each(emails_to_update, function(email) {
```
shane-tomlinson, line 843: Since this is removed because there is no verified or type field, line 794 "emails_to_update" should be removed as well.

resources/static/common/js/user.js
```diff
@@ -1022,7 +972,39 @@ BrowserID.User = (function() {
...
     },
-
+    /**
+     * Checks for outdated certificates and clears them from storage.
+     * Returns original info or null if email info is now invalid.
+     * @param {string} email - Email address to check.
+     * @param {object} info - Output from addressInfo callback
+     * @return {object} or null
+     */
+    checkEmailIssuer: function(email, info) {
```
shane-tomlinson, line 972: Thanks for adding the test @ozten!

resources/static/common/js/user.js
```diff
@@ -1022,7 +972,39 @@ BrowserID.User = (function() {
...
+        try {
+          prevIssuer = jwcrypto.extractComponents(identity.cert).payload.iss;
+        } catch (e) {
+          // error parsing the certificate!  Maybe it's of an old/different
+          // format?  just delete it.
+          helpers.log("Looking for issuer, error parsing cert for"+ email +":" + e);
+          clearCert(email, identity);
+          return null;
+        }
+        if (prevIssuer && info.issuer !== prevIssuer) {
```
shane-tomlinson: If I understand this correctly, this happens on an email by email basis?

If a user has two emails from the same domain that just got upgraded from secondary->primary, only the email that a user selected will have its cert updated?

Does this happen for silent assertions from the .watch API as well?

resources/static/common/js/user.js
```diff
@@ -1323,27 +1298,6 @@ BrowserID.User = (function() {
     },
 
     /**
-     * Check if the user has any secondary addresses.
-     * @method hasSecondary
-     * @param {function} onComplete - called with true if user has at least one
-     * email address, false otw.
-     * @param {function} onFailure - called on XHR failure.
-     */
-    hasSecondary: function(onComplete, onFailure) {
```
shane-tomlinson, line 1332: Nice cleanup

resources/static/common/js/user.js
```diff
@@ -1394,6 +1348,14 @@ BrowserID.User = (function() {
       } else {
         complete(onFailure, "user is not authenticated");
       }
+    },
+
+    /**
+     * Mark the transition state of this user as having been completed.
+     * @method usedAddressAsPrimary
+     */
+    usedAddressAsPrimary: function(email, onComplete, onFailure) {
```
shane-tomlinson, line 1352: This needs a test

resources/static/dialog/js/misc/state.js
```diff
@@ -192,6 +192,12 @@ BrowserID.State = (function() {
       complete(info.complete);
     });
 
+    handleState("transition_no_password", function(msg, info) {
+      self.resetPasswordEmail = info.email;
```
shane-tomlinson: Does this only occur during a password reset?

resources/static/dialog/js/misc/state.js
```diff
@@ -216,10 +222,18 @@ BrowserID.State = (function() {
 
     handleState("user_confirmed", handleEmailConfirmed);
 
+    handleState("upgraded_primary_user", function (msg, info) {
```
shane-tomlinson, line 234: New states in general should have tests just to make sure things are operating as expected.

resources/static/dialog/js/misc/state.js
```diff
@@ -294,7 +315,7 @@ BrowserID.State = (function() {
 
     handleState("email_chosen", function(msg, info) {
       var email = info.email,
-          idInfo = storage.getEmail(email);
+          record = storage.getEmail(email);
```
shane-tomlinson, line 327: Much better name!

resources/static/dialog/js/misc/state.js
```diff
@@ -302,19 +323,28 @@ BrowserID.State = (function() {
         complete(info.complete);
       }
 
-      if (!idInfo) {
+      if (!record) {
         throw new Error("invalid email");
       }
 
-      mediator.publish("kpi_data", { email_type: idInfo.type });
+      mediator.publish("kpi_data", { email_type: info.type });
```
shane-tomlinson, line 339: Since emails no longer have a stored type, we will now have to do a user.addressInfo to get the type to send to the KPI data.

shane-tomlinson, line 339: Scratch that comment, the type comes in from info, not the record.

shane-tomlinson, line 339: Finally, the KPI backend will have to be updated to handle these new states.

resources/static/dialog/js/misc/state.js
```diff
@@ -302,19 +323,28 @@ BrowserID.State = (function() {
...
+
+      if (info.state && 'offline' === info.state) {
+        redirectToState("primary_offline", info);
+      }
+      else if (info.type === "primary") {
+        // issuer MUST have changed... clear certs
+        if ("transition_to_primary" === info.state && record.cert) delete record.cert;
```
shane-tomlinson: Is this already taken care of by user.checkEmailIssuer? If not, should it be taken care of using that instead of having two places that delete record.cert?

resources/static/dialog/js/misc/state.js
```diff
@@ -302,19 +323,28 @@ BrowserID.State = (function() {
...
 
-      if (!idInfo) {
+      if (!record) {
         throw new Error("invalid email");
       }
 
-      mediator.publish("kpi_data", { email_type: idInfo.type });
+      mediator.publish("kpi_data", { email_type: info.type });
+
+      if (info.state && 'offline' === info.state) {
```
shane-tomlinson, line 341: These new sub-states need tests.

resources/static/dialog/js/modules/actions.js
```diff
@@ -148,6 +148,10 @@ BrowserID.Modules.Actions = (function() {
       startService("verify_primary_user", info);
     },
 
+    doUpgradeToPrimaryUser: function(info) {
```
shane-tomlinson, line 151: Tests needed for both of these new actions.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -68,11 +79,19 @@ BrowserID.Modules.Authenticate = (function() {
```
shane-tomlinson, line 78: Line 69 shows:

```
if(info && info.type) {
 onAddressInfo(info);
}
else {
```

Since types are no longer stored on the email address, I believe this can safely be removed.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -68,11 +79,19 @@ BrowserID.Modules.Authenticate = (function() {
...
       dom.removeAttr(EMAIL_SELECTOR, 'disabled');
 
-      if(info.type === "primary") {
+      // Used for side effect of clearing certificate in local storage
+      user.checkEmailIssuer(email, info);
+
+      if ("offline" === info.state) {
+        self.close("primary_offline", info, info);
+      }
+      else if("primary" === info.type) {
```
shane-tomlinson: Super pedantic comment - `if(` should be `if (`

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -68,11 +79,19 @@ BrowserID.Modules.Authenticate = (function() {
       addressInfo = info;
       dom.removeAttr(EMAIL_SELECTOR, 'disabled');
 
-      if(info.type === "primary") {
+      // Used for side effect of clearing certificate in local storage
+      user.checkEmailIssuer(email, info);
+
+      if ("offline" === info.state) {
+        self.close("primary_offline", info, info);
```
shane-tomlinson, line 83: transitionNoPassword has a test, sadly this does not.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -68,11 +79,19 @@ BrowserID.Modules.Authenticate = (function() {
...
+      }
+      else if("primary" === info.type) {
         self.close("primary_user", info, info);
       }
-      else if(info.known) {
```
shane-tomlinson, line 74: Somebody should have yelled at me for the `if(` instead of `if (` previously

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -136,6 +166,7 @@ BrowserID.Modules.Authenticate = (function() {
     var self=this;
     if (!dom.is(EMAIL_SELECTOR, ":disabled")) {
       self.publish("enter_email");
+      dom.setInner(AUTHENTICATION_LABEL, dom.getInner(EMAIL_LABEL));
```
shane-tomlinson, line 166: If the labels are already in the DOM, why not just show the new label instead of replace the contents of another label?

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -148,13 +179,18 @@ BrowserID.Modules.Authenticate = (function() {
 
     dom.setInner(PASSWORD_SELECTOR, "");
 
-    self.publish("enter_password", addressInfo);
     self.submit = authenticate;
+    var labelSelector = (addressInfo.state === "known") ? PASSWORD_LABEL : TRANSITION_TO_SECONDARY_LABEL;
+    if (labelSelector === TRANSITION_TO_SECONDARY_LABEL) {
+      dom.setInner(IDP_SELECTOR, helpers.getDomainFromEmail(addressInfo.email));
+    }
+    dom.setInner(AUTHENTICATION_LABEL, dom.getInner(labelSelector));
```
shane-tomlinson, line 184: Same comment as above w.r.t. labels.

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -60,7 +60,16 @@ BrowserID.Modules.PickEmail = (function() {
...
-      self.close("email_chosen", { email: email });
+      user.addressInfo(email, function (info) {
+        info = user.checkEmailIssuer(email, info);
+        valid = !!info;
+        if (valid)
+          self.close("email_chosen", _.extend({ email: email }, info));
+        else
+          self.notMe();
+      }, function(xhr) {
+        //TODO: Handle xhr failure
```
shane-tomlinson: Handle the xhr failure.

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -60,7 +60,16 @@ BrowserID.Modules.PickEmail = (function() {
...
     var valid = checkEmail.call(self, email);
     if (valid) {
-      self.close("email_chosen", { email: email });
+      user.addressInfo(email, function (info) {
+        info = user.checkEmailIssuer(email, info);
+        valid = !!info;
+        if (valid)
+          self.close("email_chosen", _.extend({ email: email }, info));
+        else
+          self.notMe();
```
shane-tomlinson: I am still wondering if you want to use "primary_email" here instead of notMe, that would force the email address to be re-provisioned with the updated IdP.

resources/static/test/cases/dialog/js/modules/set_password.js
```diff
@@ -10,8 +10,7 @@
       testHelpers = bid.TestHelpers,
       testElementExists = testHelpers.testElementExists,
       testElementNotExists = testHelpers.testElementDoesNotExist,
-      register = testHelpers.register,
-      controller;
```
shane-tomlinson, line 14: Good cleanup of dup variables.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -26,18 +26,29 @@ BrowserID.Modules.Authenticate = (function() {
       BODY_SELECTOR = "body",
       AUTHENTICATION_CLASS = "authentication",
       FORM_CLASS = "form",
+      AUTHENTICATION_LABEL = "#authentication_form label[for=authentication_email]",
+      EMAIL_LABEL = "#authentication_form .label.email_state",
```
shane-tomlinson, line 30: If the labels are not going to be shown but written into the DOM dynamically, is it easier just to place the strings here surrounded by a gettext?

resources/static/dialog/js/modules/authenticate.js
shane-tomlinson, line 166: OTOH, if the labels in the DOM are never shown, why not just embed the contents here and surround them with gettext?

resources/static/dialog/js/misc/state.js
```diff
@@ -216,10 +222,18 @@ BrowserID.State = (function() {
 
     handleState("user_confirmed", handleEmailConfirmed);
 
+    handleState("upgraded_primary_user", function (msg, info) {
+      user.completeTransition(info.email, function () {
```
shane-tomlinson: Unless info.complete handles it, if this XHR request fails, no error message will ever be displayed to the user. One of the bad things about putting user.\* calls inside of state.js.

resources/static/dialog/js/modules/primary_offline.js
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
shane-tomlinson, line 1: This module has no tests.

resources/static/dialog/js/modules/primary_offline.js
```diff
@@ -0,0 +1,32 @@
...
+  function startDialogOver() {
+    /*jshint validthis:true*/
+    this.close("cancel_state");
+  }
+
+  var Module = bid.Modules.PageModule.extend({
+    start: function(options) {
+      options = options || {};
+      var self = this,
+          parts = options.email ? options.email.split('@') : ['', ''];
```
shane-tomlinson: Use the generalized helpers.getDomainFromEmail here. This was added in this PR and is found in resources/static/common/js/helpers.js

resources/static/dialog/js/modules/required_email.js
```diff
@@ -121,31 +121,32 @@ BrowserID.Modules.RequiredEmail = (function() {
...
+            // authenticated to the "password" level.
+            showTemplate({
+              signin: true,
+              password: auth_level !== "password",
+              secondary_auth: secondaryAuth,
+              cancelable: options.cancelable
+            });
+            ready();
+          }
+          else if(storedEmailInfo && info.type === "primary" && storedEmailInfo.cert) {
```
shane-tomlinson, line 137: Unfortunately, this is a left over from the old required_email days. I believe this module is _only_ used to authenticate a user who is logged in at the assertion level who picks a secondary email address and must enter their password. If this is the case, this branch may be unnecessary.

resources/static/dialog/js/modules/upgrade_to_primary_user.js
```diff
@@ -0,0 +1,69 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+BrowserID.Modules.UpgradeToPrimaryUser = (function() {
+  "use strict";
+
+  var bid = BrowserID,
+      sc,
+      win,
```
shane-tomlinson, line 9: win, add, email, and auth_url are never used outside of the templates and should be removed.

resources/static/dialog/js/modules/upgrade_to_primary_user.js
```diff
@@ -0,0 +1,69 @@
...
+      self.renderDialog("upgrade_to_primary_user", {
+        email: data.email,
+        auth_url: data.auth,
+        requiredEmail: data.requiredEmail || false,
+        personaTOSPP: data.personaTOSPP,
+        siteName: data.siteName,
+        idpName: data.idpName
+      });
+
+      if (data.siteTOSPP) {
```
shane-tomlinson, line 47: Does the site TOS/PP show on the upgrade screen? Will this already be displayed from the main authentication or pick email screen? If so, does this need to be here?

shane-tomlinson, line 47: It looks like this will have already been taken care of in https://github.com/mozilla/browserid/pull/2797/files#L33R73

resources/static/dialog/js/modules/verify_primary_user.js
```diff
@@ -48,20 +58,28 @@ BrowserID.Modules.VerifyPrimaryUser = (function() {
...
-        siteName: data.siteName,
-        idpName: data.idpName
-      });
-
-      if (data.siteTOSPP) {
-        dialogHelpers.showRPTosPP.call(self);
-      }
+      user.addressInfo(email, function onSuccess(info) {
+        if (showsPrimaryTransition(info.state)) {
+          self.renderDialog("verify_primary_user", {
```
shane-tomlinson, line 63: When is "upgrade_to_primary_user" shown and when is "verify_primary_user" shown?

resources/static/dialog/views/set_password.ejs
```diff
@@ -10,7 +10,13 @@
       <ul class="inputs">
           <li>
             <% if (!password_reset) { %>
-              <%= gettext("Your email address is new to us. Please create a password to use with Persona.") %>
+              <% if (transition_no_password) { %>
+                <%- format(gettext("%(idp) no longer allows you to log into Persona with your %(idp) password. Please create a new password to use with your Persona account."), {
+                  idp: "<strong>" + escape(domain) + "</strong>"
```
shane-tomlinson, line 15: Can this lead to doubly escaped domains ever? Are emails already escaped when they are entered by the user and then checked on the backend when they are entered into the database?

resources/static/pages/js/manage_account.js
```diff
@@ -129,6 +129,9 @@ BrowserID.manageAccount = (function() {
     function changePassword() {
       user.changePassword(oldPassword, newPassword, function(status) {
         if(status) {
+          // a successful password change always puts the session in a
+          // 'password' level authentication
+          authLevel = "password";
```
shane-tomlinson, line 134: authLevel doesn't seem to actually be used anywhere anymore and should be taken out.

resources/static/pages/js/manage_account.js
```diff
@@ -164,21 +167,6 @@ BrowserID.manageAccount = (function() {
       tooltip.showTooltip("#tooltipPasswordLength");
       complete(oncomplete, false);
     }
-    else if(authLevel !== "password") {
```
shane-tomlinson, line 167: Awesome cleanup with this taken care of at a lower level.

resources/static/pages/js/manage_account.js
```diff
@@ -193,19 +181,10 @@ BrowserID.manageAccount = (function() {
...
-
-  function getSecondary() {
-    var emails = storage.getEmails();
-
-    for(var key in emails) {
-      if(emails[key].type === "secondary") {
-        return key;
-      }
-    }
+    network.withContext(function(ctx) {
```
shane-tomlinson, line 184: This check occurs in several places throughout the code, a user.hasPassword function would be a useful abstraction.

resources/static/test/cases/dialog/js/modules/verify_primary_user.js
```diff
@@ -33,97 +35,116 @@
     }
   });
 
-  test("personaTOSPP true, requiredEmail: true - show TOS/PP", function() {
+  asyncTest("personaTOSPP true, requiredEmail: true - show TOS/PP", function() {
```
shane-tomlinson, line 38: I do not believe requiredEmail is ever used anymore and this is a remnant of that era.

resources/static/test/mocks/xhr.js
```diff
@@ -152,22 +157,31 @@ BrowserID.Mocks.xhr = (function() {
...
-      "get /wsapi/address_info?email=registered%40testuser.com throttle": { type: "secondary", known: true },
-      "get /wsapi/address_info?email=registered%40testuser.com primary": { type: "primary", auth: "https://auth_url", prov: "https://prov_url" },
-      "get /wsapi/address_info?email=registered%40testuser.com mustAuth": { type: "secondary", known: true },
+      "get /wsapi/address_info?email=registered%40testuser.com valid": { type: "secondary", state: "known" },
+      "get /wsapi/address_info?email=registered%40testuser.com known_secondary": { type: "secondary", state: "known" },
+      "get /wsapi/address_info?email=registered%40testuser.com throttle": { type: "secondary", state: "known" },
+      "get /wsapi/address_info?email=registered%40testuser.com primary": { type: "primary", state: "known", auth: "https://auth_url", prov: "https://prov_url" },
+      "get /wsapi/address_info?email=registered%40testuser.com mustAuth": { type: "secondary", state: "known" },
+      "get /wsapi/address_info?email=registered%40testuser.com secondaryTransition": { type: "secondary", state: "transition_to_secondary" },
```
shane-tomlinson, line 171: Extra mocks, woot!

resources/static/test/mocks/xhr.js
```diff
@@ -223,6 +237,9 @@ BrowserID.Mocks.xhr = (function() {
       var responseKey = request.type + " " + request.url + " " + responseName,
           response = xhr.responses[responseKey],
           typeofResponse = typeof response;
+      // Unit tests busted with no feedback? Un-comment this bad boy and look for 'typeofResponse=undefined'
+      // Pull Request #2760 will automate this...
+      //console.log('responseKey=' + responseKey + ' response=' + response + ' typeofResponse=' + typeofResponse);
```
shane-tomlinson, line 242: A rebase off of dev should have this functionality built in now.

resources/views/dialog.ejs
```diff
@@ -27,7 +27,10 @@
                     <ul class="inputs">
 
                         <li>
-                            <label for="authentication_email"><%= gettext('To sign in with Persona, please enter your email address.') %></label>
+                            <label for="authentication_email"></label>
+                            <span class="label email_state"><%= gettext('To sign in with Persona, please enter your email address.') %></span>
```
shane-tomlinson, line 31: Going along with the comments up above, if these are in the DOM, why not just show them when necessary instead of copying the contents to the label?

resources/static/dialog/views/set_password.ejs
shane-tomlinson, line 15: @warner, @fmarier - Can you chime in here?

resources/static/common/js/error-messages.js
```diff
@@ -50,6 +50,10 @@ BrowserID.Errors = (function(){
       title: "Checking Script Version"
     },
 
+    usedAddressAsPrimary: {
```
shane-tomlinson: This is added but never used.

lib/wsapi/saw_idp.js
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
shane-tomlinson, line 1: I don't see this wsapi used on the front end at all, is it needed? Should it be used?

resources/static/common/js/network.js
shane-tomlinson, line 743: I do not see this function called anywhere.

lib/configuration.js
```diff
@@ -150,7 +150,11 @@ var conf = module.exports = convict({
   },
   authentication_duration_ms: {
     doc: "How long may a user stay signed?",
-    format: 'integer = 2419200000'
+    format: 'integer = 2419200000' // 4 weeks
+  },
+  idp_offline_grace_period_ms: {
+    doc: "How long should an IDP be offline before we fallback to secondary and start issuing certificates for her.",
+    format: 'integer = 2419200000' // 4 weeks
```
shane-tomlinson: I thought it was decided in the mailing list to set this to 1 week.

resources/static/pages/js/manage_account.js
shane-tomlinson, line 184: user.canSetPassword is that abstraction, it already exists.

resources/static/common/js/user.js
```diff
@@ -1051,16 +1033,9 @@ BrowserID.User = (function() {
...
-          passwordNeeded = true;
-
-      for(var key in emails) {
-        if(emails[key].type === "secondary") {
-          passwordNeeded = false;
-        }
-      }
-
-      complete(onComplete, passwordNeeded);
+      network.withContext(function(ctx) {
```
shane-tomlinson, line 1031: Instead of duplicating code, call user.canSetPassword

resources/static/dialog/js/misc/state.js
shane-tomlinson: user.completeTransition does not exist. A test would have found this. ;)

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -68,11 +79,19 @@ BrowserID.Modules.Authenticate = (function() {
...
+      user.checkEmailIssuer(email, info);
+
+      if ("offline" === info.state) {
+        self.close("primary_offline", info, info);
+      }
+      else if("primary" === info.type) {
         self.close("primary_user", info, info);
       }
-      else if(info.known) {
+      else if(hasPassword(info)) {
```
shane-tomlinson: In the notes on https://id.etherpad.mozilla.org/primary-transition-ux-swarm, it mentions there should be a screen displayed for the `transition_to_primary` state, yet here it has the user entering their password.

resources/static/test/cases/common/js/user.js
```diff
@@ -1324,19 +1295,46 @@
     );
   });
 
-  asyncTest("hasSecondary returns false if the user has 0 secondary email address", function() {
-    lib.hasSecondary(function(hasSecondary) {
-      equal(hasSecondary, false, "hasSecondary is false");
-      start();
-    }, testHelpers.unexpectedXHRFailure);
-  });
+  // JWCrypto relies on there being a random seed.  The random seed is
```
shane-tomlinson, line 1277: Nice update.

resources/static/test/cases/common/js/user.js
```diff
@@ -1324,19 +1295,46 @@
...
+          cb(cert);
+        });
+    });
+  }
 
-  asyncTest("hasSecondary returns true if the user has at least one secondary email address", function() {
-    storage.addEmail(TEST_EMAIL, { type: "secondary" });
-    lib.hasSecondary(function(hasSecondary) {
-      equal(hasSecondary, true, "hasSecondary is true");
+  asyncTest("checkEmailIssuer with changing issuer", function () {
```
shane-tomlinson, line 1300: Is there a test that makes sure this works correctly if the issuer stays the same?

shane-tomlinson:
As far as the front end portion goes, most of it looks very reasonable. 

My biggest nit is that additional QUnit tests are needed, Selenium tests do not cut it.

While unit tests won't ensure the entire system works, they help developers while they are still developing. There is at least one missing function error that would have been caught with a rudimentary unit test.

Unit tests will also help the next maintainer understand the interfaces that exist, the values that are expected, and where seemingly subtle changes can have impacts in far away places.

lib/wsapi/cert_key.js
ozten: From easy to hard:

1) We can make a read before making a write.

2) We could change `db.useOwnsEmail` to return the lastUsedAs value... which seems like a hack.

3) We could stop using `db.emailIsVerified` and use `db.emailInfo`. We can enhance the metadata returned from emailInfo to add `verified`.

4) We could introduce caching layer like memcached to make the read before the write cheap.

I don't see any issues with doing #3, I'll put that on my TODO list.

resources/static/dialog/js/modules/pick_email.js
ozten: I wasn't quite sure what to do. Other places in the code we have a failure callback handy, but not here.

Any suggestions?

**Update:** [Fixed](https://github.com/mozilla/browserid/commit/14d4bb9335e5098632c9c5033449519acff78447)

resources/static/common/js/error-messages.js
ozten: completeTransition was refactored to usedAsPrimary and that work is mid-flight.

lib/wsapi/saw_idp.js
ozten, line 1: Used from `bin/browserid`

ozten:
We want to add more unit tests. I am still learning how to read and write the frontend tests. The hope is that we can land this and continue to expand the QUnit tests during next week and write the new Selenium tests in parallel.

resources/static/dialog/js/modules/pick_email.js
ozten: [Fixed](https://github.com/mozilla/browserid/commit/14d4bb9335e5098632c9c5033449519acff78447)

resources/static/dialog/js/modules/authenticate.js
ozten: [Fixed](https://github.com/mozilla/browserid/commit/14d4bb9335e5098632c9c5033449519acff78447)

resources/static/dialog/js/modules/authenticate.js
ozten, line 74: [Fixed](https://github.com/mozilla/browserid/commit/14d4bb9335e5098632c9c5033449519acff78447)

resources/static/common/js/user.js
ozten, line 1352: I'm looking in to this.

Update: [Fixed](https://github.com/mozilla/browserid/compare/d7051d939a04...bfcfdef3cd35)

resources/static/dialog/views/set_password.ejs
seanmonstar, line 15: The email is received from an input field, and we fetch it using https://github.com/mozilla/browserid/blob/dev/resources/static/common/js/helpers.js#L12

Could the email come from the server at some point?

resources/static/dialog/js/modules/verify_primary_user.js
seanmonstar, line 63: Yea, it looks like it sucks. What we have is 2 beginning scenarions: already authed with the dialog, or not. This module is only shown if the user isn't authed. `upgrade_to_primary_user` handles the case that the user was authed before, but now must upgrade.

resources/static/dialog/js/modules/authenticate.js
seanmonstar, line 166: I'm not sure. I wasn't sure if the original label needed to stay put, so it was easiest for me to just replace the contents.

resources/static/dialog/js/modules/authenticate.js
seanmonstar, line 30: If that works, I can do that. I didn't know if `gettext` would get parsed in these files, or only in the templates.

resources/static/dialog/js/misc/state.js
seanmonstar: Indeed, looks like a missed case part of the renaming of `complete_transition` to `used_address_as_primary`

resources/static/dialog/js/misc/state.js
seanmonstar: I don't understand your question.

resources/static/common/js/user.js
seanmonstar, line 1352: There was originally a test for `completeTransition`, did that get renamed?

resources/static/common/js/network.js
seanmonstar, line 743: I'm responsible for finding the correct place to call this, that work still hasn't merged.

/whip whip

resources/static/common/js/helpers.js
seanmonstar, line 137: There isn't ;) It's indirectly tested, since I lifted it from the private function in user.js

resources/static/dialog/js/misc/state.js
ozten: I'm looking into this.

resources/static/common/js/error-messages.js
seanmonstar: Since we plan to ignore failures to this call, we probably don't need this error message.

We don't want the dialog to fall over while we do minor house-cleaning.

lib/wsapi/cert_key.js
seanmonstar: 5) Kick off a child_process to do the write?

ozten: I think the issue is the # of writes, 5 doesn't address that, right?

lib/configuration.js
lloyd: set to five days (we were debating on the list between 3, 5, and 7.)  a3572377953b99ccfb6a1acf8c16067a5362f5bd

bin/browserid
lloyd: fixed in 1270c66ccda15f2995af77a823f02d80187f5f17

lib/well-known-parser.js
lloyd: @callahad and I will be talking through all the changes in a couple hours.  I'm fine with `"enabled": false` if that strikes a balance between clarity and anti-double-negativedness

lib/well-known-parser.js
lloyd: Let's consider.  First, your university has implemented persona support:

```
{ "authentication": "/blah.html", 
   "provisioning": "/blahblah.html",
   "public-key": { "blah": "blah" } }
```

Now, the university wishes to transition to delegating their authority to some other domain (say it's a university and they've centrally implemented persona support).  So they just add a key:

```
{ "authority": [ "secure.myschool.edu", "secure.myschool.edu" ], // they messed up!  they thought they could specify multiple domains for redundancy
  "authentication": "/blah.html", 
   "provisioning": "/blahblah.html",
   "public-key": { "blah": "blah" } }
```

Is it better for:
1. us to send the user to this domain for a cert
2. us to take over and verify the email address and issue a cert.
3. us to tell the user that the idp is broken.

in my mind, the answer varies whether you're talking about verifiers or implementations.  I think implementations should do #3, and verifiers should still honor certs.

so agree with you @benadida, and think it uncovers the larger issue that our verifier and implementation should _not_ behave in the same way in their parsing of well-known docs.

/cc @callahad

lib/wsapi/address_info.js
lloyd: fixed in ed418da2097772fe387229d63873a1829a1367bd

lib/wsapi/cert_key.js
lloyd: I'm not worried about database traffic right now.  In all of our load testing it's not yet on the radar.

I _AM_ worried about the fact that the write will never succeed.  It's being performed in the `browserid` process which has read-only access to the db.

lloyd: our tests should catch this, there has been a regression, I'll fix the general problem that we're not catching writes being performed by processes which cannot write, then we can fix the underlying issue.

lloyd: now we get an error from the database layer when a process that cannot write attempts to write 55925c66383e4435e290f5a3fcb780f65c7e7a90

resources/static/common/js/error-messages.js
ozten: Good call Shane and Sean. I'll remove.

[Fixed](http://git.io/eSuO_w)

resources/static/common/js/network.js
ozten, line 743: Tests added, [fixed](https://github.com/mozilla/browserid/commit/28c275b34af4b93a48a7474e4afb94d8ac587ac1)

resources/static/dialog/js/misc/state.js
ozten, line 234: I'm looking into this.

resources/static/dialog/js/modules/authenticate.js
ozten, line 78: Could `info` be the results of an addressInfo call? Or is it always a combination of data from storage and variables set from code?

resources/static/dialog/js/modules/primary_offline.js
ozten: I'm looking into this.

Update: [Fixed](http://git.io/mV_BiQ)

resources/static/dialog/js/modules/primary_offline.js
ozten, line 1: [Fixed](http://git.io/mV_BiQ)

Also added a check for email.

lib/well-known-parser.js
warner: My two cents: `"enabled": false` kinda makes me think that you must have a `"enabled": true` to make it work at all. I'm inclined to use the rule of "the default should be short", which suggests `"disabled": true` for one side and nothing for the other.

resources/static/dialog/views/set_password.ejs
warner, line 15: I tried (and mostly failed) to follow how that field shows up, but was able to confirm at least one path that derives the domain from the email address that sean pointed to. I think that leaving the `escape()` is the right thing to do: this is a point where we convert from one type ("email address in a string") to a different type ("string of HTML which should render to something that looks like an email address"), and `escape()` is the right tool for that job. If we ever see double-escaping, we should probably fix the other thing instead.

seanmonstar, line 15: Also worth pointing out, I used `escape` here, since we usually use the auto-escape of `<%= %>`, but I needed to wrap the content in some HTML, so I had to use `<%- %>`. So in most other places, this would be possibly double escaped also.

lib/wsapi/address_info.js
```diff
@@ -26,36 +59,108 @@ exports.args = {
...
+    }
+
+    // now we have all the information we need, let's generate
+    // a response
+    var r = typeRightNow;
+
+    if (!emailKnown) {
+      r.state = "unknown";
+    } else if (r.type === 'secondary' && primarySeenRecently) {
+      r.state = "offline";
```
callahad, line 176: I think there's a race condition here that can lead to erroneously reporting "offline" when a domain explicitly disabled primary support.

Since `primary.checkSupport` and `db.getIDPLastSeen` are async, couldn't `db.getIDPLastSeen` first report that an IdP was seen recently, while `primary.checkSupport` finishes later, notices an `{"enabled": false}`, and reports a type of `"secondary"`?

In this case, the user will be shown the "offline" screen, even though the domain has explicitly disabled primary support and asked to use the fallback.

Edit: Here's a [flow chart](http://grab.by/hU14) that outlines what we're doing / want to be doing, as best as I can tell.

warner:
Quick scan of DB changes look ok to me. You might want to call out the specific command you have in mind for the DB upgrade.

Information leaks in /api/address_info : 

The new code reveals (to everybody, without authentication) whether an email address is attached to a persona account for all kinds of IDPs, whereas the old code only revealed this for secondary IDPs. The new code also reveals whether the (secondary) address has been verified or not (the old code required auth before showing this). I _think_ that's ok.. my hunch is that this is information that attackers could have deduced already.

Does the new /api/address_info do a .well-known fetch _every time_? That feels like a lot of traffic. This ties into the conversation we started last month about DNS-like expiration times for support documents (i.e. how long we're allowed to cache the response). It also reveals the time of every attempted login to the IDP, which would be less information-leaky if we only checked once every hour or day or whatever.

Looser auth level for /api/update_password: the code seems to correctly check the old password before allowing the operation, so I think the new code is as safe as the old code.

New .well-known parsing: I'm slightly in favor of using disabled=true instead of supported=false. I like the refactoring, splitting out parse-well-known.js and cleaning up the recursion to include just one function. Yay!

I might be wrong, but it seems like primary.js and `fetchWellKnown` is going to signal an error when the IDP 404s the .well-known fetch. This should be handled correctly (the IDP is signalled as a secondary), but would also emit a bunch of error messages ("primary support is misconfigured") that may obscure actual errors. Should `handleResponse` or maybe `handleProxyIDP` do a `cb(null, false)` in response to a 404 instead of an error?

The verifier certainly needs to honor the "disabled" flag. This will have to be part of the #1501 overhaul, since we don't currently check with the IDP at all.

Minor nits:
- lib/db/mysql.js line 25: typo "timestamp" not "timestame"
- lib/wsapi/address_info.js wants to use an "AsyncAND" pattern, and looks like a good use case for Deferreds or Promises or a similar device

lib/wsapi/address_info.js
warner, line 176: I think it's even weirder than that, since the call to `checkSupport` triggers an `idp_seen` event, which (eventually) updates the table that feeds `db.getIDPLastSeen`. So maybe two races.

Your flowchart looks right to me.

seanmonstar:
> Does the new /api/address_info do a .well-known fetch every time? That feels like a lot of traffic.

We've discussed with @gene1wood about being able to cache this files locally for a short time period (30s?), and that `/wsapi/list_emails` would actually start the requests. Then address_info would hopefully be able to use a cached version. Part of the goal, though, was allowing an IdP to almost instantly turn off IdP support, so we couldn't cache for too long.

resources/static/dialog/views/set_password.ejs
shane-tomlinson, line 15: @seanmonstar - that's good to know. Thanks @warner, your assessment is sound.

lib/well-known-parser.js
lloyd: ok.  this argument is compelling and the jury is split.  going with `"disabled": true`.

lib/well-known-parser.js
lloyd: @callahad and I agreed that if `authority` is present and not parsable, an implementation should consider the domain to not have support.  A verifier should ignore and verify assertions IFF `public-key` is present.  This asymmetry is important for zero user impact during transitions.

lib/wsapi/address_info.js
lloyd, line 176: it seems we need check support to return a signal if the IdP is explicitly disabled.  the test becomes `(r.type === 'secondary' && primarySeenRecently && !r.explicitlyDisabled)`.

@callahad nice find!  does this sound reasonable?

@warner I can't find the race in your reasoning.  If an idp is seen, `r.type === 'primary'`, in which case `primarySeenRecently` doesn't come into play, so the fact that it's out of date doesn't matter (and it should get up to date in ms).  Am I missing something?

lib/wsapi/saw_idp.js
lloyd, line 1: right, it's an _internal_ api for browserid to trigger a database write on dbwriter.  We do this in several cases:

1) when we need to both write the database AND contact the keysigner (like `cert_key`)
2) when we have a large api handler that only infrequently must write the db (like `authenticate_user`)
3) when browserid detects information in it's normal processing (fetching .well-known) that should be persisted (this case)

resources/static/common/js/user.js
lloyd, line 1031: was going to do that, but we'd have to call it, then invert the return value.  the code reuse removes no lines and hurts my brain.

resources/static/dialog/js/misc/state.js
lloyd, line 327: I guess we should go and be consistent throughout state.js?  Maybe after this patch?

PR #2798, opened 2012-11-29
Auth Secondary transitions

ozten:
- User already has a password
- User needs a password
- CLEANUP - network and user have a setPassword function, but
  there is no backend to support it.

### DIFF
--- a/lib/wsapi/address_info.js
+++ b/lib/wsapi/address_info.js
@@ -161,7 +161,8 @@ exports.process = function(req, res) {
     } else {
       r.state = STATE_TABLE[hasPassword][lastUsedAs][typeRightNow.type];
     }
-
+    r.state = 'transition_no_password';
+    r.type = 'secondary';
     res.json(r);
   }
 };
--- a/lib/wsapi/address_info.js
+++ b/lib/wsapi/address_info.js
@@ -161,8 +161,6 @@ exports.process = function(req, res) {
     } else {
       r.state = STATE_TABLE[hasPassword][lastUsedAs][typeRightNow.type];
     }
-    r.state = 'transition_no_password';
-    r.type = 'secondary';
     res.json(r);
   }
 };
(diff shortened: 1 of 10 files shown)


ozten:
Rebased against latest primary-transition-ux. @seanmonstar - using your set_password.ejs changes instead of mine.

lib/wsapi/address_info.js
```diff
@@ -161,7 +161,8 @@ exports.process = function(req, res) {
     } else {
       r.state = STATE_TABLE[hasPassword][lastUsedAs][typeRightNow.type];
     }
-
+    r.state = 'transition_no_password';
```
seanmonstar: more sneaky ninjas

resources/static/dialog/js/modules/set_password.js
```diff
@@ -36,6 +36,7 @@ BrowserID.Modules.SetPassword = (function() {
 
       self.renderDialog("set_password", {
         email: options.email,
+	idpName: helpers.getDomainFromEmail(options.email),
```
seanmonstar: already exists as `domain` a couple lines down. If you want this name for consistancy, just change the `domain` and usage in the template.

ozten: Missed this in the conflict resolution. I see set_password.ejs uses domain. Thanks! Removed.

resources/static/dialog/js/misc/state.js
```diff
@@ -355,6 +364,13 @@ BrowserID.State = (function() {
         }
       }
       // Anything below this point means the address is a secondary.
+      else if ("transition_to_secondary" === info.state) {
+        startAction("doAuthenticate", info);
+      }
+      else if ("transition_no_password" === info.state) {
+        self.transitionNoPassword = info.email;
```
seanmonstar, line 371: I just piggybacked on the `resetPasswordEmail` for un-authed no_password. We should probably do the same in both cases tho.

PR #2615, opened 2012-10-17
Added CEF logging at key points alongside our app logger

seanmonstar:
Originally @jedp's work from #2150, but rebased onto latest dev. All comments from that PR we're addressed, so all that's needed is a sanity check review that the rebasing didn't clobber anything.

PR #2799, opened 2012-11-29
CEF logging work, rebased again off current dev

jaredhirsch:
Same deal as @seanmonstar did with https://github.com/mozilla/browserid/pull/2615, which is to say, rebasing the work of @jedp in #2150, but making it up to date with dev.

Opened, again, so we can verify nothing untoward happened before merging in.

### DIFF
--- a/lib/cef_logger.js
+++ b/lib/cef_logger.js
@@ -1,0 +1,169 @@
+/**
+ * A thin abstraction layer over the cef library
+ *
+ * This is deliberately not a part of the logging.js library because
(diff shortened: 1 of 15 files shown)


seanmonstar:
Oh noes, needs a JSHint cleanup.

jaredhirsch:
Bah, not a fan of that particular 15 minute experience.

How can you appease jshint yet indicate in a function signature that it takes varargs? I gave up googling & trimmed the 'etc' arg from the cef_logger calls.

Issue #2768, opened 2012-11-27
IE9: mysterious java error causing tests to fail & hang for 10+ minutes


jaredhirsch:
3/5 recent failed runs were caused by this mysterious java crash.

Some recent IE9 jobs that failed due to the nefarious java plugin crash:

11/26 19:20 - #1958
- add-primary-to-primary: ARGH. java error again.
  - https://saucelabs.com/tests/8769ff0e8fbf420ba62a409ef61dd48b#
- add-primary-to-secondary: failed to click 'sign in' dialog button after entering password for verified secondary acct
- new-user-primary: clicked "verify with eyedee.me" on first try, timed out waiting to click a second time

11/26 19:04 - #1956
- new-user-secondary: mysterious java failure caused test to hang for 12 minutes!!
- https://saucelabs.com/tests/988c37ec53a24b67bf1211825f92fa49#

11/26 15:36 - #1955
- add-primary-to-secondary: test ran beautifully, mysterious java failure at end.
- https://saucelabs.com/tests/a9798be908b842b3992406a6c417f7bd#

jaredhirsch:
Spoke with @santiycr in #saucelabs on freenode. He's opening a support ticket on their side & will update me by email. I'll keep this ticket up to date.

_6a68       hey sauce. I'm seeing java SE plugin crashes on a lot of IE9 tests lately. anybody heard of such a thing?           10:55
_6a68       take a look at this test: https://saucelabs.com/tests/8769ff0e8fbf420ba62a409ef61dd48b#
10:56
_6a68       if you look at the _video_, the test runs fine up to about 50 seconds in, then the crash dialog pops up ("java platform SE binary has stopped working"), then the test hangs for another 11 minutes :-\         10:57
_6a68       I've seen this several other times in the last day; obviously it's a bummer to see tests crashing due to infrastructure failure         10:58
santiycr        sorry to hear that, _6a68           11:23
santiycr        we'll investigate           11:23
santiycr        what's your email address so we can follow up?          11:23
_6a68       santiycr: hey!          11:25
_6a68       santiycr: if it helps, I can send you links to a couple other ie9 tests that failed in the same mysterious way          11:27
_6a68       santiycr: I'm 6a68@mozilla.com          11:27
_6a68       santiycr: here's another IE9 java crash https://saucelabs.com/tests/988c37ec53a24b67bf1211825f92fa49#
11:31
[...]

PR #2803, opened 2012-11-29
Issue 2768 ie9 java errors

jaredhirsch:
See discussion in #2768. Turns out that you have to ask for the latest version of the selenium-server, and then IE9 doesn't freak out at all. Yay.


shane-tomlinson:
I feel like I prematurely merged this. Why not use the same version of Selenium for _all_ browsers? We are setting ourselves up for fail by having only one browser use this version and let the others float.

shane-tomlinson:
Opened new issue about versioning - #2819

PR #2804, opened 2012-11-29
call used_address_as_primary when returning from an idp

### LINKED ISSUE #2795: Call /wsapi/used_address_as_primary after successful primary use
blocks #2404

seanmonstar:
If the user is authed, no other network calls would have been made, and we need to call this.

If the user isn't authed, this call will fail, but we carry right on, ignoring the failure. auth_with_assertion would have to be called later, so that call can flip the bit for us in this case.

fixes #2795

### DIFF
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -269,7 +269,18 @@ BrowserID.Modules.Dialog = (function() {
       // XXX Perhaps put this into the state machine.
       self.bind(win, "unload", onWindowUnload);
 
-      self.publish("start", params);
+      function start() {
+        self.publish("start", params);
+      }
+
+      if (params.type === "primary") {
+        // at this point, we will only have type of primary if we're
+        // returning from #AUTH_RETURN. Mark that email as having been
+        // used as a primary, in case it used to be a secondary.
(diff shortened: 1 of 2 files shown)


resources/static/dialog/js/modules/dialog.js
```diff
@@ -269,7 +269,18 @@ BrowserID.Modules.Dialog = (function() {
...
+        self.publish("start", params);
+      }
+
+      if (params.type === "primary") {
+        // at this point, we will only have type of primary if we're
+        // returning from #AUTH_RETURN. Mark that email as having been
+        // used as a primary, in case it used to be a secondary.
+        user.usedAddressAsPrimary(params.email, start, start);
+      } else {
+        start();
```
ozten, line 282: Nice. `start` is the popular kid.

resources/static/test/cases/dialog/js/modules/dialog.js
```diff
@@ -249,6 +249,35 @@
...
+          controller.get(testHelpers.testOrigin, {}, function() {}, function() {});
+        }
+        catch(e) {
+          // do nothing, an exception will be thrown because no modules are
+          // registered for the any services.
+        }
+      }
+    });
+  });
+
```
ozten, line 280: Nice test!

Issue #2807, opened 2012-11-30
if PERSONA_WITH_COVER, do a graceful shutdown of bin/static so coverage_data is created


jrgm:
When I added the coverage reporting for backend, I couldn't get data for bin/static. I tried making this use shutdown.handleTerminationSignals but it seemed not to work at that time. It does now, so I'll put in a PR to add that conditionally to bin/static. (I was reminded of this while writing a test for /.well-known/browserid which is served by static).

jrgm:
... and ultimately this code was removed in GH-2951 when a non no-op listener was added for toobusy.

PR #2808, opened 2012-11-30
if PERSONA_WITH_COVER, do a graceful shutdown so cover can generate .coverage_data

jrgm:
This is the PR for mozilla/browserid#2807. Does what the summary says and otherwise has no effect if PERSONA_WITH_COVER is not set.

### DIFF
--- a/bin/static
+++ b/bin/static
@@ -12,7 +12,8 @@ path = require('path'),
 url = require('url'),
 http = require('http'),
 urlparse = require('urlparse'),
-express = require('express');
+express = require('express'),
+shutdown = require('../lib/shutdown');
 
 const
 assets = require('../lib/static_resources').all,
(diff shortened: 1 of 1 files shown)


bin/static
```diff
@@ -102,6 +103,14 @@ app.use(function(req, res, next) {
 
 app.use(express.static(static_root));
 
+// Do this only when PERSONA_WITH_COVER is set. Without a graceful shutdown,
```
shane-tomlinson, line 106: Is there a danger in always enabling this?

jrgm, line 106: No, I don't think so. But since it doesn't actually take any action, I wanted to call out why it is there at all.

shane-tomlinson, line 106: @jrgm - ok. I am cool with this change. Once I find out whether the train is open, I'll merge.

Issue #2790, opened 2012-11-28
Set a default for proxy_idps in lib/configuration.js


jrgm:
We're working around with a config change in stage and prod, but defaulting here would be best.

Maybe something like:

```
   proxy_idps: {
     doc: "A mapping of domain names to urls, which maps popular email services to shimmed IDP deployments.",
-    format: 'object { } *?',
+    format: 'object {}* = {}',
     env: 'PROXY_IDPS' // JSON text, i.e. {"yahoo.com":"yahoo.login.persona.org"}
   },
```

lloyd:
r+, make a PR

jrgm:
I realized that setting a default would change control flow in lib/primary.js (.has() would now be true not false). So, I'm just gonna change views.js to check `has` before calling `get`.

jrgm:
PR-2809 was merged for this issue and is in production for a while. Closing.

PR #2809, opened 2012-11-30
when handling request /.well-known/browserid bail out if config proxy_idps is no...t available; and tests for that

jrgm:
fixes mozilla/browserid#2790

Best reviewed with `?w=1` for the views.js part where I just ball out.


jrgm:
Review this with `?w=1`: https://github.com/jrgm/browserid/commit/351b14377fdd32c5f9f3dac241940a47d4d44d31?w=1

lloyd:
r+ @jrgm - thanks for extensive tests and providing a link to the patch that ignores whitespace.  this is solid.

Issue #2800, opened 2012-11-29
Dump process state to logfiles at startup


jrgm:
To help in debugging stage/prod or even dev/local/ephemeral issues it would be useful to have each daemon write some state at startup to its logfile.

1) From the node process module, these properties: { version, argv, env, pid, execPath }. These ones also could be dumped: {arch, platform, versions, features }

2) The value of `require('./lib/configuration').toString()`, possibly reformatted with JSON.stringify(JSON.parse(config)) to strip the new lines.

jaredhirsch:
I'm excited to mess around with this today. Feel free to steal if it seems to stagnate, though.

jrgm:
Sorry, @6a68. I'm a thief! But I just went for the minimal set of stuff. If you want to do something more, please do.

jaredhirsch:
no worries, @jrgm. maybe I'll get around to it at some point :-)

merged in your PR, closing.

jrgm:
verified stage train-2012.12.07 - the version and resolved configuration are in the {verifier,browserid,keysigner,dbwriter,static,router}.log

PR #2810, opened 2012-12-01
dump node version/path and current configuration to logs if NODE_ENV=production

jrgm:
Fixes GH-2800. I realized I was only really interested in the version and path of node.js plus the resolved configuration settings.


jaredhirsch:
r+

PR #2815, opened 2012-12-03
Keep track of the handle of the first window. Fixes #2805.

### LINKED ISSUE #2805: IE9 sucks at switching windows
We are seeing a lot [1] of test failures where
- it looks like we've logged into 123done/myfavoritebeer successfully
- IE9 cannot find the logged-in node
- the raw selenium log indicates that the element wasn't found as part of a NoSuchWindowException

My guess, based on this, is that IE9 has some bizarro way of sorting handles when it stores them internally. Either that, or it puts the newest window at the 0th, not the nth, spot in the window_handles array, and we only sometimes see this error because it's a race between browser.wwin() switching windows, and the dialog closing, reducing the length of window_handles to 1.

Here's an example of an error log snippet (full log is here [2]) that shows the element lookup failure as the message inside a NoSuchWindowException:

```
15:48:36.094 ERROR - Unexpected exception during findElement
org.openqa.selenium.NoSuchWindowException: Unable to find element with css selector == #emailList .email (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 20.01 seconds
```

Either we manually store the base window handle in every test (since every test uses the dialog and switches back to parent window at some point), or we find a clever but robust way to do this behind the scenes. One trick that could work: assign the topmost window handle when you setup the browser, then, when someone calls wwin() without arguments, pass in the parent window handle by default, instead of assuming the 0th window in the window_handles array is the right one. This is just a guess.

[1] https://gist.github.com/4173166
[2] https://saucelabs.com/jobs/7e772c3b8ebb479186a6bc3fcae657e6/selenium-server.log

jaredhirsch:
See #2805 for discussion.

Rather than track the current window as we switch from window to window during a session, I think we can get away with just grabbing a reference to the first window created at the start of a session--that'll always (I think?) be the one we want when the dialog is gone.

### DIFF
--- a/automation-tests/lib/wd-extensions.js
+++ b/automation-tests/lib/wd-extensions.js
@@ -80,8 +80,8 @@ wd.prototype.newSession = function(opts, cb) {
     browser.setImplicitWaitTimeout(timeouts.DEFAULT_TIMEOUT_MS, function(err) {
       if (err) return cb(err);
       // keep track of the ID of the first window
-      browser.windowHandle(function(err, handle) {
-        browser._parentWindow = handle;
+      self.windowHandle(function(err, handle) {
+        self._parentWindow = handle;
         return cb(err);
       });
     });
@@ -161,7 +161,7 @@ wd.prototype.wwin = function(opts, cb) {
     cb = arguments[0];
     self.windowHandles(function(err, handles) {
       if (err) return cb(err);
-      self.window(browser._parentWindow, function(err) { cb(err); }); // fire cb whether err is defined or not
+      self.window(self._parentWindow, cb); // fire cb whether err is defined or not
     });
   } else {
     self.waitForWindow(opts, cb);
--- a/automation-tests/lib/wd-extensions.js
+++ b/automation-tests/lib/wd-extensions.js
@@ -77,7 +77,14 @@ wd.prototype.newSession = function(opts, cb) {
     // procedures like find_element to succeed (we'll actually wait on the
     // *server* side for an element to become visible).  Having this be
     // the same as the global default timeout is interesting.
-    browser.setImplicitWaitTimeout(timeouts.DEFAULT_TIMEOUT_MS, cb);
+    browser.setImplicitWaitTimeout(timeouts.DEFAULT_TIMEOUT_MS, function(err) {
+      if (err) return cb(err);
+      // keep track of the ID of the first window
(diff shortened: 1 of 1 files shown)


automation-tests/lib/wd-extensions.js
```diff
@@ -153,7 +161,7 @@ wd.prototype.wwin = function(opts, cb) {
     cb = arguments[0];
     self.windowHandles(function(err, handles) {
       if (err) return cb(err);
-      self.window(handles[0], function(err) { cb(err); }); // fire cb whether err is defined or not
+      self.window(browser._parentWindow, function(err) { cb(err); }); // fire cb whether err is defined or not
```
shane-tomlinson: browser does not exist in this context, _parentWindow is on self at this point. This is causing a silent failure where all of the tests appear to pass but where all the tests continue running on sauce.

shane-tomlinson: Secondly, since the callback to self.window only calls cb with the err, can we replace the entire `function(err) { cb(err); }` with cb.

automation-tests/lib/wd-extensions.js
```diff
@@ -77,7 +77,14 @@ wd.prototype.newSession = function(opts, cb) {
     // procedures like find_element to succeed (we'll actually wait on the
     // *server* side for an element to become visible).  Having this be
     // the same as the global default timeout is interesting.
-    browser.setImplicitWaitTimeout(timeouts.DEFAULT_TIMEOUT_MS, cb);
+    browser.setImplicitWaitTimeout(timeouts.DEFAULT_TIMEOUT_MS, function(err) {
```
shane-tomlinson: We should standardize on `browser` or `self`, but not both. Having two names is probably what caused this problem.

shane-tomlinson: issue #2817 opened for this. PR #2818 fixes this replacing all `browser` instances with `self`.

jaredhirsch: Argh. This is my goof. I blatantly ignored the use of self throughout.

jaredhirsch:
One minor conflict made it a non-fast forward merge, merged origin into topic branch.

If this seems unclean I can remove that last commit and just merge topic branch into dev, resolving the conflict at that time instead.

PR #2818, opened 2012-12-04
replace browser with self to keep the name consistent.

### LINKED ISSUE #2817: wd-extensions: pick browser or self, not both
Having two names for the same thing is causing problems - see https://github.com/mozilla/browserid/pull/2815#discussion_r2305080

shane-tomlinson:
@6a68 - mind reviewing this?

fixes #2817


jaredhirsch:
Nice cleanup.

Issue #2822, opened 2012-12-05
Cannot create an account on Prod using a GMail address after address was removed from Persona


shane-tomlinson:
STR)

1) open the manage page for a Persona account with a gmail address.
2) remove gmail address from current Persona account
3) visit 123done.org
4) If signed in, sign out
5) Open Persona dialog
6) If signed in, sign out
7) Enter gmail address which will create a new account
8) Enter password and verification password
9) Visit gmail and open the verification link
10) Watch the error message display

Expected:
Account is created

This error does NOT occur on stage.

Looking at the network console for https://login.persona.org/wsapi/email_for_token?token=aZOdhGIZjvHvsYAiZVHqURQz7Q5vNOfsDfThHMIBFrBcsI3C
### Request Headers

```
GET /wsapi/email_for_token?token=aZOdhGIZjvHvsYAiZVHqURQz7Q5vNOfsDfThHMIBFrBcsI3C HTTP/1.1
Host: login.persona.org
Connection: keep-alive
Accept: application/json;text/plain
X-Requested-With: XMLHttpRequest
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_2) AppleWebKit/537.17 (KHTML, like Gecko) Chrome/24.0.1312.32 Safari/537.17
Content-type: application/x-www-form-urlencoded
Referer: https://login.persona.org/verify_email_address?token=aZOdhGIZjvHvsYAiZVHqURQz7Q5vNOfsDfThHMIBFrBcsI3C
Accept-Encoding: gzip,deflate,sdch
Accept-Language: en,en-US;q=0.8,de;q=0.6
Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.3
Cookie: browserid_state=woYgw4Rdw6vDq0PDvsOawpDCrAjDkXPCkGQ.wpbDk8OyOQ1zU1nDi3PDgTTCrsKXwrXDhsKqwrRiSsOge2zCtlkJw4jDrsOhw7HDhsKQw6jCjMK9LMOhwq05JMOdHsODwq1AbAnDiw.1353950933486.2419200000.w5jCj8O_AAzDn27DhcORdsKOQ8OnXG8NQsOFw5vCrwTCgMKcwrPCl1xQcMKAbsKWOA
```
### Response Headers

```
HTTP/1.1 200 OK
Cache-Control: no-cache, max-age=0
Content-Type: application/json; charset=utf-8
[...]

PR #2824, opened 2012-12-05
Dev hotfix issue2822

lloyd:
fix for issue #2822 in dev


shane-tomlinson:
@lloyd, @benadida - I would like somebody else to look at the SQL query, but the tests look good. The tests fail as expected when using the previous version of mysql.js and always pass when using the JSON adapter.

jrgm:
r+, closes #2822

Issue #2825, opened 2012-12-05
email_for_token should do less


lloyd:
in diagnosis of issue #2822, we found that when the database was in an unexpected state, users were unable to create an account in persona.  (specifically, one who _staged_ long ago but did not _verify_ would be blocked).

If `email_for_token` was less aggressive about verifying consistency, these users would have been to work through the problem without asking for assistance.

`email_for_token` should be simplified to do less work without any other behavioral changes - because if email for token detects problems, like there's no password but their should be - this adds no security, no improved error messages, nothing.

lloyd:
@fmarier or @warner, you seem like fantastic minds to review this?

PR #2826, opened 2012-12-05
simplify email_for_token, improving readability and removing useless con...

lloyd:
...sistency checking - issue #2825


fmarier:
Looks fine to me. r+

PR #2827, opened 2012-12-06
Make sure *all* errors are reported.

### LINKED ISSUE #2745: run-all.js sometimes returns with status "0" even though vows exits with code 2
exiting with code 2
<

0/0 tests passed

> > Exiting tests with code: 0

shane-tomlinson:
@6a68 - mind reviewing?

I tried using vows_process.on('uncaughtException', callback) but I was never getting the message. Looking at the vows code, they have an uncaughtException handler which prints the results - could this be masking the parent process from receiving this message?

1) the vows JSON reporter reports exceptions differently to passing results. The results aggregator must account for this or else some failures are not counted.
2) If a test has an asynchronous exception, the JSON reporter never gets a record of the error. Instead, errors messages must be fetched from stderr and reported approrpriately.

fixes #2745

### DIFF
--- a/automation-tests/lib/results-aggregator.js
+++ b/automation-tests/lib/results-aggregator.js
@@ -20,8 +20,84 @@ function ResultsAggregator() {
 
 utils.inherits(ResultsAggregator, events.EventEmitter);
 
+/*
+ * Asynchronous exceptions come in on stderr. These errors are not reported
+ * in the JSON reporter and must be parsed manually from the stderr output.
+ * Async errors take the form of:
+ *
+ *    /file_path:lineno
+ *    line_causing_problem
+ *          ^ (arrow to location)
+ *    TypeOfError & StackTrace
+ *
+ * Example of an error:
+ *     /Users/stomlinson/development/browserid/automation-tests/tests/sign-in-test.js:45
+ *     o['b'];
+ *     ^
+ *     ReferenceError: o is not defined
+ *         at runner.run.throw an async error (/Users/stomlinson/development/browserid/automation-tests/tests/sign-in-test.js:45:11)
+ *         at process.startup.processNextTick.process._tickCallback (node.js:244:9)
+ *
+ *
+ * Second example:
+ *     timers.js:103
+ *     if (!process.listeners('uncaughtException').length) throw e;
+ *     ^
+ *     ReferenceError: o is not defined
+ *         at Object.runner.run.throw an async error [as _onTimeout] (/Users/stomlinson/development/browserid/automation-tests/tests/sign-in-test.js:45:11)
+ *         at Timer.list.ontimeout (timers.js:101:19)
+ *
+ * Each line of the exception is reported to stderr as its own message. This
+ * means that once an error is detected on stderr, subsequent messages must be
+ * appended until the entire message is captured.
+ *
+ * To fix this (ver brittle):
+ * Look for the test_path on a line.
+ * Capture and append each message until the message with "Error: "
+ */
+ResultsAggregator.prototype.parseErrorLine = function(msg) {
+  /*
+   * look for a message that contains the test path, if the line starts with
+   * the path, there is an asynchronous error.
+   */
+  msg = msg && msg.trim();
+  if (!msg) return;
+
+  if (/https:\/\/saucelabs\.com\/tests\//.test(msg)) {
+    this.addInterestingURL(msg);
+  }
+
+  if (!this.captureLines && /\.js:\d+/.test(msg)) {
+    /*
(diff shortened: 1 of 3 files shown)


automation-tests/lib/results-aggregator.js
```diff
@@ -31,29 +107,46 @@ ResultsAggregator.prototype.parseLine = function(msg) {
...
-    } else if (msg["0"] === 'end') {
+    /*
+     * If there is a synchronous exception, no "context" message will be
+     * sent, this means the spec name must be fetched from the error message.
+     */
+    } else if (msg["0"] === 'error' && msg["1"].context) {
+      self.current.name = msg["1"].context;
+      self.handleError(msg["1"].exception);
+    // vow is sent when a vow runs and completes normally.
+    } else if (msg["0"] === 'vow') {
```
shane-tomlinson, line 121: This is pulled from down below, I reordered the message processing to be handled in the same order as they are printed in stdout.

automation-tests/lib/vows_harness.js
```diff
@@ -51,14 +52,16 @@ module.exports = function(spec, mod, opts) {
...
 
-  suite.addBatch({
-    "cleanup": {
-      topic: function() {
-        if (opts.cleanup) opts.cleanup(this.callback);
-      },
-      "done": function() { }
-    }
-  });
+  if (opts.cleanup) {
```
shane-tomlinson, line 55: Only add a cleanup vow if there is in fact cleanup to do.

automation-tests/scripts/run-all.js
```diff
@@ -179,20 +179,23 @@ function startTesting() {
...
-      if (aggregator) aggregator.parseLine(data.toString());
-      if (stdOutReporter) stdOutReporter.report(data.toString());
+      var msg = data.toString().trim();
+
+      if (aggregator) aggregator.parseLine(msg);
+      if (stdOutReporter) stdOutReporter.report(msg);
     });
 
     testProcess.stderr.on('data', function (data) {
-      data.toString().split("\n").forEach(function(line) {
```
shane-tomlinson, line 187: This was a roundabout way of doing things, so I handle everything on a single line.

automation-tests/scripts/run-all.js
```diff
@@ -179,20 +179,23 @@ function startTesting() {
...
-        line = prettifyLine(line, function(url) {
-          if (aggregator) aggregator.addInterestingURL(url);
-        });
-        // output the line to console when we're in xunit mode
-        if (line && args.output === 'xunit') {
-          stdErrReporter.report(testName + ' | ' + platform + ' | ' + line + "\n");
-        }
-      });
+      // remove any leading newline characters and trim the rest of the output.
+      var line = prettifyLine(data.toString().replace(/^[\r\n]+/, '').trim());
```
shane-tomlinson, line 190: Instead of passing a callback which handles interesting URLs to prettifyLine, I am letting aggregator.parseErrorLine handle them directly.

automation-tests/scripts/run-all.js
```diff
@@ -201,10 +204,11 @@ function startTesting() {
     });
   }
 
-  function prettifyLine(line, onInterestingURL) {
+  function prettifyLine(line) {
     // decolorize
-    line = line.replace(/\x1b\[([0-9]{1,3}((;[0-9]{1,3})*)?)?[m|K]/, "");
+    line = line.replace(/\x1b\[([0-9]{1,3}((;[0-9]{1,3})*)?)?[m|K]/g, "");
```
shane-tomlinson, line 209: Search and replace _all_ colorized instances

automation-tests/lib/vows_harness.js
jaredhirsch, line 55: Sigh. Technically, every test has opts.cleanup. If you look at the current tests, they all have the same boilerplate at the end:

```
{
    suiteName: path.basename(__filename),
    cleanup: function(done) { testSetup.teardown(done) }
});
```

Another consequence of using vows' wonky approach to flow control, it's not clear to me how we extract the opts upward while still retaining a link to the done callback.

jaredhirsch:
This is a pretty sad situation on the whole, fingers crossed this latest hack lets us forget it once and for all.

I'm going to find a timebox to just try mocha + q + wd instead of vows + wd's internal use of caolan/async + wd.

jaredhirsch:
Kicking off some IE9 tests now--we'll see if stability seems improved

lloyd:
Q hurts my brain.  Seq less so.

-- lloyd (thumb-typing)

ozten:
+1 on replacing Q and standardizing on a flow control library that a majority of the devs are okay with.

Issue #2838, opened 2012-12-07
"unknown email address" thrown during initial signup with primary email address


lloyd:
STR:
1. create a new IDP at https://testidp.org
2. set the idp environment to https://login.dev.anosrep.org
3. use the sample email to login via http://dev.123done.org
4. keep doing #3 until it doesn't work
5. notice "Uncaught Error: unknown email address"

This exception is called right after list_emails is invoked.

lloyd:
I can reproduce this on dev every time.  I cannot reproduce this on an ephemeral instance I deployed with the exact same sha - transition.personatest.org

lloyd:
Here's what I know:
1. Occurs on dev, minified or not
2. Does not occur on ephemeral instance
3. Occurs on Linux chrome or osx chrome
4. Does not occur in incognito mode
5. Appears that the problem correlates with failed cert key and logout requests from comm iframe
6. Occurs on dev with or without minification

How can this be?   Next step is to disable bigtent on dev and dig into comm iframe

shane-tomlinson:
@lloyd - I am unable to reproduce this error as specified in Chrome/OSX. Is there a voodoo incantation that needs to be done?

lloyd:
No, it was a flakey temperamental error I spent hours trying to isolate, and upon restart it vanished.  

I don't know if this is a ghost or not.

ozten:
I'm also unable to reproduce.

Steps - Did you use 'click' or 'noauth' in your testidp.org?
Did you log out of the dialog and close it, or just 'this is not me' and re-enter your addresss.

shane-tomlinson:
Closing down as cannot reproduce, feel free to re-open if found again!

PR #2842, opened 2012-12-08
don't use storedid.type in getAssertion (bug that would prevent proper p...

lloyd:
...rovisioning of primary IdPs.  also, remove dead code, unnecesary network requests, and cosmetic improvments (comments and indentation).  mitigates issue #2838

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -186,18 +186,6 @@ BrowserID.User = (function() {
       // data.
       storage.setReturnTo("");
 
-      // To avoid too many address_info requests, returns from each
-      // address_info request are cached.  If the user is doing
-      // a addressVerificationPoll, it means the user was registering the address
-      // and the registration has completed.  Because the status is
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
I am a worried that the unit tests did not catch this problem. Before merging, a new test should be written to make sure that we do not re-introduce this bug.

shane-tomlinson:
After examining the code, I am happy with this change, but I still cannot reproduce the initial problem.

lloyd:
Yeah, I wish the problem which this attempts to address was easier to understand and reproduce.  I was getting reliable reproduction on dev at one moment, and then it disappeared the next.  Because shane's review is promising, and because I swear that after the introduction of this change I could no longer reproduce the initial problem that seemed very persistent, I feel like we might as well merge.  

This train is already risky, and @jrgm is going to give it his worst...  

We do shave a network call, and we do fix the last remaining code that is relying on .type from localstorage.  Clicking the green button, but not completely satisfied.

Issue #2834, opened 2012-12-07
[train-2012.12.07/dev] after 'this is not me', must enter same password for both secondaries on same account.


jrgm:
STR

1) set up so that you have two secondary emails on the same account.
2) Go to an RP and pop the dialog.
3) Click 'this is not me', enter one of the secondary addresses, enter the password, complete signin to the RP.
4) Sign out of the RP, and pop the dialog again.
5) Pick the other secondary address.

Expected: on prod/stage, you are not challenged to enter the password.

Actual: on dev, you are challenged to enter the password again for the account.

Or, is my expected wrong?

lloyd:
your expectation is right, this looks like a bug to me.  trying to reproduce...

jrgm:
Works for me on http://transition.myfavoritebeer.org/

jrgm:
verified stage train-2012.12.07. Fixed with PR-2843.

PR #2843, opened 2012-12-08
when an email is picked, transition to email_chosen which itself decides...

lloyd:
... whether authenticaiton is neccesary - issue #2834

### DIFF
--- a/resources/static/dialog/js/modules/pick_email.js
+++ b/resources/static/dialog/js/modules/pick_email.js
@@ -61,15 +61,22 @@ BrowserID.Modules.PickEmail = (function() {
     var record = checkEmail.call(self, email);
     if (!! record) {
       dialogHelpers.refreshEmailInfo.call(self, email, function (info) {
-	record = checkEmail.call(self, email);
-        if (record.cert)
+        record = checkEmail.call(self, email);
+        if (record.cert) {
           self.close("email_chosen", info);
-        else if ("transition_no_password" === info.state)
+        }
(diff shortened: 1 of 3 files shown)


resources/static/dialog/js/modules/pick_email.js
```diff
@@ -67,7 +67,7 @@ BrowserID.Modules.PickEmail = (function() {
         else if ("transition_no_password" === info.state)
           self.close("transition_no_password", info);
         else if ("secondary" === info.type)
```
shane-tomlinson: There should be tests for each of those cases.

shane-tomlinson:
@lloyd - I have added tests and added some {}, would you mind re-reviewing to make sure I didn't change any logic?

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -61,15 +61,22 @@ BrowserID.Modules.PickEmail = (function() {
     var record = checkEmail.call(self, email);
     if (!! record) {
       dialogHelpers.refreshEmailInfo.call(self, email, function (info) {
-	record = checkEmail.call(self, email);
-        if (record.cert)
+        record = checkEmail.call(self, email);
```
shane-tomlinson, line 64: Does record need to be fetched both here and on line 61?

lloyd, line 64: sure don't think so.  Only if refreshEmailInfo can change the record?

Issue #2839, opened 2012-12-07
Cancel button non-functional when primary IDP is broken


lloyd:
reproduced by the `broken-primary.js` automated test, currently in the `primary-automation-tests` branch.

ozten:
Thanks for fixing my broken code!

This also affects non-auth users. See pull request for details.

shane-tomlinson:
Adding @ozten's STRs for posterity:

STR

1) Create a new IdP https://testidp.org/
2) Log in with the primary
3) Log out
4) Update IdP's well known to {}
5) Attempt to log in
6) Note offline copy and click 'Cancel'

Actual: Nothing happens

Expected: Return to Auth screen

PR #2844, opened 2012-12-08
fix non-functional cancel button for broken primaries

lloyd:


shane-tomlinson:
Link to #2839

shane-tomlinson:
Does this fix a non-functional cancel button for broken primaries or just avoid it all together by going to the "primary not available" screen?

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -61,8 +61,10 @@ BrowserID.Modules.PickEmail = (function() {
     var record = checkEmail.call(self, email);
     if (!! record) {
       dialogHelpers.refreshEmailInfo.call(self, email, function (info) {
-	record = checkEmail.call(self, email);
-        if (record.cert)
+        record = checkEmail.call(self, email);
+        if ("offline" === info.state)
```
shane-tomlinson: Going along with https://github.com/mozilla/browserid/pull/2843#discussion_r2358831, this should have a test as well as have some {}s.

lloyd:
Avoids it all together by fixing screen transitions.

-- lloyd (thumb-typing)

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -61,8 +61,10 @@ BrowserID.Modules.PickEmail = (function() {
     var record = checkEmail.call(self, email);
     if (!! record) {
```
shane-tomlinson, line 62: Does the record need to be checked for here? No action is taken if a record is not present.

shane-tomlinson:
@lloyd - I rebased this off of branch issue3834 and added a new test. Mind reviewing?

lloyd:
r+ with shane's improvments

PR #2848, opened 2012-12-10
Fixes cancel button issue related to issue #2839

ozten:
No Auth user has a non-functioning cancel button, also.

STR

1) Create a new IdP https://testidp.org/
2) Log in with the primary
3) Log out
4) Update IdP's well known to {}
5) Attempt to log in
6) Note offline copy and click 'Cancel'

Actual: Nothing happens

Expected: Return to Auth screen

I see that this is because offline primary shows the error. Clicking cancel we do go back into the authentication state, but the error screen is still shown. This fix always hides the error screen when restarting the authentication flow. Not sure if that is the right fix.

### DIFF
--- a/resources/static/dialog/js/modules/authenticate.js
+++ b/resources/static/dialog/js/modules/authenticate.js
@@ -226,6 +226,7 @@ BrowserID.Modules.Authenticate = (function() {
       currentHint = null;
       dom.setInner(CONTENTS_SELECTOR, "");
       dom.hide(".returning,.start");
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@ozten - I am closing this down and taking what you have done here and expanding on it. The problem you found and the approach is correct, my change includes making sure to hide the delay and wait screens as well. Would love a review - PR #2851

ozten:
Awesome, thanks for your help.

Issue #2828, opened 2012-12-06
front-end tests - new messages output when running


jrgm:
Haven't looked into this much, but there are two messages that are new after the big landing of GH-2608 and friends.
1. actions called={"doAuthenticate":true}
2. request for 'post /wsapi/used_address_as_primary valid' is not mocked in resources/static/test/cases/xhr.js

shane-tomlinson:
This second one actually points out a problem where even if used_address_as_primary returns an error, no user facing action is taken and the dialog is started as normal. See https://github.com/mozilla/browserid/blob/2ec862608909115586f5df967fe266cdcae701a9/resources/static/dialog/js/modules/dialog.js#L298

lloyd:
@shane-tomlinson this is by design.  We didn't want to block the user for a non-fatal error.

shane-tomlinson:
@seanmonstar, @lloyd - is this the desired behavior? if used_address_as_primary fails, the dialog will continue on its merry way. 

It looks like this path can be taken if either: 1) the user is not authenticated or 2) the xhr request fails.

It seems like these should be handled by two distinct paths, the first can occur if the user cancels out of the authentication process with the primary. In this case, the user should be allowed to enter in a new address. 

Case 2 seems like we should show an error message.

lloyd:
If the XHR request fails, it means we were unable to tell the server the information it needs to turn off the transition.  So the user will see the screen again the next time.

If we display an error, the user will see an error, won't get to log in, and will see the transition screen again next time they try.

I think continuing regardless of whether the call succeeds is right

-- lloyd (thumb-typing)

shane-tomlinson:
I understand the end goal, but I am going to examine this closer - silent failures make me _extremely_ uncomfortable.

seanmonstar:
This is the desired effect. However, if you look at network.usedAddressAsPrimary, you'll see that it shouldn't ever make the request unless the dialog thinks its authenticated.

Reasoning was:
- If they are authenticated, then we won't call auth_with_assertion, or any other wsapi call, so we have no way of knowing if the primary was successful. So, we need to call this wsapi.
- If they aren't authenticated, this call shouldn't happen, because auth_with_assertion is needed, and that can mark the status for us.

In the case that the request fails, we definitely **do not** want to show an error message. This would prevent them from logging in, when error could be as benign as "Server Too Busy".

shane-tomlinson:
[...]

PR #2852, opened 2012-12-11
Issue 2846 ie8 failures

### LINKED ISSUE #2846: [ie8] Two test failures and an exception on IE8
This is with either login.dev.anosrep.org or transition.personatest.org. The exception happens well after the test failures.

```
failures: 
 [ 'dialog/js/modules/primary_offline: start controller with `add: false` and XHR error displays error screen (1, 0, 1)',
   'dialog/js/modules/set_password: create with transition_no_password (1, 1, 2)' ] }
```

Exception:

```
Can't move focus to the control because it is invisible, not enabled, or of a type that does not accept the focus. 
jquery-1.7.1.min.js, line 3 character 3117
```

shane-tomlinson:
@ozten - can you review this?

Fix IE8 errors
- Added a new test helpers - testElementTextContains.
- Use the new helper instead of text.indexOf to check for text in the DOM.
- Remove extra variables that were not used.
- Fix the thrown exception when focusing the password field in the unit tests.
- Remove the console.log message for `actions called={"doAuthenticate":true}`

fixes #2846
Partial fix for #2828

### DIFF
--- a/resources/static/test/cases/dialog/js/misc/state.js
+++ b/resources/static/test/cases/dialog/js/misc/state.js
@@ -480,7 +480,6 @@
       type: 'secondary',
       state: 'transition_to_secondary'
     });
-    console.log('actions called=' + JSON.stringify(actions.called));
     equal(actions.called.doAuthenticate, true, "doAuthenticate called");
   });
 
--- a/resources/static/test/cases/dialog/js/modules/set_password.js
+++ b/resources/static/test/cases/dialog/js/modules/set_password.js
@@ -10,6 +10,7 @@
       testHelpers = bid.TestHelpers,
       testElementExists = testHelpers.testElementExists,
       testElementNotExists = testHelpers.testElementDoesNotExist,
+      testElementTextContains = testHelpers.testElementTextContains,
       register = testHelpers.register;
 
   function createController(options) {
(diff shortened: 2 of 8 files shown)


resources/static/test/cases/dialog/js/modules/set_password.js
```diff
@@ -63,9 +64,9 @@
       email: "transition@password.no",
       transition_no_password: true
     });
-    var msg = $("#set_password .inputs li").text();
-    ok(msg.indexOf("no longer allows"), "transition message shown");
```
shane-tomlinson, line 67: String.prototype.indexOf returns -1 if a value is not found. The index for this text in IE8 was 0.

resources/static/test/cases/dialog/js/misc/state.js
```diff
@@ -480,7 +480,6 @@
       type: 'secondary',
       state: 'transition_to_secondary'
     });
-    console.log('actions called=' + JSON.stringify(actions.called));
```
ozten, line 483: Doh. Thanks.

ozten:
I'm r+, but you might want to do the following to the test suite:

1) In the testing framework, detect that we're using ie8 by simulating this failure and setting a `ie8` flag.

2) For all browsers, run through the steps, to catch dom exceptions, but

3) For tests which ie8 will get an empty string, skip checking for the expected string. Alternatively have a check for empty string, you're call.

shane-tomlinson:
@ozten - Thanks for the review, I made two changes from your suggestions.

1) in testElementTextContains and testElementTextEquals, make sure there are no explosions if the elements to get the text from are not found.

2) Look more closely at the cause of the focus exception in IE8. The cause was due to a timing error where complete was called before the focus occurred. When complete called, the DOM got torn down before focus occurred. Since the element was no longer in the DOM, focus excepted. Since I moved complete into the showHint callback already, the problem was already side stepped and the try/catch was unnecessary. I removed it. This means no IE8 specific testing for the exception.

ozten:
Thanks for jumping on 2852, great work!

PR #2855, opened 2012-12-12
Bug 820446 - issue with bugzilla and persona login

jedp:
Monday 2012-12-10 while rolling out train-2012.11.23 to production there was
an issue with login to bugzilla with persona.

jrgm theorized that this was a string/int comparison error
(http://irclog.gr/#show/irc.mozilla.org/identity/187967)

git bisecting confirms that he is correct, and that the problem crept in here:
https://github.com/mozilla/browserid/commit/d494afbbce176697998f3b39764200d8732ad256#L28L77

### DIFF
--- a/lib/verifier/certassertion.js
+++ b/lib/verifier/certassertion.js
@@ -35,7 +35,8 @@ logger.debug("This verifier will accept assertions issued by " + HOSTNAME);
 //   it might be strangely formed.
 function compareAudiences(want, got) {
   function normalizeParsedURL(u) {
-    if (!u.port) u.port = u.protocol === 'https:' ? 443 : 80;
+    // return string because url.parse returns string
+    if (!u.port) u.port = u.protocol === 'https:' ? '443' : '80';
     return u;
   }
 
@@ -155,4 +156,5 @@ function verify(assertion, audience, successCB, errorCB) {
     }, errorCB);
 }
 
+exports.compareAudiences = compareAudiences;
 exports.verify = verify;
--- a/tests/verifier-test.js
+++ b/tests/verifier-test.js
@@ -15,7 +15,9 @@ config = require('../lib/configuration.js'),
 jwcrypto = require('jwcrypto'),
 http = require('http'),
 querystring = require('querystring'),
-path = require('path');
+path = require('path'),
+url = require('url'),
+compareAudiences = require('../lib/verifier/certassertion').compareAudiences;
 
 var suite = vows.describe('verifier');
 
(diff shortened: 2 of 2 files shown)


lib/verifier/certassertion.js
```diff
@@ -35,7 +35,8 @@ logger.debug("This verifier will accept assertions issued by " + HOSTNAME);
 //   it might be strangely formed.
 function compareAudiences(want, got) {
   function normalizeParsedURL(u) {
-    if (!u.port) u.port = u.protocol === 'https:' ? 443 : 80;
+    // return string because url.parse returns string
+    if (!u.port) u.port = u.protocol === 'https:' ? '443' : '80';
```
ozten, line 39: > url.parse returns string

Ya, that is a bit odd.

Another way to roll, would be `parseInt`, but your solution is solid.

lib/verifier/certassertion.js
```diff
@@ -155,4 +156,5 @@ function verify(assertion, audience, successCB, errorCB) {
     }, errorCB);
 }
 
+exports.compareAudiences = compareAudiences;
```
ozten, line 159: Thanks for making that unit testable.

tests/verifier-test.js
```diff
@@ -34,6 +36,36 @@ const TEST_EMAIL = 'someuser@somedomain.com',
...
+  'audiences': {
+    'http://fakesite.com and http://fakesite.com:80': matchesAudience(true),
+    'https://fakesite.com and https://fakesite.com:443': matchesAudience(true),
+    'http://fakesite.com:8000 and http://fakesite.com:8000': matchesAudience(true),
+    'https://fakesite.com:9000 and https://fakesite.com:9000': matchesAudience(true),
+
+    'http://fakesite.com:8100 and http://fakesite.com:80': matchesAudience(false),
+    'https://fakesite.com:9100 and https://fakesite.com:443': matchesAudience(false),
+    'http://fakesite.com:80 and http://fakesite.com:8000': matchesAudience(false),
+    'https://fakesite.com:443 and https://fakesite.com:9000': matchesAudience(false)
```
ozten, line 64: Very cool.

ozten:
Travis went to crazy town on the MYSQL tests. I'll run locally.
https://travis-ci.org/mozilla/browserid/jobs/3622187

ozten:
Passes test locally. I also ran the tests against the old code and verified the new tests fail:

```
audiences http://fakesite.com and http://fakesite.com:80
✗ should match
  » true == false

audiences https://fakesite.com and https://fakesite.com:443
✗ should match
  » true == false
✗ Broken » 85 honored ∙ 2 broken (10.819s)
```

jrgm:
"""Travis went to crazy town on the MYSQL tests. I'll run locally.
https://travis-ci.org/mozilla/browserid/jobs/3622187"""

Cool new feature: load that link, sign in with github and if you have commit privs, switch to the build tab, click the control widget and then 'restart build'. It will try again.

PR #2856, opened 2012-12-12
Train 2012.11.23  hotfix bug820446

ozten:


jrgm:
So I reviewed production logs that @gene1wood provided, for the particular message in the ~5m window on 2012-12-10 that show a 'verification failure, audience mismatch'. All but one (which was a valid mismatch of literal IP vs. domain) were of the type where the default port for the given scheme was reported as a 'port mismatch'.

(Side note: there are inevitably a lot of other messages of this mismatch at other times as people experiment).

I'm good with this change, merging. Thanks @jedp,  @ozten

PR #2864, opened 2012-12-13
Replace implicit waits with explicit waits.

### LINKED ISSUE #2857: replace implicit with explicit waits
I think we should be able to replace implicit with explicit waits under the covers of our extension methods wclick, wfind, etc.

Interesting, related: https://blog.mozilla.org/webqa/2012/07/12/how-to-webdriverwait/

shane-tomlinson:
@6a68 - Could you review this one?
- Set the default implicit wait to 0.
- Increase the DEFAULT_TIMEOUT_MS to 40s to give webdriver more time to find elements.

fixes #2857

### DIFF
--- a/automation-tests/lib/wd-extensions.js
+++ b/automation-tests/lib/wd-extensions.js
@@ -229,51 +229,15 @@ wd.prototype.click = function(which, cb) {
 // great, click it. If not, move on.
 wd.prototype.wclickIfExists = function(opts, cb) {
   var self=this;
-  // This function assumes the implicit wait when entering wclickIfExists is
-  // the same as the DEFAULT_IMPLICIT_WAIT_MS. Calling this function allows us
-  // to set the implicit wait and only send the request over the wire if the
-  // implicit wait actually needs updated. If the default is set to 0, no wire
-  // protocol request will be made.
-  var currImplicitWaitMS = timeouts.DEFAULT_IMPLICIT_WAIT_MS;
-  function setImplicitWaitTimeout(implicitWaitMS, done) {
-    if (implicitWaitMS !== currImplicitWaitMS) {
-      self.setImplicitWaitTimeout(implicitWaitMS, function(err) {
-        if (!err) currImplicitWaitMS = implicitWaitMS;
-        done(err);
-      });
+  self.wclick(opts, function(err, el) {
+    // These two errors mean the element does not exist (or is not shown)
+    // and we can move on without failing. Any other failures should be
+    // propagated.
+    if (err && /timeout hit/.test(err) || /window gone/.test(err)) {
+      err = null;
     }
-    else {
-      done(null);
-    }
-  }
 
-  // webdriver has a problem where if you search for an element that is
-  // contained in a window that has closed itself, no response is returned. To
-  // avoid this, set the implicit wait timeout to 0, try the click, if the
-  // timeout hit or window gone exceptions are thrown, things are ok,
-  // just move on.
-  setImplicitWaitTimeout(0, function() {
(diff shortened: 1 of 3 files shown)


automation-tests/lib/wd-extensions.js
```diff
@@ -230,12 +229,30 @@ wd.prototype.click = function(which, cb) {
...
+    }
+  }
+
   // webdriver has a problem where if you search for an element that is
   // contained in a window that has closed itself, no response is returned. To
   // avoid this, set the implicit wait timeout to 0, try the click, if the
   // timeout hit or window gone exceptions are thrown, things are ok,
   // just move on.
-  self.setImplicitWaitTimeout(0, function() {
+  setImplicitWaitTimeout(0, function() {
```
jaredhirsch: not sure I understand why this was changed?

automation-tests/lib/wd-extensions.js
```diff
@@ -249,7 +266,7 @@ wd.prototype.wclickIfExists = function(opts, cb) {
       // setImplictWaitTimeout fails if the 'window gone' error is
       // returned from wclick.
       if (!/window gone/.test(err)) {
-        self.setImplicitWaitTimeout(timeouts.DEFAULT_TIMEOUT_MS, function(err) {
+        setImplicitWaitTimeout(timeouts.DEFAULT_IMPLICIT_WAIT_MS, function(err) {
```
jaredhirsch: here as well?

automation-tests/lib/wd-extensions.js
shane-tomlinson: I am calling the local version of setImplicitWaitTimeout - the local version will only set the implicit timeout if it is different than the current setting. This allows us to re-introduce the implicit timeout in configuration without changing code.

jaredhirsch: Ah, ok. Can we call it something like "lazySetImplicitWaitTimeout" so it's clearly something different? I'm afraid I would not have caught this if skimming in a hurry.

automation-tests/lib/wd-extensions.js
```diff
@@ -230,12 +229,30 @@ wd.prototype.click = function(which, cb) {
 // great, click it. If not, move on.
 wd.prototype.wclickIfExists = function(opts, cb) {
   var self=this;
+  // This function assumes the implicit wait when entering wclickIfExists is
+  // the same as the DEFAULT_IMPLICIT_WAIT_MS. Calling this function allows us
+  // to set the implicit wait and only send the request over the wire if the
+  // implicit wait actually needs updated. If the default is set to 0, no wire
+  // protocol request will be made.
+  var currImplicitWaitMS = timeouts.DEFAULT_IMPLICIT_WAIT_MS;
+  function setImplicitWaitTimeout(implicitWaitMS, done) {
```
shane-tomlinson: I am quite OK with taking this code out if we no longer plan on using implicit timeouts. I was being cautious and giving us space to add it back in using configuration. Less code is better code.

jaredhirsch:
r+ 

for reference, from IRC:

stomlinson      _6a68: what do you think about removing it all together?            7:51
_6a68       stomlinson: You mean, adding a little more network traffic, but making our code less clever?            7:52
_6a68       stomlinson: or do you mean, removing the implicit timeout toggling, since we won't be using implicit timeouts anymore           7:53
stomlinson      _6a68: #2           7:57
_6a68       stomlinson: +1          7:57
_6a68       it's really just cruft at this point, and if we need to bring it back, we'll know which commit to revert if we chuck it all at once         7:58
stomlinson      _6a68: ripping it out then.

PR #2877, opened 2012-12-17
Reenable all platforms

jaredhirsch:
Let's re-enable all platforms for the home stretch, and see how things look across teh board


jaredhirsch:
aight, /me pulls the trigger

PR #2878, opened 2012-12-18
unverified email for b2g

seanmonstar:
An implementation for B2G to ask allow users to log in even if
Persona has not verified that the user owns the email address.

Starts with navigator.id.request({ allowUnverified: true });
When the assertion is returned to the RP, and if they are using
the Persona verifier, must be sure to include an allowUnverified=true
parameter to the verifier, otherwise it will be rejected.

Behind the scenes, when the dialog tries to stage the user, instead
the user is created immediately, an email is added in our email table,
and a stagedEmail token is created and send to the user. When
allowUnverified is true, our wsapi will also check for a staged email/pass,
and use that to allow the user to login.

### DIFF
--- a/bin/keysigner
+++ b/bin/keysigner
@@ -88,7 +88,7 @@ try {
 
 
 // and our single function
-app.post('/wsapi/cert_key', validate(certKey.args), function(req, resp) {
+app.post('/wsapi/cert_key', validate({ 'email': 'email', 'pubkey': 'pubkey', 'ephemeral': 'boolean', 'unverified': { type: 'boolean', required: false } }), function(req, resp) {
   var startTime = new Date();
   var forceIssuer = req.params.forceIssuer;
   var opts = {};
@@ -101,7 +101,8 @@ app.post('/wsapi/cert_key', validate(certKey.args), function(req, resp) {
   }
   cc.enqueue(_.extend(opts, {
     pubkey: req.params.pubkey,
-    email: req.params.email,    
+    email: req.params.email,
+    unverified: !!req.params.unverified,
     validityPeriod: (req.params.ephemeral ? config.get('ephemeral_session_duration_ms') : config.get('certificate_validity_ms')),
     hostname: HOSTNAME
   }), function (err, r) {
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -80,6 +80,9 @@ pre {
     </li><li>
       <input type="checkbox" id="forceAuthentication">
       <label for="forceAuthentication">Force Authentication</label><br />
+    </li><li>
+      <input type="checkbox" id="allowUnverified">
+      <label for="allowUnverified">Allow Unverified</label><br />
     </li>
   </ul>
     <button class="assertion">Get an assertion</button>
@@ -143,7 +146,8 @@ function checkAssertion(assertion) {
     data: {
       assertion: assertion,
       audience: window.location.protocol + "//" + window.location.host,
-      forceIssuer: $('#forceIssuer').attr('checked') ? "issuer.domain" : undefined
+      forceIssuer: $('#forceIssuer').attr('checked') ? "issuer.domain" : undefined,
+      allowUnverified: $('#allowUnverified').attr('checked') ? true : false
     },
     success: function(data, textStatus, jqXHR) {
       var old = $(".loginEvents > pre").text() + "\n";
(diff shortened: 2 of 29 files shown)


example/rp/index.html
```diff
@@ -80,6 +80,9 @@
     </li><li>
       <input type="checkbox" id="forceAuthentication">
       <label for="forceAuthentication">Force Authentication</label><br />
+    </li><li>
+      <input type="checkbox" id="allowUnverified">
+      <label for="allowUnverified">Allow Unverified</label><br />
```
jedp, line 85: helpful to have this example

lib/db/json.js
```diff
@@ -207,11 +207,20 @@ exports.emailToUID = function(email, cb) {
...
+exports.userOwnsEmail = function(uid, email, staged, cb) {
   sync();
+  if (typeof staged === "function") {
+    cb = staged;
+    staged = false;
+  }
   var m = jsel.match(":root > object:has(:root > .id:expr(x=" + ESC(uid) + ")):has(.emails > ." + ESC(email) + ")", db.users);
+  var owned = m && m.length === 1;
+  if (!owned && staged) {
+    m = jsel.match(":root > object:has(:root > .existing_user:expr(x=" + ESC(uid) +")):has(:root > .email:expr(x=" + ESC(email) + "))", db.staged);
```
jedp, line 219: jsonselect ftw

lib/verifier/certassertion.js
```diff
@@ -89,7 +90,8 @@ function compareAudiences(want, got) {
 // assertion is a bundle of the underlying assertion and the cert list
 // audience is a web origin, e.g. https://foo.com or http://foo.org:81
 // forceIssuer is a hostname or `undefined` for normal BID protocol
-function verify(assertion, audience, forceIssuer, successCB, errorCB) {
+// allowUnverified is boolean to check for email or unverified-email
+function verify(assertion, audience, forceIssuer, allowUnverified, successCB, errorCB) {
```
jedp, line 94: i like that you're explicit about the parameters.  in this case, the long arglist is good for making things really clear

seanmonstar, line 94: and i was thinking that line makes my codey sense tingle.

resources/static/common/css/style.css
```diff
@@ -481,7 +479,6 @@ footer .help {
   background-image: url("/common/i/grain.png"),     -ms-linear-gradient(top, #dadee1, #c7ccd0);
   background-image: url("/common/i/grain.png"),      -o-linear-gradient(top, #dadee1, #c7ccd0);
   background-image: url("/common/i/grain.png"),         linear-gradient(top, #dadee1, #c7ccd0);
-  */
```
jedp, line 484: awesome.  thanks.

tests/unverified-email-test.js
```diff
@@ -0,0 +1,195 @@
...
+    },
+    "works swimmingly": function(err, r) {
+      assert.strictEqual(r.code, 200);
+      assert.isString(r.body);
+    },
+    "provides the unverified cert": function(err, r) {
+      unverified_cert = r.body;
+      assert.lengthOf(unverified_cert.split('.'), 3);
+      var principal = JSON.parse(new Buffer(unverified_cert.split('.')[1], 'base64').toString()).principal;
+      // should not be an email property, only unverified-email
```
jedp, line 118: splendid

jedp:
Wow, this looks great, Sean.  I always love your clean code and attention to details.

Since you're writing additional questions in https://id.etherpad.mozilla.org/b2g-unverified-email, my impulse is to merge this baby so we can start testing it and working through the edge cases together.

I'm just going to wait for Travis to finish doing his thing ...

bin/keysigner
```diff
@@ -88,7 +88,7 @@ try {
 
 
 // and our single function
-app.post('/wsapi/cert_key', validate(certKey.args), function(req, resp) {
+app.post('/wsapi/cert_key', validate({ 'email': 'email', 'pubkey': 'pubkey', 'ephemeral': 'boolean', 'unverified': { type: 'boolean', required: false } }), function(req, resp) {
```
ozten, line 91: Hmm, this broke my code. I think it was a merge resolution error.

I think `certKey.args` is better as it reduces duplication. If we keep the code this way, I'll need `, 'forceIssuer': {type: 'hostname', required: false}` added back.

seanmonstar, line 91: oh... i removed that because it's not passing `allowUnverified`, but `unverified`. Otherwise, the keysigner process would need another db request...

PR #2889, opened 2012-12-20
fix cache poisioning introduced by borked production push

lloyd:
Here's the full summary of what this does, copied from email to dev-identity@:

there was an event recently where we were running some number of our servers in an odd state.  They had updated resources but were not restarted.  Shane did a pro-active assessment of impact on this list, and subsequent to that we received a report of a broken dialog in issue 2863.

Skipping some parts and zooming ahead - the explanation from Shane is that we might see the screenshots in that issue if NEW html was served with OLD css.

In auditing etagify and cachify (the libraries that we built and use to generate good cache headers) - the only path I can see where this were be plausible is if _some_ of the processes in a single data center were updated while others weren't.  In this case we could get an html request routed to a server serving NEW resources with embedded "cachified" urls.  If those cachified urls were then fetched from a server that was not updated with new resources, it could serve OLD data under the NEW cachified url -- Poisioning the cache for css/js resources.

Gene: Was it the case that some servers in a single data center were updated while others were not?

All of this aside, we can push a change that will invalidate browser caches for all of the resources that we serve, and I feel fairly confident that this will resolve any issues.  To accomplish this we simply need to change combined css/js in a way that will alter minified output (cannot be whitespace only, which would be lost in minification).

Now the question is, should we?  Given that we have 3 internal reproductions of the issue and at least one external, should we craft a tiny commit to alter our resources and guarantee that this does not adversely affect users that are not speaking up?

If we don't create and push this fix this week, then it seems like we just live with adverse affects.  

For completeness, a user workaround is to force reload the dialog or clear their cache.


jedp:
Clever fix, @lloyd  :)

jrgm:
Oops, missed https://github.com/mozilla/browserid/pull/2889#commitcomment-2336939 between my first comment and my r+, but that still stands. We can consider whether to change the the html5shiv at all.

shane-tomlinson:
I used scripts/capture_deployed_resources.js to capture resources of the current prod and an ephemeral instance with the proposed hotfix. I then manually compared the old vs new SHAs of the external includes, two resources may need to be updated for completeness - authenticate_with_primary.js and relay.js.

authenticate_with_primary.js is made of auth_with_idp/main.js and lib/winchan.js 

Winchan was last updated on oct 23, 2012, main.js on Aug 17, 2012. 

relay.js is composed of relay/relay.js

This was last updated Jan 10, 2012. There is no danger here.

Otherwise r+.

shane-tomlinson:
Nice work btw @lloyd!

shane-tomlinson:
One final comment - Since Winchan was updated between Oct 12 and Nov 23rd, we should force an update to authenticate_with_primary.js.

callahad:
@shane-tomlinson to merge after that final update.

lloyd:
changed the commit to not modify `html5shim.js` because it hasn't changed in ages so no poisioning is possible with this file.

Also tweaked winchan.js to cause a new md5 to be generated for `authenticate_with_primary.js` which did change in the vulnerability window.

all feedback addressed, this is ready for merge and deployment if @jrgm and @shane-tomlinson are on board.

PR #2894, opened 2012-12-21
Back to implicit waits

jaredhirsch:
It turns out that switching from implicit to explicit waits didn't significantly improve test stability[1]. 

So let's get rid of it.

[1] https://gist.github.com/4348207


jaredhirsch:
Shane argued correctly that explicit waits are what we need lots of. So, we'll leave the code as-is.

PR #2898, opened 2012-12-21
[b2g] The verifier service should be explicit about unverified email

ozten:
Steps to Reproduce
1) Sign in with a new secondary address and allow unverified
2) create password
3) Note success of assertion

Actual

```
{"status":"okay","email":"unv@foo.com","audience":"192.168.186.138:10001", "expires":1356117306992,"issuer":"192.168.186.138"}
```

Expected

```
{"status":"okay","unverified-email":"unv@foo.com","audience":"192.168.186.138:10001", "expires":1356117306992,"issuer":"192.168.186.138"}
```

Note: this requirement is specified in https://bugzilla.mozilla.org/show_bug.cgi?id=794634#c2

### DIFF
--- a/bin/verifier
+++ b/bin/verifier
@@ -7,6 +7,7 @@
 require('../lib/baseExceptions').addExceptionHandler();
 
 const
+_ = require('underscore'),
 util = require("util"),
 path = require('path'),
 url = require('url'),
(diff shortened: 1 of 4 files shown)


seanmonstar:
the test in unverified-email-test.js needs to be adjusted also

ozten:
So weird, I had green locally. Maybe I'm missing something...

I'll fix now.

PR #2907, opened 2013-01-04
authenticate with unverified email and allowUnverified

### LINKED ISSUE #2902: [b2g][unverified] fatal error when trying to reuse unverified email address
STR

1) Using a current unagi build on a phone, configure /data/local/user.js to enable native navigator.id:

```
pref("dom.identity.enabled", true);
pref("toolkit.identity.debug", true);
pref("layout.css.report_errors", false);
```

(The layout.css.report_errors isn't necessary, but seriously, the noise about CSS "declaration dropped" is useless information).
2) Use the firefox browser on the phone and go to `https://marketplace-dev.allizom.org`
3) Click on `Log in/Register` to bring up the trusted UI.
4) Enter an email address that has not been used before; then enter a password twice. You will be authenticated.
5) (Also an email to confirm that address will be sent out; do not verify that email address).
6) Now logout of marketplace-dev.
7) Click `Log in/Register` to bring up the dialog again.
8) Click `This is not me`. Re-enter the same email address from step 4) and a password twice.

Expected: this unverified user should be restaged again

Actual: "We are very sorry. The server is under extreme load."

```
Action: Creating Account
Network Info: POST /wsapi/stage_user
Response Code: 503
Response Text/Error Type: Service Unavailable: database unavailable
```

In the dbwriter.log on notoriousb2g.personatest.org:

```
unexpected database failure: Error: Duplicate entry 'blahblah@gmail.com' for key 'address' -- at exports.createUnverifiedUser
(/home/app/code/lib/db/mysql.js:680:13)
```

seanmonstar:
fixes #2902

### DIFF
--- a/resources/static/dialog/js/modules/authenticate.js
+++ b/resources/static/dialog/js/modules/authenticate.js
@@ -39,15 +39,18 @@ BrowserID.Modules.Authenticate = (function() {
   }
 
   function hasPassword(info) {
+    /*jshint validthis:true*/
     return (info && info.email && info.type === "secondary" && 
-      (info.state === "known" || info.state === "transition_to_secondary" ));
+      (info.state === "known" ||
+       info.state === "transition_to_secondary" ||
+       info.state === "unverified" && this.allowUnverified));
   }
 
   function initialState(info) {
     /*jshint validthis: true*/
     var self=this;
     self.submit = checkEmail;
-    if (hasPassword(info)) {
+    if (hasPassword.call(self, info)) {
       addressInfo = info;
       enterPasswordState.call(self, info.ready);
     }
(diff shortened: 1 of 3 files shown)


resources/static/dialog/js/modules/authenticate.js
```diff
@@ -85,13 +88,13 @@ BrowserID.Modules.Authenticate = (function() {
         self.close("primary_user", info, info);
       }
       else if (!!self.forceIssuer && 'default' !== self.forceIssuer) {
-	if (hasPassword(info)) {
+        if (hasPassword.call(self, info)) {
           enterPasswordState.call(self);
-	} else {
+        } else {
```
ozten, line 93: Thanks for the cleanup.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -39,15 +39,18 @@ BrowserID.Modules.Authenticate = (function() {
...
+       info.state === "transition_to_secondary" ||
+       info.state === "unverified" && this.allowUnverified));
   }
 
   function initialState(info) {
     /*jshint validthis: true*/
     var self=this;
     self.submit = checkEmail;
-    if (hasPassword(info)) {
+    if (hasPassword.call(self, info)) {
```
ozten, line 53: Ah, good catch.

resources/static/test/cases/dialog/js/modules/authenticate.js
```diff
@@ -114,6 +114,21 @@
     });
   });
 
+  asyncTest("allowUnverified with an unverified email declared in options - show password field", function() {
+    controller.destroy();
```
ozten, line 118: Excellent.

resources/static/dialog/js/modules/authenticate.js
seanmonstar, line 53: It wasn't a bug before. I just needed access to `this` inside hasPassword, to check for `allowUnverified`

PR #2916, opened 2013-01-08
Allow logout to be called before .request or .watch.

shane-tomlinson:
A slightly simpler solution to #2529 which pushes the majority of the logic into communication_iframe/start.js

### DIFF
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -92,8 +92,18 @@
   });
 
   chan.bind("logout", function(trans, params) {
-    if (loggedInUser != null) {
+    // The assumption is that the intention behind a call to
+    // navigator.id.logout is to clear any session, whether it's locally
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@yourcelf and @lloyd - could you two review this?

@lloyd - in include.js, I had to change the invocation of the `logout` message from `.notify` to `.call`. I understand why you would not want to pass a full transaction to handlers invoked using .notify, but is there a reason to not pass an object with at least the origin?

Does the change from .notify to .call have unintended consequences?

yourcelf:
Yup, I can confirm that this indeed does exactly what I was after, and looks like a better solution.  Thanks!!

lloyd:
closed in favor of #2928 (yeah, I'm rediculous - woulda just updated this PR but it's in @shane-tomlinson 's fork)

PR #2920, opened 2013-01-09
Test dialog behavior of IdP disabling support

ozten:
Two main tests.

1) log in with a primary, disable primary support, attempt to log in, set new password, confirm password, login with password

2) disabled primary, setup new user (vanilla secondary)

### DIFF
--- a/automation-tests/tests/idp_transition/primary_shuts_down_single_email.js
+++ b/automation-tests/tests/idp_transition/primary_shuts_down_single_email.js
@@ -1,0 +1,142 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const
+assert = require('../../lib/asserts.js'),
+CSS = require('../../pages/css.js'),
+CreateIdP = require('../../lib/testidp.js').CreateIdP,
+path = require('path'),
+persona_urls = require('../../lib/urls.js'),
+restmail = require('../../lib/restmail.js'),
+runner = require('../../lib/runner.js'),
+testSetup = require('../../lib/test-setup.js');
+
+var browser, testUser, testIdp;
+
+runner.run(module, {
+  "setup": function(done) {
+    testSetup.setup({browsers: 1, testidps: 1}, function(err, fixtures) {
+      if (fixtures) {
+        browser = fixtures.browsers[0];
+        var llIdP = fixtures.testidps[0];
+        testIdp = new CreateIdP(llIdP.idp);
+        testUser = llIdP.getRandomEmail();
+        noAuthTestUser = llIdP.getRandomEmail();
+      }
+      done(err);
+    });
+  },
+  "create a new selenium session": function(done) {
+    testSetup.newBrowserSession(browser, done);
+  },
+  "load 123done and wait for the signin button to be visible": function(done) {
+    browser.get(persona_urls["123done"], done);
+  },
+  "click the signin button": function(done, el) {
+    browser.wclick(CSS['123done.org'].signinButton, done);
+  },
+  "switch to the dialog when it opens": function(done) {
+    browser.wwin(CSS["persona.org"].windowName, done);
+  },
+  "Happy, healthy primary": function(done) {
+    testIdp.putWellKnown(testIdp.getNoAuth(), true, function(err, resp, body) {
+      testIdp.putEnv(persona_urls['persona'] + '/', done);
+    });
+  },
+  "Sign in": function(done) {
+    browser.chain({onError: done})
+      .wtype(CSS['dialog'].emailInput, testUser)
+      .wclick(CSS['dialog'].newEmailNextButton, done);
+  },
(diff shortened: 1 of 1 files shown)


automation-tests/tests/idp_transition/primary_shuts_down_single_email.js
```diff
@@ -0,0 +1,142 @@
...
+  },
+  "verify we're signed in to 123done": function(done) {
+    browser.chain({onError: done})
+      .wwin()
+      .wtext(CSS['123done.org'].currentlyLoggedInEmail, function(err, text) {
+        done(err || assert.equal(text, testUser));
+       });
+  },
+  "The IdP disables support": function(done) {
+    testIdp.putWellKnown({disabled: true}, false, done);
```
shane-tomlinson, line 62: Instead of using putWellKnown({disabled: true},...), I asked @lloyd to add a testIdp.disableSupport(cb), which is a bit clearer. Can you make use of that here?

automation-tests/tests/idp_transition/primary_shuts_down_single_email.js
```diff
@@ -0,0 +1,142 @@
...
+    browser.get(persona_urls["123done"], done);
+  },
+  "click the signin button": function(done, el) {
+    browser.wclick(CSS['123done.org'].signinButton, done);
+  },
+  "switch to the dialog when it opens": function(done) {
+    browser.wwin(CSS["persona.org"].windowName, done);
+  },
+  "Happy, healthy primary": function(done) {
+    testIdp.putWellKnown(testIdp.getNoAuth(), true, function(err, resp, body) {
```
shane-tomlinson, line 45: For clarity, I asked @lloyd to add a testIdp.enableSupport. You should be able to replace this line with:

```
testIdp.enableSupport(false, function(err, resp, body) {
```

automation-tests/tests/idp_transition/primary_shuts_down_single_email.js
```diff
@@ -0,0 +1,142 @@
...
+testSetup = require('../../lib/test-setup.js');
+
+var browser, testUser, testIdp;
+
+runner.run(module, {
+  "setup": function(done) {
+    testSetup.setup({browsers: 1, testidps: 1}, function(err, fixtures) {
+      if (fixtures) {
+        browser = fixtures.browsers[0];
+        var llIdP = fixtures.testidps[0];
```
shane-tomlinson, line 24: What does llIdp stand for?

automation-tests/tests/idp_transition/primary_shuts_down_single_email.js
```diff
@@ -0,0 +1,142 @@
...
+      .wtype(CSS['dialog'].existingPassword, 'password')
+      .wclick(CSS['dialog'].returningUserButton)
+      .wwin(done);
+  },
+  "Check testUser is logged in again": function(done) {
+      browser.wtext(CSS['123done.org'].currentlyLoggedInEmail, function(err, text) {
+        done(err || assert.equal(text, testUser));
+      });
+  },
+  "noAuthTestUser tries to log in": function(done) {
```
shane-tomlinson, line 110: What purpose does testing the noAuthTestUser have? It seems like this should already be covered in other tests. Is the noAuthTestUser on the same domain as the disabled IdP from above? If so, that makes sense, if not it seems like this can be safely removed as test duplication.

shane-tomlinson, line 110: Ahha. After reading the PR comment, this test makes more sense. Perhaps put the comment you wrote in the PR comment here for clarity?

shane-tomlinson:
I dig it! MORE TESTS!

Additional comments:
- We already have an idp-transition directory, this adds a new idp_transition directory. Can you move the test file from idp_transition to idp-transition and replace the `_` with `-` in the filename? That way when we are running individual tests from the command line, we do not have to remember "ah, yeah, primary_shuts_down_single_email uses _ instead of -"
- When running with `./scripts/run-all.js -i 10 -t primary_shuts_down_single_email`, 10 out of 10 runs failed with the following error: 

```
primary_shuts_down_single_email - vista_firefox_16 (1 failure(s), 0 warnings):
  15: Check testUser is logged in again, errors:
    *   timeout hit
  >> https://saucelabs.com/tests/baebe8e2114246eea8774a4f5d33a25c
```

Thanks for the additional tests @ozten!

ozten:
Sorry I didn't notice `idp_transition` which updating. Lame.

Sauce Labs - Do you have the test id, so I can watch the video and look at the protocol interactions?

shane-tomlinson:
`https://saucelabs.com/tests/baebe8e2114246eea8774a4f5d33a25c` should get you the info that you need. It looks like the dialog may never close. I am wondering if the "Is this your computer" screen is being displayed after line 102, which would cause this symptom.

If this is the case, adding a `.wclickIfExists(CSS['dialog'].notMyComputerButton)` before the `.wwin(done);` should fix it.

shane-tomlinson:
Yeah, adding `.wclickIfExists(CSS['dialog'].notMyComputerButton)` before the `.wwin(done);` on line 103 fixes the problem. 10 out of 10 passes.

The problem is the "Is this your computer" screen is being shown when the user tries to log in a second time with their password.

PR #2923, opened 2013-01-09
delete staged entries when canceling an Account

### LINKED ISSUE #2919: [b2g] account_cancel of unverified user has 503 error on foreign key constraint
STR
1. With marketplace sign up as a new unverified user
2. Open a new tab and load notoriousb2g.personatest.org
3. Sign in with that user's email address and password.
4. In the admin UI, click on 'cancel your account'.

Expected: all relevant rows are removed in database and UI shows the account is gone.
Actual: UI shows 'We are very sorry...' with details of POST /wsapi/account_cancel and 503 code.

In the server logs: 

```
{"level":"warn","message":"database is down, cannot process request: Error: Cannot delete or update a parent row:
 a foreign key constraint fails (`browserid`.`staged`, CONSTRAINT `staged_ibfk_1` FOREIGN KEY (`existing_user`) 
 REFERENCES `user` (`id`))","timestamp":"2013-01-09T00:36:35.355Z"}
```

The email row is gone, but it chokes on removing the user row due to the fkey on staged.

seanmonstar:
like it says...

fixes #2919

### DIFF
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -561,6 +561,15 @@ exports.cancelAccount = function(authenticated_uid, cb) {
   }
 
   if (i < db.users.length) {
+    var m = jsel.match(":root > object:has(.existing_user:expr(x=" + ESC(authenticated_uid) + "))", db.staged);
+    if (m.length) {
+      m.forEach(function(staged) {
+        var token = db.stagedEmails[staged.email];
+        delete db.stagedEmails[staged.email];
+        delete db.staged[token];
+      });
+    }
+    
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@seanmonstar - Does a parallel change be made for the json adapter for consistency?

seanmonstar:
@shane-tomlinson it probably should! just didn't remember, because the tests passed with JSON db.

lloyd:
nice catch @shane-tomlinson - you were right, I was wrong!

lloyd:
r+

PR #2928, opened 2013-01-10
Allow logout to be called before .request or .watch.

lloyd:
A slightly simpler solution than #2916 which adds a feature to jschannel so we don't have to use `.call` just to get origin provided to callee.


lloyd:
@shane-tomlinson With this tweak, I'm r+.  Despite the nitpicky complaint about the comment being overly complex.

lloyd:
@shane-tomlinson MERGE IT!

PR #2929, opened 2013-01-11
upgrade to awsbox 0.3.5 (and node.js 0.8.17)

lloyd:
as discussed on identity-staff@ - this updates us to node.js 0.8.17 in dev (awsbox) and in production (`browserid.spec`)

### DIFF
--- a/lockdown.json
+++ b/lockdown.json
@@ -21,7 +21,7 @@
     "0.0.5": "971b8995078d83c80f2372f134c496e71b293a46"
   },
   "awsbox": {
-    "0.3.4": "6bbd87f39df830eca333806604499728cef2e1a3"
+    "0.3.5": "0972aa4b3acb0ac82ca1c586fe3ef0357b07c5f6"
   },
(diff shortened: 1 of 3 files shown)


lloyd:
@jrgm or @6a68 - what do you think?

jrgm:
Looks good to me, and I was about to merge it. I see that package.json remains at "node": ">= 0.8.11" which is fine for travis which is @0.8.15. Do we want to bump that to 0.8.17 when travis moves to that? (which can be a separate issue from this one).

lloyd:
@jrgm my thinking goes like this - package.json would constrain what version developers (or travis) can use.  We don't care as much as long as prod, stage, and dev are all on the same version 0.8.17 and development occurs on newish node.

seanmonstar:
And awsbox will select the same version?

lloyd:
yeah, the version of awsbox that this PR moves us to is locked to 0.8.17 which effectively moves us to 0.8.17 in dev.

PR #2930, opened 2013-01-11
Create a generic tooltip type within the module to make the overall thing easier to grok.

shane-tomlinson:
- Prevent tooltips from being closed prematurely if it replaces an old tooltip.
- Add a "done" callback to the dom-jquery animation functions.
- Use the done in the tooltips to work towards the goal of remove direct jQuery calls.

### DIFF
--- a/resources/static/common/js/lib/dom-jquery.js
+++ b/resources/static/common/js/lib/dom-jquery.js
@@ -391,7 +391,7 @@ BrowserID.DOM = ( function() {
          * @param {function} [done] called when animation completes
          */
         fadeOut: function( elementToFade, animationTime, done ) {
-          return jQuery( elementToFade ).fadeOut( animationTime );
+          return jQuery( elementToFade ).fadeOut( animationTime, done );
         },
 
         /**
@@ -400,7 +400,7 @@ BrowserID.DOM = ( function() {
          * @param {selector || element} elementToStop
          */
         stopAnimations: function( elementToStop ) {
-          return jQuery( elementToStop );
+          return jQuery( elementToStop ).stop();
         }
     };
 
--- a/resources/static/common/js/lib/dom-jquery.js
+++ b/resources/static/common/js/lib/dom-jquery.js
@@ -199,14 +199,7 @@ BrowserID.DOM = ( function() {
         * @return {boolean} true if the element has the attribute, false otw.
         */
         hasAttr: function( element, attrName ) {
-            var el = jQuery( element )[ 0 ],
-                val = null;
-
-            if (el) {
-              val = el.getAttribute( attrName );
-            }
-
-            return val !== null;
+            return typeof jQuery( element ).attr( attrName ) !== "undefined";
         },
 
         /**
@@ -362,9 +355,10 @@ BrowserID.DOM = ( function() {
          * @method slideDown
          * @param {selector || element} elementToSlide
          * @param {number} [animationTime]
+         * @param {function} [done] called when animation completes
          */
-        slideDown: function( elementToSlide, animationTime ) {
-          return jQuery( elementToSlide ).slideDown( animationTime );
+        slideDown: function( elementToSlide, animationTime, done ) {
+          return jQuery( elementToSlide ).slideDown( animationTime, done );
         },
 
         /**
@@ -372,9 +366,10 @@ BrowserID.DOM = ( function() {
          * @method slideUp
          * @param {selector || element} elementToSlide
          * @param {number} [animationTime]
+         * @param {function} [done] called when animation completes
          */
-        slideUp: function( elementToSlide, animationTime ) {
-          return jQuery( elementToSlide ).slideUp( animationTime );
+        slideUp: function( elementToSlide, animationTime, done ) {
+          return jQuery( elementToSlide ).slideUp( animationTime, done );
         },
 
         /**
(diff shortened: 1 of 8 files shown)


resources/static/common/js/lib/dom-jquery.js
```diff
@@ -392,9 +388,19 @@ BrowserID.DOM = ( function() {
          * @method fadeOut
          * @param {selector || element} elementToFade
          * @param {number} [animationTime]
+         * @param {function} [done] called when animation completes
          */
-        fadeOut: function( elementToFade, animationTime ) {
+        fadeOut: function( elementToFade, animationTime, done ) {
```
seanmonstar, line 393: `done` callback was added, but not used.

resources/static/common/js/lib/dom-jquery.js
```diff
@@ -392,9 +388,19 @@ BrowserID.DOM = ( function() {
...
           return jQuery( elementToFade ).fadeOut( animationTime );
+        },
+
+        /**
+         * Stop any animations that are occurring on an element.
+         * @method stopAnimations
+         * @param {selector || element} elementToStop
+         */
+        stopAnimations: function( elementToStop ) {
+          return jQuery( elementToStop );
```
seanmonstar: this simply returns the jQuery object. Is it missing a method call that would stop the animations?

resources/static/common/js/tooltip.js
```diff
@@ -47,27 +73,37 @@ BrowserID.Tooltip = (function() {
...
 
   function animateTooltip(complete) {
-    var displayTimeMS = calculateDisplayTime(tooltip.text());
-
-    bid.Tooltip.shown = true;
-    tooltip.fadeIn(ANIMATION_TIME, function() {
-      hideTimer = setTimeout(function() {
-        tooltip.fadeOut(ANIMATION_TIME, complete);
+    /*jshint validthis: true*/
+    var self = this,
```
seanmonstar: if this function is always called with `this` as an instance of `Tooltip`, why not make it a property of the prototype? Will help read it when we come back to this file later

seanmonstar:
Besides the comments in-line, it looks like this fixes the original concern, since now each `Tooltip` can manage it's own references. Though, I likely won't around to review this later, so I'm un-assigning myself.

jaredhirsch:
I'm happy to look at this if you need additional feedback, @shane-tomlinson.

shane-tomlinson:
Excellent suggestions @seanmonstar - updated with your notes as well as a few other changes. I am still not happy with the overall readability with concern to the BrowserID.Tooltip singleton vs the Tooltip type, but I am unsure how to fix this up without changing the overall tooltip interface, something I do not want to do right now. Should these be split into two files? 

I thought about making a Tooltip a PageModule, but then there are some funky dependency situations with validation.js. Ugh.

@6a68 - dude yeah, all input and review is appreciated!

resources/static/common/js/tooltip.js
```diff
@@ -11,20 +11,62 @@ BrowserID.Tooltip = (function() {
...
+
+      return displayTimeMS;
+    }
+  });
 
-  function anchorTooltip(target) {
+  function anchorTooltip(tooltip, target) {
     target = $(target);
     var targetOffset = target.offset();
     targetOffset.top -= (tooltip.outerHeight() + 5);
```
jaredhirsch: Nice to have: remove magic numbers here and the next line (frustratingly hidden by github's diff viewer).

resources/static/common/js/tooltip.js
```diff
@@ -46,33 +88,40 @@ BrowserID.Tooltip = (function() {
...
   // showTooltip(tooltipText, tooltipAnchor, [complete])
   function showTooltip(tooltipText, tooltipAnchor, complete) {
+    // Only one tooltip can be shown at a time, see issue #1615
+    removeAttachedTooltip();
+
+    onlyAttachedTooltip = new Tooltip();
+    var tooltipConfig = getTooltipConfig(tooltipText, tooltipAnchor, complete);
+    var displayTimeMS = onlyAttachedTooltip.start(tooltipConfig);
+
+    bid.Tooltip.shown = true;
```
jaredhirsch: Whoops! Meant to say that here:

It seems like shown is only used by unit tests? Maybe it could be set and unset inside the generic Tooltip's start and stop methods?

resources/static/common/js/tooltip.js
```diff
@@ -11,20 +11,62 @@ BrowserID.Tooltip = (function() {
...
+
+
+  // Tooltip generic object type.
+  var Tooltip = function() {};
+  _.extend(Tooltip.prototype, {
+    start: function(options) {
+      var self = this;
+
+      self.tooltipEl = renderer.append("body", "tooltip", {
+        contents: options.text
```
jaredhirsch, line 29: Does renderer manage escaping the text for us? If not, might want to escape the text here

resources/static/common/js/tooltip.js
jaredhirsch: Wait, maybe I'm misreading this. You're setting properties on bid.Tooltip, the singleton, not Tooltip the generic object. Right?

jaredhirsch:
R+. Looks good overall, few stylistic points.

Arguably might improve readability to move anchorTooltip and calculateDisplayTime inside the Tooltip object. I'll defer to you there.

Feel free to merge in after addressing (or ignoring) my points above.

resources/static/common/js/tooltip.js
shane-tomlinson, line 29: I just added a test for this - the answer is yes, it does escape the text.

Issue #2669, opened 2012-11-01
put TOS/PP above [sign up] button


skinny97214:
As shown >> http://cl.ly/image/2x09122p2m0i

PR #2933, opened 2013-01-14
Issue 2669 tos pp above button

shane-tomlinson:
This is a two part TOS/PP update.

First, the Persona TOS/PP text is _only_ supposed to be shown to unauthenticated users before they enter their email address. Because it is never shown anywhere else, I removed all other Persona TOS/PP references.

Second, the request in #2669 is to place the Persona TOS/PP text above the submit button. This is done in both the dialog and the main site /signin page.

### DIFF
--- a/resources/static/dialog/js/modules/required_email.js
+++ b/resources/static/dialog/js/modules/required_email.js
@@ -146,19 +146,13 @@ BrowserID.Modules.RequiredEmail = (function() {
             // 2) Authenticated user who does not control address.
             // 3) Unauthenticated user.
             if(info.type === "primary") primaryInfo = info;
-            
+
             if (info.type === "primary" && info.authed) {
               // this is a primary user who is authenticated with their IdP.
               // We know the user has control of this address, give them
               // a chance to hit "sign in" before we kick them off to the
               // primary flow account.
-
-              // Show the Persona TOS/PP to any primary user who is authed with
-              // their IdP but not with Persona.  Unfortunately, addressInfo
-              // does not tell us whether a primary address already has an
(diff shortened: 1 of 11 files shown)


shane-tomlinson:
@skinny97214 and @ozten - can you two review this?

A trial run can be done on http://tospp-above-submit.123done.org and https://tospp-above-submit.personatest.org

@ozten - I am asking you to review this because some of the removed code deals with the upgrade_to_primary code. My impression is that the Persona TOS/PP is only ever supposed to be displayed on the signin and authentication screens. I want to make sure that I am correct in this assumption and have not broken important functionality.

ozten:
Thanks for improving this. Looks good to me.

I'll wait for Skinny's ok, before merging.

shane-tomlinson:
Thanks @ozten!

```
stomlinson 17:51 skinny: the other thing I was working on today is putting the 
                 Persona TOS/PP text above the submit button and just 
                 wanted to verify - the Persona TOS/PP text is *only* supposed 
                 to be shown on 1) the authentication screen of the dialog if the 
                 user is not authenticated and 2) the /signin screen on the main 
                 site if the user is not authenticated.
stomlinson 17:51 skinny: the reason I ask is because there are remnants of old 
                 Persona TOS/PP lying around and I'd like to remove them if 
                 they are not needed.
skinny     17:51 stomlinson: that sounds right to me
```

@skinny97214 say's the assumption is correct - a verification here would cement it.

lloyd:
skinny, PING.  we need your final approval on this for merge monday.

skinny97214:
approved!

PR #2934, opened 2013-01-14
Issue 2710 wsapi ver lock

shane-tomlinson:
READY FOR FINAL REVIEW!

Large portions of @ozten's original code from f2adc7aad1c3321f398dd83e9349610ba8514897 as well as an updated version of @lloyd's PR #2284 to turn version() into an asynchronous function are used.

The asynchronous version() is used because generate_code_version was using the randomly generated SHA that is returned before git has finished fetching the code version from disk.

Instead of placing the SHA into the wsapi URL, it is placed into the BrowserID-Version HTTP header.


ozten:
Looks good. I think having this be asynchonous instead of returning a random value is a huge improvement.

lloyd:
let's merge #2284 first on monday, then rebase this.  preliminarily it looks good.

jaredhirsch:
@lloyd @ozten @shane-tomlinson bump! anything holding this up?

shane-tomlinson:
It is completely broken on ephemeral instances. New help. Description contains info.

Jared Hirsch notifications@github.com wrote:

@lloyd @ozten @shane-tomlinson bump! anything holding this up?

---

lloyd:
only thing holding it up from my side is lack of time to review.  I'm eager to land the monster PR then hope to contribute to this one.

Aside from general review, the thing that is most important to me is to land this initially with zero behavioral change, just report a statsd counter when frontend and backend code doesn't match.

callahad:
Oh man. I'm so sorry to do this. But as per [RFC 6648](http://tools.ietf.org/html/rfc6648), we should name the header `BrowserID-Version`, and _not_ use the `X-` prefix. This is especially important because we actually are, explicitly, aiming for future standardization.

lib/wsapi_client.js
```diff
@@ -82,31 +91,36 @@ exports.get = function(cfg, path, context, getArgs, cb) {
...
   }
 
-  var headers = { };
-  injectCookies(context, headers);
-
-  if (typeof getArgs === 'object')
-    path += "?" + querystring.stringify(getArgs);
-
-  var timingHandle = startRequest();
-  meth.get({
```
shane-tomlinson, line 92: Removing the tweaker from the picture.

lib/wsapi_client.js
```diff
@@ -82,31 +91,36 @@ exports.get = function(cfg, path, context, getArgs, cb) {
...
+      'X-BrowserID-Version': code_version
+    };
+    injectHeaders(context, headers);
+    injectCookies(context, headers);
+
+    if (typeof getArgs === 'object')
+      path += "?" + querystring.stringify(getArgs);
+
+    var timingHandle = startRequest();
+    meth.get({
```
shane-tomlinson, line 105: Only to put him back again.

lib/static/views.js
```diff
@@ -100,6 +100,17 @@ exports.setup = function(app) {
     next();
   });
 
+  app.get('/build/code_version.js', function(req, res) {
```
ozten: From a maintenance perspective, it is confusing that have both a route and a `scripts/generate_code_version.js` that generates a static file.

We should choose one or the other... no?

Perhaps this route could be a middleware and would call `scripts/generate_code_version.js` if there is no code_version.js on disc?

scripts/generate_code_version.js
```diff
@@ -0,0 +1,62 @@
+const
+fs = require('fs'),
+path = require('path'),
+mkdirp = require('mkdirp'),
+logger = require('../lib/logging').logger,
+version = require('../lib/version.js');
+
+module.exports = function(cb) {
```
ozten: Currently the signature of the callback is no parameters.

Change that to have `error` as the first parameter. There are conditions where this code can fail. We should bubble those failures up.

scripts/compress
```diff
@@ -4,75 +4,79 @@ var
...
-computecluster = require('compute-cluster');
+computecluster = require('compute-cluster'),
+version = require('../lib/version.js');
 
 const staticPath = path.join(__dirname, '..', 'resources', 'static');
 
 var langs = config.get('supported_languages');
 
-var all = resources.all(langs);
+generateCodeVersionJS(function() {
```
ozten: See note about adding err to callback function signature.

lib/static/views.js
shane-tomlinson: I did it that way initially and that felt yuck too. I'll have another look tomorrow. Thanks.

scripts/generate_code_version.js
```diff
@@ -0,0 +1,62 @@
...
+        if (err.code === "ENOENT") {
+          // file doesn't exist? no problem. No need to compare versions, just
+          // write it out.
+          return writeCodeVersion();
+        }
+
+        // Something more serious is happening, uh oh.
+        var msg = "Error reading " + codeVersionPath + ": " + String(err);
+        console.error(msg);
+        return logger.error(msg);
```
ozten: Use callback and pass `msg` back.

tests/post-limiting-test.js
```diff
@@ -22,29 +23,52 @@ suite.options.error = false;
 
 start_stop.addStartupBatches(suite);
 
+function addCodeVersionHeader(path, code_version, headers) {
+  // All wsapi calls require the BrowserID-Version header
+}
```
ozten: Unused function? If so remove.

lib/wsapi.js
```diff
@@ -138,6 +139,21 @@ function checkCSRF(req, resp, next) {
...
 
+function checkCodeVersion(req, resp, next) {
+  version(function(expectedCodeVersion) {
+    if (req.url.indexOf(WSAPI_PREFIX) === -1) return next();
+    var requestedCodeVersion = req.headers['browserid-version'];
+
+    if (requestedCodeVersion !== expectedCodeVersion) {
+      logger.warn("Code version mis-match: " + req.url + " expected: " + expectedCodeVersion + " received: " + requestedCodeVersion);
+
+      statsd.increment("wsapi_code_mismatch." + req.url);
```
ozten, line 150: Shouldn't we respond with a 400 or other "bad client request' HTTP status code?

tests/session-context-test.js
```diff
@@ -99,6 +99,20 @@ suite.addBatch({
   }
 });
 
+suite.addBatch({
+  "code version mismatch using session_context API": {
+    topic: wsapi.get('/wsapi/session_context', {}, {
+      headers: {
+        "Browserid-Version": "INVALID_SHA"
```
ozten, line 106: Why is session_context exempt from version lock?

ozten:
Great work!

tests/session-context-test.js
shane-tomlinson, line 106: It's not. The code that I snagged from you returned a 400 response if the version was incorrect. @lloyd asked that the behavior be changed so if a version mismatch occurs, the wsapi still responds normally but a statsd log entry is added.

This test makes it easy to test for changes to this behavior. When we want to disallow version mismatches, change 200 to whatever the new expected response code is.

tests/post-limiting-test.js
shane-tomlinson: Good call. That is left over from a refactor.

scripts/generate_code_version.js
shane-tomlinson: Good call.

scripts/compress
shane-tomlinson: On it.

lib/wsapi.js
shane-tomlinson, line 150: @lloyd asked for the behavior that you wrote to be changed until we have a better understanding of how frequently this is occurring. All he wants for now is to tick a statsd counter, but continue to process the wsapi normally.

lib/static/views.js
shane-tomlinson: I think I have a cleaner solution for this now.

ozten:
Awesome! r+

PR #2936, opened 2013-01-15
Issue 1232 password reset

### LINKED ISSUE #1232: password reset flow is confusing
Shane's addition:
We have seen a lot of confusion around the password reset flow. Many users do not understand why we are asking for their password when they click "forgot password".

Skinny's original:
Currently
1. get user's email
2. send verification email
3. set password

Desired
1. get user's email
2. create password
3. send verification email

shane-tomlinson:
This PR changes the location the user enters their password for a password reset. For the normal password reset flow, a user now enters their password post-verification.

fixes #1232

### DIFF
--- a/lib/db/mysql.js
+++ b/lib/db/mysql.js
@@ -356,13 +356,15 @@ exports.authForVerificationSecret = function(secret, cb) {
       var o = rows[0];
 
       // if there is a hashed passwd in the result, we're done
-      if (o.passwd) return cb(null, o.passwd, o.existing_user);
+      if (o.passwd) return cb(null, o.passwd, o.existing_user, true);
 
       // otherwise, let's get the passwd from the user record
       if (!o.existing_user) return cb("no password for user");
 
       exports.checkAuth(o.existing_user, function(err, hash) {
-        cb(err, hash, o.existing_user);
+        // fourth parameter indicates that there was no
+        // password in the stage table
+        cb(err, hash, o.existing_user, false);
       });
     });
 };
@@ -458,11 +460,11 @@ exports.completeConfirmEmail = function(secret, cb) {
   });
 };
 
-exports.completePasswordReset = function(secret, cb) {
+exports.completePasswordReset = function(secret, password, cb) {
   getAndDeleteRowForSecret(secret, function(err, o) {
     if (err) return cb(err);
 
-    if (o.new_acct || !o.passwd || !o.existing_user) {
+    if (o.new_acct || (!password && !o.passwd) || !o.existing_user) {
       return cb("this verification link is not for a password reset");
     }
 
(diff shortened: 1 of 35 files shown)


lloyd:
r- This change breaks the flow where a primary IdP shuts down:

https://saucelabs.com/tests/ca14d9f302ff4fe4896a1d4210b1e61d

that's in `automation-tests/tests/idp-transition/primary-shuts-down-single-email.js`

lib/db/mysql.js
```diff
@@ -536,9 +538,13 @@ exports.userOwnsEmail = function(uid, email, cb) {
...
 exports.stageEmail = function(existing_user, new_email, hash, cb) {
   secrets.generate(48, function(secret) {
+    // if we are staging a password reset in the new flow,
+    // we nullify any previously staged password
+    // this will clear out stale attempts from the old flow
+    var overwrite_password = (hash === null);
     // overwrite previously staged users
     client.query('INSERT INTO staged (secret, new_acct, existing_user, email, passwd) VALUES(?,FALSE,?,?,?) ' +
-                 'ON DUPLICATE KEY UPDATE secret=VALUES(secret), existing_user=VALUES(existing_user), new_acct=FALSE, ts=NOW(), passwd=VALUES(passwd)',
+                 'ON DUPLICATE KEY UPDATE secret=VALUES(secret), existing_user=VALUES(existing_user), new_acct=FALSE, ts=NOW()' + (overwrite_password ? ', passwd=VALUES(passwd)' : ''),
```
lloyd, line 547: this _looks_ right to me.  poses some challenges testing.  @jrgm should stage password resets on old software then verify on new software to test this properly.

lib/static/views.js
```diff
@@ -211,7 +211,7 @@ exports.setup = function(app) {
 
 
   app.get("/reset_password", function(req,res) {
-    renderCachableView(req, res, 'confirm.ejs', {title: _('Reset Password')});
+    renderCachableView(req, res, 'reset_password.ejs', {title: _('Reset Password')});
```
lloyd, line 214: nice catch.

lloyd:
general note: we should remove transition code in 1 train's time (a couple weeks).  we should capture somewhere all of the code (and records in the staged table) that should be purged in order to fully move to the new flow.

lloyd:
update.  I'm now reviewing this and PR 2961 in parallel.  I have one open issue which is upon verification it looks like the verification landing page is non-functional.  This can be reproduced by deploying  0a9b877 and then running primary-shuts-down-single-email.js.

If I can resolve this with another commit I'm comfortable merging and making this a focus of DDT next week.

lloyd:
in the case described by the previous comment, the frontend receives a 401 response to `complete_email_confirmation`.

lloyd:
So I believe the fact that transition from primary is now broken is due to the fact that we re-use the same landing page `/confirm`, which calls `complete_email_confirmation` instead of `complete_transition`.  trying a minimal patch.

lloyd:
this issue is now fixed in dedfc418f80567ecd04914c833990cb9805ae7ab.

Onto the next issue, but first I'm going to close down both of these PRs and open a new one.

PR #2944, opened 2013-01-16
Show the "This account cannot be verified" tooltip if the user enters a verification password < 8 chars.

### LINKED ISSUE #2943: Unhelpful error message displayed if user enters in a short password on email verification
STR

1) In browser 1, stage a new user
2) In browser 2, open verification link
3) Enter a password that is too short - in this case it was "pass"
4) Check out the very unhelpful error message.

error-message-post-verification-wrong-password

More desirable:

Give the user another chance to enter their email address.

shane-tomlinson:
fixes #2943

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -128,6 +128,13 @@ BrowserID.User = (function() {
   }
 
   function completeAddressVerification(completeFunc, token, password, onComplete, onFailure) {
+    if (password && password.length < bid.PASSWORD_MIN_LENGTH) {
+      // if the password is too short, call the onFailure callback with
+      // the "cannot authenticate" response. The backend blows up if trying to
+      // complete the verification with a password that is too short.
+      return complete(onFailure, { network: { status: 401 } });
+    }
+
     User.tokenInfo(token, function(info) {
       var invalidInfo = { valid: false };
       if (info) {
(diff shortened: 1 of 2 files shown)


resources/static/common/js/user.js
```diff
@@ -128,6 +128,13 @@ BrowserID.User = (function() {
   }
 
   function completeAddressVerification(completeFunc, token, password, onComplete, onFailure) {
+    if (password && password.length < bid.PASSWORD_MIN_LENGTH) {
```
jaredhirsch, line 131: It seems odd to me that this is the only edge case poorly handled by the server: only too-short passwords, not too-long passwords, or passwords with illegal chars. Is it possible that the better fix is to harden the server?

OTOH, if we want to be considerate users of our own server APIs, is it possible that the better fix is to validate all passwords site-wide using some common front-end util?

jaredhirsch:
R+ with caveats
1. The question of whether this is the right fix, see my comment inline. I defer to your judgment here on best vs expedient fix and all that.
2. I verified the funky error on dev, pushed this branch to an awsbox to verify that the interaction was fixed, but attempting to stage my mozilla email address threw an error I haven't seen before: http://imgur.com/STbzB. Up to you whether this is something or nothing; I strongly suspect it's nothing, since the changeset is minimal.

Otherwise looks good. All tests passing.

shane-tomlinson:
@6a68 - your points are valid, and you have found a problem. 

I am giving the location of the password validation more thought.

The problem with the error screen displayed is because the domain name contains an "_" which I believe is officially invalid though it is supported by DNS. Filed #2948, pile on there!

PR #2947, opened 2013-01-17
Issue 2891 email normalization

jaredhirsch:
NOT READY TO BE MERGED YET. JUST LOOKING FOR INPUT.

This is in a bit of a weird state currently.

Initially, I just threw the normalize call inside the address_info wsapi, added unit tests for two of the email normalization bugs to address_info tests, and got them to pass.

After getting that to pass, I looked for a more natural place to globally normalize emails, and decided to expand on what lib/validate is already doing. Since lib/validate just throws on validation error, killing the current request with an error, I thought I'd add a separate sanitize step after validation succeeds. This is barely implemented in the current code.

Is wsapi.js the right place to add sanitization? Is it too implicit, too magical? I think the alternative is to call normalizeEmail inside each wsapi that handles email--a step that could easily be forgotten. I'm cool either way, or open to other suggestions.

Assuming the normalization code is in the right place, I'd next like to return display emails by adding a display_email field to the DB, returning it where user info is returned by the server (IIRC list_emails and address_info wsapi output). If a user has no display_email, just let it be the same as the email field.

To insert un-normalized display_emails into the DB, I think I'd need to add a display_email or raw_email param to the signup endpoints, and add this to the lib/validate set of types. The idea is to validate, but not normalize, display_emails.

Once this is done, I'll ensure all the dialog and site code handles display emails appropriately and add tests.

After that, finally, I'll add unit tests for the bugs caused by assertion case mismatch to be certain we've fixed that bug.

This is taking a while, but it's a great intro bug, and I think I'm catching on--hope to get this fully done tomorrow or Friday. Am I on the right track? Comments welcome.

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -19,9 +19,32 @@ check = require('validator').check;
 
 var hostnameRegex = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$|^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$/;
 
+// This could easily be inlined in its current form, but our choice of 
+// normalization algorithm has been contentious to date. extracted to call
+// attention to it. see #2891 or:
+// https://groups.google.com/d/topic/mozilla.dev.identity/tvQdyLbfyH8/discussion
+var normalizeEmail = function(addr) { return addr.toLocaleLowerCase() }
+
 var types = {
-  email: function(x) {
-    check(x).isEmail();
+  email: function(address) {
+    // return normalized email or false
+    // pulled over from client-side resources/common/js/validation.js
+    if (typeof(address) !== "string")
+      return false;
+    // Original gotten from http://blog.gerv.net/2011/05/html5_email_address_regexp/
+    // changed the requirement that there must be a ldh-str because BrowserID
+    // is only used on internet based networks.
+    var parts = address.split("@");
+
+    return (/^[\w.!#$%&'*+\-\/=?\^`{|}~]+@[a-z\d\-]+(\.[a-z\d\-]+)+$/i).test(address)
+           // total address allwed to be 254 bytes long
+           && address.length <= 254
+           // local side only allowed to be 64 bytes long
+           && parts[0] && parts[0].length <= 64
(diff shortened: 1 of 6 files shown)


jaredhirsch:
Assigned to @shane-tomlinson but would love to get input from other folks as well

lib/validate.js
```diff
@@ -19,9 +19,34 @@ check = require('validator').check;
...
+// Original gotten from http://blog.gerv.net/2011/05/html5_email_address_regexp/
+// changed the requirement that there must be a ldh-str because BrowserID
+// is only used on internet based networks.
+var emailRegex = /^[\w.!#$%&'*+\-\/=?\^`{|}~]+@[a-z\d\-]+(\.[a-z\d\-]+)+$/i;
+
+// This could easily be inlined in its current form, but our choice of 
+// normalization algorithm has been contentious to date. extracted to call
+// attention to it. see #2891 or:
+// https://groups.google.com/d/topic/mozilla.dev.identity/tvQdyLbfyH8/discussion
+var normalizeEmail = function(addr) { return addr.toLocaleLowerCase() }
```
shane-tomlinson, line 31: We want the lower case of the user's locale, correct? If so, will this return the lower case of the server's locale?

lib/validate.js
```diff
@@ -19,9 +19,34 @@ check = require('validator').check;
...
+    if (parts[0] && parts[0].length > 64) throw 'email address local part is longer than 64 chars';
+    if (parts[1] && parts[1].length > 253) throw 'email address domain part is longer than 253 chars';
+}
+
 var types = {
   email: function(x) {
-    check(x).isEmail();
+    validateEmail(x);
+  },
+  display_email: function(x) {
```
shane-tomlinson, line 48: What is the difference between email and display_email?

shane-tomlinson:
@6a68, @lloyd - I am wondering if in wsapi.js:402, if it is possible to add another level of indirection after validate that is "normalize" that could have a table similar to the validation table.

For example:

```
wsapis[operation].validate(req, resp, function() {
  // this does normalization on the request parameters, updating normalized parameters
  // directly on the request.
  wsapis[operation].normalize(req, resp, function() {
    if (wsapis[operation].i18n( {
    ...
    }
  });
});
```

That way each individual wsapi handler would not have to remember to do normalization, addressing your very valid concern.

lib/validate.js
jaredhirsch, line 31: Dang. You're right

lib/validate.js
jaredhirsch, line 48: display_email is the un-normalized email whatever the user typed in, which we can store in the DB and display on the front-end. Working on expanding this now.

jaredhirsch:
I have a good idea of how to get this task actually done. will reopen when I have a final PR.

PR #2949, opened 2013-01-17
Issue 2943 auth password length range

shane-tomlinson:
@6a68 - mind reviewing? This is a whole series of updates prompted by your comments in mozilla/browserid#2944. I have moved the password length verification into the validation.js file and out of user.js.

I then cleaned up the tests to test both authentication password length ranges and new password length ranges, using a generic set of tests in all necessary locations.

### DIFF
--- a/resources/static/common/js/validation.js
+++ b/resources/static/common/js/validation.js
@@ -59,10 +59,34 @@ BrowserID.Validation = (function() {
     return valid;
   }
 
-  function passwordLength(password) {
-    var valid = password && (password.length >= bid.PASSWORD_MIN_LENGTH && password.length <= bid.PASSWORD_MAX_LENGTH);
+  function authenticationPassword(password) {
+    var valid = passwordExists(password)
+                && passwordLengthForAuthenticationPassword(password);
 
-    if(!valid) {
+    return valid;
+  }
+
+  function passwordLengthForAuthenticationPassword(password) {
+    var valid = password
+                && (password.length >= bid.PASSWORD_MIN_LENGTH
+                    && password.length <= bid.PASSWORD_MAX_LENGTH);
+
+    if (!valid) {
+      // If the authentication password is out of range, don't even bother to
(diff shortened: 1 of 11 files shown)


resources/static/common/js/validation.js
```diff
@@ -59,10 +59,34 @@ BrowserID.Validation = (function() {
...
+  function passwordLengthForAuthenticationPassword(password) {
+    var valid = password
+                && (password.length >= bid.PASSWORD_MIN_LENGTH
+                    && password.length <= bid.PASSWORD_MAX_LENGTH);
+
+    if (!valid) {
+      // If the authentication password is out of range, don't even bother to
+      // send the password to the backend. Fail quickly without wasting
+      // electrons.
+      tooltip.showTooltip("#cannot_authenticate");
```
jaredhirsch, line 78: Not necessarily something to deal with in this PR, but it might be cleaner to use pubsub to decouple displaying the error tooltip from this validation logic.

jaredhirsch:
R+, looks great. Tests passing. To be merged Monday.

Are there bugs in the server-side handling of password fields that should be addressed, or is the server returning the right kind of error each time?

jrgm:
See https://github.com/mozilla/browserid/blob/dev/tests/password-length-test.js for current tests of password (there are some other files with password tests as well). (Note: we make no restriction on the characters used in the password, and even accept high-bit).

resources/static/common/js/validation.js
shane-tomlinson, line 78: @6a68 - I dig having you and @seanmonstar review my code, it keeps me honest. The entire tooltip/validation system has been something that I have wanted to handle for a while. I'd love to get together with both of you and come up with a sane approach.

jaredhirsch, line 78: @shane-tomlinson - it's been super fun to get back into front-end code. Been a while. Happy to hash out idears whenever :-)

Issue #2951, opened 2013-01-18
remove no-longer-needed duplicate shutdown.handleTerminationSignals call


jrgm:
GH-2743 added a shutdown.handleTerminationSignals to shutdown toobusy. The no-op one I had registered is no longer needed (and  causes an error to be thrown at shutdown since `app.close` is called twice.

jrgm:
Closing, since PR-2952 was committed.

PR #2952, opened 2013-01-18
remove no-longer-needed call to shutdown.handleTerminationSignals

jrgm:
Fixes GH-2951

### DIFF
--- a/bin/static
+++ b/bin/static
@@ -123,14 +123,6 @@ app.use(function(req, res, next) {
 
 app.use(express.static(static_root));
 
-// Do this only when PERSONA_WITH_COVER is set. Without a graceful shutdown,
-// `cover` will not generate .coverage_data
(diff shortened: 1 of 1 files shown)


bin/static
```diff
@@ -123,14 +123,6 @@ app.use(function(req, res, next) {
...
 app.use(express.static(static_root));
 
-// Do this only when PERSONA_WITH_COVER is set. Without a graceful shutdown,
-// `cover` will not generate .coverage_data
-if (process.env.PERSONA_WITH_COVER) {
-  shutdown.handleTerminationSignals(app, function() {
-    // nothing particular to do here for static daemon, yet..
-  });
-}
-
```
ozten, line 133: Nice catch.

ozten:
Backend JSON test failed with

```
browserid (2092): error: failed to create primary user with assertion for test@example.domain: non-200 response: 503
```

PR #2954, opened 2013-01-18
Issue 2953 only one address

### LINKED ISSUE #2953: Show only the previously selected email address for a domain.

shane-tomlinson:
Only show the email address that was last selected for a particular site.

fixes #2953


shane-tomlinson:
Closing because of a bad rebase. Will re-open.

PR #2955, opened 2013-01-18
Issue 2953 only one address

### LINKED ISSUE #2953: Show only the previously selected email address for a domain.

shane-tomlinson:
Only show the last selected email address for a given site.

fixes #2953


shane-tomlinson:
@skinny97214 - would you mind doing a review of the functionality on http://only-one-address.123done.org? @lloyd and I would like to get your r+ on functionality before merge.

lloyd:
Screenshot_1_24_13_9_28_AM

lloyd:
Screenshot_1_24_13_9_29_AM-3

skinny97214:
SHIP IT! Awesome work, Shane et al.

lloyd:
reviewing and running automated tests...

a copy question: "use another email address" vs. "use a different email"

skinny97214:
Leave it as is. It's not worth the complexity even if I was totally convinced it's better.

lloyd:
the complexity is nil, fwiw.

Slightly uncomfortable by this language.  "use another email address" suggested to me upon first read "log in with two email addresses".  "another" can mean "additional" while "different" doesn't have this ambiguity....   Aaaaand it feels more human "use a different email".  See?  it's hip.

Regardless of whether I'm able to sway you, I'll do my damndest to review and merge this train.

skinny97214:
I am swayed. Mostly bc the button does something different! What do you think of "switch to another address" ?

callahad:
What about "switch to a different address?"

I wouldn't want to be so close to the "add another email address" button that people think "I already clicked that once, and now I'm stuck on this list."

"Hi, I'm from the Internet, and I can over think a plate of beans."

lloyd:
So I got on the phone with @skinny97214 and we talked through this PR.

Ultimately, seeing it viscerally, she decided that we probably don't want this patch right at this moment.  Her reasoning (which I find myself agreeing with) is:
1. Given that we have users actively complaining that "persona doesn't remember the last email I used", we're actually raising the prominence of that issue.  If we solve that first, maybe this change becomes more appealing
2. we're changing the meaning of a button that users already know about - it used to mean "add another email to my persona account", now it means "use a different email in my persona account", and then is followed by a button that means the original thing.

Sorry to be the bearer of bad news, because the patch is already done...  But on the upside actually having it written was what gave skinny the context to be decisive about it.

:/  thoughts?

jaredhirsch:
broader UX process discussion, we can take this to the public list if it
goes around too long:

Would a super lo-fi mock (10-15 min of effort) have avoided this? Examples:
- cell phone pic of napkin sketch
- screenshot of modded DOM with MS Paint-quality arrows calling out
  relevant changes
- illustrator/balsamiq, for those who can bang it out with such tools

Maybe a rule of thumb is, "give skinny the skinny on UX-affecting changes
**before hacking**, with visual aids if possible?"

Or is that more than what's needed/going to slow dev down too much?

shane-tomlinson:
@6a68 - the original hack took 30 mins and a working hack on an ephemeral instance gives a much more visceral feel than any of the other methods.

lloyd:
@skinny97214 weigh in here.  Do you think there's any super lightweight way we could have staved the wasted work?  I personally could have done a better job relaying your verdict.

Also, back to the topic at hand.  Once we fix issue #1500 do enough of your objections fall away that we can merge?

skinny97214:
Attaching screenshots to the bug instead of the pull request should help quite a bit.

I do thing Shane's original goal of reducing clutter is a great one. I don't think we can ship as is. Need to tweak button text and run a few users through to make sure there aren't any other nasty surprises.

c

jaredhirsch:
Closing per skinny's comments. Reopen if I'm too trigger-happy :-)

Issue #2903, opened 2013-01-03
Send  a longer STS max-age so that we get STS pinning


evilpie:
persona.org was originally included in the STS pinnig list included with Firefox 17. Due to some issues all sites having a max-age small than 18 weeks however were removed from that list. This means persona.org at the moment is not on the preloaded STS list. (https://bugzilla.mozilla.org/show_bug.cgi?id=786417)

Please change the max-age of the Strict-Transport-Security head on login.persona.org and persona.org. We also probably need to coordinate with Firefox so that we update the list.

jrgm:
Also noted in GH-2816.

jrgm:
Gifted to @fmarier to make this change (and closed GH-2816)

jrgm:
Also, the request above is for persona.org and *.persona.org. But should also cover legacy browserid.org, I suppose.

evilpie:
Any update on this?

lloyd:
merged. should be in production on feb 13th.  http://personatra.in/#2013.02.01

evilpie:
Great, thank you.

evilpie:
The max-age of login.persona.org is still 2592000. And we are close to shipping!

jrgm:
Ugh. So, this also needs an update in bin/static and bin/router. (bin/browserid is serving 18 weeks when queried directly, but it's moot, since router stands in front of it).

jrgm:
request in stage for any GET path returns 'Strict-Transport-Security: max-age=10886400; includeSubdomains'. Should go out to prod tomorrow.

PR #2959, opened 2013-01-22
Increase Strict-Transport-Security max-age

evilpie:
This is required to get onto the Firefox HSTS preload list. Should fix GH-2903.


shane-tomlinson:
@evilpie - thanks for the PR!

@fmarier - my tests all pass, the new time is exactly 18 weeks. Is that sufficient for platform to re-include Persona? Do we need additional work for this? Wanna hit the green button?

lloyd:
r+!  https://wiki.mozilla.org/Privacy/Features/HSTS_Preload_List - "Have a max-age larger than or equal to 18 weeks (10886400)"

Issue #2965, opened 2013-01-24
support app:// prototocol in validator for FirefoxOS


jedp:
Apparently gaia apps give their origin as app://foo instead of http://foo.  This does not pass the persona origin validator.

As a result, users will receive a red "There has been an error" message when trying to log in with unverified email and/or forced issuer.  The error message will say something about an invalid origin.

I had never noticed this, because building gaia with DEBUG=1 transforms all the app:// origins to http://, and I always build with DEBUG=1 so I can do testing.  Such is life.

jds2501:
I should note that fixing this is required to use persona in a packaged app. So this is definitely a bad bug.

PR #2966, opened 2013-01-24
Validator should accept FirefoxOS app:// origins

jedp:
Issue #2965

Tested using the navigator.mozId part of the b2g UI tests app to check regular assertion, allow unverified, and allow unverified + force issuer.

@ozten can I get your blessing on this one?  Did I miss anything?

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -48,7 +48,7 @@ var types = {
   origin: function(x) {
     /* origin regex
     /^                          // beginning
-    https?:\/\/                 // starts with http:// or https://
+    (?:https?|app):\/\/         // starts with http://, https://, or app:// (FirefoxOS)
     (?=.{1,254}(?::|$))         // hostname must be within 1-254 characters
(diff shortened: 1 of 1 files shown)


lloyd:
given this targets the b2g branch I'm r+.  Are we convinced that this is the only change required to allow an application that uses persona to run thinking it's origin is `app://example.com`?

jedp:
Thanks, @lloyd.  As far as I can tell, this is all that's necessary.  I had been building gaia (the interface layer of b2g) with DEBUG=1, which turns out to have the effect of translating all the app:// origins to http:// origins, so I never even knew app:// existed until yesterday.  With this change, I've tested getting regular assertions, unverified assertion, force-issuer, and force-issuer with unverified.  So I'm pretty confident the only issue was the validator not liking the app:// origin.

ozten:
Thanks for the quick turn around!

There has been some community interest in generalized URN. We should revisit that design when landing this on dev.

PR #2967, opened 2013-01-24
downgrade log message for when invalid url is requested from .error() to .info()

lloyd:
we reserve `.error()` for fatal or unexpected conditions.  Because anyone on the internet can request a bogus url, it's not really a big deal, just a silly robot (or human).

Just as HTTP `500` errors are a sign of something wrong with our software, so are errors in logs.


jrgm:
r+

PR #2968, opened 2013-01-24
fix a broken selenium test

lloyd:
The selenium test we have set up that is supposed to run our frontend tests on remote machines was detecting which environment to target in an odd way.

We set up automated-tests/lib/urls.js so that it can magically target different environments in (ephemeral, beta, dev, etc) in a symmetric way.

this patch makes `frontend-qunit-test.js` work like everyone else.

### DIFF
--- a/automation-tests/tests/frontend-qunit-test.js
+++ b/automation-tests/tests/frontend-qunit-test.js
@@ -8,13 +8,13 @@ const
 assert = require('../lib/asserts.js'),
 utils = require('../lib/utils.js'),
 runner = require('../lib/runner.js'),
+persona_urls = require('../lib/urls.js'),
 path = require('path'),
 testSetup = require('../lib/test-setup.js');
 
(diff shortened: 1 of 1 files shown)


jaredhirsch:
R+. Travis failed for random reason unrelated to the selenium test changes here.

jrgm:
Yeah, mine. It was an odd way but intentional because '/test' is 404 on stage or prod. But it was sucky because of the magic environment variable. 

However, with the current change this test is guaranteed to fail against stage and prod environments.

If there's a better way to skip tests based on which test environment to use, let me know. This is my WIP - https://github.com/jrgm/browserid/compare/skip-frontend-selenium-on-prod-stage

jaredhirsch:
A cleaner approach might be to split out the tests-to-ignore file into 'dev', 'staging', and 'prod', and switch on the env passed in from the command line?

lloyd:
Sorry for regressing this.

-- lloyd (thumb-typing)

Issue #2937, opened 2013-01-15
transitionNoPassword sends the reset password email instead of a more appropriate email


shane-tomlinson:
This will cause user confusion when they get an email that says:

"Forgot your password for Persona? It happens to the best of us."

shane-tomlinson:
Hey @skinny97214 - When a user who has an account with only primary email addresses and one of their addresses converts from a primary to a secondary, currently the password reset email is sent out.  While it is an edge case that we hope does not happen often, it is an area that will cause confusion when it does. Whatcha think about sending a new email?

lloyd:
+1 - this will be confusing.  This case occurs only when a primary transitions to a secondary, which is thought to be rare.  We did not bother optimizing the email text for this case as part of the initial merge because the patch was HUGE, we thought time was tight, and this seemed like something we could safely get to later.

skinny97214:
We do need a new email with custom copy. @callahad can you help draft the copy?

callahad:
@skinny97214 / @shane-tomlinson Do we need to expose the user to the fact that their email provider isn't supporting Persona? If we do, how about:

> It looks like your email provider is having trouble with their Persona support.
> 
> To continue, you need to re-confirm this email address and set up a new, Persona-specific password.
> 
> Click the link below to get started: 
> {link}
> 
> Thank you,
> The Persona Team

I'm intentionally ignoring the fact that a domain might intentionally disable Persona, and instead referring to everything as "trouble with their Persona support." Because I can't think of a better way to put it, and because the verbiage above is more likely to be the common case. And it will hopefully inspire user inquiries to the provider. :)

skinny97214:
This is a weird enough thing that it seems like there needs to be a support article with more detail that we could link to. 

Other than that it looks great. I tweaked the link copy a teeny bit.

```
It looks like your email provider is having trouble with their Persona support.

To continue, you need to re-confirm this email address and set up a new, Persona-specific password.

{Click here to set your new password}

Thank you,
The Persona Team
```

zaach:
@callahad @skinny97214 Any ideas for the subject of the email?

callahad:
Current email subject lines:

<dl>
<dt>confirm</dt>
<dd>Confirm email address for Persona</dd>
<dt>new</dt>
<dd>Confirm email address for Persona</dd>
<dt>reset</dt>
<dd>Reset Persona password</dd>
</dl>

For this case, re-using "Confirm email address for Persona" seems reasonable.

[...]

PR #2969, opened 2013-01-25
Improve language in email when an IdP drops support AND improve password reset flow

lloyd:
I'm combining PR #2936 and PR #2961 

These address issue #1232 and issue #2937 

It sucks to conflate these two things, but because the first PR introduces a regression (makes it so that a user cannot use an email address if the provider disables), I don't think it can be merged independantly.

### DIFF
--- a/automation-tests/pages/css.js
+++ b/automation-tests/pages/css.js
@@ -49,11 +49,12 @@ module.exports = {
     forgotPassword: 'a#forgotPassword',
     choosePassword: 'div#set_password input#password',
     verifyPassword: 'input#vpassword',
+    postVerificationPassword: 'input#password',
     // the button you click on after typing and re-typing your password
     createUserButton: 'button#verify_user',
-    resetPasswordButton: 'button#password_reset',
     returningUserButton: 'button.returning',
     verifyWithPrimaryButton: 'button#verifyWithPrimary',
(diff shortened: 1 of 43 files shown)


lloyd:
@shane-tomlinson last known bug in this pile of patches happens when trying to run selenium tests which excercise the path when an IdP disables support.  

Concrete issue is that the dialog does not progress to waiting at the moment a user chooses their password.  Have a look at this video: https://saucelabs.com/tests/5138cdb6088f4df894914d0fe83d9058

look carefully at second 42 - the "done" button displays a spinner, then becomes clickable again and the dialog does not progress to a waiting page.  At second 50 you can see the dialog pop back to the foreground.

jaredhirsch:
Grabbing this PR unless you have someone specific in mind to review it, @lloyd.

lloyd:
I've already reviewed, as far as I'm concerned a final rerun of selenium tests and its ready for merge.

-- lloyd (thumb-typing)

lloyd:
r- @shane-tomlinson pointed out today that we cannot rollback with this patch.

Strategy is to break out the UI changes that would cause the new password flow to take affect and only commit the new landing page and server apis.  This would allow us to merge the actual UI changes in a subsequent PR and roll them out with a rollback path.

shane-tomlinson:
Closing in favor of #2995

PR #2974, opened 2013-01-28
fix frontend qunit selenium tests - they cannot run against stage or prod

lloyd:
Because frontend tests are disabled on prod and stage, we cannot actually run them in those environments.  It is, however, useful to run frontend tests via selenium in ephemeral and dev environments as it allows us to excercise the tests on all browsers.

Previously, @jrgm had it so that tests would be run against dev when targeting stage or prod.  I broke that, this fixes it by just disabling them when they cannot be run anyway.

### DIFF
--- a/automation-tests/config/tests-to-ignore.js
+++ b/automation-tests/config/tests-to-ignore.js
@@ -3,8 +3,19 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // These are tests to ignore
+// XXX extract duplication if this file gets significantly longer
 
-exports.tests_to_ignore = [
+exports.dev = [
+  "public-terminals.js",
+  "remove-email.js"
+];
+exports.stage = [
+  "frontend-qunit-test.js",
+  "public-terminals.js",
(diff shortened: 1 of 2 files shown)


lloyd:
review with whitespace disabled: https://github.com/mozilla/browserid/commit/53d5272718035aa59648b071b9f0d61f2b72240d?w=1

jaredhirsch:
I'm not crazy about putting knowledge of ignorability right in the test. What about doing something like https://gist.github.com/4660172 instead?

lloyd:
r- on lloyd's approach, r+ on _6a68's.

jaredhirsch:
Discussed alternative soln on IRC:

lloyd       _6a68: I like your solution better.         3:17
_6a68       lloyd: w00t         3:17
lloyd       _6a68: force push over my branch?
...
_6a68       lloyd: ok, will force-push in approx 5 min          3:19
lloyd       _6a68: rad

lloyd:
testing...

lloyd:
r+

PR #2975, opened 2013-01-28
disable request throttling when running under travis, as travis boxes ARE toobusy

lloyd:
An unexpected side affect of merging toobusy was that under travis our tests started occasionally failing.  I believe this is because travis boxes are running at high load.

Increasing the delta between test and prod environments is not great, but I don't see a better solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ notifications:
     - jrgm@mozilla.com
 
 env:
- - WHAT_TESTS=front MYSQL_USER=root
- - WHAT_TESTS=back_mysql MYSQL_USER=root
- - WHAT_TESTS=back
+ - WHAT_TESTS=front MYSQL_USER=root DISABLE_REQUEST_THROTTLING=true
+ - WHAT_TESTS=back_mysql MYSQL_USER=root DISABLE_REQUEST_THROTTLING=true
+ - WHAT_TESTS=back DISABLE_REQUEST_THROTTLING=true
 
 mysql:
   adapter: mysql2
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -187,7 +187,8 @@ var conf = module.exports = convict({
   },
   disable_request_throttling: {
     doc: "Disable 'server too busy' responses when servers are too busy.",
-    format: 'boolean = false'
+    format: 'boolean = false',
+    env: 'DISABLE_REQUEST_THROTTLING'
   },
   certificate_validity_ms: {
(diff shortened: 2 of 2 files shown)


jrgm:
So re: the delta. Yeah not optimal, although at least the module will be loaded before being disabled. (I guess another approach would be to set the maximum_event_loop_lag to some much higher threshold, which would mean that the timer has to fire (if the daemons stay up for >0.5 seconds, which is probably ~all of the test runs).

lloyd:
@jrgm do you feel strongly enough about this alternate approach that I should change it?

lloyd:
fwiw, quick look at the last 3 travis failures (over the last 50 builds) suggest that this is a real problem, causing 4% of our tests to fail erroneously.

lloyd:
https://gist.github.com/4660152

jaredhirsch:
I must be missing something here. I see a config variable being set, but I don't see it being read/used anywhere. Does this live in the travis config someplace?

lib/configuration.js
```diff
@@ -187,7 +187,8 @@ var conf = module.exports = convict({
   },
   disable_request_throttling: {
     doc: "Disable 'server too busy' responses when servers are too busy.",
-    format: 'boolean = false'
+    format: 'boolean = false',
+    env: 'DISABLE_REQUEST_THROTTLING'
```
lloyd, line 191: this line specifies that configuration for disable_request_throttling can be pulled out of the environment

.travis.yml
```diff
@@ -22,9 +22,9 @@ notifications:
     - jrgm@mozilla.com
 
 env:
- - WHAT_TESTS=front MYSQL_USER=root
- - WHAT_TESTS=back_mysql MYSQL_USER=root
- - WHAT_TESTS=back
+ - WHAT_TESTS=front MYSQL_USER=root DISABLE_REQUEST_THROTTLING=true
+ - WHAT_TESTS=back_mysql MYSQL_USER=root DISABLE_REQUEST_THROTTLING=true
+ - WHAT_TESTS=back DISABLE_REQUEST_THROTTLING=true
```
lloyd, line 27: these three lines are updated to set the environment variable that configuration.js will defer to for throttling config

jaredhirsch:
bah, never mind. grepped for the wrong thing.

jaredhirsch:
looks great. r+

jrgm:
Given real failures in test at the moment, I think we should just take this as is. I don't feel that strongly about the alternate.

PR #2977, opened 2013-01-29
Fix selenium tests in stage

lloyd:
this is a cherry pick of four fixes to selenium tests.  Merging this will make the `train-2013.01.18` branch consistent, with selenium tests that are expected to pass.

no stage update is necessary, as these are only automated test changes.

### DIFF
--- a/tests/lib/secondary.js
+++ b/tests/lib/secondary.js
@@ -1,5 +1,4 @@
-var start_stop = require('./start-stop.js'),
-wsapi = require('./wsapi.js'),
+var wsapi = require('./wsapi.js'),
 wcli = require('../../lib/wsapi_client.js'),
 jwcrypto = require("jwcrypto");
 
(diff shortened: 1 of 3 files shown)


tests/lib/secondary.js
```diff
@@ -9,7 +7,7 @@ exports.create = function(opts, cb) {
   opts.email = opts.email || 'testuser@example.com';
   opts.password = opts.password || opts.pass || 'password';
   opts.site = opts.site || 'http://rp.example.com';
-  opts.fetchVerificationLinkCallback = opts.fetchVerificationLinkCallback || start_stop.waitForToken;
+  opts.fetchVerificationLinkCallback = opts.fetchVerificationLinkCallback || require('./start-stop.js').waitForToken;
```
jaredhirsch, line 10: :-S

This require being nestled lower in the file was surprising, makes me a little squeamish. 

Why do we need to do this?

Is there any other way we can conditionally require at the top of the file? s'ok if not.

lloyd, line 10: as always nice catch.  I'm squeamish too.

I'm hacking around the fact that we have `automation-tests/package.json` which is a subset of the node modules we install at the top level `package.json`.

The motivation for this was to npm install a more minimal set of packages for automation testing, and to be able to be a little more fancy free with what we depend on in automation tests - both good goals.

When I went and started reusing backend testing libraries (secondary.js) which were written more with the assumption that ALL packages were available, I inadvertently made it so automation tests failed, because they indirectly relied on stuff not in `automation-tests/package.json`

what you see here is a _minimal_ patch to undo the damage I did.  But it is a minimal hack - we should figure out a better long term solution (I hope it involves a bit of restructuring of tests) and figure out how testing libraries can use production libraries to share code, but we still can not install the world in order to run tests....

jaredhirsch, line 10: Ah, ok. That makes sense.

R+, tests are passing fine for me on sauce. Merging in.

PR #2978, opened 2013-01-29
upgrade toobusy - we run on win32 again

lloyd:
the delta between toobusy is not high (only toobusy.cc is really relevant): https://github.com/lloyd/node-toobusy/compare/v0.1.1...v0.2.2?w=1

@seanmonstar - verify you can install with this?


seanmonstar:
It works \o/

jrgm:
r+, also compiles for me on win32.

lloyd:
r+ and no merge, I merge.

PR #2984, opened 2013-02-01
mark password reseted emails as verified

### LINKED ISSUE #2925: [b2g] email not verified on password reset for unverified email
STR
1. Sign in with marketplace with a previously unused email address to be an unverified user.
2. Log out of marketplace, and begin sign in again. 
3. Click this is not me and re-enter the same email address.
4. When prompted for the password, click on 'Forgot password?'
5. Enter password times two, and load the reset link in the browser.
6. Enter the password again and complete the reset.

At the end of this sequence, there will no longer be a row in staged for this user's email address. The `verified` column of `email` is still set to `0`.

seanmonstar:
if an email is not verified, and the user goes through the
reset password flow, then the email really should be verified.

fixes #2925

### DIFF
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -132,21 +132,6 @@ exports.emailIsVerified = function(email, cb) {
   });
 };
 
-exports.emailInfo = function(email, cb) {
-  sync();
-  var m = jsel.match(".emails .?", [ email ], db.users);
-  if (!m || !m.length) return process.nextTick(function() {
-    cb(null, null);
-  });
-  var response = {
-    type: m[0].type || "secondary"
-  };
-  // if the email is known, let's determine if the account has a password
-  m = jsel.match(":root > object:has(.emails > .?) > .password", [email], db.users);
-  response.hasPassword = m.length ? (typeof m[0] === 'string') : false;
-  process.nextTick(function() { cb(null, response); });
-};
-
 exports.emailType = function(email, cb) {
   sync();
   var m = jsel.match(".emails ." + ESC(email), db.users);
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -436,6 +436,8 @@ exports.completePasswordReset = function(secret, cb) {
         Object.keys(emails).forEach(function(email) {
           if (email !== o.email && emails[email].type === 'secondary') {
             emails[email].verified = false;
+          } else if (email === o.email) {
+            emails[email].verified = true;
           }
         });
         flush();
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -223,10 +223,19 @@ function addEmailToAccount(userID, email, type, cb) {
   });
 }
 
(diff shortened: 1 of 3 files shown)


jaredhirsch:
I haven't touched B2G code before, but I'm guessing this can be verified in the usual way: spin up an awsbox with this code on board, verify the bug is fixed?

How can I repro the existing bug, to verify it's actually broken in prod?

seanmonstar:
STR are in the associated issue: #2925 

You could check the database like it says if you really wanted, but they way I tested locally is that the first time, the local example RP receives an assertion with `unverified-email` field, and then after resetting your password (and hence confirming your email), the second assertion should have the standard `email` field.

shane-tomlinson:
@seanmonstar - would you mind adding some tests to this so we can be sure we don't break your behavior later?

seanmonstar:
@shane-tomlinson Ah, I got caught!

So I just spent some time writing a test for it, ran into email throttling as I'm asking for a password reset just after staging the user. Is there a way to tell the system to **not** throttling me in this instance?

lloyd:
yeah!  you can fiddle this: https://github.com/mozilla/browserid/blob/dev/lib/configuration.js#L217

are you writing a backend test?  if so, it's just `config.set()`

lloyd:
see `double-stage-updates-password.js` for an example of disabling throttling

seanmonstar:
done! tests added!

shane-tomlinson:
@seanmonstar - I am manually testing this with a fresh JSON db. Following the steps in #2925, once I complete all the steps the email is marked as "verified" under "users", but the email is still listed in "staged" as well.

It seems like once the user verifies the password reset, the staged email should no longer be in the staging table.

```
{
  "users": [
    {
      "id": 1,
      "lastPasswordReset": 1360749434,
      "password": "$2a$12$Uk.sFi3rJaOAEIkxvVWN/OEqcjw95Bn1/qtcoEegQh3TqsQx.HZHe",
      "emails": {
        "testuser@mailinator.com": {
          "type": "secondary",
          "verified": true
        }
      }
    }
  ],
  "stagedEmails": {},
  "staged": {
    "mdFUJ7ESJuuVjMSIRFVAd9VZkH1QAL47rbF0L7YXVJFFYX45": {
      "type": "add_email",
      "existing_user": 1,
      "email": "testuser@mailinator.com",
      "passwd": "$2a$12$VJyfbwYoJc9Y9h5lF8isYu80qymuQyYQwhoOCXr0OHeUXOTWIFqnK",
      "when": 1360749081632
    }
  },
  "idp": {}
}
```

seanmonstar:
It looks like there was nothing to clean out the `staged` table in the JSON db, like there is `getAndDeleteRowForSecret` in the MySQL db. I'll add one in.

seanmonstar:
@shane-tomlinson how's that?

PR #2987, opened 2013-02-04
Broken version js

lloyd:
an error was introduced in version.js this train when it was massaged to be async all the time. this fixes it

### DIFF
--- a/scripts/browserid.spec
+++ b/scripts/browserid.spec
@@ -2,7 +2,7 @@
 
 Name:          browserid-server
 Version:       0.2013.02.01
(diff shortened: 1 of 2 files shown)


scripts/browserid.spec
```diff
@@ -2,7 +2,7 @@
 
 Name:          browserid-server
 Version:       0.2013.02.01
-Release:       1%{?dist}_%{svnrev}
```
jaredhirsch, line 5: Do we want to bump this or Version?

lloyd, line 5: this is the right thing to bump...  Version == train

Issue #2462, opened 2012-09-11
Use a custom dialog for "cancelling" an account (language paradox) 


jrgm:
Okay, this is a maybe a nitpick, and experienced users are probably not terribly confused. But anyways the meaning of 'cancel' is overloaded in an English localized browser... "To not cancel your account, click cancel" ;-)

![img](https://img.skitch.com/20120911-t4w7k374fnceku694kdim42ugs.png)

I have no idea how this plays out in other languages (L10N), so probably just getting away from window.confirm is the way to go (where a custom html dialog just presents 'Yes' and 'No').

shane-tomlinson:
@skinny97214, @jrgm - This is a good catch.

Could a language change work here instead of using an entirely new dialog?

jrgm:
A language change would work in English, but some other word translated to another language might have the same way of being overloaded. But yeah, a whole custom dialog to present Yes/No is a pain.

skinny97214:
This is a case where we want a modal with a confirmation. Agree the wording is bad. Would "delete" be better than "cancel?"

pnkfelix:
+1 to skinny97214's suggest of using "delete' rather than "cancel".  While jrgm's point is valid that there is a risk of the same overloading happening again in translation, there is still the very strong argument one should not delay in putting in a fix for a very real current problem solely because of a hypothetical instance of the _same problem_.

PR #2990, opened 2013-02-05
change language in account cancel dialog to remove ambiguity

lloyd:
issue #2462 - change a word.

Click the green button, close a bug.


ozten:
Nice!

Issue #2985, opened 2013-02-04
Yahoo BigTent loops after authentication


ozten:
Steps to Reproduce

1) Go to http://beta.123done.org/
2) Open HttpFox or another tool to see HTTP requests
3) Complete log in with a yahoo.com email address

Expected:
Dialog closes and your logged in

Actual
Dialog hangs on last screen

Note HTTP requests:
A loop is present. `auth_with_assertion`, `list_emails`, `session_context`, and then back to `auth_with_assertion`.

This does not reproduce in [dev](http://dev.123done.org) nor in my local dev instance.

ozten:
Certificate issuer looks good, matching `yahoo.login.anosrep.org`.

BigTent portion of the flow appears to complete successfully.

After attempted login, localStorage has 

```
({'eozten@yahoo.com':{created:"2013-02-04T17:01:02.856Z", updated:"2013-02-04T17:01:06.773Z"}})
```

It is missing several fields! With a successful auth, localStorage would also have `cert`, `priv` and `pub`.

ozten:
The loop is after provisioning, we loop over have_email, auth_with_assertion, and  list_email again and again.

```
https://login.anosrep.org/wsapi/session_context
authenticated: false
csrf_token: "4AjqL0nbvAoONr8T4hfjyQ=="
data_sample_rate: 0
domain_key_creation_time: 1322071714847
has_password: false
random_seed: "4X456JnwSX/jdJ4tNgi/fuJ6i4ZDw6BXNBgPvNqw04w="
server_time: 1359680197687
{"csrf_token":"4AjqL0nbvAoONr8T4hfjyQ==","server_time":1359680197687,"authenticated":false,"has_password":false,"domain_key_creation_time":1322071714847,"random_seed":"4X456JnwSX/jdJ4tNgi/fuJ6i4ZDw6BXNBgPvNqw04w=","data_sample_rate":0}

https://login.anosrep.org/wsapi/have_email?email=eozten%40yahoo.com
email_known: true

https://login.anosrep.org/wsapi/address_info?email=eozten%40yahoo.com
{"auth":"https://yahoo.login.anosrep.org/authentication","prov":"https://yahoo.login.anosrep.org/provision","type":"primary","issuer":"yahoo.login.anosrep.org","state":"known"

https://yahoo.login.anosrep.org/provision
[...]

PR #2991, opened 2013-02-05
We should log expiration issues. A root cause of Issue #2985

ozten:
We don't give the IdP developer any feedback when we see expired or certs without an iat field.

See Issue #2985 for details.

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -45,7 +45,13 @@ BrowserID.User = (function() {
 
           // or if it was issued before the last time the domain key
           // was updated, it's invalid
-          if (!cert.payload.iat || cert.payload.iat < creationTime) {
+          if (!cert.payload.iat) {
+            helpers.log('Data Format ERROR: expected cert to have iat ' +
+              'property, but found none, marking expired');
+            return true;
+          } else if (cert.payload.iat < creationTime) {
+            helpers.log('Certificate issued ' + cert.payload.iat +
+              ' is before creation time ' + creationTime + ', marking expired');
             return true;
           }
 


resources/static/common/js/user.js
```diff
@@ -45,7 +45,13 @@ BrowserID.User = (function() {
...
           // or if it was issued before the last time the domain key
           // was updated, it's invalid
-          if (!cert.payload.iat || cert.payload.iat < creationTime) {
+          if (!cert.payload.iat) {
+            helpers.log('Data Format ERROR: expected cert to have iat ' +
+              'property, but found none, marking expired');
+            return true;
+          } else if (cert.payload.iat < creationTime) {
+            helpers.log('Certificate issued ' + cert.payload.iat +
+              'before creation time' + creationTime + ', marking expired');
```
lloyd: spacing is wonky (numbers smashed up against text)

lloyd:
r+ - travis failure is orthogonal.

I like the minimalist solution to make discovering a problem next time faster - without boiling the ocean.

consider fixing the formatting eyesore discussed in diff, then click the green button.

ozten:
Goodcalllloyd.fixed!

resources/static/common/js/user.js
```diff
@@ -45,7 +45,13 @@ BrowserID.User = (function() {
 
           // or if it was issued before the last time the domain key
           // was updated, it's invalid
-          if (!cert.payload.iat || cert.payload.iat < creationTime) {
+          if (!cert.payload.iat) {
+            helpers.log('Data Format ERROR: expected cert to have iat ' +
+              'property, but found none, marking expired');
+            return true;
+          } else if (cert.payload.iat < creationTime) {
+            helpers.log('Certificate issued ' + cert.payload.iat +
```
shane-tomlinson, line 53: @ozten, @lloyd - I know this has already been merged, but additionally, would this info be useful in the KPI stream so that we can keep track of how frequently this occurs in the wild?

ozten, line 53: I guess so, do we have an etherpad of proposed new data elements for the roadmap?

shane-tomlinson, line 53: I have no new proposed events other than this. Does knowing this data provide any value?

PR #2995, opened 2013-02-06
Issues1232and2937transition

shane-tomlinson:
@lloyd - Could you review this one?

I added one commit that can be later reverted that adds the transition code necessary to rollback gracefully.

Major updates from the previous flow are to allow rollback from this intermediate stage to the current code as well as from the desired end flow to the intermediate flow.
- All the user/network infrastructure is left in place to split off transitionToSecondary from resetPassword. requestTransitionToSecondary is changed to call /wsapi/stage_reset.  The other two transitionToSecondary calls are left in place without any changes to handle rollback from the final flow to this intermediate flow.
- The stageReset all have a password parameter added.
- The forgot page does the old flow that accepts a password/vpassword.
- The state machine forces a user to enter their password in the dialog before verification.

All other code remains the same which should allow rollback.

### DIFF
--- a/automation-tests/pages/css.js
+++ b/automation-tests/pages/css.js
@@ -49,11 +49,12 @@ module.exports = {
     forgotPassword: 'a#forgotPassword',
     choosePassword: 'div#set_password input#password',
     verifyPassword: 'input#vpassword',
+    postVerificationPassword: 'input#password',
     // the button you click on after typing and re-typing your password
     createUserButton: 'button#verify_user',
-    resetPasswordButton: 'button#password_reset',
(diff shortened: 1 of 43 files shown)


lloyd:
```
Running 14 suite(s) on 1 platform(s), 2 at a time against https://bigass.personatest.org
>>........................<>............<>............<>..<>.........................<>.....<>..................<>.............<>....<>.......<>.............<>....................<>..........................<...<

198/198 tests passed
```

lloyd:
manually tested that we can rollback, re-skimmed code.  r+

PR #2996, opened 2013-02-06
Use connect-fonts to serve up browser/locale specific fonts to reduce overall download size.

shane-tomlinson:
This is try number two, using the new connect-fonts.

1) An update to connect-cachify is needed so that resources that are not in the root path can be looked up in a table - mozilla/connect-cachify#27
2) core font functionality is in connect-fonts which could use some review.
3) fonts are located in connect-fonts-opensans. Fonts for several Euro-centric locales have been added.

This should help significantly reduce the font footprint for the majority of our users as well as remove the fonts from the repo itself.


lib/static_resources.js
```diff
@@ -113,16 +113,21 @@ var dialog_js = und.flatten([
   ]]);
 
 exports.resources = {
-  '/production/dialog.css': [
+  '/production/:locale/dialog.css': [
+    '/:locale/opensans-regular/fonts.css',
+    '/:locale/opensans-bold/fonts.css',
+    '/:locale/opensans-light/fonts.css',
     '/common/fonts/fonts_common.css',
```
fmarier: This file was removed in your branch, should it not be removed from this array too?

lib/static_resources.js
```diff
@@ -113,16 +113,21 @@ var dialog_js = und.flatten([
...
     '/common/fonts/fonts_dialog.css',
```
fmarier: Ditto

tests/static-resource-test.js
```diff
@@ -51,13 +51,16 @@ suite.addBatch({
       assert.equal(files[minFile].length,
                    res[minRes].length);
       // Non-localized files underneath stay the same
+      /*
```
fmarier: Is this intended to be commented-out?

tests/static-resource-test.js
```diff
@@ -51,13 +51,16 @@ suite.addBatch({
...
       [0, 1, 2, 3, 4, 5, 6, 7].forEach(function (nonLocalizedIndex) {
         assert.equal(files[minFile][nonLocalizedIndex],
                      res[minRes][nonLocalizedIndex]);
       });
 
+      */
       // testing :locale has been replaced
       var localeIndex = res[minRes].indexOf('/i18n/:locale/client.json');
+      console.log(localeIndex);
```
fmarier: Did you forget to remove this?

lib/static_resources.js
shane-tomlinson: Great catch. Removing.

tests/static-resource-test.js
shane-tomlinson: I had to change the tests to make them pass. Updated this morning, a new version is pushed.

fmarier:
Looks good to me

jrgm:
I've noted two issues in `https://github.com/shane-tomlinson/connect-fonts-opensans`:

https://github.com/shane-tomlinson/connect-fonts-opensans/issues/2 
https://github.com/shane-tomlinson/connect-fonts-opensans/issues/1

jrgm:
On the other hand, in stage, I've run a selenium script to show me the dialog and main site with FF19 in the ~fully-translated languages from config/l10n-prod.json and don't see any obvious issues. A set of (mostly) western european languages are getting the ~16K woff file in firefox which matches up with what connect-fonts-opensans provides. The CJK languages get the default woff, as does Russian (Cyrillic).

Issue #1132, opened 2012-02-16
Improve string extraction for JavaScript and EJS


ozten:
As part of localization, we have to pull all of the "strings" out of the code and templates. Strings are just copy, labels, and other words shown to users.

Our current script `scripts/extract_po.sh` is okay, but lacks the following features:
- Robust JS support
- Gettext Comments in JS
- Robust EJS support
- Gettext Comments in EJS

If you read `extract_po.sh` you'll be horrified to find that we parse JS files via `xgettext` in Perl mode. Similarly, we EJS files are parsed via `xgettext` in PHP mode.

Ideally we'd either use [Translate Toolkit](http://translate.sourceforge.net/wiki/), patch Translate Toolkit with Node.js support, patch GNU Gettext (good luck with that) or ... maybe you have some ideas.

Gettext comments mean that I can do this 

```
/* l10n: This %s will be the user's name */
var msg = "Hello %s, welcome back!";
```

After string extraction, the po file has not only our string, but a special comment, so the localizer knows what `%s` is supposed to mean.

@ozten is happy to mentor first time contributors on this issue.

mathjazz:
When I extract new strings using extract_po.sh, I get different POTs generated if I leave existing POTs in the templates directory than if I empty the old POTs.

The difference is that strings that we remove for the new releae don't get removed from the POT if I keep the old POTs, but they do get removed if I empty the POTs before running extract_po.sh. So what I do now is every time before I run extract_po.sh, I empty messages.pot and client.pot in the templates folder.

(Should I file another bug?)

ozten:
@mathjazz  - please file a new bug. You basically want to make client.pot and messages.pot an empty file at the beginning of extract_po.sh, right? I can do that ASAP.

zaach:
I started a proof-of-concept [here](https://gist.github.com/8bbc336a5b261d628934).

/cc @ozten

shane-tomlinson:
@zaach - you finished this work up, didn't you?

zaach:
@shane-tomlinson Yes, it's currently being used in the standalone i18n-abide. #1322 tracks when we'll pull that into browserid (post beta1).

shane-tomlinson:
Closing this in favor of #1322, thanks @zaach

PR #3003, opened 2013-02-12
Integrated i18n-abide. Removed lib/i18n.js. Fixes #1322.

### LINKED ISSUE #1322: Break I18n L10n code out into a stand alone module
In order to reuse our I18n / L10n code in anothe rnode app, we should make a proper module out of it.

1) Create a functioning repo out of our code

https://github.com/mozilla/i18n-abide

2) Keep Mozilla specific code, config, and docs

3) Remove all other code and add i18n-abide as a dependency

Test plan - This should not affect or provide any new functionality.

### LINKED ISSUE #2644: small bug in lib/i18n.js assuming two character language codes
I was looking at the bug for changing the way we handle 'en' and noticed this small issue.

https://github.com/mozilla/browserid/blob/dev/lib/i18n.js#L187

There it assumes the language tag is 2 bytes, but >2 is possible, e.g., 'lij' Ligurian, which we have in config/l10n-prod.json.

ozten:
- Fixes #1322.
- Fixes #2644 and [Bug# 806649](https://bugzilla.mozilla.org/show_bug.cgi?id=806649)
- Removes eval usage
- Fixes format API drift `%(foo)` versus `%(foo)s`

I've already talked to mathjazz about re-writing the .po files for that last format fix. I have an automated script to replace them.


shane-tomlinson:
@ozten - I updated lockdown.json so that the Travis tests pass.

Some questions:

The change from %(variableName) to %(variableName)s leaves me scratching my head - what does the extra `s` signify? Is the extra `s` the standard? If not, does it make more sense to remove the `s` and change the % to a new tag so that meaning is immediate?

Is there a way to make sure all the strings are properly extracted and that strings are found for all the locales?

shane-tomlinson:
@ozten - i18n-abide looks awesome; another node module contributed by the Mozilla Persona team. 

I updated lockdown.json so the tests can run on travis.

The change to add the `s` leaves me scratching my head - is the appended `s` the standard? If not, would it make more sense to change the `%` to something else and ditch the `s` so that there is less to think about?

I tried running

```
CONFIG_FILES=./config/production.json node ./scripts/check_l10n_config.js 
```

and received this output:

```
Checking  1 languages from  ./config/production.json
path.exists is now called `fs.exists`.
Language  en  doesn't exist. Expected /Users/stomlinson/development/browserid/locale/en/LC_MESSAGES/messages.po
FAIL
```

Directions in the I18N directory on obtaining the SVN repo and compiling the .po/.mo/.json files would be handy. Now that these tools are part of i18n-abide, I am not sure how to compile translations! Is there a script to do this automatically? If not, should there be one?

Is there a way to check to make sure strings are properly extracted and substituted, esp when running on ephemeral instances with compressed resources?

bin/static
```diff
@@ -67,6 +68,8 @@ app.use(i18n.abide({
   disable_locale_check: config.get('disable_locale_check')
 }));
 
+i18n_check();
```
shane-tomlinson, line 71: Will this check fail twice if there is a problem since it is in both bin/browserid and bin/static?

lib/wsapi.js
```diff
@@ -40,6 +41,8 @@ var abide = i18n.abide({
   disable_locale_check: config.get('disable_locale_check')
 });
 
+i18n_check();
```
shane-tomlinson, line 44: A third time!

ozten, line 44: Yes, this is not new logic. In the `dev` branch look at the [check for client.json](https://github.com/mozilla/browserid/blob/dev/lib/i18n.js#L60). This is something that doesn't make sense in i18n-abide (yet). So I've kept the code and moved it into i18n_check.

ozten:
The original design for the `format` function is [documented here](https://github.com/mozilla/browserid/blob/dev/lib/i18n.js#L241). The very first time we did localization... I don't think we used the format function, or it was used properly... but at some point we went in and changed the semantics from `%(foo)s` to `%(foo)`.

This breaks the API of i18n-abide which BigTent uses and some people in the wild are using.

`%(foo)s` is a standard from the Python world. Our localizaers are used to seeing these strings from our other webapps and services.

shane-tomlinson:
@ozten - thanks for the comment about the trailing `s`. That makes more sense.

ozten:
> If not, would it make more sense to change the % to something else and ditch the s so that there is less to think about?

We can change from the named object syntax `%(username)s to the`%s`, but one nice property is for localizers is that these are self documenting.

Would you rather translate

```
Persona lets you use your %s account to sign into sites like %s.
```

or 

```
Persona lets you use your %(emailProvider)s account to sign into sites like %(aWebsite)s.
```

We [don't support gettext comments](https://github.com/mozilla/browserid/issues/1132) yet. So localizers have no context what the string is about. At least `emailProvider` and `aWebsite` is some documentation to help them.

ozten:
> I updated lockdown.json so that the Travis tests pass.

Awesome! Sorry about that.

ozten:
> I tried running
> CONFIG_FILES=./config/production.json node ./scripts/check_l10n_config.js 

This is not a new script. This works "as well" as it does today. I'm not sure anyone uses this script.

ozten:
Good call on missing references to creating JSON files.

I've updated docs/I18N.md with developer focused notes.

Neither `dev` nor this pull request use MO files. That was changed several months ago.

This pull request doesn't introduce any new tooling. This is the same flow we currently do.

shane-tomlinson:
> Neither dev nor this pull request use MO files. That was changed several months ago.

There is a script in the scripts directory called `compile_mo.sh`, is this still needed? There is also `check_po.sh` and `extract_po.sh`.

I have created a helper script that pulls the svn repo, compiles the json files, and compresses the templates based on this update, but I am not sure whether this is the right place to put this. Is there value in adding it to this PR?

shane-tomlinson:
I am comfortable with this this PR. Tested locally and with ephemeral instances. 

r+, will merge when given the green light that the repo is open.

@mathjazz - the updates in this PR are going to cause a lot of fuzzy mis-matches in the locales due to the new `s`'s found in EJS variables.

For example:
`%(variableName)` is now `%(variableName)s` so that our scheme stays consistent with the rest of Mozilla.

mathjazz:
Thanks, @shane-tomlinson.

I think I talked to @ozten about this some time ago, so we can use a script to replace the variable names.

PR #3012, opened 2013-02-14
Remove query string parameters for xhr urls in KPI. Fixes #3008

### LINKED ISSUE #3008: kpi stuff is reporting emails - blocker of KPI shipping
example: 

```
  "xhr./wsapi/user_creation_status?email=lloyd%40hilaiel.com",
```

this is no good.  perhaps we should wash away query data.  Any ideas on how we can fix this at a more fundamental level and reduce chances of future data leak?

ozten:
Removes query string parameters.

Changes `xhr.` to `xhr_complete`, in the future we can easily take `xhr_error`, `xhr_delay`, etc.

### DIFF
--- a/resources/static/common/js/modules/interaction_data.js
+++ b/resources/static/common/js/modules/interaction_data.js
@@ -33,7 +33,7 @@ BrowserID.Modules.InteractionData = (function() {
       dom = bid.DOM,
       sc;
 
-  function handleXhr(msg, data) {
+  function removeGetData(msg, data) {
     if (msg && data.network && data.network.type && data.network.url) {
       return msg + "." + data.network.type + data.network.url.split('?')[0];
     } else {
@@ -101,7 +101,7 @@ BrowserID.Modules.InteractionData = (function() {
     password_submit: "authenticate.password_submitted",
     authentication_success: "authenticate.password_success",
     authentication_fail: "authenticate.password_fail",
-    xhr_complete: handleXhr
+    xhr_complete: removeGetData
   };
 
   function getKPIName(msg, data) {
--- a/resources/static/common/js/modules/interaction_data.js
+++ b/resources/static/common/js/modules/interaction_data.js
@@ -33,6 +33,14 @@ BrowserID.Modules.InteractionData = (function() {
       dom = bid.DOM,
       sc;
 
+  function handleXhr(msg, data) {
+    if (msg && data.network && data.network.type && data.network.url) {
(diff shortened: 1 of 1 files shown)


resources/static/common/js/modules/interaction_data.js
```diff
@@ -33,6 +33,14 @@ BrowserID.Modules.InteractionData = (function() {
       dom = bid.DOM,
       sc;
 
+  function removeGetData(msg, data) {
+    if (msg && data.network && data.network.type && data.network.url) {
+      return msg + "." + data.network.type + data.network.url.split('?')[0];
```
shane-tomlinson, line 38: Changing from "xhr." to "msg" means that the returned message will be "xhr_complete." instead of just "xhr.". Is this intended? Otherwise, looks good.

ozten, line 38: This was intentional. I thought we might, in the future we might want the other types of xhr events (error, delay, etc).

shane-tomlinson, line 38: Has @kparlante been updated to this change? I'm not sure whether she is logging XHR requests or not.

Right now, this message is triggered for both XHR success and error, no matter how long the request takes. Is the semantic meaning that you are going for the same either way?

ozten, line 38: @kparlante - will 'xhr' versus 'xhr_complete' and 'xhr_error' be a problem?

kparlante, line 38: The current dashboard does not look at any xhr events, so no, not a problem.

shane-tomlinson, line 38: I'm all for the change then. I would actually like to make use of the [Resource Timing API](http://www.w3.org/TR/2011/WD-resource-timing-20110524/) for much more precise measurements in browsers that support it. http://calendar.perfplanet.com/2012/an-introduction-to-the-resource-timing-api/

shane-tomlinson:
I am now seeing this in the stream:

```
"xhr_complete.GET/wsapi/user_creation_status",
```

This is what we expect. Nice work @ozten. As soon as we are given the green light to merge, I'm going to merge this.

r+

resources/static/common/js/modules/interaction_data.js
ozten, line 38: Cool, I hadn't heard of the Resource Timing API. We can add that later, right?

ozten:
Thanks for deploying and testing!

PR #3013, opened 2013-02-15
Resolve issue caused by old code served under new SHA: Fixes #3010

### LINKED ISSUE #3010: Persona login window shows "emails" and "success" rather than the emails addresses to select
STR:

1) I've signed in with persona on one site (e.g. etherpad).
2) I go to log into a different site

Expected Results:

Show me the dialog with the two email addresses that I've got listed.

Actual Results:
 I get the normal dialog, however, the radio buttons to select my two email addresses are "emails" and "success" rather than the actual email address.

You can see an example here:

Screen Shot 2013-02-14 at 13 49 28

The only possible relevant item on the error console was:

Timestamp: 14/02/2013 14:00:26
Warning: WARN addons.manager: Exception calling callback: [Exception... "Component returned failure code: 0x804b000a (NS_ERROR_MALFORMED_URI) [nsIIOService.newURI]"  nsresult: "0x804b000a (NS_ERROR_MALFORMED_URI)"  location: "JS frame :: resource://gre/modules/NetUtil.jsm :: NetUtil_newURI :: line 198"  data: no]
Source File: resource://gre/modules/NetUtil.jsm
Line: (null)

On the site I was logged into last night, I had to log out fully, then mess around a lot, and eventually was able to log back in again.

However, it appears to have yet again reverted to this state.

If I visit https://login.persona.org/ then I also see "emails" and "success".

ozten:
This is the same fix as applied in 1e2647f14afacc8e1122615d17e5222128942d9c

Fixes #3010


fmarier:
r+

It would be nice to have a mechanism to change hashes without having to make these types of no-op changes.

jrgm:
So, I don't see a change to resources/static/common/js/lib/html5shim.js which was part of the previous hack (see comments at https://github.com/mozilla/browserid/commit/1e2647f14afacc8e1122615d17e5222128942d9c#commitcomment-2336912)

jrgm:
Hmm, well, you have to click to show the inline notes from me and @lloyd about html5shiv.js

shane-tomlinson:
This looks good to me, but there has to be an easier way to force cache busting than this.

lloyd:
An idea, we could make static `process.exit(1)` with an error when files on disk change.

This is inefficient, however I think the performance hit will be marginal and it would guarantee that we're not ever updating files on disk without stopping and restarting the daemon.

If we did this, hopefully it would obviate user facing impact as well as eliminate the need for us to pave the cowpath of having to fiddle etags so frequently.

PR #3015, opened 2013-02-15
issue 2903, increase STS to 10886400 for bin/router and bin/dbwriter

jrgm:
In GH-2903, STS was increased to 18 weeks for bin/browserid, but where it really had to be changed was in bin/router. I changed bin/dbwriter for consistency. 

(But hmm, why is there no Strict-Transport-Security in bin/static, et al.)


lloyd:
r+

PR #3018, opened 2013-02-18
Revert "Adding transition code to the passwordReset/transitionNoPassword flow to handle rollback."

shane-tomlinson:
This reverts commit 7a2cea449ae8d97dcc5caa071f6642b416c7d30a.

This implements the full password reset flow.

### DIFF
--- a/automation-tests/pages/css.js
+++ b/automation-tests/pages/css.js
@@ -52,9 +52,6 @@ module.exports = {
     postVerificationPassword: 'input#password',
     // the button you click on after typing and re-typing your password
     createUserButton: 'button#verify_user',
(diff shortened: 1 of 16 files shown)


lloyd:
deploying to run automated tests...

lloyd:
I love how you had the discipline to implement in two phases and how you executed.  All tests are clean (front, back, and selenium), skimming the code there's nothing significantly different from the last time I looked.  Manually walking through the reset flow it works as I'd expect on the same device and on different devices.

r+ - fantastic work shane.

jrgm:
password reset flows in multiple browsers/platforms and tests are working for train-2013.03.01 (checked this a long time ago; just noting now).

PR #3019, opened 2013-02-18
L10n helpers - make it easy to update locales locally and on ephemeral instances.

shane-tomlinson:
@ozten, would you mind reviewing this? This is an addon to mozilla/browserid#3003.
- Add an l10n-update.js helper script to make updating from the svn repo easy.
- Before compression on remote instances, run l10n-update.js and compress all locales that are enabled in aws.json


ozten:
Hawt! r+

If we should merge to dev... merge away.

PR #3024, opened 2013-02-19
Disable ejs escaping of already-escaped fields. Fixes #3007.

### LINKED ISSUE #3007: siteName is double escaped
STR:

1) In the repo, edit example/rp/index.html:186 to be "Persona Test / Relying Party"
2) Load up the dialog

Screen Shot 2013-02-13 at 16 52 14

jaredhirsch:
We already escape rp info [inside dialog.js](https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/dialog.js#L250), but our buddy [ejs](https://github.com/visionmedia/ejs#features) was double-escaping because we used `<%=`, not `<%-`, in our tags. Whoops!

TODO: is this potentially a problem elsewhere in the codebase? There's a funky asymmetry between relying on ejs to escape templated stuff, and relying on our js to escape stuff inserted by js. Should I look deeper before calling this double-escaping escapade donezo?

### DIFF
--- a/resources/static/dialog/views/rp_info.ejs
+++ b/resources/static/dialog/views/rp_info.ejs
@@ -8,7 +8,7 @@
 
 
 <% if(siteName) { %>
-  <h2 id="rp_name"><%= siteName %></h2>
+  <h2 id="rp_name"><%- siteName %></h2>
 <% } %>
 
 <% if(siteName) { %>


jaredhirsch:
@lloyd - there's a series of sanitization/escaping steps starting here: https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/dialog.js#L194. 

sitename is escaped at line 250.

hmm, hostname isn't escaped--it comes straight out of the User object, not sure if that's trustworthy.

shane-tomlinson:
@lloyd, @6a68 - hostname is taken from WinChan. We discussed whether this needs to be escaped, but the hostname is used to generate an assertion so we did not. The other argument against it was this field is populated by the browser from its handling of postMessage.

Perhaps when displaying hostname, we should continue to escape it?

jaredhirsch:
@shane-tomlinson updated, now I guess it's just a one-liner?

Issue #3037, opened 2013-02-22
sign-in-test selenium test frequently errors with "Error: funky getVerifiedUser response"


shane-tomlinson:
This is our #1 selenium test failure. It frequently causes me to cry in anguish.

```
383/384 tests passed, here are your failures:
sign-in-test - vista_firefox_16 (1 failure(s), 0 warnings):
  1: setup, errors:
    *   Error: funky getVerifiedUser response
```

PR #3040, opened 2013-02-22
Fix the sign-in-test selenium test.

shane-tomlinson:
Only use one personatestuser email. Since we were always signing in an existing user, only one user is really needed.

Along with lloyd/personatestuser.org#38, this handles #3037


jaredhirsch:
r+ :-)  nice work

PR #3041, opened 2013-02-22
Bail if persona iframe can't set cookies.

### LINKED ISSUE #2999: onlogout called immediately after onlogin if 3rd party cookies are disabled
A user reported an oddity today that may affect many Chrome users.

If 3rd party cookies are disabled, onlogout is called immediately after onlogin.

The following test page shows the error.

```
<html>
<body>
<fieldset>
    <legend>Persona Login</legend>
    <button onclick="login_persona()">Login</button>
</fieldset>
<script>
window.login_persona = (function() {
  navigator.id.watch({
    loggedInUser: null,
    onlogin: function (assertion) {
      console.info ("Checking an assertion from Persona.");
    },
    onlogout: function () {
      console.info ("Persona thinks I am or should be logged out.");
    }
  });

  return function() {
    console.info ("User requested login");
    navigator.id.request();
  }
}());
</script>
</body>
</html>
```

To both users and developers, it appears that Persona is operating as it should until the onlogout function is invoked with no clear indication why.

Instead of leaving users clueless, we should show them how to whitelist the Persona domain.

Relate to:
#982, #989

jaredhirsch:
PR is ready for final review.
- Show cookies_disabled screen on Chrome, FF18 if third-party cookies are busted.
- Still lets Android 2.2 stock browser & IE8 get the job done.
  - I really, really dislike the IE8 sniff. Suggestions welcome.

Closes #2999.


resources/static/communication_iframe/start.js
```diff
@@ -22,13 +30,14 @@
...
-  network.cookiesEnabledOverride = true;
-
-  var chan = Channel.build({
-    window: window.parent,
-    origin: "*",
-    scope: "mozid_ni"
-  });
+  // network.cookiesEnabledOverride = true;
+  network.cookiesEnabled(function(cookiesEnabled) {
+    // XXX per old comment above, could also be "if (!cookiesEnabled && !isIE)"
```
jaredhirsch: I'm particularly curious if we need to check that we're not on IE before complaining about third-party cookies not working.

jaredhirsch:
currently messing with this using an awsbox and android handsets on 2.2 and 4.2.2

jaredhirsch:
signin works fine on stock android browser 2.2. This is without third-party cookies disabled, I'm assuming the concern was to ensure it'll still actually work ;-)

jaredhirsch:
Augh, dammit. Breaks IE8. Going to use our browser sniff to get around this for IE.

Screen Shot 2013-02-22 at 11 05 44 AM

jaredhirsch:
Current plan: unbreak IE, ensure Android 2.2 still works, then rebase so that it looks like one pretty commit.

jaredhirsch:
ok, seems to work with IE8 but also warn chrome/ff18 third-party disabled browsers about cookies (using existing disabled cookies dialog screen--not a great UX but better than fake login).

about to rebase & update.

jaredhirsch:
gah, messed up the rebase. trying again...

jaredhirsch:
nope, the rebase is fine. not sure why those other commits showed up in the sequence of events here.

jaredhirsch:
ugh. closing, will reopen as pristine PR.

Issue #2831, opened 2012-12-07
Make browserid send kpi data to kpiggybank awsbox instance


jedp:
In order to send KPI data confidentially to a kpiggybank server on AWS, there are probably three parts to this issue:
1. Update `lib/wsapi/interaction_data.js` to create https connections (not http as now).  
2. Any webhead configurations, such as updating the proxy config to allow the communication.  (The urls for kpiggybank's api all begin with `wsapi/interaction_data`, if that helps at all; or perhaps it's better to go by IP address.)
3. Any changes to `lib/configuration.js` to use the http_proxy. 

There are other issues that will have to be addressed for kpiggybank, in particular setting up instances, getting ssl certs, and ensuring that read queries are restricted only to privileged apps (like kpi-dashboard).  We can track these elsewhere.

@6a68 , @gene1wood , and I talked this over on IRC.  Crucially, there is this note to bear in mind:

http://irclog.gr/#show/irc.mozilla.org/identity/186352

gene1wood:
jedp, will _all_ calls from the webheads to kpiggybank be POSTs against the specific url path "/wsapi/interaction_data"? Or will the webheads initiate "Stream Changes" calls to kpiggybank?

When we'd talked we had called out the specific target of https://kpiggybank.hacksign.in/ . I'll use that IP in my proxy rules

gene1wood:
Here's the change I'd like to make to enable this. :

```
Index: idproxy-stage.conf
===================================================================
--- idproxy-stage.conf  (revision 54058)
+++ idproxy-stage.conf  (working copy)
@@ -37,6 +37,11 @@
 acl bigtent_microsoft1 url_regex ^https://oauth.live.com/token$
 acl bigtent_microsoft2 url_regex ^https://apis.live.net/v5.0/me?access_token=.*$

+# KPIggyBank connections
+acl kpipath url_regex ^/wsapi/interaction_data$
+# to_dev_kpiggybank is the IP address of kpiggybank.hacksign.in
+acl to_dev_kpiggybank 23.20.153.45/32
+
 # Allow verifier requests over SSL.
 http_access allow bidcert ssl get

@@ -51,6 +56,9 @@
 http_access allow bigtent_microsoft1 post
 http_access allow bigtent_microsoft2 get

+# Allow KPIggyBank calls
+http_access allow to_dev_kpiggybank kpipath post
+
 # Reject any requests to internal networks.
 http_access deny to_localnet
 http_access deny to_localhost
```

[...]

PR #3042, opened 2013-02-22
KPI testing: added interaction_data activity to load_gen

kparlante:
Perhaps don't merge this just yet, but first attempt at adding capability for load testing kpi data push

### DIFF
--- a/bin/load_gen
+++ b/bin/load_gen
@@ -102,8 +102,10 @@ var activity = {
     probability: (1.0 / (40 * 56))
   },
   "interaction_data" : {
-    // post interaction data to KPI server, use the proposed sample rate
-    probability: .01
+    // post interaction data to KPI server
+    // All non-include_only activities have the potential
+    // to send interaction_data, multiply by sample rate
+    probability: (9 / 40.0) * .1
   }
 };
 
--- a/bin/load_gen
+++ b/bin/load_gen
@@ -100,6 +100,10 @@ var activity = {
   "change_pass": {
     // users change their passwords once every two months
     probability: (1.0 / (40 * 56))
+  },
+  "interaction_data" : {
+    // post interaction data to KPI server, use the proposed sample rate
+    probability: .01
   }
 };
 
--- a/lib/load_gen/activities/interaction_data.js
+++ b/lib/load_gen/activities/interaction_data.js
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+ * this file is the "interaction_data" activity, which simulates 
+ * the load of the client passing on interaction data for kpi metrics */
+
+var client = require('../../wsapi_client.js');
+
+var blob = { data:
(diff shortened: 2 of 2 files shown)


jrgm:
Hey @kparlante, thanks for doing this. Some comments:
- Instead of a hard-coded value for interaction_data.probability, maybe tie that to some proportion of the number of signin activities planned.
- maybe also check that the JSON from /wsapi/interaction_data is `{success: true}`.
- could someday also follow the pattern in other activities to pick a user (`userdb.getExistingUser()`), wrap cb, pick a context that may already have state, release user etc. But skip this point for this PR.

kparlante:
Good points @jrgm, I've cleaned this up. 

Perhaps this shouldn't be a separate activity but instead a function in common.js. We could then add it to all relevant activities (which is I think everything but include_only.js), configurable by rate.

jrgm:
@kparlante thanks for the update and looks good to me.

On the question of making interaction_data just something that happens on any activity that brings up the dialog just does, I'd thought about that. But signin in real world should be expected to dominate signup, reset_pass, change_pass, etc. So just tying this to signin seems fine to me (but did you mean `0.9`, not `0.8`).

Anyways, this tool could use an audit of which wsapi calls are made for each activity, as things have changed. But topic for another day.

shane-tomlinson:
I am not sure how to run load_gen correctly. I set up an instance and tried to point load_gen at it, but I get 400 response codes. 

I have read docs/LOAD_GENERATION.md, but it is not giving me anything obvious. Is there a configuration option that I have to flip on?

kparlante:
@jrgm: The logic for using 9 / 40 instead of 8 / 40 was that it should happen about as often as everything-that-is-not-include_only, or 100% - include_only. Or we could tie it to signin, agreed. If the tool gets an audit/overhaul, sure move it to be part of the other activities.
@shane-tomlinson: yeah, I'm not sure if it would be better to use the config for the sample rate or make it configurable from load_gen options -- one or the other is probably preferable to tweaking load_gen

shane-tomlinson:
I am not sure what number I can go up to, when I peg it at 500000 users on a c1.medium instance, I start to see:

```
4rh 235sn 0iy 1cs)        (6 503s)
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: connect EMFILE
error: (signin) unknown: Error: getaddrinfo ENOENT
[1241ms] 367567.20   395222.47   396398.77   212 R, 232 S (3sp 3rs 0al 5rh 199sn 0iy 2cs)    (10 ERRORS!)     (67 503s)
[1501ms] 407831.76   397744.32   396970.42   
```

This is intermittent, but somewhere is running out of file descriptors. I am unsure if this is on my local system or if this is reported from the remote end.

jrgm:
Shane's issue is that osx has max files 250 by default, and I could easily see that being transiently surpassed during a run.

jrgm:
Also re: size, I run tests @250K and add more instances to drive up to millions. The reason is that the number of users grows over time and at a certain point a single client core can't stay on schedule.

jrgm:
verified in stage (sort of; see GH-3064, since in stage the current config leads to a bit of a loopback).

Hey @kparlante is there a test instance of the collector that we can file a ticket to get stage pointed at for initial load testing.

kparlante:
@jrgm: I have two test instances of the collector right now (kpiggybank): kpiggybank-stage.personatest.org and kpiggybank-secure.personatest.org. Anyone can use kpiggybank-stage, access to kpiggybank-secure is whitelisted. Stage should point at the latter collector, and I'll need to configure it properly. Setting up stage properly also involves configuring the proxy, assuming it matches production. Details in #2831

jrgm:
@kparlante - thanks for the info. Can we chat in IRC about stuff that needs to be done to finish stage in scl2 to point at https://kpiggybank-stage.personatest.org?

PR #3043, opened 2013-02-22
Bail if 3rd party cookies disabled. Closes #2999.

### LINKED ISSUE #2999: onlogout called immediately after onlogin if 3rd party cookies are disabled

jaredhirsch:
Chrome and FF18 currently will not allow any third-party to access localStorage
  or cookies in a third-party iframe. Safari is smarter and allows this for third-
  parties after the user interacts with their site (via throwing the popup).

  Added an exception for IE8, as it disallows cookie setting but will actually
  allow network requests to go through. Not thrilled about UA sniffing, though.

  Verified does the right thing on Chrome, FF18, IE8, and android 2.2 stock handset.

See also discussion in bug #2999.

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -908,8 +908,12 @@
       }
     }
 
+    function checkThirdPartyCookies() {
+      if (thirdPartyCookiesDisabled) return "THIRD_PARTY_COOKIES_DISABLED";
+    }
+
     function isSupported() {
-      reason = explicitNosupport() || checkLocalStorage() || checkPostMessage() || checkJSON();
+      reason = explicitNosupport() || checkLocalStorage() || checkPostMessage() || checkJSON() || checkThirdPartyCookies();
 
       return !reason;
     }
@@ -1006,6 +1010,9 @@
       }
     }
 
+    // used in interalRequest to warn the dialog if third-party cookies
+    // are disabled. we get this info from the communication_iframe.
+    var thirdPartyCookiesDisabled;
 
     // this is for calls that are non-interactive
     function _open_hidden_iframe() {
@@ -1059,6 +1066,10 @@
 
           commChan.bind('match', function(trans, params) {
             if (observers.match) observers.match();
+          });
+
+          commChan.bind('cookiesDisabled', function(trans) {
+            thirdPartyCookiesDisabled = true;
           });
 
           if (defined(loggedInUser)) {
@@ -1148,6 +1159,8 @@
       checkRenamed(options, "tosURL", "termsOfService");
       checkRenamed(options, "privacyURL", "privacyPolicy");
 
+      options.thirdPartyCookiesDisabled = thirdPartyCookiesDisabled;
+
       if (options.termsOfService && !options.privacyPolicy) {
         warn("termsOfService ignored unless privacyPolicy also defined");
       }
@@ -1179,6 +1192,11 @@
         url = "unsupported_dialog";
 
         if(reason === "LOCALSTORAGE_DISABLED") {
+          url = "cookies_disabled";
+        }
+        // XXX we could add a more specific case to cookies_disabled
+        if(reason === "THIRD_PARTY_COOKIES_DISABLED") {
+          console.log('discovered third party cookeis disabled, redirecting dialog to error page');
           url = "cookies_disabled";
         }
 
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -15,20 +15,30 @@
 
   network.init();
 
-  // Do not check to see if cookies are supported in the iframe.  Just
-  // optimistically try to work by running network requests.  There are
-  // cases (especially in IE) where our checks will fail but our actual
-  // requests will not.  issue #2183
-  // (NOTE: if we want to try to improve failure modes for users with
-  //  a "disable 3rd party cookies"-like preference set in their browser,
-  //  we may need to re-visit this)
-  network.cookiesEnabledOverride = true;
-
   var chan = Channel.build({
     window: window.parent,
     origin: "*",
     scope: "mozid_ni"
   });
+
+  // In order to warn the user about third-party cookies being disabled,
+  // try to set cookies from this third-party iframe & phone home if it's
+  // not possible. We'll alert the user on dialog open.
+  //
+  // IE is the exception; it'll fail to set cookies, but still actually allow
+  // network requests. In that case, we'll just ignore the cookie check.
+  // See #2183 for more (also see earlier revisions of this file).
+  // TODO do we need to make exceptions for other browsers?
+  network.cookiesEnabled(function(cookiesEnabled) {
+    if (!cookiesEnabled && 
+        navigator.appName !== 'Microsoft Internet Explorer') {
+      chan.notify({ method: 'cookiesDisabled' })
+    }
+  })
+
+  if (navigator.appName === 'Microsoft Internet Explorer') {
+    network.cookiesEnabledOverride = true;
+  }
 
   var remoteOrigin;
 


resources/static/include_js/include.js
```diff
@@ -1181,6 +1194,11 @@
         if(reason === "LOCALSTORAGE_DISABLED") {
           url = "cookies_disabled";
         }
+        // XXX we could add a more specific case to cookies_disabled
+        if(reason === "THIRD_PARTY_COOKIES_DISABLED") {
+          console.log('discovered third party cookeis disabled, redirecting dialog to error page');
+          url = "cookies_disabled";
```
shane-tomlinson, line 1200: What about "cookies_disabled#thirdparty"?

shane-tomlinson:
start.js has a jshint error:

```
 run jshint on the static directory
1047 ✗ should have no jshint warnings
1048 » expected [ 'Missing semicolon. /home/travis/build/mozilla/browserid/resources/static/communication_iframe/start.js:35', 'Missing semicolon. /home/travis/build/mozilla/browserid/resources/static/communication_iframe/start.js:37' ] to have 0 element(s) // assert.js:49
```

shane-tomlinson:
Give this a try in FF Nightly and Safari with clean profiles. I am wondering if this will cause a pre-mature failure in those browsers for the first time visit to a Persona RP. Since the Persona dialog will not have yet been opened, they may cause a failure in the cookie check.

resources/static/include_js/include.js
jaredhirsch, line 1200: is that a thing that already exists, or a thing I should add?

jaredhirsch:
dang, good call. I'll look into it.

IIRC it worked beautifully in Safari with third-party cookies disabled--but maybe that's because I hadn't used a totally fresh profile.

I wonder if 3p cookies become legal on click or on load of the first-party page? (Probably the latter?)

If we can't establish the relay until after the dialog has loaded, it seems like our options aren't great:
- delay dialog render time: wait to render the dialog until phoning home & asking the comm_iframe to do the check
- interrupt the user later: show the dialog, do the check async while rendering the dialog, then jump to a "3p cookies fail" screen if the check failed--presumably while the user is typing

In either case, it seems like we can't use the cookies_disabled screen, we need to load all the stuff in the dialog, in order to send signals back and forth.

@shane-tomlinson do you see any cleaner/simpler ways to handle the "no third-party till first-party page load" case?

seanmonstar:
I was thinking the same thing: we don't _need_ cookies in the iframe until after they've used us as a 1st party (opening the dialog), but if we check for cookies before, and they've never used us before, then the iframe will blow up pre-maturely.

jaredhirsch:
OK, so, I'm running code on etagify.personatest.org (reusing an awsbox from blog post work :-P).

Safari works fine, FF nightly fails with the cookies disabled error.

jaredhirsch:
okay, I've gone around on this for a couple of hours.

Suppose we wait till the dialog is opened, then did the iframe 3rd party cookie check, and found that it failed.

How can we get the info back to the dialog, so that the dialog shows an informative error screen?
- can't use winchan, because it only fires once when opening the dialog
- can't use window.opener tricks, because of cross-domain restrictions between dialog and RP page

I think we'd have to create a persona.org trusted iframe on the RP page, and use that to establish a channel between the dialog and the RP page. This gives us a pocket of trustedness on an untrusted page.

I need to take a break for this for a while--either will get back to it tonight or tomorrow. Comments please!

seanmonstar:
We already have an iframe on the RP, the `communication_iframe`...

shane-tomlinson:
Ugh, this is tricky.

Things we know:
- If we check for 3rd party cookies from the communication_iframe naively, we receive an error for first time Firefox users who have not yet opened the dialog. cookies will work fine for these users once they have opened the dialog.
- To get around this, we only want cookies to be checked for users who have already opened the dialog.
- We need a way to know the user has opened the dialog. Store a bit in localStorage?
- Chrome blocks access to localStorage in an iframe if 3rd party cookies are disabled.

Thoughts of alternate approaches:
- Use loggedInUser from navigator.id.watch as the bit of info needed to indicate a user has visited the Persona dialog before. Unfortunately, it only means a user _may have_ opened the dialog before. If the user has a session with the site before the site integrates Persona, the site may set the loggedInUser field without the user ever having visited Persona. This leads to incorrect failures. Can we make this work?
- Performance destroying approach: use two distinct domains to load the dialog. First, load the dialog on one domain. Have it set a cookie. Redirect to a second domain which takes care of the bulk of the work. Request a resource that uses the set cookie from the first domain and see if the cookie makes it through. I think a kitten was just killed for thinking of this.
- Could we do an interplay between the dialog and the iframe? 
  1) iframe checks and believes it's cookies are disabled, takes no action.
  2) dialog opens, user has now visited the site.
  3) dialog messages to iframe "hey, go check a cookie"
  4) iframe requests a special cookie checking wsapi from the server - maybe just session_context. since the dialog is open, the user should have a session cookie which gets sent to the wsapi call.
  5) server responds with "I got your cookie" or "who are you?"
  6) iframe sends whether the server got the cookie back to dialog
  7) the dialog says "yay! and continues" or "dude, 3rd party cookies are disabled and we need them for the .watch API."

Is this wack? If we only do it for the .watch API, we can avoid this check for the internal API or for RPs that use .get

jaredhirsch:
@seanmonstar: Oh right, that's slightly embarrassing. I guess it's good I took a break when I did.

jaredhirsch:
@shane-tomlinson so, actually, when the dialog is opened, the RP page sends the iframe the 'dialog_running' signal, so I think that could be used to tell the iframe to try the cookie check and postmessage failure results to the window with the __persona_dialog handle?

jaredhirsch:
@shane-tomlinson whoa, I need to read your comments more carefully, that's good stuff. I didn't realize we only need third-party cookies for watch().

I was thinking the cookie check would just be, basically, what network.cookiesEnabled does: try to set a cookie or access localstorage locally from the communication_iframe. I'm not sure we need a server roundtrip, unless I'm overlooking something, which is very possible at this point.

So the chain of events might be:
- dialog opens
- RP page sends 'dialog_running' message to communication_iframe
- communication_iframe attempts to set cookies (possibly just reusing network.cookiesEnabled with errback?)
- if cookie setting fails, communication_iframe sends a postmessage with "sorry bro, 3p cookies disabled"
- dialog uses renderError() to bail in the middle of user interaction

I need to double-check that winchan will let me send the window handle from RP page to communication_iframe, but otherwise I think this seems doable.

jaredhirsch:
ok, @shane-tomlinson and I talked through some possibilities--right now we're going to let this sit for a day and see if the current plan makes sense (use RP page as relay between iframe + dialog, and deal with IE8 workaround separately).

jaredhirsch:
this has been deprioritized in favor of Q1 commitments. will reopen in april.

PR #3052, opened 2013-02-26
Unit tests to make sure fonts are working as expected.

shane-tomlinson:
@jrgm or @6a68, mind reviewing?
- Upgrade to the newest version of connect-fonts.
- Split off respondsWith into its own file so that it can be shared.
- For each locale, check whether all of the fonts.css requests are being served.
- For each served fonts.css, check to make sure embedded fonts are fetchable.

### DIFF
--- a/lib/wsapi_client.js
+++ b/lib/wsapi_client.js
@@ -98,7 +98,8 @@ exports.get = function(cfg, path, context, getArgs, cb) {
     injectHeaders(context, headers);
     injectCookies(context, headers);
 
-    if (typeof getArgs === 'object')
+    // skip the query string if getArgs is null (null is an object)
+    if (typeof getArgs === 'object' && getArgs !== null)
       path += "?" + querystring.stringify(getArgs);
 
     var timingHandle = startRequest();
--- a/tests/fonts-request-test.js
+++ b/tests/fonts-request-test.js
@@ -1,0 +1,176 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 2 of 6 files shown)


lib/wsapi_client.js
```diff
@@ -98,7 +98,8 @@ exports.get = function(cfg, path, context, getArgs, cb) {
     injectHeaders(context, headers);
     injectCookies(context, headers);
 
-    if (typeof getArgs === 'object')
+    // skip the query string if getArgs is null (null is an object)
+    if (typeof getArgs === 'object' && getArgs !== null)
```
jaredhirsch, line 102: you could do if (getArgs && typeof getArgs === 'object')? Slightly less verbose...

shane-tomlinson, line 102: I dig it.

tests/fonts-request-test.js
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * The basic flow:
+ * Get a list of all production resources for all locales
+ * Search through the list of resources for all /fonts.css
+ * Check to make sure all /fonts.css are fetchable.
+ * Peak into each fonts.css file to look for embedded url(...)s
+ * Fetch each embedded URL to make sure the font is served.
+ */
+
+// Vows can only handle up to 50 tests per batch. After 50, it just
```
jaredhirsch, line 33: wow, that's depressing

jaredhirsch:
r+ looks good, works well. merging in.

tests/fonts-request-test.js
shane-tomlinson, line 33: I searched the vows code and cannot find a hard limit, but when I added 100 tests to a batch, only the first 50 would run. It was janked.

Issue #3034, opened 2013-02-22
[ie8] main site: password fields not shown until mouse moved after the click


jrgm:
STR
1. Using IE8, sign in with a known secondary address at https://login.persona.org/
2. In the Account Manager UI, for the password section, click 'edit' **but be very careful not to move the mouse after the click**

Expected: UI changes to reveal two password entry inputs
Actual: password inputs are not shown
1. Now move the mouse: the password inputs will be shown.

(Linking to GH-3005, which may or may not be related.)

jrgm:
Yeah, looks good in dev with IE8.

Issue #3005, opened 2013-02-13
[android4.2.1/stock] Main site, change password doesn't show password input boxes


jrgm:
Sorry, I've been burnt out of clicking on mobile.

STR:
1) login to https://login.{persona,anosrep}.org
2) On the accounts page, click on 'edit' for the password.

Expected: two password fields are visible for password entry
Actual: no editable fields are shown; nothing in `adb log` of interest, probably some style/dom thing.

Fails on Android 4.2.1; stock browser 4.2.1-533553

Works on 4.2.1 for current Chrome, FF, FFBeta, FFAurora, FFNightly. 
And on android 2.3.20 (droid2/verizon) for **stock version 2.2***, and  
FF, FFBeta, FFAurora, FFNightly.

shane-tomlinson:
@jrgm - on my Nexus 4 with 4.2.1, I _only_ have Chrome by default, I have no stock browser. How can I go about installing that?

jrgm:
@shane-tomlinson - hmm, my 4.2.1 phone is also a Nexus, but one that I've had for >1 year with several OTA kernel updates. I may just have a legacy stock browser here. I'll ask some folks here in MV about this in (PST) morning if there's a way to get the stock browser installed on your phone.

callahad:
I'll try to reproduce

callahad:
Aaand I am able to reproduce this.

callahad:
For anyone without a device, this is also reproducible in the Android emulator.

Issue #2296, opened 2012-08-15
IE8: Old Password/New Password fonts on Account Manager page are jagged


shane-tomlinson:
STR

1) In IE8, load up the account manager with an account containing at least one secondary
2) Click the edit button for the "Change Password" section.
3) Check out those label fonts

![Jagged Fonts in IE8](https://img.skitch.com/20120815-xs9dc6at9aprf2dtw8y76f527f.jpg)

Issue #1837, opened 2012-06-22
Android 2.3 Native: possible to focus hidden change password fields


shane-tomlinson:
![Android 2.3 Native with hidden password fields focused](https://img.skitch.com/20120622-bs3mm6p23k64xcuafnbkdwnsyf.png)

csuciu:
I am unable to reproduce on Samsung Galaxy R - Android 2.3 using stock browser and FF mobile.

@shane-tomlinson Are you still seeing this issue

PR #3053, opened 2013-02-26
Instead of using a CSS fadein to show the change password fields, use display: none and display: block.

shane-tomlinson:
This fixes the following problems:
- Fields not shown in IE8 until mouse movement [#3034]
- Jagged text in IE8 [#2296]
- Able to focus hidden password fields in Android 2.3 Stock Browser [#1837]

And hopefully (untested):
- Password fields not shown in Android 4.2.1 Stock Browser [#3005]

### DIFF
--- a/resources/static/pages/css/style.css
+++ b/resources/static/pages/css/style.css
@@ -305,20 +305,11 @@ button.delete:active {
 }
 
 .showedit {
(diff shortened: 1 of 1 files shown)


callahad:
r+, and verified on Android 4.2.1 Browser.

PR #3054, opened 2013-02-26
Throw an error if navigator.id.request is called before navigator.id.watch

### LINKED ISSUE #3023: Detect against mis-use of navigator.id.request before the watch API
As reported in https://bugzilla.mozilla.org/show_bug.cgi?id=842784 is likely that misuse of `request` and `watch` end up changing the behavior of the dialog.

This shouldn't be possible.

If we can, we should just make things work.

If we detect that watch's state isn't setup when we get a call to request, we should log an error to let a developer know.

Basically, we need to validate if this hypothesis for 842784 is accurate and what is the best behavior for users and developer feedback.

shane-tomlinson:
fixes #3023

### DIFF
--- a/resources/static/include_js/include.js
+++ b/resources/static/include_js/include.js
@@ -1239,6 +1239,10 @@
       request: function(options) {
         if (this != navigator.id)
           throw new Error("all navigator.id calls must be made on the navigator.id object");
+
+        if (!observers.login)
+          throw new Error("navigator.id.watch must be called before navigator.id.request");
+
         options = options || {};
         checkCompat(false);
         api_called = "request";


jaredhirsch:
R+, throws as it should. Merging.

Thinking about testing this feature, I am starting to like the idea of using the local test RP, or maybe a set of local RP pages: it would make selenium testing much faster and more reliable (no network traffic). The annoyances might be 
- modifying the system's `/etc/hosts` so we could effectively simulate cross-domain stuff
- setting up certs locally, to enable testing of https vs http behavior

@shane-tomlinson any thoughts on whether this seems like a useful approach to explore?

shane-tomlinson:
@6a68 - I think we could generalize your approach and throw up a new test RP that has several different pages with a variety of errors. A local version, as you say, would be hot. Using selenium, we could visit each page in turn and see if the expected error is thrown.

resources/static/include_js/include.js
```diff
@@ -1239,6 +1239,10 @@
       request: function(options) {
         if (this != navigator.id)
           throw new Error("all navigator.id calls must be made on the navigator.id object");
+
+        if (!observers.login)
+          throw new Error("navigator.id.watch must be called before navigator.id.request");
```
ozten, line 1244: Awesome! Thanks for the quick turn around.

PR #3055, opened 2013-02-26
update README, minimum node required is 0.8.11

timkuijsten:
The minimum required version of node is not 0.6.17 but 0.8.11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ If you get stuck and need help, you can find the core team on our [public mailin
 
 BrowserID needs the following dependencies before it can run:
 
-* node.js (>= 0.6.17)
+* node.js (>= 0.8.11)
 * libgmp3
 * g++
 


jaredhirsch:
Thanks!

callahad:
Thanks Tim! I wonder what sort of forcing function we could build to make sure the readme gets updated when we bump Node versions...

timkuijsten:
Hi Dan, I was thinking about it too, I only have experience generating docs with my own node scripts. Would that be a feasible option? otherwise maybe a build script or something like that?

ozten:
We could write a vows test that makes sure the README, package.json, and scripts/browserid.spec all agree.

seanmonstar:
@ozten that'd be awesome! Except... for a while, we had the package.json and browserid.spec differ, since we wanted the latest version on servers, but travis-ci didn't quite have it yet.

Issue #1971, opened 2012-07-09
include.js should be generated


seanmonstar:
include.js has both JSChannel and WinChan as dependencies, but those are also include in common/js/lib. include.js received some patches that fixed a few things in WinChan (at the least), that never got upstream. To prevent this, the include.js should get generated from the lib files, ensuring that there is no source duplication (and divergence).

shane-tomlinson:
We had another issue (#483) opened for this, but I closed it as "won't fix".  I still think there is a lot of value in this.

Also link to #1199

shane-tomlinson:
link to #1749

callahad:
FWIW, as of today, dev (ef06d2c2ffd0ffb8e44924a3257b511e2adb2579) has identical copies of jschannel in include.js and the resources folder, modulo differences in whitespace.

winchan slightly differs. digging in a little more...

shane-tomlinson:
#3056 is merged, closing.

callahad:
reopening so I can find it when I get home -- we haven't completely fixed the duplication yet, just made it fixable...

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

Issue #1199, opened 2012-02-27
include.js, embedded winchan needs updating, breaks IE9 testing on localhost


mmullens:
resources/static/include_js/include.js:

does not embed the latest version of winchan which causes inability to find relay frame while developing browserid enabled site on localhost and testing on localhost with IE9 (and probably others).  Issue is that the relay frame is never found.  

The fix is to embed the latest version of winchan in include.js.  There are some significant fixes in latest winchan that should be kept in sync with include.js since it is being embedded.

Thanks,
Matt

shane-tomlinson:
@mmullens - is this still an issue?

mmullens:
It is difficult for me to check at the moment, but I think that it is.  I checked about a month or so ago and there were differences in findRelay and somewhere in the onOpen method if memory serves.  Basically I took winchan out of include.js and compared it to winchan on github, found the differences, patched it, and then things worked.  I apologize that I can't be of more help right now, I should have uploaded the diff while I had it for ease of investigation.

But yes, if it is generated, then that is much easier if you expect many changes to winchan in the future.  Thank you for all of your work on this system, beautiful piece of work!

jaredhirsch:
are we good to close this?

jedp:
Should be fixed now.  Please reopen if necessary.

Issue #483, opened 2011-10-25
break include.js into distinct components (for testing), combine in build script


shane-tomlinson:
This is to facilitate unit testing.

ianb:
FWIW, here's what I use to include the JSChannel code into OWA's include.js: https://github.com/mozilla/openwebapps/blob/develop/site/tools/inline_code.py

(though I check in the result)

shane-tomlinson:
thanks @ianb

shane-tomlinson:
Closing as won't fix.

shane-tomlinson:
Re-opening with @callahad's push to make this happen.

callahad:
Closing back down -- include.js is now broken out into its constituent components and combined at build time. We just have to do some cleaning to make it awesome...

PR #3056, opened 2013-02-27
Generate include js

callahad:
This is a first step toward #1199 and #1971.

Reviewer: It should be easiest to follow this if you look at each commit individually -- I took care to make sure the progression was natural.


fmarier:
r+

Thanks a lot for breaking this up in multiple commits, it was really easy to review!

seanmonstar:
My only thought is if there's a way to preserve the history onto the `_include.js` file, so we can `git blame` and understand when something looks crazy.

callahad:
I don't believe there's an easy way to do that which preserves the readability the commit sequence :(.

Boo for Git only doing heuristic-based rename tracking.

seanmonstar:
This sounds nuts, and forget it if the value doesn't seem there, but would this work?

Starting from the top: rename include.js to _include.js in a commit, then create the _files, deleting lines from _include.js in another commit. Then, all the left over lines still have history?

callahad:
That would probably get us there, but at the cost of muddling the actual commit history (versus the blame history), which makes it harder to verify that the snippets do indeed yield a result identical to the original file.

(In b83eb67, the previous include.js is still in the repo, so any mismatch on generation would show up...)

The `_include.js` snippet is less than 430 SLOC -- I'm inclined to say good idea, but not worth it. Or at least, you're welcome to do it if you think it'd be significantly valuable.

shane-tomlinson:
@callahad - This is HOT! I like this approach and have wanted it for a long time. Aside from the benefits you point out, a split up include.js should allow us to write front end unit tests that exercise various code paths [see #483].

Some suggestions:
- include.js is only generated in postinstall.js. If we are editing include.js and it needs to be re-built, there should be a mechanism to automatically do this. Perhaps in lib/static/views.js, where it fetches include.js if you check to see if the served resource needs to be rebuilt. We do this for other resources; see the /build/code_version.js route for an example. Calling generate_include.js from within a route probably necessitates making it asynchronous.
- Our generated resources have been placed into the `build` subdirectory of resources/static. If you place include.js into there, no need to add include.js to .gitignore and no danger of a stale version being served up by static middleware if a component was updated.
- Perhaps split BrowserSupport out into its own module. We have one in resources/static/common/js/browser-support.js and maybe these can be shared.
- The component resources should all have the license at the top. All but the first will be stripped during minification.
- Can the winchan.js and jschannel.js be shared with the resources already in resources/static/common/js/lib? It would be great if we could make updates to either of these in one place and have the changes reflected everywhere.
- The test in tests/jshint-test.js may need to be updated (I have not run this)
- To facilitate unit testing, it would be nice to be able to call some initialization function in include.js that allows us to mock in components like WinChan and JSChannel. In the built version, the initialization function would be called with the real WinChan and JSChannel.

callahad:
@shane-tomlinson My immediate goal was to end up with an on-disk representation (including paths, etc) that was identical to before this pull request. All of those suggestions sound like solid "next steps." I'm going to go ahead and merge this, then work on another pull request to handle the shared components.

jrgm:
no problems noted in stage train-2013.03.01

PR #3059, opened 2013-02-27
b2g, picl/native merge - WIP 95% done

jaredhirsch:
NOT READY BUT NEEDS COMMENTS SUCKA!!!

```
  remaining conflicts:
    lib/db/mysql.js
    resources/static/common/js/user.js
    resources/static/include_js/_jschannel.js
    resources/static/test/mocks/xhr.js
```

Please take a look at those files and suggest how to resolve :-)


jedp:
Also, we have notes in this etherpad: https://id.etherpad.mozilla.org/the-big-b2g-merge

callahad:
An omnibus merge with ~3000 lines of churn and which introduces numerous new product features scares me, from a review standpoint, especially as we're getting ready to announce Beta 2.

Is it possible to split this into a series of smaller pull requests, one per feature?

jaredhirsch:
Hey Dan - Uh, you're welcome to submit an alternate pull request ;-)

I, for one, don't fancy that merging piecemeal will actually produce a more
reliable end product. (I'm happy to be proved wrong, though.)

jedp:
@callahad, yes.  Diving in and doing a full-on merge seemed like the best way to see what the hard bits are, and what the overall difficulty of the merge will be.  (Overall, I have to say, I'm amazed it was so straightforward.)
@6a68 and I were thinking the plan forward might be something like this:
1. resolve merge of remaining four files
2. deploy this branch to an awsbox and QA it as we would both for dev and for b2g
3. when QA is satisfied, push to native-persona.org - now we have a path for pushing ongoing dev fixes and speed-ups back to b2g (all that make-us-fast goodness is badly needed on b2g)
4. split this PR into four or five PRs, one for each of the main features being merged, as you suggest

But we can't do anything until the conflicts in those files are resolved.

lib/db/mysql.js
```diff
@@ -486,9 +491,21 @@ exports.completePasswordReset = function(secret, password, cb) {
         function(err) {
           if (err) return cb(err);
 
+<<<<<<< HEAD
           // update the password!
           exports.updatePassword(uid, password || o.passwd, true, function(err) {
```
jedp, line 496: The `password || o.passwd` was added in commit @84346b20; I presume the corresponding changes want to be made to the b2g branch lines below (`addEmailToUser` and `updatePassword`).  Is that correct?

jaredhirsch:
Let me rephrase minus the snark.

If we break into, say, 4 smaller features, we incur QA costs for each
feature independently, then the final integration pass, and we have to
worry that we've mutated history incorrectly in doing the rather complex
split along the way.

If we just throw the whole blob into testing and verify that all the
features, website and native, work as expected, then we're done all at
once, or we file individual bugs to fix individual features.

The latter approach seems more tractable to me. I'm open to discussion
though :-)

resources/static/common/js/user.js
```diff
@@ -1215,13 +1299,77 @@ BrowserID.User = (function() {
                 else {
                   complete(null);
                 }
+=======
```
jedp, line 1302: I think the main narrative here is:
- @50e88dcc `forceIssuer` was added to `getAssertion`
- @8c599a80 some production hotfixes were merged from dev to b2g
- @95d38165 update to handle expired primary certs

Would love to have @shane-tomlinson or @ozten take a look at the merge here

resources/static/include_js/_jschannel.js
```diff
@@ -613,3 +613,698 @@
     };
   })();
 
+<<<<<<< HEAD:resources/static/include_js/_jschannel.js
```
jedp, line 616: Commit @71ff754c saw a big refactor of `jschannel`; I think we just need to reflect those changes in b2g.

Note that the test for the existence of of a native implementation should not be for `navigator.id`, but for `navigator.mozId`.

callahad:
My primary concern is the feasibility of actually _reviewing_ a 3,000 line patch that does so much.

I haven't _tried_ to do that, yet. Just a quick reaction to the size + list of things in the etherpad.

callahad:
Case in point, when we try to "verify that all the features, website and native, work as expected," well, do we actually have a list of "all the features" to verify against?

jedp:
One more thing I want to note about this: We're calling this a b2g merge, but this has necessary features for other native clients.  On picl, the native persona implementation depends on exactly this set of patches.  So this is a b2g and firefox/picl PR.

jedp:
@callahad I think we have a pretty good list of all the use cases and features that need to be verified, and we have the list in triplicate: bz bugs for identity, marketplace stories that drove some of this, and marketplace QA bugs in testing these features as part of larger flows.  

One of the first things I'd like to do is take the marionette tests I wrote for b2g (marionette is selenium with firefox superpowers) and incorporate them into our selenium tests for dev.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -171,6 +211,14 @@ BrowserID.Modules.Dialog = (function() {
         startChannel.call(self);
       }
 
+      // if B2G
+      //this.click(TOSPP_SELECTOR, showTOSPP);
+      $(TOSPP_SELECTOR).live('click', function(e) {
```
shane-tomlinson, line 216: A live event is not torn down whenever the dialog object is torn down. Can we just use `this.bindEvent` or `this.click`?

resources/static/dialog/js/modules/dialog.js
```diff
@@ -146,6 +149,43 @@ BrowserID.Modules.Dialog = (function() {
   }
 
 
+  function showTOSPP(e) {
```
shane-tomlinson, line 152: This should be its own module.

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -41,7 +41,13 @@ BrowserID.Modules.PickEmail = (function() {
       return;
     }
 
-    var identity = user.getStoredEmailKeypair(email);
+    var identity;
+    if ('default' !== user.forceIssuer) {
```
shane-tomlinson, line 45: This looks like it needs a test. Would it be better to abstract getStoredEmailKeypair so that is checks forceIssuer itself and does the appropriate action?

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -38,16 +39,18 @@ BrowserID.Modules.Authenticate = (function() {
   }
 
   function hasPassword(info) {
-    return (info && info.email && info.type === "secondary" &&
-      (info.state === "known" || info.state === "transition_to_secondary" ));
+    /*jshint validthis:true*/
+    return (info && info.email && info.type === "secondary" &&
```
shane-tomlinson, line 43: These changes have no corresponding tests.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -85,7 +87,14 @@ BrowserID.Modules.Authenticate = (function() {
       else if ("primary" === info.type) {
         self.close("primary_user", info, info);
       }
-      else if (hasPassword(info)) {
+      else if (!!self.forceIssuer && 'default' !== self.forceIssuer) {
```
shane-tomlinson, line 90: There is no corresponding tests for this.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -118,6 +127,18 @@ BrowserID.Modules.Authenticate = (function() {
     }
   }
 
+  function createFxAccount(callback, forceIssuer) {
```
shane-tomlinson, line 130: Nor this.

resources/static/dialog/js/modules/actions.js
```diff
@@ -95,6 +95,17 @@ BrowserID.Modules.Actions = (function() {
...
     },
 
+    doAuthenticateWithUnverifiedEmail: function(info) {
+      var self = this;
+      dialogHelpers.authenticateUser.call(this, info.email, info.password, function() {
+        self.publish("authenticated", info);
+      });
+    },
+
+    doResetPassword: function(info) {
```
shane-tomlinson, line 105: This was removed in dbed8308624a1b930c24c46dbcb0d26d029c10f4

resources/static/dialog/js/modules/provision_primary_user.js
```diff
@@ -56,7 +54,7 @@ BrowserID.Modules.ProvisionPrimaryUser = (function() {
         throw new Error("missing config option: email");
       }
 
-      user.addressInfo(email, function(status) {
+      user.addressInfo(email, 'default', function(status) {
```
shane-tomlinson, line 57: Should addressInfo check forceIssuer itself and take the appropriate action so that passing in 'default' is not needed?

resources/static/dialog/js/misc/state.js
```diff
@@ -199,20 +205,30 @@ BrowserID.State = (function() {
       complete(info.complete);
     });
 
+    // B2G forceIssuer on primary
+    handleState("new_fxaccount", function(msg, info) {
```
shane-tomlinson, line 209: All the new_fxaccount stuff needs tests.

shane-tomlinson:
I'd like to see this split up as well. If an individual feature is not possible to merge in by itself, then at least smaller groupings. There is just no way to reliably review a 3000 line patch, and even manual testing is only going to get us so far.

It looks like on the front end, some tests that I took out in cdaa5573 have been re-added, at least for network.js and user.js. The way these tests were done was completely re-worked, but I see the old code here again. There are other areas with problems - the re-worked password reset code seems to have been reverted, the old flow has been re-introduced.

These two together make me question the overall integrity of a PR this size.

Selenium, backend, and frontend tests would make my confidence level skyrocket. 

Other things I'd like to see:
- New frontend tests for new functionality
  *\* fxaccount in `set_password.js`
  *\* All the new functionality in dialog.js relating to TOS/PP
- New selenium tests for new functionality
- Get rid of the `$.live` event handler in dialog.js and use `this.bind` or `this.click`
- Split `showTOSPP` (https://github.com/mozilla/browserid/pull/3059/files#L36R152) into its own module. dialog.js is getting awefully cluttered.
- Other comments inline. Could some user.js functions be more cleanly abstracted?

More review to come.

jedp:
@shane-tomlinson thanks for your detailed comments.  More tests are clearly needed.

Do you have any thoughts about `resources/static/common/js/user.js` lines 1302 and following?

And it sounds like you also favor breaking this PR up into components.  As soon as I have the big merge resolved, I can start focusing on that.

jaredhirsch:
@shane-tomlinson it is a bummer that a bunch of out-of-date stuff reappeared. so much for the quick and dirty.

@jedp I'm happy to help going through this branch a wee bit more carefully, we can regroup next week and take it one file at a time (or something).

jedp:
Yes, that's huge drag that dead and removed code was reanimated in the merge.  Argh.

@6a68 yes, let's try to set some time aside this week to go through this together.  I still have hopes that it can be done without too much pain.

jaredhirsch:
@jedp sounds good. we can take it commit-by-commit, or try doing the "merge
trunk into feature branch, clean up, then merge branch back" thing.

I'll do a little reading on suggested strategies for gigantic git merges.
Surely this thing has happened once or twice in the history of the kernel
;-)

jaredhirsch:
ok. clearly this experiment shows that our "F it, let's just merge" hypothesis failed.

closing the PR. we'll find a less insane approach.

PR #3061, opened 2013-02-28
generate include.js more frequently

shane-tomlinson:
@callahad - mind reviewing?

generate include.js more frequently:
- On every compress
- Every time run_locally.js is called.

Also:
- Build include.js into the resources/static/build directory like other generated resources.
- Rename generate_include.js to create_include.js for consistency
- remove resources/static/include_js/include.js from the .gitignore list


.gitignore
```diff
@@ -12,7 +12,6 @@
 /resources/static/common/js/lib/bidbundle.js
 /resources/static/common/js/lib/gobbledygook.js
 /resources/static/dialog/views/site
-/resources/static/include_js/include.js
```
callahad, line 15: Why? Oh, because it doesn't exist anymore?

callahad:
r+, merge on monday. Awesome!

.gitignore
shane-tomlinson, line 15: Exactly. If it does, I'd like to know about it. ;)

jaredhirsch:
merging now that the train's been branched :beers:

lloyd:
nit: missing a bug with rationale!  this would help me write the changelog

jaredhirsch:
for context see shane's comments in #3056: basically include.js was only getting rebuilt when postinstall.js was run. In order to make life better when hacking locally, this pull ensures that include.js is regenerated more often.

Issue #3060, opened 2013-02-27
Persona assets need to be reduced for FX Marketplace


mattbasta:
Persona is way too heavy on bandwidth right now and is contributing to some serious performance issues in Marketplace. Some issues that need to be addressed:
- A full copy of underscore.js is being included in communication_iframe.js. It's only used 34 times, and most of those are `_.extend` and `_.defer` (which could be provided standalone).
- There's a second full copy of underscore.js that's included in communication_iframe.js because you mark it as a dependency with require.js.
- Lots of uncompressable function names (e.g.: `cancelWaitForPasswordResetComplete`, `_LEGACY_extractAssertionParamsFrom`, `_20120815_serializeCertParamsInto`, etc.) which are only used less than four times (gzip doesn't save you any/many bytes) and many used only once or not at all.
- Multiple copies of printable ASCII bytes (`ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/`) in various permutations (capital/lowercase/numbers/symbols, lowercase/capital/numbers, capital/lowercase/numbers/different symbols, etc.) in communication_iframe.js. This hurts gzip and means that the code is not DRY.
- include.js is mostly composed of workarounds for IE. A separate (IE-free) version would be appreciated for folks that don't support <IE9 (or to allow for conditional comments to use the faster version for non-IE users).

There are a lot more optimizations that could be made, but this is pretty critical right now for the performance of Marketplace (and apps that use Persona in general). This is something that we hope can be resolved before June.

jaredhirsch:
Rather than suggest specific optimizations, I'd prefer we take a more holistic view. What's the current page load time delay due to include.js? How can we measure that in a repeatable, rigorous way (maybe webpagetest.org)? Once we know the delta, I'd propose the front-end folks will figure out how best to reduce the page load impact--again, by choosing optimizations and measuring them rigorously. Sound good?

jedp:
Hi, Matt, we completely agree; the current situation is not acceptable on b2g device.  We've been working hard to make persona slimmer and faster, and we're not there yet for b2g.

Here is a list of improvements (which goes beyond asset reduction) that we are currently working on: https://id.etherpad.mozilla.org/b2g-persona-performance

[...]

PR #3062, opened 2013-03-01
Replace underscore witn tinyscore.

shane-tomlinson:
Replace underscore with a super tiny replacement that I put together that only contains the functionality that we are actually using. Saves 10KB minified to the communication_iframe, which is around an 8% savings.

issue #3060


jaredhirsch:
For methods like `each`, `keys`, and `indexOf`, what if we use the native version, if available?

jaredhirsch:
My earlier comment notwithstanding, this is fine as is & I'm going to merge in.

shane-tomlinson:
@6a68 - I considered that, and then thought "having a choice will make the code larger". I'm down to revisiting though.

Issue #3063, opened 2013-03-04
make load_gen interaction_data rate configurable, including disabling


jrgm:
I know we left this as an open question when landing @kparlante's update to send interaction_data as part of load_gen. 

However, right now in the stage environment, the configuration sends /wsapi/interaction_data to localhost:80, which then loops back again via nginx to the browserid process, but the second time without cookies, so each submit leads to 200 then 403. So, needed a way to disable this, and made it configurable from the load_gen side (which I think is what we want, so we can dial this up or down independent of the server's data_sample_rate).

jrgm:
PR-3064 was merged. Closing this issue.

PR #3064, opened 2013-03-04
[load_gen] add option --interaction-data, to allow tuning/disabling of rate of interaction_data

jrgm:
Also piggybacking a small fix to error reporting on a rare error.


jrgm:
Forgot to note. Fixes GH-3063

kparlante:
Agreed, this needs to be configurable on the load_gen side. This works for me. The default could be to leave it at zero, at least until its turned on most places (like staging).

jrgm:
Hmm, yeah could go either way on the default value, but I'm leaning a bit to it being on by default.

kparlante:
Fair enough, whatever works best for you. +r

PR #3069, opened 2013-03-04
issue 3067 - don't require caller to have curry on Function prototype

jedp:
Follow the pattern of using a `complete` function within the `internal_api` callbacks so that use of `curry` is confined to the dialog.

This is important in a context like b2g or desktop firefox, where callbacks are handed from the browser to the `internal_api`; they should not need to modify the Function prototype.

### DIFF
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -142,8 +142,12 @@
    * @param {function} callback
    */
   internal.logout = function(origin, callback) {
+    function complete(status) {
+      callback && callback(status);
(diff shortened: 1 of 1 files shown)


seanmonstar:
I'd click the merge button if the target was b2g. Or maybe we're just going straight to dev now...

jedp:
Thanks, @seanmonstar - i could be thinking of this the wrong way, but i'm proposing it to dev so it can be pulled both into b2g and branch we're using for desktop firefox for picl.  I thought this would be easier, and it seems like a valid bug for dev anyway.

PR #3072, opened 2013-03-05
Cleanup config/sauce-platorms.js to use documented names, in prep for upgrading browser version numbers

jrgm:
This may not be the right time to make this change (i.e., get the selenium tests stabilized first). On the other hand, we probably don't want to "fix" something that is only broken on a version we do not support, when the currently shipping version didn't need any help.

Anyways, in looking at this, the platform names we are using don't line up with names documented at https://saucelabs.com/docs/browsers. In particular, VISTA is not a real name, and actually means 'Windows 7' (a.k.a, 'Windows 2008' or 'Windows NT 6.1').

So this PR doesn't change any of the versions that we are currently testing against; version changes can happen after this PR. 

This just normalizes the platform names to what is documented. Version upgrades can happen later. 

You can check the net effect of these changes by dropping https://gist.github.com/jrgm/5075399 into automation-tests/tests and looking at /tmp/useragent.txt before and after and see that the UAs have not changed at all (especially that the IE UA for "VISTA" was always 'NT 6.1').

### DIFF
--- a/automation-tests/config/sauce-platforms.js
+++ b/automation-tests/config/sauce-platforms.js
@@ -10,12 +10,12 @@ const platforms = {
     version:'16'
   },
   "linux_firefox_16": {
-    platform: 'LINUX',
+    platform: 'Linux',
     browserName: 'firefox',
     version: '16'
   },
   "osx_firefox_14": {
-    platform: 'MAC',
+    platform: 'Mac 10.6',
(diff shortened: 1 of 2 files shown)


jaredhirsch:
thanks for chipping away at the selenium bits.

r+, merging in

Issue #2891, opened 2012-12-20
JS exception when using mixed case in the domain name of an email address 


jrgm:
STR
1. Sign up as a new user 'examplename@Gmail.com'. Note the upper case 'G' in gmail.
2. Go to an RP, bring up the dialog, and click 'this is not me'.
3. Enter 'examplename@gmail.com' and the password. Note the lower case 'g' in gmail.

Expected: domain names are case-insensitive, so you should be signed in

Actual: javascript exception: 'Error: Error: invalid email' in dialog.js

The back end appears to be fine with either, but haven't thoroughly vetted it.

guillaumev:
Related to this, see https://groups.google.com/forum/?fromgroups=#!topic/mozilla.dev.identity/pa3Sc0kh9Jw and the fact that the website refreshes itself automatically if you are logged in with an "upper case" domain name. jrgm > would you have a test environment for me to see if this happens only on my website (in which case it would be a bug within the Persona drupal module) or if it happens on any persona enabled site ?

Thanks,

jrgm:
We have test RPs at http://123done.org and http://myfavoritebeer.org (the former uses watch/request, while the latter uses get). Or add 'beta.' to point them to use our stage environment as the identity provider login.anosrep.org.

jaredhirsch:
fixed by #3078

jrgm:
Note for testing: this issue was specifically about the domain part of an email address.

Lazy question: @6a68 does #3078 address mixed case on the domain name?

jrgm:
Reopening.

For the original STR, I no longer get  'Error: Error: invalid email' . 

However, using login.dev.anosrep.org + http://dev.123done.org, on every **second** attempt to use the somename@Gmail.com user on dev.123done.org, I get a window.alert 'failed to verify assertion: bad signature in chain'. 

This does not happen in current stage (train-2013.03.01) or on prod (train-2013.02.15). And it appears that the original issue of 'Error: Error: invalid email' was fixed somewhere along the way. 

Anyhow, I'll bump this to 5-star, and we can discuss tomorrow.

jaredhirsch:
jrgm, what browser are you using? I cannot repro this problem on chrome or FF.

jrgm:
I don't recall. Probably FF 19. Let me try again (but I really shouldn't at the moment).

jrgm:
I remember now that the window.alert was from the FF UI. 

[...]

Issue #2866, opened 2012-12-13
Weird login failures possibly related to capitalization and/or localStorage state


callahad:
See: https://support.mozilla.org/en-US/questions/943789

When the user selects their email address an "invalid email" error shows up in the console, and nothing happens on the page.

This used to exist in `resources/static/dialog/js/misc/state.js` before the big BigTent merge. That code was:

``` javascript
handleState("email_chosen", function(msg, info) {
  var email = info.email,
      idInfo = storage.getEmail(email);

  self.email = email;

  function oncomplete() {
    complete(info.complete);
  }

  if (!idInfo) {
    throw new Error("invalid email");
  }
  // ... snip ...
}
```

One of the folks involved noted in a separate email:

> So when I try to login with "Foo@example.com" that's when I get the error.
> But when I try to login with "foo@example.com" then everything works ok!

So it sounds like we should downcase sooner? Still need to reproduce and trace where this code moved since the BigTent core merge.

shane-tomlinson:
Link to #2551

jamesshore:
+1. This affected at least one user of mine in production.

It's easy to reproduce:
1. Create a test user, (e.g., `http get http://personatestuser.org/email`. My test user was `lopez19464@personatestuser.org` with password `LPrAaF2ehXweXEtj`.)
2. Log in using Persona (e.g., via the sign-in link at https://developer.mozilla.org/en-US/ )
3. (Select 'This is not me' link if you're recognized by Persona.)
4. Enter test user's email address using different capitalization and press enter. (e.g., `LOPEZ19464@personatestuser.org`.)
5. Enter password and press enter. (e.g., `LPrAaF2ehXweXEtj`.)
6. Observe 'Error: invalid email' in console.

(Reproduced on Firefox 17.0.1 on Mac OS X 10.8.2, but also observed on IE9 & Chrome 24 in Windows 7, and on iPad 3.)

![Screen Shot 2013-01-02 at 11 27 49 AM](https://f.cloud.github.com/assets/123129/39069/8bbf292c-5512-11e2-9712-bb012e54c14d.png)

callahad:
Oh, beautiful. I'm able to reproduce an error with those steps in prod, stage, and dev. Thanks, @jamesshore!

skinny97214:
What are the issues with always downcasing?

jrgm:
[...]

Issue #2551, opened 2012-10-03
email address handling in new primary user signup is case sensitive


jrgm:
STR

Either run `node ./tests/new-user/new-user-primary-test.js` or the following:
1) Go to 123done.org and click on sign in
2) enter <someemail>@eyedee.me where 'someemail' is a string with an upper case character
3) dialog fails to dismiss and login is not completed. Error console says:
    Error: uncaught exception: invalid email (FF,Chrome)

shane-tomlinson:
Link to #2891

jrgm:
re-gift upon @6a68 - another case issue.

jaredhirsch:
fixed by #3078

jrgm:
Reopening. And bleh.

So, for the manual STR, this still fails in current dev.

The automated test passes because (see above) "use lowercase alphanum for restmail email until GH-2551 is fixed
https://github.com/mozilla/browserid/blob/dev/automation-tests/lib/restmail.js#L16.

jaredhirsch:
sadly I can repro this. digging in.

jrgm:
Just remove this line
    https://github.com/mozilla/browserid/blob/dev/automation-tests/lib/restmail.js#L16
and run 
    ./scripts/run-all.js -e train0301 -p 14 --platform win7_firefox_16 -t new-user-primary-test
and that would fail (unless it's really lucky and doesn't pick any uppercase characters).

jaredhirsch:
Not totally sure but this might be a bug in eyedee.me. It downcases user names:

https://github.com/mozilla/eyedee.me/blob/master/lib/wsapi.js#L33

but we are agnostic about hanging onto the email _as typed by the user_, so my suspicion is that we're getting an assertion for the downcased name (somedude@eyedee.me), while we are expecting the version entered by the user (Somedude@eyedee.me).

Not quite sure but out of time for the day. State.js line 340 is where you can set a debugger to catch the failure + inspect the stack.

callahad:
Check out #2680 -- sounds like the same thing.

jaredhirsch:
@callahad is right--the underlying question is, if a user supplies "alice@idp" and the primary says, actually that user is "bob@idp", which one do we trust? Seems like we at least need to give the user an "are you sure you want to sign in as bob@idp" confirmation screen.

[...]

Issue #2049, opened 2012-07-16
Can't sign in in production - email case sensitivity


csuciu:
OS: Win7
Browser: FF 13.01
                 Chrome 20.0.1132

Steps:
1. Go to mozillians.org, bugzilla.mozilla.org or developer.mozilla.org
2. Press on 'Sign In' with Persona
3. Enter an email address, press 'Next', enter the password then press 'Sign In' 

Expected:
User should be signed in with the new email

Actual:
Nothing happens.

shane-tomlinson:
@csuciu - Thanks for the bug report.  I am not having a problem in any of these three sites in Win7.  Can you possibly attach a screencast/screenshots of what you are seeing, with dev tools open if possible?

csuciu:
I got this error in the console when pressing on Sign in: 
Timestamp: 7/16/2012 12:40:03 PM
Error: [Exception... "'PopupNotifications_show: invalid browser' when calling method: [nsILoginManagerPrompter::promptToSavePassword]"  nsresult: "0x8057001e (NS_ERROR_XPC_JS_THREW_STRING)"  location: "JS frame :: resource:///components/nsLoginManager.js :: <TOP_LEVEL> :: line 971"  data: no]
Source File: resource:///components/nsLoginManager.js
Line: 971

and here is the screencast: http://www.screencast.com/t/v0ubDfWHWTc

shane-tomlinson:
@csuciu - Thanks for the screencast, it is helpful.  If you disable all of your addons, do you still see the problem?

csuciu:
Yes, I'm still seeing this with all addons disable

csuciu:
Turns out that you cannot sign in if you have an account created with uppercases (e.g TEST@example.com) and you try to sign in with the same account but using lowercases (test@example.com).  
This is the issue in my case..

callahad:
This looks like a continuation of the issue in #1350. We need to take a stance on this.

seanmonstar:
With the influx of people trying it on at Beta launch, it seems likely we would have more instances of this case. beta1-req? I'd at least for what stance we want to take.

lloyd:
@callahad @vthunder if you want this to get fixed by b1, please provide concrete guidance on approach

jaredhirsch:
fixed by #3078

PR #3078, opened 2013-03-05
Case sensitivity phase 1

jaredhirsch:
I think this looks ready, and all the tests pass.

As discussed [on the list](https://groups.google.com/forum/#!msg/mozilla.dev.identity/tvQdyLbfyH8/YSwiNkW5Kj8J) back in January (wow, it's been a while), phase 1 of fixing email normalization is to be consistent about applying the existing rules. Phase 2 is deciding if we want to change the normalization scheme, and making any changes. This patch covers phase 1.

Because our handling of email case has been inconsistent, we have a number of bugs that should be fixed by this patch: #2891, #2866, #2049, #2551, #1104.

There are a number of related bugs suggesting one approach or another, or suggesting that we explicitly define our approach, and those aren't addressed yet.


automation-tests/tests/sign-in-test.js
```diff
@@ -68,10 +69,10 @@ runner.run(module, {
   "mfb switch to the dialog when it opens": function(done) {
     browser.wwin(CSS["persona.org"].windowName, done);
   },
-  "mfb sign in with personatestuser account": function(done) {
+  "mfb sign in with allcaps personatestuser - signin is case insensitive": function(done) {
```
lloyd, line 72: updating selenium tests to excercise this, priceless.

automation-tests/tests/sign-in-test.js
```diff
@@ -81,6 +82,23 @@ runner.run(module, {
       .wtext(CSS['myfavoritebeer.org'].currentlyLoggedInEmail, function(err, text) {
         done(err || assert.equal(text, testUser.email));
       });
+  },
+
+  // Go to main site and make sure sign in is case insensitive
+  "go to main site, sign in using all caps - sign in is case insensitive": function(done) {
```
lloyd, line 88: because the account was created with upper case, shouldn't this time we try logging in with lower case to verify case insensitivity?

recall, phase one does not change what the user types in at signup time.

lib/db/json.js
```diff
@@ -138,13 +138,21 @@ exports.emailType = function(email, cb) {
 
 exports.emailInfo = function(email, cb) {
   sync();
+  // emailInfo is case insensitive, so search the database and find the
+  // "canonical" or normalized form of the email address
+  db.users.forEach(function(user) {
+    Object.keys(user.emails).forEach(function(foundEmail) {
+      if (foundEmail.toLowerCase() === email.toLowerCase()) email = foundEmail;
```
lloyd, line 145: I added this, it's inefficient as heck, but all of json.js is - doesn't matter.

resources/static/dialog/js/misc/helpers.js
```diff
@@ -138,24 +138,25 @@
...
-          complete(callback, true);
-        }
-        else {
-          self.publish("stage_email", { email: email });
-          complete(callback, true);
-        }
-      }, self.getErrorDialog(errors.addressInfo, callback));
-    }
+    // go get the normalized address and then do the rest of the checks.
+    user.addressInfo(email, function(info) {
```
lloyd, line 142: Footnote: here's a case where we possibly add a network request.  practically speaking I don't think we'll really see this hit.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -34,7 +34,13 @@ BrowserID.Modules.Authenticate = (function() {
       currentHint;
 
   function getEmail() {
-    return helpers.getAndValidateEmail(EMAIL_SELECTOR);
+    // If available, use the email sent back by addressInfo because it contains
+    // the normalized email. If not, fetch from the DOM.
+    // Email should only be fetched from the DOM if the email is new OR
+    // if the user has edited the email field after the email address was
+    // checked.
```
lloyd, line 41: interesting decision to NOT update what the user types in.  I guess they'll see the canonical version on the next screen /me goes to test that via eph instance

I think I like that decision, changing what's in the box _during_ the transition could feel clumsy...

resources/static/pages/js/forgot.js
```diff
@@ -21,24 +21,27 @@ BrowserID.forgot = (function() {
...
-        var tooltipEls = {
-          throttle: "#could_not_add",
-          invalid_email: "#not_registered",
-          primary_address: "#primary_address"
-        };
-
-        var tooltipEl = tooltipEls[info.reason];
-        if (tooltipEl) {
-          tooltip.showTooltip(tooltipEl);
+    user.addressInfo(email, function(info) {
```
lloyd, line 24: another case where this theoretically could be an extra round trip - but again, considering it's a password reset, we should hit cache.

tests/address-info-test.js
```diff
@@ -78,6 +78,116 @@ suite.addBatch({
...
+      "returns known": function(e, r) {
+        assert.isNull(e)
+        var r = JSON.parse(r.body);
+        assert.equal(r.state, "known");
+        assert.equal(r.normalizedEmail, "Test2@Capitalize.Example.Com");
+      }
+    }
+  }
+});
+
```
lloyd, line 190: travis likes it: https://travis-ci.org/mozilla/browserid/builds/5266805

lloyd:
code review r+ - now running automated tests and performing some manual testing.

lloyd:
Running 15 suite(s) on 1 platform(s), 5 at a time against https://case.personatest.org

> > > > > ...............................................<>.<>...................<>..<>...<>.............................<>..............<>...<>.......<>...........<>.............................................<....<......<.....<......<

217/217 tests passed

lloyd:
yeah, r+.  :facepunch:

jaredhirsch:
:beers:

jaredhirsch:
marked #2891, #2866, #2049, #2551, #1104 as fixed.

callahad:
WOOOOOHOOOO! Awesome work Jared! :apple:

resources/static/pages/js/forgot.js
shane-tomlinson, line 24: if the address was already checked, then the address info will already be cached.

lloyd, line 24: got it.  here at least we're not talking extra round trips then.

Issue #2339, opened 2012-08-16
raiseProvisioningFailure requires undocumented argument with mandated wording 


nmalkin:
I think I just discovered an issue with either our implementation of provisioning for IdPs, or with our documentation, or with me.

For background, when we load the provisioning page from the IdP, we ask them to check the user's identity and, if in cases where authentication is needed, [call navigator.id.raiseProvisioningFailure()](https://developer.mozilla.org/en-US/docs/Persona/Implementing_a_Persona_IdP#The_Provisioning_Page).

In our example code, we call it like this: `navigator.id.raiseProvisioningFailure('user is not authenticated as target user')`, but our documentation doesn't say anything about the argument to the function.

Now take a look [at this line](https://github.com/mozilla/browserid/blob/2143dfa7ec540e2f3d39e3b7058753eea5467afd/resources/static/common/js/user.js#L508) in BrowserID code. It seems to be that we **require** the message to be _user is not authenticated as target user_.

This is supported by some experimental IdP code that I wrote (not on Github yet), where any provisioning-failure message not matching the one above results in the login process failing with the error, "_Action: Checking Address Info._"

Is there some deep reason for this? Is it somehow part of the protocol? If so, we should definitely update our documentation. Otherwise, I see no reason at all for the second part of that _if_-statement to be there.

(On a related note, why do we even have the `error.msg` field if we never use it?)

Note that this issue affects at least one of our IdPs: eyedee.me. In two places, it [calls raiseProvisioningFailure](https://github.com/lloyd/eyedee.me/blob/master/static/browserid/provision.html#L41) with a "forbidden" message.
But hostedpersona [seems to use](https://github.com/ozten/hostedpersona/blob/master/views/provisioning.ejs#L15) the (single) correct message everywhere, so maybe @ozten knows something I don't.

ozten:
+1 this bite me too

ozten:
https://github.com/mozilla/browserid/issues/1207

nmalkin:
(Repeating for internal linking:)

This issue affected [someone on the mailing list](http://groups.google.com/group/mozilla.dev.identity/browse_thread/thread/ea00cf0bac16b1ca/3e0f360e37a9a24b?lnk=gst&q=raiseprovisioningfailure#3e0f360e37a9a24b), was filed as #1182, where @ozten [discovered](https://github.com/mozilla/browserid/issues/1182#issuecomment-4146610) what I just did, and filed #1207, which is open but has had no discussion.

[...]

PR #3081, opened 2013-03-07
don't require that an IdP supply a specific text string to indicate that the user is not authenticated

lloyd:
this is for issue #2339.  As part of the fix I wanted to ensure that errors not originating from the primary got shown under extended error information in the dialog.

Note to reviewer: I am not convinced my method of propagating error information is How We Do It.  Please review.

### DIFF
--- a/example/primary/provision.html
+++ b/example/primary/provision.html
@@ -19,7 +19,7 @@
     $.get('/api/whoami')
       .success(function(who) {
         if (user != who) {
-          return navigator.id.raiseProvisioningFailure('user is not authenticated as target user');
+          return navigator.id.raiseProvisioningFailure('not authed');
         }
 
         // Awesome!  The user is authenticated as who we want to provision.  let's
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -449,7 +449,6 @@ BrowserID.User = (function() {
      * @param {function} [onFailure] - called on failure
      */
     provisionPrimaryUser: function(email, info, onComplete, onFailure) {
-
       User.primaryUserAuthenticationInfo(email, info, function(authInfo) {
         if (authInfo.authenticated) {
           persistEmailKeypair(email, authInfo.keypair, authInfo.cert,
@@ -529,14 +528,16 @@ BrowserID.User = (function() {
           complete(userInfo);
         },
         function(error) {
-          if (error.code === "primaryError" && error.msg === "user is not authenticated as target user") {
+          // issue #2339 - in case an error is raised we don't care
+          // about the specific error code.
+          if (error.code === "primaryError") {
             var userInfo = _.extend({
               authenticated: false
             }, info);
             complete(userInfo);
           }
           else {
-            onFailure(info);
+            onFailure($.extend(info, { detailedError: error }));
           }
         }
       );
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -537,7 +537,7 @@ BrowserID.User = (function() {
             complete(userInfo);
           }
           else {
-            onFailure($.extend(info, { detailedError: error }));
+            onFailure($.extend(info, { action: { message: error }}));
           }
         }
       );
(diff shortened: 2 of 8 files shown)


example/primary/provision.html
```diff
@@ -19,7 +19,7 @@
     $.get('/api/whoami')
       .success(function(who) {
         if (user != who) {
-          return navigator.id.raiseProvisioningFailure('user is not authenticated as target user');
+          return navigator.id.raiseProvisioningFailure('not authed');
```
lloyd, line 22: a tweak to the built-in primary to change the not-authed string

resources/static/common/js/user.js
```diff
@@ -529,14 +528,16 @@ BrowserID.User = (function() {
           complete(userInfo);
         },
         function(error) {
-          if (error.code === "primaryError" && error.msg === "user is not authenticated as target user") {
+          // issue #2339 - in case an error is raised we don't care
+          // about the specific error code.
+          if (error.code === "primaryError") {
```
lloyd, line 533: `.code` of primaryError implies we actually received a response from the primary - that their javascript loads ok and calls the functions we provide.  As long as that is the case we assume that the user is not authenticated.

Recall the flow here:
1. dialog load
2. 1st provisioning attempt
3. if fails, direct to authentication screen
4. upon redirect back, 2nd provisioning attempt.  

So this change means if `.raiseProvisioningFailure` is called on the first attempt, we'll always go to authentication.  Before, this was only if the IdP provided the exact string `user is not authenticated as target user`

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -78,7 +78,13 @@ BrowserID.Modules.Authenticate = (function() {
     else {
       showHint("addressInfo");
       user.addressInfo(email, onAddressInfo,
-        self.getErrorDialog(errors.addressInfo));
+                       function(info) {
+                         var ai = errors.addressInfo;
+                         if (info.detailedError) {
+                           ai = $.extend(ai, { message: info.detailedError });
+                         }
+                         self.getErrorDialog(ai)();
```
seanmonstar: this doesn't pass the error as `lowLevelInfo`, and so we may miss out on network info?

lloyd: I think I just need to pass `info` along, as that would be what used to happen.  right?  `self.getErrorDialog(ai)(info)`?

shane-tomlinson:
@lloyd - I have been unable to get an error message to appear no matter what I do.

in `example/primary/provision.html`, I added the following line on line 17:

```
    return navigator.id.raiseProvisioningFailure('ghetto failure');
```

I expected this to show an error screen after authenticating with the example IdP, but so far have been unable to make this occur.

resources/static/dialog/js/modules/authenticate.js
shane-tomlinson: I think you can revert this change and simplify displaying the message by changing detailedError=>message in https://github.com/mozilla/browserid/pull/3081/files#L1R540

lloyd:
@shane-tomlinson so with this change _never_ show an error to the user when the IdP calls `.raiseProvisioningFailure()`

Why?  Because if the primary is having trouble provisioning, _they_ will explain the problem to the user while the auth page is showing.

In cases where the primary is more hard core broken (any case where `fail()` is called here with an error other than `primaryError` https://github.com/mozilla/browserid/blob/dev/resources/static/common/js/provisioning.js) this means that the primary is too broken to catch their own errors, and we _will_ display an error msg to the user.

make sense?

resources/static/dialog/js/modules/authenticate.js
shane-tomlinson: I am wrong, my suggestion does not work. I need to give this some though because this seems like a common enough scenario where it should be taken care of automatically.

lloyd: Yeah, seems like we should have an easy way to surface specific errors to make user error reports actionable.

Should that block this pr?

-- lloyd (thumb-typing)

shane-tomlinson: I have a minimal change (22 insertions, 20 deletions - with tests) that makes defining action.message much easier.

shane-tomlinson:
One final check we should make - ensure that any text that comes from the IdP is escaped before they XSS users.

shane-tomlinson:
r+ from my side, a review of the changes I made would be great.

shane-tomlinson:
+1. I dig it.

PR #3085, opened 2013-03-11
Run `npm install` in the automation-tests directory in scripts/postinstall.js

### LINKED ISSUE #3084: automation-tests dependencies are not installed automatically
This leads to confusion for people like myself and @jbonacci

shane-tomlinson:
fixes #3084

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -24,3 +24,12 @@ function node(script) {
 }
 
 node('./generate_ephemeral_keys.js');
+
+console.log(">>> Installing automation-tests dependencies");
+// install automation-test dependencies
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@6a68 - mind reviewing this?

jaredhirsch:
works for me, nicely done.

lloyd:
so outcome of this is we're installing more software in more places that we don't want to.  on deployer, on aws instances we spin up.  rather than understanding why this installation of addition stuff is so flakey and debugging it, I wonder if we can restate the problem that this PR was designed to solve and figure out a creative solution that gets us all the good without the bad.

Issue #3095, opened 2013-03-13
automation run-all.js: tell the user when they pick an unknown --platform


jrgm:
I've made this mistake (typo, forgot) and started run-all.js with a `--platform` that is not specified in automation-tests/config/sauce_platforms.js.

jrgm:
So, no, doesn't need expanding. Generally, this defaults sanely or whines about the wrong thing. Just this (optional) flag, would just say it was going to run and then silently exit without explaining "hey, I don't know that value for --platform".

`./scripts/run-all.js --help` will show you the options. `./scripts/run-all.js --list-platforms` shows the platforms. Mostly stable on win7_chrome and {win7,linux}_firefox_16. YMMV.

jrgm:
PR-3096 merged. Closing.

PR #3096, opened 2013-03-13
automation run-all.js: tell the user when they pick an unknown --platform

jrgm:
So that if I typo, forget or switch branches, it tells me what I got wrong.


jrgm:
Fixes GH-3095

lloyd:
R+

PR #3101, opened 2013-03-14
Work with 3rd party cookies disabled on chrome (issues 2340+2999)

jaredhirsch:
:warning: work in progress, don't touch that green button.

taking a break for a little sleep + thought I'd show @shane-tomlinson what I've got so far

### DIFF
--- a/bin/browserid
+++ b/bin/browserid
@@ -25,6 +25,7 @@ heartbeat = require('../lib/heartbeat.js'),
 logger = require('../lib/logging.js').logger,
 shutdown = require('../lib/shutdown'),
 primary = require('../lib/primary'),
+addP3PHeader = require('../lib/p3p.js'),
 forward = require('../lib/http_forward.js'),
 toobusy = require('../lib/busy_middleware.js');
 
@@ -94,6 +95,10 @@ if (config.get('scheme') == 'https') {
     next();
     });
 }
+
+// Add P3P headers to please IE8
+app.use(addP3PHeader);
+
 
 // #4 - prevent framing of everything.  content underneath that needs to be
 // framed must explicitly remove the x-frame-options
--- a/lib/p3p.js
+++ b/lib/p3p.js
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var policy = 'CP="This is not a P3P policy, but Mozilla deeply cares about ' +
+             'your privacy. See http://www.mozilla.org/persona/privacy-policy ' +
+             'for more."';
+
+// add 'P3P' headers so that IE8, with default security settings, will allow
+// us to set third-party cookies. See #2340
+module.exports = function(req, res, next) {
+  res.on('header', function() {
+    res.setHeader('P3P', policy);
+  });
+  next();
+};
--- a/resources/static/common/js/network.js
+++ b/resources/static/common/js/network.js
@@ -684,7 +684,7 @@ BrowserID.Network = (function() {
      * Check if the user's cookies are enabled
      * @method cookiesEnabled
      */
-    cookiesEnabled: function(onComplete, onFailure) {
+    cookiesEnabled: function(onComplete) {
       var enabled;
       try {
         // NOTE - The Android 3.3 and 4.0 default browsers will still pass
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -24,6 +24,21 @@
   //  we may need to re-visit this)
   network.cookiesEnabledOverride = true;
 
+  // 1. try network.cookiesEnabled check.
+  // 2. see if we can read the cookie set by the dialog.
+  // if both checks fail, third-party cookie functionality is disabled,
+  // and we want to fall back gracefully by ceding session management
+  // to the RP. this means that a user might be logged out of persona.org
+  // but still have active logins across the web, if they have third-
+  // party cookies disabled in their browser.
+  // TODO if the user has cleared cookies/cache, but 3p cookies are accessible.
+  //      this check will always fail. Will that actually break login?
+  network.cookiesEnabled(function(err) {
+    if (err && document.cookie.indexOf("third-party-readable") == -1) {
+      // iframe falls back + agrees with whatever the RP page says.
+    }
+  });
+
   var chan = Channel.build({
     window: window.parent,
     origin: "*",
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -15,14 +15,16 @@
 
   network.init();
 
-  // Do not check to see if cookies are supported in the iframe.  Just
-  // optimistically try to work by running network requests.  There are
-  // cases (especially in IE) where our checks will fail but our actual
-  // requests will not.  issue #2183
-  // (NOTE: if we want to try to improve failure modes for users with
-  //  a "disable 3rd party cookies"-like preference set in their browser,
-  //  we may need to re-visit this)
-  network.cookiesEnabledOverride = true;
+  // TODO remove/update this bit:
+    // Do not check to see if cookies are supported in the iframe.  Just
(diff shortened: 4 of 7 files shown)


jaredhirsch:
@lloyd ping

bin/browserid
```diff
@@ -95,6 +96,10 @@ if (config.get('scheme') == 'https') {
     });
 }
 
+// Add P3P headers to please IE8
+app.use(addP3PHeader);
```
shane-tomlinson, line 100: Will the proxy forward this header?

lib/p3p.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var policy = 'CP="This is not a P3P policy, but Mozilla deeply cares about ' +
```
shane-tomlinson, line 5: meh. Is there some combination of real values we could use? This makes me very uncomfortable.

resources/static/communication_iframe/start.js
```diff
@@ -15,14 +15,30 @@
...
-  // Do not check to see if cookies are supported in the iframe.  Just
-  // optimistically try to work by running network requests.  There are
-  // cases (especially in IE) where our checks will fail but our actual
-  // requests will not.  issue #2183
-  // (NOTE: if we want to try to improve failure modes for users with
-  //  a "disable 3rd party cookies"-like preference set in their browser,
-  //  we may need to re-visit this)
-  network.cookiesEnabledOverride = true;
+  // TODO remove/update this bit:
+    // Do not check to see if cookies are supported in the iframe.  Just
```
shane-tomlinson, line 19: This comment looks like it needs updated.

resources/static/common/js/network.js
```diff
@@ -684,7 +684,7 @@ BrowserID.Network = (function() {
      * Check if the user's cookies are enabled
      * @method cookiesEnabled
      */
-    cookiesEnabled: function(onComplete, onFailure) {
```
shane-tomlinson, line 687: We have no other precedent for using only onComplete as the error return. We should keep consistency.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -171,6 +171,11 @@ BrowserID.Modules.Dialog = (function() {
         startChannel.call(self);
       }
 
+      if (document.cookie.indexOf('third-party-readable') === -1) {
```
shane-tomlinson, line 174: Can this be pushed into network.js with a clear note about why and when it can be used?

shane-tomlinson:
I have several concerns.

First, I am confused about the overall goal. I see what you are trying to do, but calling "onmatch" for disabled 3rd party cookies is a departure from semantics. onmatch indicates "we agree with you." Instead of "we agree", we should be indicating "we have no idea". If onmatch is that mechanism, we should still notify the RP somehow. @callahad - can you weigh in?

How will this affect first time visitors to sites that use .watch? Since an attempt will never have been made to set the cookie, it won't yet exist. Will onmatch be called?

lloyd:
**high level goal**:  In the case of google chrome (or any browser) with third party cookies disabled completely, this means that the communication_iframe is an island, and cannot have authenticated communication with our servers.

What is the ideal behavior in this scenario?  [About a year ago](http://lloyd.io/doing-more-with-browserid) we concluded it was to do the best we can, and not prevent the user from logging in, nor require them to change their browser settings.  Concretely, I believe optimal behavior is:
- disable session management features which require authenticated communication from the iframe
  1. No `logout` will occur from the iframe
  2. No automatic login will occur from the iframe

To make this work right, if we find that we cannot have auth'd comm with our servers from the iframe we should do two things:
1. Always agree with the site's belief of user state (disable session management)
2. issue assertions with a short suggested session duration

2 can happen later and is tricky.

If you agree with those high level goals, the question becomes, how do we optimally determine that we cannot have authenticated communication with the server in the iframe?  Specifically, how do we disambiguate between the case where the user is not auth'd to persona, and the case where they are, but our cookies are being blocked.

lloyd:
The one thing I don't like about this PR is that the cookie check is not sufficiently granular.  We don't care if we can set cookies from javascript, we just care if we can read them.  What if we add another cookie to `session_context` 'third_party_cookies_blocked: false' - can we significantly simplify the code and get rid of the P3P header requirement this way?

shane-tomlinson:
@lloyd - @seanmonstar and I had that same suggestion many moons ago. We could simplify the cookieCheck in network.js by reading the same cookie. We would have to have another check to see if we can write to localStorage though.

shane-tomlinson:
> To make this work right, if we find that we cannot have auth'd comm with our servers from the iframe we should do two things:
> 1. Always agree with the site's belief of user state (disable session management)

A blanket agreement with the RP could be unexpected if the RP relies on Persona to take care of session management. Some sort of notification that we cannot determine the users status would allow RPs to decide whether they should continue with a user they believe to be authenticated. If we don't know, the notification would allow the RP to log the user out themselves and force the user to open the dialog again.

lloyd:
@shane-tomlinson r+ - I think this was just an oversight.  pending travis I'll merge.

lib/p3p.js
callahad, line 5: I've gone back and forth on that with Jared. I'm alright with this approach since:
1. Legal has reviewed and approved this strategy.
2. P3P is a huge mess of XML. Crafting a comprehensive policy is time prohibitive.
3. As of one year ago, [5% of the top 10,000 web sites (Alexa) set invalid P3P headers](http://www.nikcub.com/posts/facebook-also-doesnt-honor-p3p).
4. Most importantly, P3P is dead. This policy is only interpreted by IE8.

lloyd, line 5: I'm ok with the P3P hack, but not yet convinced that this is the best way to solve the bug.  See below where I ask if just sending a new cookie from session_context that's a forever cookie simplifies and improves implementation.

jaredhirsch, line 5: what @callahad said, except also, I'm just the messenger for @benadida, who got this approach approved by legal.

resources/static/communication_iframe/start.js
jaredhirsch, line 19: heh, yeah. work-in-progress PR :-)

resources/static/common/js/network.js
jaredhirsch, line 687: That's fine, but the existing function doesn't call onFailure if it errors out.

So I guess the right fix would be to update cookiesEnabled to call onFailure, and update callers to pass an errback?

jaredhirsch:
OK, so it seems like a better cookie check would
- set an extra cookie in `session_context`
- check for that
- attempt to write/read localstorage

@shane-tomlinson I agree, it would be great to tell RPs that Persona session management has been disabled by user security settings. What's the right way to do that? I'm not sure whether overloading the onmatch callback or adding another callback makes for the smallest, simplest API possible.

shane-tomlinson:
@lloyd and I started a new branch that we are working in. 2999newcookie. This branch may be mothballed and portions stolen depending on the outcome of that branch.

jaredhirsch:
Ah, ok. In that case, how about I close this PR and have a look there.

shane-tomlinson:
@6a68, @lloyd - for the record, I don't really care about the form of the notification as long as it is closely related in time to whatever callback is called. Whether it is onmatch with a flag passed to the callback that says "we aren't sure" or a new callback entirely, both serve the same purpose. I want the RP to be able to make an informed decision on how to treat the session. An RP could force a user to re-authenticate for instance. 

I agree that falling back to .get is an better than not allowing users to sign in at all, but the semantics of .watch and .get are not interchangeable. .get means, definitively and without question, "the RP takes care of all session management." .watch tells a site that they can rely on Persona to tell them when a user has logged out. This proposal removes that promise. onmatch will no longer be a definitive source of truth. This is why I want a notification.

An assertion that suggests a short lived session is a good start, but it is too far from the onmatch callback. The two are too decoupled.

The proposal, as it stands, gets us closer to allowing users with 3rd party cookies disabled to sign in. This is a huge win. I feel that it needs a tiny bit more to allow RPs to make the choice that is right for them.

Issue #3091, opened 2013-03-12
uncaught exception in communication_iframe.js (possibly bigtent related)


jrgm:
I don't have the particular STR at the moment, but I wound up in a state where loading beta.123done.org, the signin button would not show and the spinner just kept spinning. 

In the error console I get 

```
Timestamp: 3/12/13 12:18:24 PM
Error: TypeError: i is not a function
Source File: https://static.login.anosrep.org/v/2172a9e451/production/communication_iframe.js
Line: 4
```

I'll leave this browser in this state for a while if there any more debug info I can pull out of it let me know.

karlht:
Calling @edmoz, this is the same problem you ran into, yes?  Did we determine that it was does to a typo in the hostname?

edwindotcom:
I've reproduced this two separate times deleting anosrep.org cookies fixes this. but I don't have specific STR right now.

ozten:
Playing around, I haven't reproduced. Looking forward to STR.

jrgm:
So, I went back to the browserid I had set aside and it was no longer having this issue (possibly as some session somewhere had expired). So nothing left to pull out of debugger.

lloyd:
@lloyd to theorize with an eye toward indicators of this happening in production

karlht:
Just tickled this again ... trying to pare it down to a minimal case.  Theory so far:
1. The user has interacted with BigTent before (i.e. They have a history with persona with at least one successful BigTent login with a yahoo.com address.)
2. The user is not currently logged in.

But I can't get it to repro consistently.  More as I get it.

lloyd:
browser?

lloyd:
so this is train-2013.03.01, right?

edwindotcom:
I'm on Mac 10.8, Firefox Beta (20.0).  I believe this was train-2013.03.01.  Karl mentioned he repro'd one time and is working on steps to repro.

karlht:
I'm on beta (FF 20.0b5).  And I'm looking at the problem right now -- I
just can't seem to get the steps to repro from a clean browser.

lloyd:
STR for a problem that may be the same, may not, but exhibits the same symptoms:
1. create a persona acct with a single yahoo.com email via beta.123done.org
2. on the login.anosrep.org domain, localStorage.removeItem('emails')
3. remove all cookies on beta.123done.org
4. reload beta.123done.org

lloyd:
```
[...]

PR #3106, opened 2013-03-15
fix certificate provisioning for primary email addresses from within com...

lloyd:
...munication iframe - issue #3091


lib/static_resources.js
```diff
@@ -159,6 +159,7 @@ exports.resources = {
     '/common/js/xhr_transport.js',
     '/common/js/xhr.js',
     '/common/js/network.js',
+    '/common/js/provisioning.js',
```
ozten, line 162: Nice find. Good call.

PR #3108, opened 2013-03-18
Issue 3104 rp api kpi

### LINKED ISSUE #3104: rp_api missing from kpi blob
I don't see it when testing against my ephemeral browserid instance. I've been testing by logging into/out of the 123done site.

@skinny97214 is interested in knowing how many RPs are using .get vs .watch

shane-tomlinson:
@kparlante, @seanmonstar - wanna review this?

I broke it into two commits. The first is code cleanup. The second fixes the bug.

fixes #3104


kparlante:
Excellent. I tested this and can now see rp_api showing up in the collector. +r

PR #3112, opened 2013-03-18
Fix the IE8 404 problem for background images.

### LINKED ISSUE #1736: (ie8) 404 requests for background-image
I was looking at logs on stage for a different reason, but noticed these 404 responses:

GET /v/d99a08c589/i/grain.png),%20linear-gradient(top,%20rgba(113,%20126,%20137,%200),%20rgba(113,%20126,%20137,%200.2) HTTP/1.0 404

The UA for those requests is always IE8. The 404's are not new to the current train.

shane-tomlinson:
@6a68 - mind a bit of review?

Add a #iefix to all background-image: url(''s that contain two parts. fixes IE8 404'ing when requesting these images.

For a bit of background, see http://www.fontspring.com/blog/the-new-bulletproof-font-face-syntax.

They use ?#iefix syntax, but the ? is unnecessary.

fixes #1736

### DIFF
--- a/resources/static/common/css/style.css
+++ b/resources/static/common/css/style.css
@@ -264,12 +264,12 @@ button::-moz-focus-inner, .button::-moz-focus-inner {
     padding: 6px 45px 7px 10px;
     background-color: #4eb5e5;
     background-image: url("/common/i/button-arrow.png");
-    background-image: url("/common/i/button-arrow.png"), -webkit-gradient(linear, left top, left bottom, from(#4eb5e5), to(#3196cf));
-    background-image: url("/common/i/button-arrow.png"), -webkit-linear-gradient(top, #4eb5e5, #3196cf);
-    background-image: url("/common/i/button-arrow.png"),    -moz-linear-gradient(top, #4eb5e5, #3196cf);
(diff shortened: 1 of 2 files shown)


jaredhirsch:
@shane-tomlinson Is there a visual component to this fix, or is it just fixing server 404s as mentioned in #1736?

I don't see any visual differences in the buttons on IE8 when looking at login.dev.anosrep.org, for instance.

jrgm:
This was originally about 404 in the logs, which were tedious but tolerable. I doubt there was any visual effect, given that this is linear-gradient in ie8, ...

@shane-tomlinson thanks for cleaning these out!

shane-tomlinson:
@6a68 - Yeah, @jrgm is correct. This is purely to keep IE8 from making the extra requests which cause 404's. There should be 0 user facing changes. If you use Fiddler or the likes, the different between pre and post will be fewer requests made.

jaredhirsch:
Cool! Thanks for the info, guys.

Here's a fiddler2 screen cap, showing the 404 in the current site:

Screen Shot 2013-03-20 at 5 21 37 PM

And here's a screen cap of an awsbox running this branch, showing no 404s:

Screen Shot 2013-03-20 at 5 35 02 PM

so, looks solid. merging :surfer:

PR #3113, opened 2013-03-18
Use the newest connect-fonts-opensans which correctly uses the PostScript name of every font.

shane-tomlinson:
In OSX, local fonts are searched for by their Postscript names instead of their True Type Font names.


jaredhirsch:
Merging because this is just a dependency, but: I couldn't see any differences in the CSS `@font-face` declaration when I opened the dialog from the local RP. Am I missing something?

At any rate, the files really were served up, dunno if from my local system or not. Merging. Ping me if you want me to take a closer look.

PR #3115, opened 2013-03-19
Fix the escaped HTML in the cookies_disabled screen.

### LINKED ISSUE #2979: "Cookies Disabled" screen shows escaped HTML when invoked from /sign_in
STR

1) Load up Safari on Windows XP or OSX
2) Visit 123done with cookies enabled.
3) Disable Cookies.
4) Click "Sign in with Persona"
5) Meh.

cookies disabled safari

shane-tomlinson:
fixes #2979


jaredhirsch:
works locally for me. not the cleanest approach, but I can't think of anything better, and this is basically a one-off.

Issue #2308, opened 2012-08-15
Android - /signin page - cookie check does not fail until after user types password


shane-tomlinson:
STR

1) On an android device, disable browser cookies.
2) Visit the main site /signin page
3) Enter unknown secondary address
4) Enter password
5) Enter verify password
6) Submit form
7) "Persona requires cookies to remember you" is finally shown.

This has been known about for a long time but not documented.

More user friendly would be if the cookie check failed as soon as the user visited the page like it does in other browsers.

PR #3116, opened 2013-03-19
add a js accesible cookie set by session_context to help users log in when 3rd party cookies are disabled...

shane-tomlinson:
I am starting a PR to get discussion started.

The new, tentative approach is:

1) Set a cookie on the server that can be checked on the client to see if cookies are enabled.
2) Take the correct action in the communication_iframe.

I did initial work on the client side and #2308 kept manifesting. If we are going to start transferring JS readable cookies, it may be possible to take care of both of these at once.

I have started testing an alternate approach which is to set the cookie in the client before session_context, have the server attempt to read the cookie value and send a field in its response: "cookies_enabled". This works so far in all browsers but led me down a deep dark black hole of code cleanup. I am trying to come up with a minimal changeset to make this work.

The downside to trying to set a cookie in the client is P3P on IE8. Ugh. 

Once we know definitively whether cookies can be used in the current context, we can take the appropriate action with the .watch API.


shane-tomlinson:
Adding @lloyd's comments from #3101:

> high level goal: In the case of google chrome (or any browser) with third party cookies disabled completely, this means that the communication_iframe is an island, and cannot have authenticated communication with our servers.
> 
> What is the ideal behavior in this scenario? About a year ago we concluded it was to do the best we can, and not prevent the user from logging in, nor require them to change their browser settings. Concretely, I believe optimal behavior is:
> 
>    disable session management features which require authenticated communication from the iframe
>        No logout will occur from the iframe
>        No automatic login will occur from the iframe
> 
> To make this work right, if we find that we cannot have auth'd comm with our servers from the iframe we should do two things:
> 1. Always agree with the site's belief of user state (disable session management)
> 2. issue assertions with a short suggested session duration
> 
> 2 can happen later and is tricky.
> 
> If you agree with those high level goals, the question becomes, how do we optimally determine that we cannot have authenticated communication with the server in the iframe? Specifically, how do we disambiguate between the case where the user is not auth'd to persona, and the case where they are, but our cookies are being blocked.

shane-tomlinson:
Adding my own comments from #3101:

> @6a68, @lloyd - for the record, I don't really care about the form of the notification as long as it is closely related in time to whatever callback is called. Whether it is onmatch with a flag passed to the callback that says "we aren't sure" or a new callback entirely, both serve the same purpose. I want the RP to be able to make an informed decision on how to treat the session. An RP could force a user to re-authenticate for instance.
> 
> I agree that falling back to .get is an better than not allowing users to sign in at all, but the semantics of .watch and .get are not interchangeable. .get means, definitively and without question, "the RP takes care of all session management." .watch tells a site that they can rely on Persona to tell them when a user has logged out. This proposal removes that promise. onmatch will no longer be a definitive source of truth. This is why I want a notification.
> 
> An assertion that suggests a short lived session is a good start, but it is too far from the onmatch callback. The two are too decoupled.
> 
> The proposal, as it stands, gets us closer to allowing users with 3rd party cookies disabled to sign in. This is a huge win. I feel that it needs a tiny bit more to allow RPs to make the choice that is right for them.

shane-tomlinson:
Could the solution to #2308 and the IE8 P3P problem be a hybrid solution? I do not know what the form of that solution would look like yet, or even whether it is possible.

(1/3 baked thought)
- First, try to set a cookie from the client. This will fail in IE8 but succeed in Android.
- Then, on the server, try to read the cookie and send back a response to session_context whether it can see the cookie. Also have it set a cookie.
- Finally, on the client, read the "can we see your cookie" value sent by the server. If it is "no", then try to read the cookie value set by the server.

Unfortunately, Android 3.2 and 4.0 with Cookies disabled will still see the cookie sent from the server.

shane-tomlinson:
A new approach fixes #2308 while allowing us to continue on with our goal.

With #2308 in mind, we need a new approach to know if cookies are enabled in the current context. Android 3.2 and 4.0 continue to report that cookies are enabled because they can read the "can_set_cookies" cookie even if cookies are disabled. Even though Android can read the cookie when cookies are disabled, the cookies are never sent to the backend. This means we can use the backend to reliably tell us whether cookies are enabled or disabled for the current context.

If session_context checks for the existence of a set cookie, then it knows definitively whether cookies are enabled in the current context and can send a response back to the front end.

I made a revised attempt at this. Every time a top level page is requested, the server responds with a set-cookie header that contains the "can_set_cookies" cookie. The cookie must have an expiration or else IE8 treats it as a session cookie and relaxes its rules.  Setting a cookie on the backend like this will not affect caching or etagging and can be safely done. 

When session_context is called, it sends the cookie back to the server. wsapi/session_context checks for the existence of the "can_set_cookie" cookie and responds with "cookie: true" if the cookie is read and "cookies: false" if not.

I will experiment with setting this cookie from JS to see if the same result occurs.

In the communication iframe, before attempting to get the assertion, I check whether cookies are enabled. If cookies are not enabled, instead of calling `onmatch`, I am calling `onready`. This still allows an RP to continue, it has a signal that cookies are disabled because none of onlogin, onlogout, nor onmatch are called, but can still continue. onlogout is no longer called if cookies are disabled.

shane-tomlinson:
This brings up new problems in Chrome with 3p cookies disabled.:

1) navigator.id.logout does not work because there is no cookie that can be used to sign the user out.
2) the provisioning iframe is going to fail, again, because there are no cookies that can be used to authenticate with the IdP.

shane-tomlinson:
The solution as it stands sets the cookie from the client before the request to /wsapi/session_context is made. The server then checks for the cookie. If it finds it, we know cookies are enabled in the context making the call to session_context. If the cookie is not found, cookies are disabled. A "cookies" field is added to the response, which is then checked on the client. On the client, if response.cookies is true, we know definitively, even on Android 3.2 and 4.0, whether cookies are enabled in the current context.

PHEW.

jaredhirsch:
Also true on Android 2.2 and 4.2.1 stock browser

shane-tomlinson:
I have fixed the logout behavior. navigator.id.logout can be called even if 3rd party cookies are disabled. .logout will then call the site's .onlogout callback, allowing them to sign the user out of their site.

jaredhirsch:
just to keep things tied together: this pull request should fix #2999. cc @toolness

toolness:
Awesome, this is great!! Thanks, everyone, for helping resolve this issue. :)

shane-tomlinson:
@6a68, @lloyd - I'd love to get a review of this. 123done.org/Chrome act oddly when 3rd party cookies are disabled, I am trying to find out why it acts differently from the other browsers.

jaredhirsch:
I can take a look

jaredhirsch:
@shane-tomlinson It looks like, when 3rd party cookies are disabled, login completes but nothing is written to login.persona.org localstorage in current dev. The 2999newcookie branch, on the other hand, does the right thing:
1. Go to 2999newcookie.123done.org in a clean Chrome profile with third-party cookies disabled
2. sign in
3. reload the page

After step 3, on regular 123done.org, you'll see the 'sign in with persona' button again. On the branch, after reload, you correctly appear to be logged in.

I think this might be good to merge? Unless you have other questions about Chrome's behavior?

lloyd:
r- - it looks like all interaction with primary IdPs is broken by this PR: http://pastebin.mozilla.org/2293089

seanmonstar:
And yet the tests pass. Looks like we need a test that this fails!

lloyd:
@seanmonstar once I figure out root cause I'll report it and we'll see if we could find something like this in a unit test (as opposed to selenium / browser based)

lloyd:
So let's find the root cause of the failure, rebase against current dev, and target a merge on monday.

This slippage is my fault for not jumping in with review earlier.  :poop:

jaredhirsch:
Closing. See #3230 for the next chapter in the saga.

shane-tomlinson:
I am re-opening this branch since it has context with it.

shane-tomlinson:
rebased

shane-tomlinson:
@lloyd, @6a68, @seanmonstar - I suspect the test failures are a red herring due to Sauce hitting its max number of concurrent browsers. I ran the Selenium tests often while developing, and am running them now without a problem:

```
➜  automation-tests git:(2999newcookie) ✗ ./scripts/run-all.js -e p3-cookie -p 5
adding add-primary-to-primary.js
adding add-primary-to-secondary.js
adding cancel-account.js
adding change-password-test.js
adding frontend-qunit-test.js
adding health-check-tests.js
adding broken-primary.js
adding primary-shuts-down-single-email.js
adding primary-starts-up-single-email.js
adding new-user-primary-test.js
adding new-user-secondary-test.js
ignoring public-terminals.js
adding remove-email.js
adding reset-password-test.js
adding returning-user.js
adding sign-in-test.js
Running 15 suite(s) on 1 platform(s), 5 at a time against https://p3-cookie.personatest.org
>>>>>.........................<>.........<>....<>..............................<><>....................<>.......<>..................................<>..........<>.....<>................................<....................<...<.<..<

217/217 tests passed
```

If Sauce hits its max number of concurrent browsers and cannot open another browser, it errors in the manner seen in the pastebin.

Run the tests with -p 5.

I already have an ephemeral instance set up - p3-cookie.personatest.org

lloyd:
This wasn't a sauce issue, I manually tested after seeing the sauce errors.  VM was 2999newcookie.  I'll retest and re-review the rebased branch.

-- lloyd (thumb-typing)

jaredhirsch:
@shane-tomlinson I cannot sign into p3-cookie.123done.org using an eyedee.me address :-\

shane-tomlinson:
@6a68 - is this with 3rd party cookies enabled or disabled? Primaries in general are broken with 3rd party cookies disabled - see #3169

jaredhirsch:
@shane-tomlinson: ooh, you're right! I did have third-party cookies disabled. This works fine with third-party cookies enabled.

@lloyd I'd say this is good to merge in. can you still repro any problems?

resources/static/communication_iframe/start.js
```diff
@@ -50,28 +41,38 @@
...
-        // when the site already has the user logged in and does not want
-        // the resources or cost required to generate an assertion
-        if (assertion) chan.notify({ method: 'login', params: assertion });
-        loggedInUser = email;
-      } else if (loggedInUser !== null) {
-        // only send logout events when loggedInUser is not null, which is an
-        // indicator that the site thinks the user is logged out
+    network.cookiesEnabled(function(enabled) {
+      if (!enabled) {
+        // cookies are disabled, call onready and do nothing more.
```
lloyd, line 46: isn't this a contract violation?  If you pass undefined, we claim we'll call one of either onlogin or onlogout if you supply `undefined` to watch.

callahad, line 46: Where is that contract specified? I'd be down with fixing that instead.

shane-tomlinson, line 46: I agree with @callahad's suggestion. I think in this case, we should update the doc, for the reasons stated in my first comment here.

lloyd, line 46: @callahad convinced me.  I like saying nothing when we don't know better, and the only fallout is that websites that pass `undefined` will not automatically gain support for chrome users with 3rd party cookies disabled.  This feels right.

lloyd:
r+

jaredhirsch:
\o/

PR #3121, opened 2013-03-20
B2g picl merge

shane-tomlinson:
DO NOT MERGE! This branch is to do review and generate discussion.


bin/keysigner
```diff
@@ -90,14 +92,24 @@ try {
 
 
 // and our single function
-app.post('/wsapi/cert_key', validate({ 'email': 'email', 'pubkey': 'pubkey', 'ephemeral': 'boolean' }), function(req, resp) {
+app.post('/wsapi/cert_key', validate({ 'email': 'email', 'pubkey': 'pubkey', 'ephemeral': 'boolean', 'unverified': { type: 'boolean', required: false }, 'forceIssuer': {type: 'hostname', required: false}}), function(req, resp) {
   var startTime = new Date();
-  cc.enqueue({
+  var forceIssuer = req.params.forceIssuer;
```
shane-tomlinson, line 97: I forget, what happens with forceIssuer?

bin/verifier
```diff
@@ -135,13 +147,11 @@ function doVerification(req, resp, next) {
         rp: audience
       });
     } else {
-      resp.json({
+      resp.json(_.extend(r.success, {
```
shane-tomlinson, line 150: Are there any fields that come from success that we do not want passed to the client? Should there be a filter to prevent any new fields from automatically being passed?

config/b2g.json
```diff
@@ -0,0 +1,3 @@
+{
+  "authentication_duration_ms": 315360000000
```
shane-tomlinson, line 2: A comment about this magic number would be awesome.

example/rp/index.html
```diff
@@ -68,15 +68,21 @@
...
-    </li>
     </li><li>
       <input type="checkbox" id="siteLogo">
       <label for="siteLogo">Supply Site Logo</label><br />
-    </li>
     </li><li>
       <input type="checkbox" id="returnTo">
       <label for="returnTo">Supply returnTo</label><br />
-    </li>
+    </li><li>
```
shane-tomlinson, line 80: I dig updated example RPs. And nice cleanup of the HTML.

lib/configuration.js
```diff
@@ -280,6 +280,10 @@ var conf = module.exports = convict({
   measure_dom_loading: {
     doc: "Measure the time it takes the DOM to start loading by embedding an inline script in the document HEAD",
     format: 'boolean = false'
+  },
+  forcable_issuers: {
+    doc: "Hostnames which this Persona instance will issue identies for. Used in b2g only.",
```
shane-tomlinson, line 285: nit - identies should be identities

lib/db/json.js
```diff
@@ -339,69 +352,70 @@ function getAndDeleteRowForSecret(secret, cb) {
...
-          var hash = o.passwd;
-          if(e || typeof hash !== 'string') return cb(e, o.email, o.existing_user);
-
-          // a hash was specified, update the password for the user
-          exports.emailToUID(o.email, function(err, uid) {
-            if(err) return cb(err, o.email, o.existing_user);
-
-            exports.updatePassword(uid, hash, false, function(err) {
-              cb(err || null, o.email, o.existing_user);
+    dbutils.withType(o.email, function (type) {
```
shane-tomlinson, line 354: What happens here if I have an email address and have verified it, then you come along and claim to own it to under the unverified email flow?

shane-tomlinson, line 354: Is type here needed because an unverified email could be either primary or secondary and we cannot assume it is secondary like we used to assume?

lib/db/json.js
```diff
@@ -339,69 +352,70 @@ function getAndDeleteRowForSecret(secret, cb) {
...
-          else createAccount();
-        });
-      } else {
-        createAccount();
-      }
+    dbutils.withType(o.email, function (type) {
+      exports.emailKnown(o.email, function(err, known) {
+        function createAccount() {
+          var emailVal = {};
+          emailVal[o.email] = { type: type, verified: true };
```
shane-tomlinson, line 390: Here you add a verified: true flag, but not in completeConfirmEmail, is this expected?

lib/db/json.js
```diff
@@ -427,6 +441,8 @@ exports.completePasswordReset = function(secret, password, cb) {
         Object.keys(emails).forEach(function(email) {
           if (email !== o.email && emails[email].type === 'secondary') {
             emails[email].verified = false;
```
shane-tomlinson, line 442: We had unverified emails before? Will DBs have to be updated to deal with these?

shane-tomlinson, line 442: Are all emails in the DB expected to have the verified field? What about emails we started with nearly 2 years ago? If not, would it be better to add an "unverified" flag for new entries so that old email entries are not affected?

lib/db/json.js
```diff
@@ -587,6 +603,25 @@ exports.cancelAccount = function(authenticated_uid, cb) {
   process.nextTick(function() { cb(null); });
 };
 
+exports.createUnverifiedUser = function(email, hash, cb) {
```
shane-tomlinson, line 606: Is there an unverified add, or is that taken care of by the stage function?

lib/static/views.js
```diff
@@ -71,6 +71,12 @@ function isProductionEnvironment() {
     ].indexOf(config.get('public_url')) !== -1;
 }
 
+const X_FRAME_ALLOWED = [
+  '/communication_iframe',
+  '/relay',
+  '/tos',
+  '/privacy'
+];
```
shane-tomlinson, line 79: Does /about need added to this list?

shane-tomlinson, line 79: While we are at it, can we add '/test/index.html' to this list?

lib/wsapi/address_info.js
```diff
@@ -100,48 +104,63 @@ exports.process = function(req, res) {
...
+    };
+    questionAnswered();
+  } else {
+    primary.checkSupport(domain, function(err, r) {
+      questionsToAnswer--;
+
+      if (!err && r && r.urls) {
+        typeRightNow = r.urls;
+        typeRightNow.type = 'primary';
+        typeRightNow.issuer = r.authoritativeDomain;
```
shane-tomlinson, line 123: This override the issuer that comes in from the request, does this interact with primaries as we expect it to? What is the expected interaction for primaries if an issuer is specified in the request?

resources/static/common/js/storage.js
```diff
@@ -88,6 +102,14 @@ BrowserID.Storage = (function() {
     return {};
   }
 
+  function getForceIssuerEmails(forceIssuer) {
```
shane-tomlinson: I understand why you are doing it this way, is it possible to combine this series of functions with the ones that already exist? What are the downsides to doing so?

resources/static/common/js/user.js
```diff
@@ -120,8 +121,9 @@ BrowserID.User = (function() {
 
     // stagingStrategy is a curried function that will have all but the
     // onComplete and onFailure functions already set up.
-    stagingStrategy(function(staged) {
-      var status = { success: staged };
+    stagingStrategy(function(status) {
```
shane-tomlinson, line 115: Do all of the staging strategies now return an object?

resources/static/common/js/user.js
```diff
@@ -839,20 +893,41 @@ BrowserID.User = (function() {
...
           var client_emails = _.keys(issued_identities);
 
-          var emails_to_add = _.difference(server_emails, client_emails);
-          var emails_to_remove = _.difference(client_emails, server_emails);
-          var emails_to_update = _.intersection(client_emails, server_emails);
+          var emails_to_add_pair = [_.difference(server_emails, client_emails)];
+          var emails_to_remove_pair = [_.difference(client_emails, server_emails)];
+          var emails_to_update_pair = [_.intersection(client_emails, server_emails)];
+
+          if (!! User.forceIssuer && 'default' !== User.forceIssuer) {
```
shane-tomlinson: Instead of creating pairs (my head is ready to explode, it's too small to handle this!), would it be easier to run the old code with each possible issuer? Once for the default, once for User.forceIssuer?

resources/static/common/js/user.js
```diff
@@ -1157,85 +1239,91 @@ BrowserID.User = (function() {
...
-            createAssertion(storedID);
-          }, 0);
+        var storedID,
+            assertion,
+            self=this;
+
+        if ('default' === forceIssuer)
+          storedID = storage.getEmail(email);
+        else
+          storedID = storage.getForceIssuerEmail(email, forceIssuer);
```
shane-tomlinson: I feel like there should be a way to just call storage.getEmail(email, forceIssuer) and have it do the right thing.  There are a lot of if/else's running around that could be collapsed.

resources/static/common/js/user.js
```diff
@@ -1157,85 +1239,91 @@ BrowserID.User = (function() {
...
+          prepareDeps();
+          if (storedID.priv) {
+            // parse the secret key
+            // yield to the render thread!
+            setTimeout(function() {
+              createAssertion(storedID);
+            }, 0);
+          }
+          else {
+            // TODO what will the type of forceIssuer email addresses be?
```
shane-tomlinson, line 1343: That's a very good question. Should it be a secondary since there are no primaries in B2G?

resources/static/common/js/user.js
```diff
@@ -984,7 +1064,7 @@ BrowserID.User = (function() {
      *        if type is secondary.
      * @param {function} [onFailure] - Called on XHR failure.
      */
-    addressInfo: function(email, onComplete, onFailure) {
+    addressInfo: function(email, issuer, onComplete, onFailure) {
```
shane-tomlinson: If issuer is already stored at User.forceIssuer, do we need to pass issuer here? Can the code be made easier to understand and more transparent by not passing it? Can we set User.forceIssuer and forget about it from there on out and just have all calls to user.js "do the right thing"?

resources/static/common/js/user.js
```diff
@@ -1157,85 +1239,91 @@ BrowserID.User = (function() {
...
-              sk,
-              function(err, signedAssertion) {
-                assertion = jwcrypto.cert.bundle([idInfo.cert], signedAssertion);
-                storage.site.set(audience, "email", email);
-                complete(assertion);
-              });
-          }, 0);
-        }, onFailure);
-      }
+    getAssertion: function(email, audience, forceIssuer, onComplete, onFailure) {
```
shane-tomlinson: Same comment about forceIssuer here as with addressInfo, can we forgo passing forceIssuer if User.forceIssuer is already set? We end up with two sources of truth so to speak.

bin/keysigner
jedp, line 97: forceIssuer disables primary lookup and forces the assertion to be issued by the (whitelisted) issuer.  The product case is for marketplace not to have to kick their users over to third-party auth screens in the payment flow.

example/rp/index.html
jedp, line 80: Agreed.  @seanmonstar is always so good about that.

lib/db/json.js
jedp, line 354: I can't claim your email address unless I know the password.  If anything, the danger is the other way around: I mistype my email address without realizing, and eventually you end up verifying it and claim the identity I thought I had.

lib/db/json.js
jedp, line 390: Excellent catch, Shane.  I'm not sure; let's check that.

lib/db/json.js
jedp, line 606: I believe that's it - unverified emails are basically just staged emails that haven't been verified yet

lib/static/views.js
jedp, line 79: sounds good

lib/wsapi/address_info.js
jedp, line 123: When forceIssuer is used, primary lookup is disabled and it falls right over to the issuer we specify.  So the primaries are not involved at all.

resources/static/common/js/user.js
jedp: Yes, it looks like there's an opportunity to refactor and collapse a bit

resources/static/common/js/user.js
jedp, line 1343: Yes, I think we should have a forceIssuer summit soon.  There are still a number of open questions we punted on.  I'll put something on the calendar.

resources/static/common/js/user.js
jedp: That would be great.

resources/static/common/js/user.js
```diff
@@ -1137,12 +1217,14 @@ BrowserID.User = (function() {
      * server a keypair for the given email address.
      * @method syncEmailKeypair
      * @param {string} email - Email address.
-     * @param {string} [issuer] - Issuer of keypair.
+     * @param {string} forceIssuer - IdP that should back the assertion.
```
shane-tomlinson: does forceIssuer need to be added to the function signature?

resources/static/dialog/js/modules/dialog.js
```diff
@@ -13,6 +13,9 @@ BrowserID.Modules.Dialog = (function() {
       helpers = bid.Helpers,
       win = window,
       startExternalDependencies = true,
+      TOSPP_SELECTOR = ".tospp a",
```
shane-tomlinson: I think we should start to make a separate module for CSS selectors that can be shared across other modules.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -146,6 +149,43 @@ BrowserID.Modules.Dialog = (function() {
   }
 
 
+  function showTOSPP(e) {
```
shane-tomlinson: The TOS/PP stuff should move out of here and into its own module so that each can be thought about and verified independently.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -171,6 +211,14 @@ BrowserID.Modules.Dialog = (function() {
         startChannel.call(self);
       }
 
+      // if B2G
+      //this.click(TOSPP_SELECTOR, showTOSPP);
+      $(TOSPP_SELECTOR).live('click', function(e) {
```
shane-tomlinson: Let's use dom.bindEvent or this.bindEvent instead of live events.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -257,6 +305,27 @@ BrowserID.Modules.Dialog = (function() {
           user.setReturnTo(returnTo);
         }
 
+        // forceIssuer is used by the Marketplace to disable primary support
+        // and replace fxos.login.persona.org as the issuer of certs
+        if (paramsFromRP.forceIssuer) {
+          // TODO check for valid domain
+          params.forceIssuer = paramsFromRP.forceIssuer;
```
shane-tomlinson: This should be scrubbed if it hasn't been already.

resources/static/dialog/js/modules/provision_primary_user.js
```diff
@@ -56,7 +54,7 @@ BrowserID.Modules.ProvisionPrimaryUser = (function() {
         throw new Error("missing config option: email");
       }
 
-      user.addressInfo(email, function(status) {
+      user.addressInfo(email, 'default', function(status) {
```
shane-tomlinson: 'default' on this line is not future proof.

A user who makes it to this screen now should already have User.forceIssuer set to 'default'. If the restriction barring primaries on B2G is ever lifted, we are going to have to remember to change this line. 'default' (I believe) can be safely set to User.forceIssuer.

resources/static/dialog/js/modules/required_email.js
```diff
@@ -32,7 +32,7 @@ BrowserID.Modules.RequiredEmail = (function() {
     var self = this;
 
     function getAssertion() {
-      dialogHelpers.getAssertion.call(self, email, callback);
+      dialogHelpers.getAssertion.call(self, email, 'default', callback);
```
shane-tomlinson: 'default' again.

resources/static/dialog/js/misc/state.js
```diff
@@ -343,9 +367,18 @@ BrowserID.State = (function() {
 
     handleState("email_chosen", function(msg, info) {
       var email = info.email,
-          record = storage.getEmail(email);
+          record;
 
-      self.email = email;
+      // qunit tests won't have run start state... reinit selfIssuer
+      self.forceIssuer = self.forceIssuer || 'default';
```
shane-tomlinson: Since forceIssuer is stored in user.js, can the local copy be removed and that one used so there is a single source of truth?

config/b2g.json
jedp, line 2: Yes.

I'd prefer to make it 1324512000000, myself.  But either way, it doesn't account for leap years.

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -125,7 +132,8 @@
       // User must be authenticated to get an assertion.
       if(authenticated) {
         user.setOrigin(origin);
-        user.getAssertion(email, user.getOrigin(), function(assertion) {
+        var forceIssuer = 'default';
```
shane-tomlinson: Does user.forceIssuer need to be set at all?

resources/static/dialog/js/misc/state.js
```diff
@@ -117,6 +117,10 @@ BrowserID.State = (function() {
       self.hostname = info.hostname;
       self.siteName = info.siteName || info.hostname;
       self.siteTOSPP = !!(info.privacyPolicy && info.termsOfService);
+      self.forceIssuer = user.forceIssuer = (!!info.forceIssuer ? info.forceIssuer : 'default');
```
shane-tomlinson: Some QUnit tests around these new flows would be awesome.

shane-tomlinson:
This is _MUCH_ cleaner than the first go!

Overall thoughts:

1) Let's get the tests passing.
2) There are some areas of cleanup, especially on the front end that I would like to see.
2a) Use user.forceIssuer is the source of truth for the issuer. Set this early. Remove issuer from user.js function signatures and from being passed around everywhere on the front end.
2b) storage.js could be cleaned up to remove the specialized functions which use issuer.
3) MOAR QUnit tests!
4) MOAR Selenium tests! I would like to see the expected flows codified so that we can flesh out some of the intricacies.
5) Let's break up the features for merge back to dev.

jedp:
Awesome.  Much rejoicing.

Once the tests are passing, and b2g and dev functionality is secure, I'll branch this branch and re-deploy b2g so that we can get all the speedups and other goodness that have landed in dev.

For front-end tests, I have some selenium flows for b2g here: https://github.com/jedp/gaia-ui-tests/commit/6d8674d6418f072943d9bead8703c182c64e9c81 I can work on pulling these into the persona selenium tests.

I'll make a list of commits by feature that we can use to make the job of breaking features up more manageable.

@shane-tomlinson thank you so much for your time and thought in this.  Cheers!

jedp:
Pardon the craziness -- rebase dev introduced conflicts

resources/static/common/js/user.js
```diff
@@ -258,12 +260,50 @@ BrowserID.User = (function() {
    * @method persistEmailKeypair
    * @param {string} email - Email address to persist.
    * @param {object} keypair - Key pair to save
+   * @param {string} forceIssuer - IdP that should back the assertion.
+   *                   Value is a hostname or the string 'default'.
    * @param {function} [onComplete] - Called on successful completion.
    * @param {function} [onFailure] - Called on error.
    */
   function persistEmailKeypair(email, keypair, cert, onComplete, onFailure) {
+    User.addressInfo(email, User.forceIssuer, function(info) {
```
shane-tomlinson: This needs to be looked at to make sure that caching does not bite us. Only one request is made per address, addressInfo caches responses. If an address's info was checked pre-verification, cached with "state=unverified" and then the address is persisted post-verification, email_obj.unverified will be true.

A better way is to call persistEmailKeypair with an "unverified" flag that is guaranteed to be correct.

.awsbox.json
```diff
@@ -9,7 +9,7 @@
     "bin/static"
   ],
   "env": {
-    "CONFIG_FILES": "$HOME/code/config/production.json,$HOME/code/config/aws.json,$HOME/config.json"
+    "CONFIG_FILES": "$HOME/code/config/production.json,$HOME/code/config/aws.json,$HOME/code/config/b2g.json,$HOME/config.json"
```
seanmonstar, line 12: we want to merge in this b2g.json config file here?

jedp, line 12: right, it should become superfluous.  thanks, good catch.

jedp, line 12: but it's not immediately obvious to me what the right solution for `authentication_duration_ms` is

shane-tomlinson:
Closing this down because we broke it into its component parts.

PR #3124, opened 2013-03-20
Added logging of KPI Backend response

### LINKED ISSUE #3100: interaction_data should log response of the forward of data
Stage wasn't quite configured right and the squid proxy was returning 403, which could be seen on pencil graphs. However, interaction_data does not make note of the failure and it should.

All the logs had to say was:

```
{"level":"info","message":"127.0.0.1 - - [Wed, 13 Mar 2013 21:40:13 GMT] \"POST /wsapi/interaction_data HTTP/1.1\" 200 16 \"-\" \"-\"","timestamp":"2013-03-13T21:40:13.753Z"}
{"level":"debug","message":"Simulate write to KPI Backend DB - {\n  \"event_stream\": [ ... etc. ...
{"level":"debug","message":"Using proxy for KPI","timestamp":"2013-03-13T21:40:13.756Z"}
{"level":"debug","message":"sending request to KPI backendhttps://kpiggybank-stage.personatest.org/wsapi/interaction_data","timestamp":"2013-03-13T21:40:13.757Z"}
```

kparlante:
This should fix #3100, @shane-tomlinson could you review?

@jrgm asked if the simulated write to backend could be turned off in production -- is that already handled by using logger.debug?

### DIFF
--- a/lib/wsapi/interaction_data.js
+++ b/lib/wsapi/interaction_data.js
@@ -60,7 +60,6 @@ var store = function (kpi_json, cb) {
         }
       };
       kpi_req = http.request(options);
-      logger.debug("Using proxy for KPI");
     } else {
       options = {
             hostname: db_url.host,
@@ -78,15 +77,22 @@ var store = function (kpi_json, cb) {
       
       var protocol = (db_url.scheme === 'https') ? https : http;
       kpi_req = protocol.request(options);
-      logger.debug("Not using proxy for KPI");
     }
+    
+    kpi_req.on('response', function(res) {
+      if (res.statusCode !== 201) {
+        logger.warn('KPI Backend (or proxy) response code is not 201: ' + res.statusCode);
+      } else {
+        logger.info('interaction data successfully posted to KPI Backend');
+      }  
+    });
     
     kpi_req.on('error', function (e) {
       // TODO statsd counter
       logger.error('KPI Backend request error: ' + e.message);
     });
 
-    logger.debug("sending request to KPI backend" + config.get('kpi_backend_db_url'));
+    logger.debug("sending request to KPI backend: " + config.get('kpi_backend_db_url'));
     kpi_req.write(post_data);
     kpi_req.end();
 


jrgm:
re: debug and logging in production

So, debug messages are enabled in production. You could condition that message on `if (config.get('env') === 'production') {`.

lib/wsapi/interaction_data.js
```diff
@@ -78,15 +77,22 @@ var store = function (kpi_json, cb) {
...
       var protocol = (db_url.scheme === 'https') ? https : http;
       kpi_req = protocol.request(options);
-      logger.debug("Not using proxy for KPI");
     }
     
+    kpi_req.on('response', function(res) {
+      if (res.statusCode !== 201) {
+        logger.warn('KPI Backend (or proxy) response code is not 201: ' + res.statusCode);
+      } else {
+        logger.info('interaction data successfully posted to KPI Backend');
```
shane-tomlinson, line 86: Do we log successful responses we make in other areas? If not, do we need to here? (it looks like we do, I just want to make sure)

lib/wsapi/interaction_data.js
```diff
@@ -78,15 +77,22 @@ var store = function (kpi_json, cb) {
       
       var protocol = (db_url.scheme === 'https') ? https : http;
       kpi_req = protocol.request(options);
-      logger.debug("Not using proxy for KPI");
     }
     
+    kpi_req.on('response', function(res) {
+      if (res.statusCode !== 201) {
+        logger.warn('KPI Backend (or proxy) response code is not 201: ' + res.statusCode);
```
shane-tomlinson, line 84: This shows up when running in dev mode locally, where we aren't really expecting KPIs to be enabled. It makes me wonder whether the default for configuration.js->kpi_backend_db_url should be empty and we have to explicitly enable KPIs locally if we want them.

shane-tomlinson:
Outside of my other comments, I am happy with this. Both front and backend tests pass. I see the warning message when a local KPI server is not running. I see the correct log message when an external KPI server is specified. @jrgm, does this meet the need? If so, want to hit the green button?

lib/wsapi/interaction_data.js
jrgm, line 86: On the router we do, but indirectly: the router notes that it's going to forward to whichever other host, and then the common access_log style response back to the client is logged. In this case, I don't mind having the positive response logged.

lib/wsapi/interaction_data.js
jrgm, line 84: @shane-tomlinson By "KPIs to be enabled" I'm assuming you mean "the collector is available". I was pondering whether we should have a tiny daemon locally, similar to the way we have the tiny proxy instead of squid - so that the normal production code paths get exercised in dev.

jrgm:
Yeah, we can settle what the default behaviour in dev should be for the collector later, and also quieting the dump of the data; that information my be interesting in the short term (but ultimately redundant).

jrgm:
Here's an idea (possibly bad) to make the dev proxy act as the kpi collector by default; can be overridden. https://github.com/jrgm/browserid/commit/433b9c26f18c78399bb07b06f8124011c800d9b5

lib/wsapi/interaction_data.js
shane-tomlinson, line 84: @jrgm - Love it.

Issue #2340, opened 2012-08-17
P3P headers needed for IE8/some variants of IE9


shane-tomlinson:
http://msdn.microsoft.com/en-us/library/ms537341%28v=vs.85%29.aspx

See #2183, #2192

seanmonstar:
Quote from @benadida:

> In a brief conversation with Jishnu on Friday, he said he was supportive of the P3P approach that goes something like "this is not a P3P policy, but Mozilla deeply cares about your privacy, go <here> for more."

jaredhirsch:
So we'd literally send down:

```
P3P: CP="This is not a P3P policy, but Mozilla deeply cares about your privacy. See 
http://www.mozilla.org/persona/privacy-policy for more."
```

An example of Google's workaround: http://blogs.msdn.com/b/ie/archive/2012/02/20/google-bypassing-user-privacy-settings.aspx

shane-tomlinson:
Closing it, p3p headers are merged.

PR #3132, opened 2013-03-22
add P3P headers to endpoints IE might care about

jaredhirsch:
Reopened. @shane-tomlinson or @lloyd, mind taking a look?

I split this off from the main blob of third-party cookie discussion.

Add P3P headers to IE family browsers (a little searching suggests IE9 and, presumably, IE10 also care about this header).

Do I need to add tests to confirm this header is appended if the user-agent header contains an IE-like string?

### DIFF
--- a/bin/browserid
+++ b/bin/browserid
@@ -25,6 +25,7 @@ heartbeat = require('../lib/heartbeat.js'),
 logger = require('../lib/logging.js').logger,
 shutdown = require('../lib/shutdown'),
 primary = require('../lib/primary'),
+addP3PHeader = require('../lib/p3p.js'),
 forward = require('../lib/http_forward.js'),
 toobusy = require('../lib/busy_middleware.js');
 
@@ -94,6 +95,10 @@ if (config.get('scheme') == 'https') {
     next();
     });
 }
+
+// Add P3P headers to please IE8
+app.use(addP3PHeader);
+
 
 // #4 - prevent framing of everything.  content underneath that needs to be
 // framed must explicitly remove the x-frame-options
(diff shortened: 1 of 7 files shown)


jaredhirsch:
this should fix #2340

jaredhirsch:
relevant MSDN page with some discussion of IE privacy policy:

http://msdn.microsoft.com/en-us/library/ms537343(v=vs.85).aspx#privacy_preference_settings

jaredhirsch:
@lloyd @shane-tomlinson thoughts on this'n?

shane-tomlinson:
@6a68 - I do not think this is needed anymore, esp with the 3rd party cookie changes in #3116. IE allows a 3rd party site to send cookies as soon as the user visits that site.

jaredhirsch:
I'll test with an IE8 VM to see if this affects first-time user experience. If it doesn't, I'll close.

jaredhirsch:
After rebasing against current dev, I can't produce any problems with first-time experience on IE8 or IE9. CLosing.

jaredhirsch:
Reopening because #3417.

jaredhirsch:
@TeeTeeHaa (affected user) confirms this patch fixes #3417

shane-tomlinson:
@6a68 - lockdown seems to not like this change, see the travis logs.

jaredhirsch:
@shane-tomlinson Fixed the lockdown error, working for me. Mind taking another look?

jaredhirsch:
I can't figure out why the travis build failed. The tests run fine on my local machine and the failing tests aren't affected by my patch.

jrgm:
There. I made you green.

jaredhirsch:
thanks @jrgm!

somebody clicky!

PR #3134, opened 2013-03-25
Refactor network.js & user.js to move user related context to user.js

shane-tomlinson:
It always frustrated me that so much user related context bled into network.js This is an effort to clean that up by moving a lot of the user related context from network.js into user.js This allowed a lot of simplification to occur of many network calls because there is only one place to take care of user related stuff.

Many direct calls to network.js were removed and replaced with equivalent calls to user.js. user.js should be the one low level interface to rule them all. This lays the groundwork for a future simplification of context by moving it into its own module.

Note, no Selenium tests were modified in this. They all continue to pass.

### DIFF
--- a/resources/static/common/js/modules/development.js
+++ b/resources/static/common/js/modules/development.js
@@ -83,7 +83,7 @@ BrowserID.Modules.Development = (function() {
   }
 
   function forceIsThisYourComputer() {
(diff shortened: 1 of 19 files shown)


jaredhirsch:
r+ nice! all tests passing.

PR #3142, opened 2013-03-27
Add separate selenium config for local runs, including multiple FF channels. Fixes #3118, #3138.

jaredhirsch:
Selecting the --local option, or setting PERSONA_NO_SAUCE=1 in the
  environment, will now switch to a different config file. The set of
  supported platforms will be totally generic.

Example:

``` bash
]$ ./scripts/run-all.js --local --list-platforms
5 platforms configured:
  * firefox
  * chrome
  * opera
  * safari
  * ie
]$ ./scripts/run-all.js --list-platforms
9 platforms configured:
  * win7_firefox_16
  * linux_firefox_16
  * osx_firefox_14
  * win7_chrome
  * winxp_ie_8
  * win7_ie_9
  * win8_ie_10
  * linux_opera_12
  * osx_safari_5
```

Updated to include an example of how to specify an additional firefox platform locally. Specifying multiple channels for chrome is a bit tougher--I couldn't figure out the syntax in 30 min of twiddling--but we really only need multiple firefoxen to close out the bug :-)

### DIFF
--- a/automation-tests/config/local-platforms.js
+++ b/automation-tests/config/local-platforms.js
@@ -11,6 +11,16 @@ const platforms = {
     browserName: 'firefox',
     version: ''
   },
+  /* if you wish to add a second firefox, give it a unique name and specify
+     the path to the binary, which will vary across systems.
+
+  "firefox-nightly": {
+    platform: 'ANY',
+    browserName: 'firefox',
+    version: '',
+    firefox_binary: '/Applications/FirefoxNightly.app/Contents/MacOS/firefox-bin'
+  },
+  */
   "chrome": {
     platform: 'ANY',
     browserName: 'chrome',
--- a/automation-tests/config/local-platforms.js
+++ b/automation-tests/config/local-platforms.js
@@ -1,0 +1,43 @@
+//
+// By default, we assume local platforms just have default locations.
+//
+// You must have installed ChromeDriver to use Chrome.
+// Ditto for all Webdrivers except Firefox.
+// see the Webdriver wiki for more: https://code.google.com/p/selenium/w/list
+//
+const platforms = {
+  "firefox": {
+    platform: 'ANY',
+    browserName: 'firefox',
+    version: ''
+  },
+  "chrome": {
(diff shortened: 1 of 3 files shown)


jaredhirsch:
The example firefox-nightly platform is now in `/automation-tests/config/local-platforms.json` and should fix #3138.

jaredhirsch:
@shane-tomlinson mind having a look?

automation-tests/config/local-platforms.js
```diff
@@ -0,0 +1,53 @@
...
+    version: ''
+  },
+  /* if you wish to add a second firefox, give it a unique name and specify
+     the path to the binary, which will vary across systems.
+
+  "firefox-nightly": {
+    platform: 'ANY',
+    browserName: 'firefox',
+    version: '',
+    firefox_binary: '/Applications/FirefoxNightly.app/Contents/MacOS/firefox-bin'
```
shane-tomlinson, line 23: This may make windows users like @seanmonstar very sad. Can this be made to work cross OS?

shane-tomlinson:
Aside from the comment I already made about cross OS behavior, looking good. I get some ghetto behavior out of non-Firefox browsers locally (they stall on startup or when requesting dev.login.persona.org), but that is likely a separate issue.

automation-tests/config/local-platforms.js
jaredhirsch, line 23: That's just the example for MacOS, the idea is to illustrate that you use the firefox_binary desiredCapability to indicate the alternate path. Mac surprisingly needs the binary inside the package; on Windows it's just `foo\bar\blah\firefox.exe`.

There are example paths for all three platforms on the webdriver wiki, maybe I should add those as well? Any thoughts on how to document this so it's easy-to-understand?

See here, second table on the page: https://code.google.com/p/selenium/wiki/FirefoxDriver

jaredhirsch:
@shane-tomlinson I updated the example in the config file to try to make it clearer. Helpful?

PR #3144, opened 2013-03-27
Dynamically shrink footer text to fit on one line.

### LINKED ISSUE #3129: Footer text wraps on to two lines in certain languages.
As reported by a list member:

FF - German: http://f.666kb.com/i/cck4ck1idf6b6kp3q.jpg (broken design)
FF - Russian: http://f.666kb.com/i/cck4dbtk85yd77tw6.jpg (broken design)
FF - Spanish: http://f.666kb.com/i/cck4dpjhxmjkl9sly.jpg (broken design)
IE - German: http://f.666kb.com/i/cck4e5p4ftzqz27me.jpg (broken design)
Chrome - German: http://f.666kb.com/i/cck4elrl9chkxbb4m.jpg (broken design)

jaredhirsch:
Fixes #3129--now updated. @shane-tomlinson, mind taking another look?

orig description:

Not done yet--I'm not quite sure where I can plug this into the dialog startup flow such that it **always** runs when the dialog loads, but is in run in a context where we can loop with setTimeout (so that the UI can redraw the footer text between measurements).

@shane-tomlinson any thoughts on a good place to put this code?

here's a fiddle showing the resize thingy in action: http://jsfiddle.net/vFM2v/

### DIFF
--- a/resources/static/dialog/js/misc/screen_size_hacks.js
+++ b/resources/static/dialog/js/misc/screen_size_hacks.js
@@ -118,6 +118,18 @@
         boundingRectEl.css("top", favIconHeight + "px");
     }
 
+    // this can be used to keep the footer text on one line, #3129.
+    function resizeFooterText(cb) {
+      function shrinkFooter() {
+        var footerText = $('#footerText');
+        if (footerText.width() < $('footer').width()) return;
+        var newFontSize = parseInt(footerText.css('fontSize'), 10) - 1 + 'px';
+        footerText.css('fontSize', newFontSize);
+        setTimeout(shrinkFooter); // let UI loop run before measuring again
+      }
+      shrinkFooter();
+    }
+
     // The mobile breakpoint is 640px in the CSS.  If the max-width is changed
     // there, it must be changed here as well.
     if($(window).width() > 640) {
@@ -126,7 +138,7 @@
     else {
       mobileHacks();
     }
-
+    resizeFooterText();
     scrollableEl.css("position", "");
   }
 
--- a/resources/views/dialog_layout.ejs
+++ b/resources/views/dialog_layout.ejs
@@ -48,7 +48,7 @@
       </div>
 
       <footer>
-<%- format(gettext('<strong>Persona.</strong> Simplified sign-in, built by a non-profit. <a %s>Learn more&rarr;</a>'), [" href='/about' target='_blank'"]) %>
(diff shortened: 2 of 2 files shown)


shane-tomlinson:
@6a68 - the best place for this is probably in dialog/js/misc/screen_size_hacks.js

shane-tomlinson:
@6a68 - Have you experimented with allowing the footer to stretch over two lines and resize the main contents area?

This footer stuff is a pain. There has to be an easier way.

jaredhirsch:
@shane-tomlinson OK, I'll put it in screen_size_hacks.

The bummer about resizing around the UI loop iterations is that, on sluggish systems, it might look choppy (I haven't previewed this in IE8 or FXOS, for instance).

Good point about reducing footer magic--we could totally play around with downsizing the main area to make room for a wrapped footer. This would arguably look less insane on mobile handsets in portrait mode, which I just thought of: resizing the footer to fit on one line will probably look comically small. OTOH, if the footer pushes the contents up, on a mobile handset it might totally obscure the input form, sigh.

I guess I will stand up an awsbox with the changes put in screen_size_hacks and see if the experience is better. Maybe I should do a screencast for @skinny97214 to give input?

shane-tomlinson:
A screen cast would help everyone I think. 

I was wondering about portrait mode on a mobile device as well. At one point, @skinny97214 had some great ideas on getting rid of the footer altogether. Maybe we should revisit that?

What if the footer was allowed to scroll off the screen if the content grew past a certain point?

jaredhirsch:
ooh, I should requestAnimationFrame perhaps

skinny97214:
Here's a mock from the FxOS design team for sign in to Marketplace and eventually, the first-run experience. This is the trusted UI frame, FYI. We have a postage stamp size to deal with. Most users have no hope of seeing the footer since the keyboard covers it up.

Screen Shot 2013-03-28 at 1 47 09 PM

shane-tomlinson:
@skinny97214 - you had ideas at one point about removing the footer from the dialog as well. Are you still considering this?

skinny97214:
Not in the near term until we get more resources. If we did a more comprehensive redesign in the direction smartell wanted to take us, I think that would be high on the chopping block to remove.

jaredhirsch:
updated to latest dev and moved the resize call into screen_size_hacks

shane-tomlinson:
@6a68 - thanks for the update. I tested with Russian on "shrink_footer" 123done.org/personatest.org but am not seeing the line shrink as expected. The instance is still up if you want to give it a try.

shane-tomlinson:
r+ Manually merging with one tiny CSS change.

Nicely done @6a68.

PR #3149, opened 2013-03-28
Rename BrowserID-Version header to BrowserID-git-sha

### LINKED ISSUE #3125: is 'BrowserID-Version' a misleading name for this http header
I was looking at some code, and I had this thought earlier but didn't file the issue. Anyways, is 'BrowserID-Version' the right name for this header.

A naive reading would be that this is the public protocol version (e..g, 1.0, 2.0). I think it could be something better, possibly more suggestive that this is an implementation detail. 
'BrowerID-current-commit-sha'?

Feel free to close this as BIKESHED.

shane-tomlinson:
@jrgm - wanna review this?

fixes #3125


jrgm:
Actually, there's a few places in tests/ that still refer to BrowserID-Version.

shane-tomlinson:
@jrgm - updated. Thanks for finding those misses.

jrgm:
r+

PR #3150, opened 2013-03-28
Simplify the storage.js interface

shane-tomlinson:
- Remove setLoggedIn and getLoggedIn
- Remove signInEmail namespace

Data is now handled by the storage.site.set/get functions with approprite namespace keys. This is to make where this info is stored consistent with other site specific info.


jaredhirsch:
sorry this is sitting a while. I'll try to get to it after looking at urgent bugs :surfer:

jaredhirsch:
r+

PR #3158, opened 2013-03-31
Remove stray underscore leftover. Might fix #3153...

### LINKED ISSUE #3153: [dev,train-2013.03.28] expired certificates are not being invalidated.
STR
1. Start with a brand new profile.
2. Go to dev.123done.org, and use the dialog to signin with an @yahoo.com email address.
3. After the dialog dismisses, do nothing for >5 minutes.
4. Then logout of dev.123done.org and try to signin again with that @yahoo.com address.

Expected: Signed in.
Actual: alert is shown with 'failed to verify assertion: assertion has expired'.

Basically from looking at localStorage, the certificate for that email is not invalidated when it has expired. And then that expired certificate is used for the assertion.

This does not reproduce with train-2013.03.01. I expect the refactor of user.js/network.js is where the invalidation was dropped.

This, of course, is 5-star and blocks train-2013.03.28.

jaredhirsch:
Still working on #3153 but this seems like a bug either way.

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -64,7 +64,7 @@ BrowserID.User = (function() {
         var emails = storage.getEmails();
         var issued_identities = {};
         prepareDeps();
-        _(emails).each(function(email_obj, email_address) {
+        _.each(emails, function(email_obj, email_address) {
           try {
             email_obj.pub = jwcrypto.loadPublicKeyFromObject(email_obj.pub);
           } catch (x) {


jrgm:
This makes sense to me, but could we get someone else to review. /cc @jedp, @seanmonstar 
See background on GH-3153

shane-tomlinson:
@6a68 - You did well to diagnose this, I took this PR and made a new one that has a test for the failing condition. Closing this in favor of #3168, could you review? 

This looks good to me.

Issue #3160, opened 2013-04-01
npm install on clean tree produces ominous error messages


jrgm:
And actually, I think this gets cleaned up since postinstall.js gets run again and the second time automation-tests npm install is ok. But I don't want to ship train-2013.03.29 with this splat.

```
> node ./scripts/postinstall.js

>>> Installing automation-tests dependencies
*** Generating ephemeral keys used for testing ***
npm http GET http://127.0.0.1:49428/vows/0.6.4
npm http GET http://127.0.0.1:49428/request/2.12.0
npm http GET http://127.0.0.1:49428/optimist/0.3.5
npm http GET http://127.0.0.1:49428/underscore/1.4.2
npm http GET http://127.0.0.1:49428/minimatch/0.2.6
npm http GET http://127.0.0.1:49428/q/0.8.10
npm http GET http://127.0.0.1:49428/mkdirp/0.3.4
npm http 404 http://127.0.0.1:49428/request/2.12.0
npm ERR! registry error parsing json
npm http 404 http://127.0.0.1:49428/vows/0.6.4
npm ERR! registry error parsing json
npm http 404 http://127.0.0.1:49428/optimist/0.3.5
npm ERR! registry error parsing json
npm ERR! SyntaxError: Unexpected token p
npm ERR! package installation disallowed by lockdown
npm ERR!     at Object.parse (native)
npm ERR!     at RegClient.<anonymous> (/usr/local/node-0.8.17/lib/node_modules/npm/node_modules/npm-registry-client/lib/request.js:226:23)
npm ERR!     at Request.init.self.callback (/usr/local/node-0.8.17/lib/node_modules/npm/node_modules/request/main.js:120:22)
npm ERR!     at Request.EventEmitter.emit (events.js:99:17)
npm ERR!     at Request.<anonymous> (/usr/local/node-0.8.17/lib/node_modules/npm/node_modules/request/main.js:648:16)
npm ERR!     at Request.EventEmitter.emit (events.js:126:20)
npm ERR!     at IncomingMessage.Request.start.self.req.self.httpModule.request.buffer (/usr/local/node-0.8.17/lib/node_modules/npm/node_modules/request/main.js:610:14)
npm ERR!     at IncomingMessage.EventEmitter.emit (events.js:126:20)
npm ERR!     at IncomingMessage._emitEnd (http.js:366:10)
npm ERR!     at HTTPParser.parserOnMessageComplete [as onMessageComplete] (http.js:149:23)
[...]

PR #3161, opened 2013-04-01
undo PR 3085 in train-2013.03.29

jrgm:
i.e., shipping a build that claims to fail is bad karma. So don't automatically try to `npm install browserid/automation-tests`

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -24,12 +24,3 @@ function node(script) {
 }
 
 node('./generate_ephemeral_keys.js');
-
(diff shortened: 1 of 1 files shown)


jrgm:
Link to GH-3085 and GH-3160

jrgm:
Crap. I mistakenly made this a PR against the dev branch. I meant to target this at train-2013.03.29. Closing (although it still needs a fix in the dev branch so it's doesn't claim to fail. It also should be listening for the 'exit' event of the child_process and aborting the build if that is not 0).

PR #3162, opened 2013-04-01
undo PR 3085 in train-2013.03.29

jrgm:
i.e., shipping a build that claims to fail is bad karma. So don't automatically try to npm install browserid/automation-tests

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -24,12 +24,3 @@ function node(script) {
 }
 
 node('./generate_ephemeral_keys.js');
-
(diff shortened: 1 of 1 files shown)


jrgm:
Link to GH-3160

jaredhirsch:
I think this is a legit problem but not the right fix--see #3164, I think it was missing lockdown deps.

jrgm:
reopening. This is to get the current train branch deployed on stage.Your other PR is targetted at the dev branch.

jaredhirsch:
ah. r+ then

PR #3163, opened 2013-04-01
Individually revert all the accidentally-added commits to train-2013.03.01.

jaredhirsch:
This is a lot more verbose than trying to undo everything at once, but it's also a lot clearer what happened and how it was fixed.

Maybe just use this?


jaredhirsch:
@edmoz I was feeling suspicious about the amount of magic involved in `git reset` tricks, so I just reverted them all individually via

```
for i in `git rev-list 0d62ce67^..899ae402`; do git revert --no-edit -s $i; done
```

I'm fairly certain this covers **everything**.

jaredhirsch:
@lloyd @jrgm second opinion? this is a fix for an accidental push to the wrong remote (mozilla instead of edmoz). We've tried messing with `git reset --keep` but it didn't seem like it was capturing all the changes, see ed's earlier PR #3159.

edit: got the PR #s wrong.

jrgm:
Here's an idea. Cut a branch off the sha for train-2013.03.01, then swap the names of the two trains. Or am I on crack.

jrgm:
Actually, I am on crack. But **if** we have to ship a hotfix for train-2013.03.01, I might prefer to branch from 0d62ce67 to train-2013.03.01-redux and ship from there. (And then just delete this branch).

jaredhirsch:
@jrgm that's cool. in that case, mind if we just merge this in anyway? I don't want to orphan any existing branches, regardless of whether we hotfix release from this or another branch.

jrgm:
@6a68 - yeah, I guess merge this in.

edwindotcom:
Thanks guys. @jrgm we went down a few different paths and decided this one was the safest.

jaredhirsch:
arright. :beers: this was fun

PR #3168, opened 2013-04-02
Issue 3153 invalid certs

### LINKED ISSUE #3153: [dev,train-2013.03.28] expired certificates are not being invalidated.

shane-tomlinson:
@6a68 and @jrgm - can you review my second commit? The first commit that @6a68 did looks good.
- Remove stray underscore leftover. fixes #3153 
- Add a proper failure callback to cleanupIdentities.

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -64,7 +64,7 @@ BrowserID.User = (function() {
         var emails = storage.getEmails();
         var issued_identities = {};
         prepareDeps();
-        _(emails).each(function(email_obj, email_address) {
+        _.each(emails, function(email_obj, email_address) {
           try {
             email_obj.pub = jwcrypto.loadPublicKeyFromObject(email_obj.pub);
           } catch (x) {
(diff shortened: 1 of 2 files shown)


jaredhirsch:
looks good, tests pass. r+ :cake:

jrgm:
So @6a68, this PR was for a merge in to dev. But we need this on train-2013.03.29.

PR #3171, opened 2013-04-02
Issue 3148 xhr polling

### LINKED ISSUE #3148: [kpi] we are recording all xhr polling in interaction_data (regressed from issue 1619)
In GH-1619, we stopped recording repetitive xhr polling. For example, if the /sign_in dialog is left up for several minutes, it will poll every 3 seconds waiting for e.g., `https://login.dev.anosrep.org/wsapi/email_addition_status?email=...`. 

A blob with all this data will exceed our 10K cap on POSTs, and the data will be discarded, with 413 returned.

However, trying this with dev.123done.org tonight, and leaving the dialog up for a long time, shows that we have started accumulating the polling xhr again.

It looks like this now:

```
    ...
    [
      "xhr_complete.GET/wsapi/email_addition_status",
      560839,
      223
    ],
    [
      "xhr_complete.GET/wsapi/email_addition_status",
      564064,
      219
    ],
    [
      "xhr_complete.GET/wsapi/email_addition_status",
      567285,
      240
    ],
    ...
```

I've checked that this is still rejected as 413 for POST /wsapi/interaction_data and that the localStorage.interaction_data is reset (but need to check this more with other browsers). (Side note: I need to check #1620 for the working offline case, but that's not this bug).

Anyways, @shane-tomlinson, we need a way to coalesce the xhr polling.

shane-tomlinson:
@jrgm and @kparlante - could you review this?

There are two commits, the first is a test for the change that @ozten did a little while ago to strip GET data from the submitted URL.

The second is to consolidate consecutive XHR requests to the same URL into one entry.

@kparlante - when the event starts with xhr_complete, there may be a 4th item in the event's array - the number of consecutive times the xhr event was called. If the URL was only called once, the count is omitted.

fixes #3148

### DIFF
--- a/resources/static/common/js/modules/interaction_data.js
+++ b/resources/static/common/js/modules/interaction_data.js
@@ -31,6 +31,7 @@ BrowserID.Modules.InteractionData = (function() {
       storage = bid.Storage,
       complete = bid.Helpers.complete,
       dom = bid.DOM,
+      REPEAT_COUNT_INDEX = 3,
       sc;
 
   function removeGetData(msg, data) {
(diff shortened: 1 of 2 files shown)


kparlante:
+r looks good to me and worked when I tested it

PR #3182, opened 2013-04-03
Use the compile-json from i18n-abide instead of the one from the svn repo

### LINKED ISSUE #2492: Confirmation for removing last email always displays in English
When trying to remove my email from Persona (https://translate.personatest.org/) I get:

 Removing the last address will cancel your Persona account.
 Are you sure you want to continue?

This is actually translated for Estonian but never displays. Currently also seeing the same for pl and it for example.

shane-tomlinson:
@zaach - could you review this one to make sure I am using i18n-abide correctly?

Fixes the problem of "Removing the last address will cancel your Persona account" always being in English.

fixes #2492

### DIFF
--- a/scripts/l10n-update.js
+++ b/scripts/l10n-update.js
@@ -28,7 +28,7 @@ const svnRepo =
 const localePath = path.join(__dirname, '..', 'locale');
 
 // where compile script is located.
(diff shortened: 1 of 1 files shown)


zaach:
This looks good @shane-tomlinson, r+.

PR #3183, opened 2013-04-03
Fix button width in "Is this your computer" not keeping up with screen resize.

### LINKED ISSUE #2066: "Remember You" button text overflows on desktop when resized to mobile device size
STR
1) On a desktop browser, open the dialog to the email picker for an account that is older than one minute.
2) Open debug menu, select "force is this your computer"
3) Select an email and click next.
4) Resize dialog so that it is mobile size.

Button Text Overflow

shane-tomlinson:
jQuery changed the way to remove a manually set CSS style. We did not keep up when we updated jQuery. Now we do.

fixes #2066


jaredhirsch:
r+

PR #3186, opened 2013-04-03
Issue 3173 double escape rp

### LINKED ISSUE #3173: siteName is double escaped when redirecting to a primary.
From a user on our list:

> fyi - the slash in a site name now displays correctly, however, that same problem exists in BigTent for Yahoo.  For example, siteName "a/b" shows up as shown here:
> 
> yahoo.com makes this easy
> Persona lets you use your yahoo.com account to sign into sites like a&#x2F;b.
> Once you verify your account there, you will be signed in to a&#x2F;b.

Using the example RP:

Screen Shot 2013-04-02 at 17 02 29

shane-tomlinson:
The RP name was being doubly escaped in several locations before the user is redirected to the primary. Since the RP name is escaped when it comes in in dialog.js, a second escaping is unnecessary.

Because the same code affects IdP names, I am now escaping the IdP name when it leaves user.js, where it comes into the system.

fixes #3173.

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -1072,7 +1072,7 @@ BrowserID.User = (function() {
             User.isUserAuthenticatedToPrimary(normalizedEmail, info,
                 function(authed) {
               info.authed = authed;
-              info.idpName = getIdPName(info);
+              info.idpName = _.escape(getIdPName(info));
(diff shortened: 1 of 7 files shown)


jaredhirsch:
r+ reviewed yesterday forgot to merge

PR #3190, opened 2013-04-03
Train 2013.03.29  add underscore fix

jaredhirsch:
Cherry-picking the contents of #3168 into the current train. @jrgm have at it. Diffs look identical to me.


jrgm:
changes are the same I think. r+, merging into train-2013.03.29 and respinning.

Issue #3170, opened 2013-04-02
.onmatch called after .onlogin if user opens dialog.


shane-tomlinson:
STR

1) Visit development RP
2) If logged in, click logout
3) Force refresh the page (CTRL-SHIFT-R)
4) Open dialog
5) Log in
6) .onmatch is called after .onlogin

Before Login:
![Screen Shot 2013-04-02 at 13 09 43](https://f.cloud.github.com/assets/848085/328118/6a0e86ca-9b8e-11e2-8e5c-bf02169c5f76.png)

After Login:
![Screen Shot 2013-04-02 at 13 09 59](https://f.cloud.github.com/assets/848085/328119/7139ddaa-9b8e-11e2-8154-c8928f819fb0.png)

callahad:
Doh! Need more unit tests in that part of the code.

I wonder why that's happening....

shane-tomlinson:
I did a little digging, and the sequence of events is as follows:

1) Dialog opens, user signs in. Assertion is sent back to include.js
2) include.js sees the response contains an email and messages the communication_iframe with the "loggedInUser" message.  This message is asynchronous and will be delivered sometime in the future.
3) include.js messages the communication_iframe with the "dialog_complete" message. This message is asynchronous and will be delivered sometime in the future.
4) start.js continues on its way, it sees there is a login callback. It calls the login callback with the assertion.
5) the communication_iframe receives the "loggedInUser" message with the updated email. It sets its currently tracked email to the new email.
6) the communication_iframe receives the "dialog_complete" message. It is told to "checkAndEmit". checkAndEmit checks whether the user is authenticated. It sees the user is authenticated to the same user as in step 5. It calls onmatch.

shane-tomlinson:
@callahad - What is the expected behavior if:

1) User visits a site, site sends signedInUser: undefined.
2) Persona sees the user is signed in, generates an assertion and calls onlogin
3) User then opens dialog and selects a different address

Options:
1) .onlogout is called for first address, then .onlogin for new address
2) .onlogin is called for new address.

This can be tested with the current development RP.

callahad:
Thanks for the nice debugging!

In that scenario, I believe the expected behavior is 1:
1. Automatic `onlogin` for first address.
2. Automatic `onready`
3. Manual `onlogin` for second address.

PR #3194, opened 2013-04-04
Prevent .onmatch from being called if an assertion is generated in the dialog.

shane-tomlinson:
@lloyd - I would love to pow-wow on this. My "fix" is any time an assertion is generated by the dialog, do not then attempt to generate an assertion in the communication_iframe.

I am not convinced this is correct, but the behavior has survived IE8 with every combination of primary/secondary, single or multiple browsers that I could think of.

issue #3170

### DIFF
--- a/resources/static/communication_iframe/start.js
+++ b/resources/static/communication_iframe/start.js
@@ -117,11 +117,13 @@
     pause = true;
   });
 
-  chan.bind("dialog_complete", function(trans, params) {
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
This is failing the "broken-primary" selenium test. a self r-

shane-tomlinson:
.onmatch has more than just this problem. Closing until we can get them all out of the way at once.

PR #3195, opened 2013-04-04
Add a basic performance test to the selenium suite.

shane-tomlinson:
@fmarier, @lloyd, @kparlante, @jedp - this could use some review!

Releasing this to the wild to get review on the approach.
- Must be manually turned on for it to run.
- Requires an update to kpiggybank to be able to fetch the "last" result - https://github.com/mozilla/kpiggybank/pull/18
- Adds a new KPI that should probably go into the event stream - ready_time_ms. This is the amount of time it takes for the communication_iframe to load.

### DIFF
--- a/automation-tests/config/tests-to-ignore.js
+++ b/automation-tests/config/tests-to-ignore.js
@@ -7,17 +7,20 @@
 
 var testsToIgnore = {
   dev: [
-    "public-terminals.js"
+    "public-terminals.js",
+    "can-interact-test.js"
   ],
 
   stage: [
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
@fmarier - could you do a review on this?

fmarier:
I like your defensive approach to using this new `ready_time_ms` variable. If we stick to this kind of approach, we'll be able to introduce new things while making `include.js` cacheable.

shane-tomlinson:
I am closing this until I get further review. Will issue a new PR against my own repo to get feedback.

PR #3203, opened 2013-04-04
Add KPI load testing docs

jaredhirsch:
These used to live at https://gist.github.com/kparlante/5054439/
    and might be useful, so let's add 'em to the docs folder


jaredhirsch:
@kparlante thoughts?

Issue #3208, opened 2013-04-08
need tool to check a verifier can properly handle an assertion


jrgm:
In turning up AWS, where I have a verifier but not a functioning webhead, I needed a way to post an assertion the verifier and check the result. This would also be useful during deployments of trains as a check.

jrgm:
Closed by PR-3209. Verified.

jrgm:
Forgot to hit close. Doing so now.

PR #3209, opened 2013-04-08
builds on get_assertion to post to a verifier and check the response; useful for aws and other buildout

jrgm:
fixes https://github.com/mozilla/browserid/issues/3208

@lloyd can you check my math; mostly all stolen from you on get_assertion except the post to /verify.

### DIFF
--- a/scripts/verify-assertion.js
+++ b/scripts/verify-assertion.js
@@ -1,0 +1,166 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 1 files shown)


jaredhirsch:
@jrgm what would you think of adding a corresponding file to the /docs folder with some minimal overview/theory of operation?

jrgm:
So, there are 50 odd tools in ./scripts. Should each of those have a separate document in ./docs, or is `--help, -h` the right place for adhoc tools?

lloyd:
r+

PR #3213, opened 2013-04-09
Add tabindex attributes to the TOS/PP links.

### LINKED ISSUE #3082: placeholder issue - remove lowercasing in automation-tests/lib/restmail.js when GH-2551 is fixed
Putting this as an explicit bug, instead of a comment buried in the code. 

When GH-2551 is fixed, that will need to include removing the workaround at
https://github.com/mozilla/browserid/blob/d0f8ef43a74b110d4f91a991d4a42a7d6ca7afc5/automation-tests/lib/restmail.js#L16

shane-tomlinson:
Allows users to tab directly from the email/password field to the submit button.

fixes #3082


shane-tomlinson:
Closing because mis-labeled the branch

PR #3215, opened 2013-04-09
Force focus of either a button or an input if available when any screen is shown.

### LINKED ISSUE #2981: <Your email provider> makes this easy screen is not keyboard accessible.
Before being redirected to the IdP, the "Your email provider makes this easy" screen does not have a key selected, it is not possible to hit "enter"

shane-tomlinson:
If a button or input box is not available on a screen, no change from the current behavior.

fixes #2981

### DIFF
--- a/resources/static/common/js/lib/dom-jquery.js
+++ b/resources/static/common/js/lib/dom-jquery.js
@@ -322,7 +322,22 @@ BrowserID.DOM = ( function() {
          * @param {selelector || element} elementToFocus
          */
         focus: function( elementToFocus ) {
-          jQuery( elementToFocus ).focus();
+          try {
+            jQuery( elementToFocus ).focus();
+          } catch(e) {
+            // IE8 is awesome and can except if an element it doesn't believe
+            // can be focused is attempted to be focused.
+            if (window.console) window.console.log(String(e));
+          }
+        },
+
+        /**
+         * Blur an element
+         * @method blur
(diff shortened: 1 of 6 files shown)


jaredhirsch:
We're doing this in the name of better keyboard accessibility, right?

So, if we make sure that every form's most important element has tabindex=1, then I think a simpler fix would be, [inside PageModule.renderForm](https://github.com/mozilla/browserid/blob/dev/resources/static/common/js/modules/page_module.js#L62-L66), to just do `dom.focus( $("[tabindex=1]") )`.

My concern with this patch is that, every time you change screens, it [traverses the DOM contents 2x](https://github.com/mozilla/browserid/pull/3215/files#L2R29) to find focusable inputs and buttons. This might impact performance on low-end devices or older computers (not sure but suspicious).

Another approach: what if, as part of the dialog API, we nominate a specific input/button to always receive focus on load? That keeps the logic explicitly in the JS (less magical than implicitly fishing for for focusable els) and avoids the DOM traversals.

jaredhirsch:
@shane-tomlinson forgot to name-check you in my comments.

shane-tomlinson:
@6a68 - I dig your suggestion, I'll have a good/deep look at this unless you want to have a look.

shane-tomlinson:
At @6a68's great analysis, I am closing this for a better solution.

PR #3217, opened 2013-04-09
set agent:false for requests

jrgm:
A few other places where we should be agent:false /cc @lloyd


lloyd:
r+ - extremely low risk - may be a significant performance improvement.  Let's keep an eye on the loadgen breaking point this time around.

lloyd:
```
[16:08:20] <jrgm>    lloyd: yep. Click that button.
```

Issue #3221, opened 2013-04-10
check_primary_support fails silently unless ephemeral keypair present


callahad:
This is the second report I've seen of check_primary_support stalling and never returning info: https://twitter.com/Lukasaoz/status/320967071934603264

Lukasa:
I got this failure on Mac OS X 10.8.2 (Mountain Lion), using Node version v0.10.3 (which is the version you get if you install from Homebrew).

jrgm:
Works for me on either hand-rolled 0.10.3 or 0.8.17 on Mac OS X 10.7.5 (which isn't a "no you're wrong", but just "need a better way to narrow this down").

Lukasa:
To be totally clear, I'm not very familiar with Node at all, so it's quite possible that I've done something stupid. I don't think I did though :smile:.

When I stepped through this in the debugger, execution went basically as I expected to [here](https://github.com/mozilla/browserid/blob/dev/lib/primary.js#L149), and then it never hit the callback [here](https://github.com/mozilla/browserid/blob/dev/lib/primary.js#L107).

seanmonstar:
Could it be like we've seen just recently on the list, an SSL error? Maybe
one of you guys have the cert cached on your Mac, and the others don't?

Lukasa:
Do you have a link to the issue on the list? I'm not subscribed to it so I can't really say.

callahad:
@Lukasa A quick way to rule out sean's suggestion would be to try to run check_primary_support against a known-good primary, like eyedee.me. Does that fail in the same way for you?

Lukasa:
It does, yes.

callahad:
Cool. I suspect a node 0.10 issue, since most of us are still on 0.8... but haven't had a chance to actually try upgrading myself

Lukasa:
When I get some spare time, I'll see if installing 0.8 fixes my issue. =)

jrgm:
Yeah, I missed that the "fail" was that it never returns. Missed that. With 0.10.3 I see that. The answer comes back okay, but it never exits. Or actually, if you wait long enough, it will exit. Anyways, looking at the code, I'm not sure how this ever worked ;-). It's registering listeners on the request not the response.

Lukasa:
Hmm. Installing v0.8 hasn't fixed the issue, and neither has updating to the latest HEAD (which presumably includes #3185. Any further thoughts about areas I might look into?

callahad:
[...]

PR #3185, opened 2013-04-03
Fix check_primary_support to read responses

robottaway:
Hi been working on implementing and IDP for my site, was using the check_primary_support script and found it was not reading the responses when checking for script tags in the provision and auth endpoints. May be worth taking a look at my fixes to make it work for me.

-rob

PR #3226, opened 2013-04-10
listen for data and end on the response not the request

jrgm:
I had to do something during the meetings ;-). Anyways, this reworks check_primary_support to listen on right objects for certain events, and pulls the logic for checkResource into the end event handler of the request. Hope I didn't change what was intended to work (although did allow for using browserid.org instead of login.persona.org, although maybe I should back that out).

Fixes https://github.com/mozilla/browserid/issues/3221


jrgm:
Oh, and now I see @robottaway put in a PR to address the same stuff - https://github.com/mozilla/browserid/pull/3185

Closing this PR then.

PR #3234, opened 2013-04-11
Asynchronously load bidbundle for performance wins!

shane-tomlinson:
@fmarier - could you review?
- Add a CryptoLoader module that loads jwcrypto on demand.
- Update the build script to replace '/production/bidbundle.js' references in the JS with the cachified version.
- Force bidbundle.js to be compressed before everything else.

I have a bundle of numbers to show performance improvements: http://www.pastebin.mozilla.org/2294467

Furthering the goals set out in issue #3060


resources/static/common/js/user.js
```diff
@@ -1072,65 +1063,70 @@ BrowserID.User = (function() {
...
-                function(authed) {
-              info.authed = authed;
-              info.idpName = _.escape(getIdPName(info));
+      else {
+        network.addressInfo(email, function(info) {
+          // update the email with the normalized email if it is available.
+          // The normalized email is stored in the cache.
+          var normalizedEmail = info.normalizedEmail || email;
+          info.email = normalizedEmail;
+          User.checkEmailIssuer(normalizedEmail, info, function(info) {
```
fmarier: This line uses two different variables called "info". It's a bit confusing and I'm afraid it might lead to mistakes later on.

How about renaming the inner "info" to something else? issuerInfo?

resources/static/common/js/user.js
```diff
@@ -313,10 +308,6 @@ BrowserID.User = (function() {
 
 
   function onContextChange(msg, context) {
-    // seed the PRNG
-    prepareDeps();
-    jwcrypto.addEntropy(context.random_seed);
```
fmarier, line 318: So now we only add entropy the first time that `jwcrypto` is loaded.

Is that a significant change from before? How often does `onContextChange` fire?

shane-tomlinson, line 318: /wsapi/context_info was only ever called once, and then cached. onContextChange was called the one time /wsapi/context_info was returned. This should be a net 0 change.

resources/static/common/js/user.js
shane-tomlinson: Good idea.

resources/static/common/js/crypto-loader.js
```diff
@@ -0,0 +1,70 @@
...
+      network = bid.Network,
+      RECHECK_DELAY_MS = 50;
+
+  function addScript(src) {
+    var script = document.createElement("script");
+    script.setAttribute("src", src);
+    document.head.appendChild(script);
+  }
+
+  function waitUntilExists(checkFor, context, done) {
```
jaredhirsch, line 24: @shane-tomlinson Not a huge deal for this specific issue, but a lot of this file is a one-off implementation of a deferred. If we have jQuery on the page, we could use $.Deferred to separate the app logic from the flow control logic.

I took an hour timebox and [implemented a version](https://github.com/6a68/browserid/blob/be9a7b6ed5523a94b38e341877641220dda88901/resources/static/common/js/crypto-loader.js) that uses Deferreds--not necessarily for this patch, but just to stimulate conversation. (I ran out of time before I had a chance to test my code--I can go another hour if you want to dig in.)

jaredhirsch, line 24: @fmarier pointed out in IRC that we're trying to remove jQuery from the front-end code as much as possible.

Without laboring too much on this promises thing, just want to mention that [Q](https://github.com/kriskowal/q) is ~3KB minified + gzipped.

shane-tomlinson, line 24: @6a68 - if either the deferreds can be abstracted into their own standalone unit so that we can replace their backing implementation (the jQuery issue), I'm down for replacing this code.

lloyd:
what's the bottom line of those numbers?  can you summarize the complexity/performance tradeoff?

jaredhirsch:
@lloyd bidbundle is ~24KB minified + gzipped. Pulling that out of the initial time-to-content should produce savings depending on how many kbps your connection supports.

As for the empirical data, I've munged @shane-tomlinson's numbers but haven't yet graphed them. Here's my work so far: https://gist.github.com/6a68/5408839

jaredhirsch:
@lloyd quickly updating with the differences in the averages from @shane-tomlinson's test runs:

### ~12% improvement with empty cache on 2G and 3G.

#### This is over 2 seconds saved on 2G, not quite a full second on 3G.

### no significant change with primed cache.

here are the numbers from the updated gist:

``` javascript
// doing simple math, dodging calculation of SD of the differences,
// we can see that the improvement is significant on unprimed cache,
// and there is no significant change with primed cache.

// let's look at time to 'can_interact':
// 2g, empty cache: 19950 - 17665 = 2285, 11% faster (over two seconds faster!)
// 3g, empty cache: 6692 - 5826 = 866, 13% faster
//
// there is no net change with prime
// 2g, primed cache: 3091 - 3140 = -49, actually 1.5% slower--except that our margin of error is certainly bigger than this
// 3g, primed cache: 1381 - 1373 = 8, 
```

resources/static/common/js/crypto-loader.js
jaredhirsch, line 24: @shane-tomlinson and I discussed this yesterday. I'll open a separate issue for the inclusion of deferreds here and avoid blocking the PR on that count.

lloyd:
r+ - generally I'm nervous at this point about further effort into performance.  I think it will be pretty trivial to get CDN up, and I wonder how that might stack up.  here we're talking about 140 lines of code for a 12% improvment in some cases.

I'm not going to stop this tho.  I'm just asking us to step back and explain what our goal is with performance.  When is it good enough?  Where are we right now?

PR #3235, opened 2013-04-11
B2g dev force authentication

shane-tomlinson:
@jedp, @ozten - The first one! Can you review?

This is the initial checkin for merging forceAuthentication into dev. I have _only_ merged the feature as it stood before, with some QUnit tests. I have not yet done any work to place this option under an "experimental" namespace.

### DIFF
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -259,8 +259,8 @@ BrowserID.Modules.Dialog = (function() {
 
         // forceAuthentication is used by the Marketplace to ensure that the
         // user knows the password to this account. We ignore any active session.
-        if (paramsFromRP.forceAuthentication &&
-            true === paramsFromRP.forceAuthentication) {
+        if (paramsFromRP._experimental_forceAuthentication &&
+            true === paramsFromRP._experimental_forceAuthentication) {
           params.forceAuthentication = true;
         }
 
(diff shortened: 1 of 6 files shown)


jedp:
I was talking over the experimental namespace idea with Ben yesterday.  I like the namespace, which we can do with no modification of the b2g platform code (since we've built the internal api to pass arbitrary arguments through).  But if we add something like navigator.id.enableExperimentalThings(), then we do have to touch the b2g code, and that's opening a huge can of worms at this point.

I'll bring the idea to the public list now, but I propose we rename `forceAuthentication` to something like `_experimental_forceAuthentication`.  That seems pretty loud and clear to me.

callahad:
I'm +1 to boxing that option into `_experimental_`

resources/static/dialog/js/modules/dialog.js
```diff
@@ -257,6 +257,13 @@ BrowserID.Modules.Dialog = (function() {
           user.setReturnTo(returnTo);
         }
 
+        // forceAuthentication is used by the Marketplace to ensure that the
+        // user knows the password to this account. We ignore any active session.
+        if (paramsFromRP._experimental_forceAuthentication &&
+            true === paramsFromRP._experimental_forceAuthentication) {
+          params.forceAuthentication = true;
```
jedp, line 264: so I'm scrubbing `_experimental_forceAuthentication` (param provided by RP) to `forceAuthentication` for internal use.

shane-tomlinson, line 264: I didn't realize members starting with a "_" could be accessed directly. Nice.

Issue #3202, opened 2013-04-04
Need a way to supress alert dialogs to reduce webdriver test failures


edwindotcom:
There's a known issue with alert dialogs on opera and safari (possibly more).  If we could have a query string param such as: 123done.org?noAlert=true that would suppress these errors, it would greatly improve our test stability.  Not sure of the implementation or security issues involved.

I'm on mac10.8 with safari installed:
1. run java -jar selenium-server-standalone-2.31.0.jar
2. run this from automated_tests/
scripts/run-all.js --local --platform=safari --tests=remove-email -e stage

result: safari eats the alert dialog
on saucelabs:
https://saucelabs.com/tests/bb1d9d74805549ecbd971e56a070a9c6

On my local isolated click alert test, i get this in stdout:
selenium.common.exceptions.WebDriverException: Message: u"A modal dialog was opened. The SafariDriver does not support interacting with modal dialogs. To avoid hanging your test, the alert has been dismissed. For more information, see http://code.google.com/p/selenium/issues/detail?id=3862

jaredhirsch:
Maybe try hacking on this a bit
https://github.com/mozilla/browserid/blob/dev/automation-tests/lib/urls.js

edwindotcom:
I'd be happy with crew:automated-tests, there are only 3 open and 34 closed so there won't be too much collision.

shane-tomlinson:
@skinny97214 says that using alerts is bunk. We should create our own lightbox with a [Sandstone](https://www.mozilla.org/b/en-US/sandstone/) style.

edwindotcom:
for a one two punch, I'm going to work on adding another option to run-all.js --ignore-test=remove-email that way we can omit tests per browser/os config.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #3236, opened 2013-04-11
B2g dev tos pp

shane-tomlinson:
@jedp, @seanmonstar - can I get a review on this?

This is part 2 of the B2G->Dev merge.

I added a new RP parameter: inlineTermsOfService to indicate whether the TOS/PP agreements should be shown inline. Like the other PR, I have not yet put this under any sort of experimental namespace until we agree on an approach.


jedp:
Hm.  Front-end tests seem to be failing.  I'll take a look.

jedp:
I'm getting some intermittent failures.  First time I got most of cach-header-test failing; second time everything was green.

shane-tomlinson:
I'd like to get some input from @skinny97214 on how this should look, perhaps we can do this for the normal Persona flow as well that way we aren't special casing this for B2G.

shane-tomlinson:
I asked @skinny97214 about error messages yesterday for #3202. She said that we should come up with a generic whitebox using the [Sandstone](https://www.mozilla.org/b/en-US/sandstone/) style. We could use that same whitebox here.

shane-tomlinson:
I cannot figure out why phantom on travis is crashing, but it is not locally.

shane-tomlinson:
Fixed by #3279

resources/static/dialog/css/m.css
```diff
@@ -226,5 +227,30 @@
...
+  #tosppmodal .close {
+    color:#fff;
+    display:block;
+    font-size:2em;
+    font-weight:bold;
+    height:6%;
+    padding:5px;
+  }
+
+  #tosppframe {
```
shane-tomlinson: We use tosppframe here and .lightbox below. Pick one.

resources/views/dialog.ejs
```diff
@@ -73,9 +73,9 @@
                        <%- format(gettext("By proceeding, you agree to %(site)s's <a %(terms)s>Terms</a> and <a %(privacy)s>Privacy Policy</a>."),
                                   { site: "Persona",
                                     terms:
-                                        'href="https://login.persona.org/tos" target="_new" tabindex="1"',
+                                        'href="https://login.persona.org/tos?embedded" target="_new" tabindex="1"',
                                     privacy:
-                                        'href="https://login.persona.org/privacy" target="_new" tabindex="1"' }) %>
+                                        'href="https://login.persona.org/privacy?embedded" target="_new" tabindex="1"' }) %>
```
shane-tomlinson: In systems that still open the TOS/PP text in a new tab, this will show them the text without the headers and footers.

shane-tomlinson:
Things aren't looking quite right in IE8, it's like the contents are not respecting the IFRAME width.

Screen Shot 2013-04-29 at 10 41 36

resources/static/dialog/css/m.css
```diff
@@ -129,6 +129,7 @@
 
   #formWrap {
       background-color: transparent;
+      overflow:auto;
```
shane-tomlinson: This will make it so the formWrap can scroll, I do not think we want that.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -231,6 +231,10 @@ BrowserID.Modules.Dialog = (function() {
           self.publish("kpi_data", { rp_api: rpAPI });
         }
 
+        if (paramsFromRP._experimental_inlineTermsOfService === true) {
```
shane-tomlinson: This needs a better boolean check.

resources/static/dialog/js/modules/inline_tospp.js
```diff
@@ -0,0 +1,91 @@
...
+    if (!self._tospp) {
+      self._tospp = renderer.append(IFRAME_PARENT_SELECTOR, "inline_tospp", {
+        no_iframe: self.options.no_iframe
+      });
+      self.click(TOSPP_CLOSE_SELECTOR, closeTOSPP, self);
+    }
+
+    // If a TOS/PP agreement is shown inline, give the host a hint of the
+    // display context, it may want to not display headers/footers.
+    var url = url + '?embedded';
```
shane-tomlinson: This is brittle if the RP already specifies query parameters.

shane-tomlinson: @jedp, @seanmonstar - what is a good way of hinting to the RP that we are showing their TOS/PP agreement in an inline context? Using query parameters?

shane-tomlinson:
The IE8 scrollbar problem is caused by the `width: 896px` declaration on line 102 of `pages/css/style.css`

resources/static/dialog/js/modules/dialog.js
shane-tomlinson: I have a better boolean check in the b2g_dev_allow_unverified branch

shane-tomlinson:
@jedp and I discussed how whether we need a mechanism to surface to the RP "hey, your TOS/PP is being shown in a mobile context and has neither JS nor links". At the moment, I am appending a `?embedded` query parameter onto the URL, but this feels hacky and dirty. Is something like this needed?

How should we handle the Persona TOS/PP links that show neither the header nor footer - do we use some sort of query parameter on the URL or come up with a new url like `https://login.persona.org/embedded_tos`?

shane-tomlinson:
Also, we could use some styling pointers.

resources/static/dialog/js/modules/dialog.js
seanmonstar: Better boolean? How do you get better than strict equality?

resources/static/dialog/js/modules/inline_tospp.js
```diff
@@ -0,0 +1,91 @@
...
+    if (!self._tospp) {
+      self._tospp = renderer.append(IFRAME_PARENT_SELECTOR, "inline_tospp", {
+        no_iframe: self.options.no_iframe
+      });
+      self.click(TOSPP_CLOSE_SELECTOR, closeTOSPP, self);
+    }
+
+    // If a TOS/PP agreement is shown inline, give the host a hint of the
+    // display context, it may want to not display headers/footers.
+    url = url + '?embedded';
```
seanmonstar: This includes URLs provided by the RP? SO it could end up being `myrp.com?tos?embedded`

resources/static/dialog/js/modules/dialog.js
shane-tomlinson: By throwing an error like: https://github.com/mozilla/browserid/pull/3307/files#L23R162

resources/static/dialog/js/modules/inline_tospp.js
shane-tomlinson: yeah, this is one of the things I was really worried about. After talking with @jedp and @lloyd, I am going to continue with this PR under the assumption that _all_ tos/pp agreements will be displayed in the iframe and that no special hint is needed.

PR #3237, opened 2013-04-11
B2G->dev merge - Update helpers.log to handle multiple arguments. Objects will be JSON.stringified.

shane-tomlinson:
@jedp - Can you review?

### DIFF
--- a/resources/static/common/js/helpers.js
+++ b/resources/static/common/js/helpers.js
@@ -64,8 +64,24 @@
     }
   }
 
-  function log(msg) {
+  function log() {
     try {
+      var msg = "";
+      var args = [].slice.call(arguments, 0);
+      _.each(args, function(arg, index) {
+        if (index > 0) msg += " ";
+
+        var type = Object.prototype.toString.apply(arg);
+        if (type === "[object String]") {
+          msg += arg;
+        }
+        else if (type === "[object Object]") {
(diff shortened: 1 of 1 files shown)


resources/static/common/js/helpers.js
```diff
@@ -64,8 +64,24 @@
...
+
+        var type = Object.prototype.toString.apply(arg);
+        if (type === "[object String]") {
+          msg += arg;
+        }
+        else if (type === "[object Object]") {
+          msg += JSON.stringify(arg, null, 2);
+        }
+        else {
+          msg += String(arg);
```
jedp, line 90: If one of the args is a callback, this will print the complete source code of the callback.  Maybe functions should just be printed as `<<function>>` or something?

resources/static/common/js/helpers.js
```diff
@@ -64,8 +64,24 @@
     }
   }
 
-  function log(msg) {
+  function log() {
     try {
+      var msg = "";
+      var args = [].slice.call(arguments, 0);
+      _.each(args, function(arg, index) {
+        if (index > 0) msg += " ";
```
jedp: A nit, but maybe `msg += ", "` ?

jedp:
@shane-tomlinson back atcha!  Changes ok?

shane-tomlinson:
hot hot, I'll check this and then merge.

PR #3238, opened 2013-04-11
B2g dev prepare internal api

shane-tomlinson:
@zaach & @jedp - Can you review this one?

This prepares the internal API for B2G. There is a set of tests that go along with this, but those must be merged once we have updated user.js to have "setIssuer"

### DIFF
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -175,6 +175,8 @@
         user = bid.User,
         storage = bid.Storage;
 
+    network.init();
+
     log('internal watch options', options);
     var remoteOrigin = options.origin;
     var loggedInUser = options.loggedInUser;
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -178,6 +178,7 @@
     log('internal watch options', options);
     var remoteOrigin = options.origin;
     var loggedInUser = options.loggedInUser;
+    user.setOrigin(remoteOrigin);
 
     function checkAndEmit() {
       log('checking and emitting');
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -209,6 +209,7 @@
       }, log);
     }
 
+    network.clearContext();
     checkAndEmit();
 
     function doReady (params) {
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -163,4 +163,78 @@
     user.logoutUser(callback, complete.curry(null));
   };
 
+  internal.watch = function (callback, options, log) {
+    if (typeof options === 'string') options = JSON.parse(options);
+    internalWatch(callback, options, log);
+  };
+
+
+  function internalWatch (callback, options, log) {
+    var bid = BrowserID,
+        network = bid.Network,
+        user = bid.User,
+        storage = bid.Storage;
+
+    // Initialize all localstorage values to default values.  Neccesary for
+    // proper sync of IE8 localStorage across multiple simultaneous
+    // browser sessions.
+    storage.setDefaultValues();
+
+
+    log('!!!!!', options);
+    var remoteOrigin = options.origin;
+    user.setOrigin(options.origin);
+    var loggedInUser = user.getOriginEmail();
+
+    log('origin: ', remoteOrigin);
+    log('loggedInUser: ', loggedInUser);
+    log('callback: ', typeof callback);
+
+    function checkAndEmit() {
+      log('checking and emitting');
+      // this will re-certify the user if neccesary
+      user.getSilentAssertion(loggedInUser, function(email, assertion) {
+        log('silent return: email, assertion ', email, assertion);
+        if (email) {
+          // only send login events when the assertion is defined - when
+          // the 'loggedInUser' is already logged in, it's false - that is
+          // when the site already has the user logged in and does not want
+          // the resources or cost required to generate an assertion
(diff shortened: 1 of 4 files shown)


resources/static/dialog/js/misc/internal_api.js
```diff
@@ -61,6 +61,13 @@
    * options.silent defaults to false.
    */
   internal.get = function(origin, callback, options) {
+    if (typeof options === 'string') {
+      // Firefox forbids sending objects across the blood-brain barrier from
+      // gecko into userland JS.  So we just stringify and destringify our
+      // objects when calling from b2g native code.
+      options = JSON.parse(options);
```
jedp: Maybe throw a `try ... catch` around this and log any weird strings that come from firefox land?

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +170,73 @@
     user.logoutUser(callback, complete.curry(null));
   };
 
+  internal.watch = function (callback, options, log) {
+    if (typeof options === 'string') options = JSON.parse(options);
```
jedp: Ditto for `try ... catch`, or refactor into a `parseOptions` function that returns object or null?

jedp:
Apart from questions about `try ... catch` around `JSON.parse`, this is looking good to me!

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +170,73 @@
...
+
+  function internalWatch (callback, options, log) {
+    var bid = BrowserID,
+        network = bid.Network,
+        user = bid.User,
+        storage = bid.Storage;
+
+    network.init();
+
+    log('internal watch options', options);
```
shane-tomlinson: Can we safely remove all of these logs before we merge?

shane-tomlinson:
@jedp - I added parseOptions, which logs the invalid JSON. Should the callback to .watch be called with some sort of error message when this happens?

shane-tomlinson:
@jedp - aside from the logs all over the place, I am comfortable with this. Should we take those out? I have made the changes for parseOptions. What more needs done for this one to get the initial PR in? The updates dealing with forceIssuer I would rather handle in that PR separately.

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +186,78 @@
...
+    } catch(e) {
+      // should we call the callback with an error message?
+      return;
+    }
+    internalWatch(callback, options, log);
+  };
+
+
+  function internalWatch (callback, options, log) {
+    var bid = BrowserID,
```
shane-tomlinson: The external dependencies should be moved to the top of the file.

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +186,78 @@
...
+        doReady();
+      }, function(err) {
+        log('silent return: err', err);
+        doLogout();
+        loggedInUser = null;
+        doReady();
+      }, log);
+    }
+
+    network.clearContext();
```
shane-tomlinson: These two function calls live on their own down here and are easy to miss, they should be moved up before checkAndEmit

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +186,78 @@
...
+    checkAndEmit();
+
+    function doReady (params) {
```
shane-tomlinson: spacing before the (

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +186,78 @@
...
+      log('doReady');
+      callback({ method: 'ready' });
+    }
+
+    function doLogin (params) {
```
shane-tomlinson: spacing before the (

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +186,78 @@
...
+      log('doLogin (with silent assertion)');
+      // Through the _internalParams, we signify to any RP callers that are
+      // interested that this assertion was acquired without user interaction.
+      callback({ method: 'login', assertion: params, _internalParams: {silent: true} });
+    }
+
+    function doLogout () {
```
shane-tomlinson: spacing before the (

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +186,78 @@
...
+      log('doLogout');
+      if (loggedInUser !== null) {
+        storage.setLoggedIn(remoteOrigin, false);
```
shane-tomlinson: This needs to be updated for the storage simplification that occurred. storage.setLoggedIn no longer exists, this is now taken care of by storage.site.set

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -163,4 +187,68 @@
     user.logoutUser(callback, complete.curry(null));
   };
 
+  internal.watch = function (callback, options, externalLog) {
+    log = externalLog || bid.Helpers.log;
+
+    try {
+      options = parseOptions(options);
+    } catch(e) {
+      // should we call the callback with an error message?
```
jedp: That would be helpful for the caller.  We callback with an object; in this case, we could `callback({error: e.toString()})` and then throw?

shane-tomlinson: I think that would be extremely useful.

jedp:
@shane-tomlinson this is looking good to me.  I made the changes to the `internal.watch` callback you suggested; it now sends the error text if there was a problem parsing the options.

Since this feature is new, I'm comfortable leaving the logs in, with the intention of removing them in the future.

What do you think?  Push the Big Green Button?

PR #3240, opened 2013-04-11
B2g dev force issuer

shane-tomlinson:
@jedp and I are pow-wowing on this.


config/local.json
```diff
@@ -17,7 +17,7 @@
 
   // whether to show the development menu.
   "enable_development_menu": true,
-
   // enable the measurement of the dom loading
-  "measure_dom_loading": true
+  "measure_dom_loading": true,
+  "forcable_issuers": ["issuer.domain"]
```
jedp: I sort of think this should be spelled `forceable_issuers` (not that "forceable" is an English word, either, but it somehow seems better to me.  Interestingly, this becomes one of those words that means two opposite things, like "oversight."  Well, by sound only, I guess.  Forcible vs Forceable)

resources/static/common/js/user.js
```diff
@@ -828,7 +875,7 @@ BrowserID.User = (function() {
      * @param {function} [onFailure] - Called on XHR failure.
      */
     requestTransitionToSecondary: function(email, password, onComplete, onFailure) {
-      User.addressInfo(email, function(info) {
+      User.addressInfo(email, user.forceIssuer, function(info) {
```
jedp: Should this be `User.forceIssuer`?

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -91,6 +90,13 @@ BrowserID.Modules.Authenticate = (function() {
       else if ("primary" === info.type) {
         self.close("primary_user", info, info);
       }
+      else if (!!self.forceIssuer && 'default' !== self.forceIssuer) {
+	if (hasPassword(info)) {
+          enterPasswordState.call(self);
+	} else {
+          createFxAccount.call(self, self.forceIssuer);
+	}
```
jedp: Some funky indentation here

resources/static/dialog/views/set_password.ejs
```diff
@@ -10,13 +10,15 @@
       <ul class="inputs">
           <li>
             <% if (!password_reset) { %>
-              <% if (transition_no_password) { %>
-                <%- format(gettext("%(idp)s no longer allows you to log into Persona with your %(idp)s password. Please create a new password to use with your Persona account."), {
+              <% if (fxaccount) { %>
+                <%= gettext("Your email address is new to us. Please create a password to use with FirefoxOS.") %>
+            <% } else if (transition_no_password) { %>
+                <%- format(gettext("%(idp) no longer allows you to log into Persona with your %(idp) password. Please create a new password to use with your Persona account."), {
```
jedp: I think this wants `%(idp)s` instead of `%(idp)` in each case?

shane-tomlinson: That it is, the fix for that hasn't been merged yet. Will do that.

config/local.json
shane-tomlinson: When I type in "forceable" to Google, it asks "Did you mean forcible?" m-w.com - http://www.merriam-webster.com/dictionary/forceable. Either way, forcable is not the future.

resources/static/common/js/user.js
```diff
@@ -909,6 +936,7 @@ BrowserID.User = (function() {
      * Sync local identities with login.persona.org.  Generally should not need to
      * be called.
      * @method syncEmails
+     * @param {string} - TODO
```
shane-tomlinson: This TODO needs a real entry.

resources/static/common/js/user.js
```diff
@@ -1284,32 +1327,33 @@ BrowserID.User = (function() {
           }, 0);
         }
         else {
-          // first we have to get the address info, then attempt
-          // a provision, then if the user is provisioned, go and get an
-          // assertion.
-          User.addressInfo(email, function(info) {
-            if (info.type === "primary") {
+          // TODO what will the type of forceIssuer email addresses be?
```
shane-tomlinson: Is it a secondary?

resources/static/common/js/user.js
```diff
@@ -24,7 +24,8 @@ BrowserID.User = (function() {
       stagedEmail,
       stagedPassword,
       userid,
-      auth_status;
+      auth_status,
+      forceIssuer = "default";
```
shane-tomlinson: Rename "forceIssuer" to just "issuer".

resources/static/common/js/user.js
```diff
@@ -265,29 +266,35 @@ BrowserID.User = (function() {
...
-    _.extend(email_obj, {
-      updated: now,
-      pub: keypair.publicKey.toSimpleObject(),
-      priv: keypair.secretKey.toSimpleObject(),
-      cert: cert
-    });
-
-    storage.addEmail(email, email_obj);
-    if (onComplete) onComplete(true);
+    // XXX This needs to be looked at to make sure caching does not bite us.
```
shane-tomlinson, line 269: Here is an XXX. Will caching bite us?

resources/static/common/js/network.js
```diff
@@ -508,14 +512,15 @@ BrowserID.Network = (function() {
      * Certify the public key for the email address.
      * @method certKey
      */
-    certKey: function(email, pubkey, onComplete, onFailure) {
+    certKey: function(email, pubkey, forceIssuer, onComplete, onFailure) {
+      var opts = 'default' === forceIssuer ? {} : {forceIssuer: forceIssuer};
```
shane-tomlinson: For clarity, maybe set a default opts object first, and then add forceIssuer if it is needed.

resources/static/common/js/user.js
```diff
@@ -1038,6 +1085,7 @@ BrowserID.User = (function() {
      * (is it a primary or a secondary)
      * @method addressInfo
      * @param {string} email - Email address to check.
+     * @param {string} issuer - Force a specific Issuer by specifing a domain. null for default.
```
shane-tomlinson: Has issuer actually been added to the function signature or is this left over?

resources/static/dialog/js/misc/state.js
```diff
@@ -228,6 +240,10 @@ BrowserID.State = (function() {
       else if (self.transitionNoPassword) {
         redirectToState("stage_transition_to_secondary", info);
       }
+      else if(self.newFxAccountEmail) {
+        startAction(false, "doStageUser", info);
+// TODO         startAction(false, "doStageResetPassword", info); ???
```
shane-tomlinson, line 248: What is the doStageResetPassword?

resources/static/dialog/js/misc/state.js
```diff
@@ -343,9 +359,10 @@ BrowserID.State = (function() {
 
     handleState("email_chosen", function(msg, info) {
       var email = info.email,
-          record = storage.getEmail(email);
+          record = user.getStoredEmailKeypair(email);
 
-      self.email = email;
+      // Maybe use a second global variable so we know which email address was chosen?
+      self.email = user.forceIssuerEmail = email;
```
shane-tomlinson: An accessor function should be used to set the forceIssuerEmail - what is it used for anyways?

resources/static/dialog/js/misc/state.js
```diff
@@ -380,6 +397,23 @@ BrowserID.State = (function() {
           redirectToState("primary_user", info);
         }
       }
+      else if (!user.isDefaultIssuer() && !record.cert) {
+        // TODO: Duplicates some of the logic in the authentication action module.
```
shane-tomlinson, line 403: If this duplicates some of the logic from the authentication module, should this be moved there?

resources/static/dialog/js/modules/dialog.js
```diff
@@ -122,6 +122,12 @@ BrowserID.Modules.Dialog = (function() {
     return returnTo;
   }
 
+  function fixupIssuer(url) {
+    // XXX do something awesome here.
```
shane-tomlinson: We should indeed so something awesome here - like make sure the issuer is legit.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -264,6 +270,14 @@ BrowserID.Modules.Dialog = (function() {
           params.forceAuthentication = true;
         }
 
+        // forceIsuser is used by the Marketplace to disable primary support
+        // and replace fxos.login.persona.org as the issuer of certs
+        if (paramsFromRP._experimental_forceIssuer) {
+          // TODO check for valid domain
```
shane-tomlinson: This comment can be removed, we are calling fixupIssuer.

resources/static/dialog/views/set_password.ejs
```diff
@@ -10,13 +10,15 @@
...
+                <%= gettext("Your email address is new to us. Please create a password to use with FirefoxOS.") %>
+            <% } else if (transition_no_password) { %>
                 <%- format(gettext("%(idp)s no longer allows you to log into Persona with your %(idp)s password. Please create a new password to use with your Persona account."), {
                   idp: "<strong>" + escape(domain) + "</strong>"
                 }) %>
-              <% } else { %>
-                <%= gettext("Your email address is new to us. Please create a password to use with Persona.") %>
               <% } %>
+            <% } else { %>
+              <%= gettext("Your email address is new to us. Please create a password to use with Persona.") %>
```
shane-tomlinson: I no longer see this message whenever signing up as a new secondary user.

resources/static/dialog/js/modules/dialog.js
shane-tomlinson: How are we going to get a list of valid issuers?

shane-tomlinson:
@jedp, @benadida 

What should the behavior in this sequence be:

Sequence 1
1) User opens dialog from RP that has forceIssuer set
2) User creates account with forceIssuer
3) Site removes forceIssuer from .request call.

OR

Sequence 2
1) User opens dialog from RP that has forceIssuer set
2) User creates account with forceIssuer
3) User visits another RP that does not have forceIssuer set

Should the generated assertion in both of these cases have the "issuer" set if the address was a secondary to begin with?

lib/verifier/certassertion.js
```diff
@@ -105,6 +106,12 @@ function verify(assertion, audience, successCB, errorCB) {
       else if (config.get('disable_primary_support')) {
         return errorCB("this verifier doesn't respect certs issued from domains other than: " +
                        HOSTNAME);
+      } else if (issuer === forceIssuer) {
+        if (config.get('forcable_issuers').indexOf(forceIssuer) === -1) {
+          return errorCB("this verifier doesn't won't force issuer for " + forceIssuer);
```
shane-tomlinson: `doesn't won't`?

lib/verifier/certassertion.js
```diff
@@ -105,6 +106,12 @@ function verify(assertion, audience, successCB, errorCB) {
...
+        } else {
+          return next(null, publicKey);
```
shane-tomlinson, line 113: @jedp - I'm confused by this. If the forceIssuer is the same as the issuer (and an allowed issuer), we call next, but if the forceIssuer is not the same as the issuer, we go on to check delegation? Does that make sense? If an RP requests forceIssuer for a specific domain and then that domain delegates off to another domain, does that break the contract with the RP?

lib/wsapi/address_info.js
```diff
@@ -178,6 +198,10 @@ exports.process = function(req, res) {
 
     if (!emailKnown) {
       r.state = "unknown";
+    } else if ('default' !== issuer && hasPassword) {
+      r.state = "known";
+    } else if ('default' !== issuer && ! hasPassword) {
```
shane-tomlinson, line 205: It took me a few seconds to work through the mental gymnastics of this logic, a comment would be useful.

resources/static/common/js/storage.js
```diff
@@ -48,12 +48,20 @@ BrowserID.Storage = (function() {
   upgradeLoggedInInfo();
   // END TRANSITION CODE
 
-  function storeEmails(emails) {
-    storage.emails = JSON.stringify(emails);
+  function emailsStorageKey(issuer) {
+    if (issuer && issuer !== "default")
+      return 'forceIssuerEmails';
```
shane-tomlinson: If there are multiple issuers, all issuers will use the same namespace. This should namespace issuers individually.

resources/static/common/js/storage.js
```diff
@@ -48,12 +48,20 @@ BrowserID.Storage = (function() {
...
+      return 'emails';
+  }
+
+  function storeEmails(emails, issuer) {
+    storage[emailsStorageKey(issuer)] = JSON.stringify(emails);
   }
 
   function clear() {
     storage.removeItem("emails");
+    storage.removeItem("forceIssuerEmails");
```
shane-tomlinson: Likewise, all forceIssuerEmail namespaces should be removed.

resources/static/common/js/user.js
```diff
@@ -265,29 +266,35 @@ BrowserID.User = (function() {
...
-    });
-
-    storage.addEmail(email, email_obj);
-    if (onComplete) onComplete(true);
+    // XXX This needs to be looked at to make sure caching does not bite us.
+    // I would rather pass in the unverified flag so that caching does not
+    // cause problems here.
+    User.addressInfo(email, function(info) {
+      var now = new Date();
+      var email_obj = storage.getEmails(forceIssuer)[email] || {
```
shane-tomlinson: Use storage.getEmail here.

resources/static/common/js/user.js
```diff
@@ -300,13 +307,13 @@ BrowserID.User = (function() {
    * @param {string} options.type - Is the email a 'primary' or a 'secondary' address?
    * @param {string} options.verified - If the email is 'secondary', is it verified?
    */
-  function persistEmail(options) {
+  function persistEmail(options, issuer) {
     storage.addEmail(options.email, {
-      created: new Date()
-    });
+      created: new Date(),
+      verified: options.verified
```
shane-tomlinson: It looks like allowUnverified has crept into here.

resources/static/common/js/user.js
```diff
@@ -300,13 +307,13 @@ BrowserID.User = (function() {
    * @param {string} options.type - Is the email a 'primary' or a 'secondary' address?
    * @param {string} options.verified - If the email is 'secondary', is it verified?
    */
-  function persistEmail(options) {
+  function persistEmail(options, issuer) {
```
shane-tomlinson: Rename this persistEmailWithoutKeypair for clarity.

resources/static/common/js/user.js
```diff
@@ -1063,29 +1111,28 @@ BrowserID.User = (function() {
...
-                  cleanedInfo.idpName = getIdPName(cleanedInfo);
-                  complete(cleanedInfo);
-                }, onFailure);
+
+      network.addressInfo(email, forceIssuer, function(info) {
+        // update the email with the normalized email if it is available.
+        // The normalized email is stored in the cache.
+        var normalizedEmail = info.normalizedEmail || email;
+        info.email = normalizedEmail;
+        User.checkEmailIssuer(normalizedEmail, info, function(issuerInfo) {
```
shane-tomlinson: rename issuerInfo back to cleanedInfo, this was a botch on my part.

resources/static/dialog/js/misc/state.js
```diff
@@ -117,6 +117,8 @@ BrowserID.State = (function() {
       self.hostname = info.hostname;
       self.siteName = info.siteName || info.hostname;
       self.siteTOSPP = !!(info.privacyPolicy && info.termsOfService);
+      var forceIssuer = (!!info.forceIssuer ? info.forceIssuer : 'default');
+      user.setIssuer(forceIssuer);
```
shane-tomlinson: For consistency, this should probably only be set if forceIssuer is not default.

@jedp - How do we handle when an RP changes its forceIssuer? If the communication_iframe has its forceIssuer set, but the dialog does not, there will be a discrepancy between their two views of the world.

resources/static/dialog/js/misc/state.js
shane-tomlinson: @jedp - I cannot find where this is supposed to be used, any ideas?

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -127,6 +132,18 @@ BrowserID.Modules.Authenticate = (function() {
     }
   }
 
+  function createFxAccount(callback) {
+    /*jshint validthis: true*/
+    var self=this,
+        email = getEmail();
+
+    if (email) {
+      self.close("new_fxaccount", { email: email, fxaccount: true }, { email: email });
```
shane-tomlinson: the "fxaccount: true" is implicit in "new_fxaccount".

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -93,6 +91,13 @@ BrowserID.Modules.Authenticate = (function() {
       else if ("primary" === info.type) {
         self.close("primary_user", info, info);
       }
+      else if (!user.isDefaultIssuer()) {
```
shane-tomlinson: This path needs a test.

resources/static/dialog/js/misc/state.js
```diff
@@ -380,6 +397,23 @@ BrowserID.State = (function() {
           redirectToState("primary_user", info);
         }
       }
+      else if (!user.isDefaultIssuer() && !record.cert) {
+        // TODO: Duplicates some of the logic in the authentication action module.
+        user.addressInfo(info.email, function (serverInfo) {
+          // We'll end up in this state again, but we want to see serverInfo.state change
+          user.resetCaches();
+          if (serverInfo.state === "transition_no_password") {
+            var newInfo = _.extend(info, { fxaccount: true });
```
shane-tomlinson, line 408: newInfo is created but then never used.

shane-tomlinson, line 408: This indicates a test is needed.

resources/static/dialog/js/modules/authenticate.js
```diff
@@ -127,6 +132,18 @@ BrowserID.Modules.Authenticate = (function() {
     }
   }
 
+  function createFxAccount(callback) {
```
shane-tomlinson, line 144: Now that there is createFxAccount, maybe createSecondaryUser should be renamed createPersonaAccount for consistency.

resources/static/dialog/js/modules/authenticate.js
shane-tomlinson: If non-default issuers completely disable primaries, this check should go above the primary check.

resources/static/dialog/js/misc/state.js
```diff
@@ -380,6 +398,23 @@ BrowserID.State = (function() {
...
+            self.newFxAccountEmail = info.email;
```
shane-tomlinson, line 409: @jedp, do we really want to create a new account here, or should some FX OS specific email be sent?

resources/static/dialog/js/misc/state.js
shane-tomlinson, line 403: This also duplicates logic found in pick_email.

These should be collapsed.

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -61,6 +62,8 @@ BrowserID.Modules.PickEmail = (function() {
     var record = checkEmail.call(self, email);
     if (!! record) {
       dialogHelpers.refreshEmailInfo.call(self, email, function (info) {
+        // XXX Why is this here? This is almost a complete duplication of
+        // the logic in state.js, and it should be there.
```
shane-tomlinson, line 66: This logic is also found in authenticate.js.

PR #3241, opened 2013-04-11
add maximum_event_loop_lag to config.json

jedp:
I was getting intermittent test failures on localhost.  This fixes that.

@shane-tomlinson can you review this?

### DIFF
--- a/config/local.json
+++ b/config/local.json
@@ -15,6 +15,9 @@
   "env": "local",
   "kpi_backend_sample_rate": 1.0,
 
+  // increase to tolerate slower server responses; useful for testing
+  "maximum_event_loop_lag": 100,
+
   // whether to show the development menu.
   "enable_development_menu": true,
 


jaredhirsch:
@jedp looking at [busy_middleware.js](https://github.com/jedp/browserid/blob/event_loop_lag/lib/busy_middleware.js#L11), it seems like a more robust fix here might be disabling toobusy altogether, by setting `disable_request_throttling` in config.json?

jrgm:
Travis runs actually `disable_request_throttling` - https://github.com/mozilla/browserid/blob/dev/.travis.yml#L24

I personally though prefer raising the threshold to disabling, as that means the event loop timer code has to continue executing.

jaredhirsch:
@jrgm thanks for the comments! merging in--better to keep closer to prod where possible.

(edited to finish my thought)

Issue #3184, opened 2013-04-03
dialog/modules/verify_primary_user and dialog/modules/upgrade_primary_user are near duplicates


shane-tomlinson:
These can be collapsed to reduce the amount of code where things can go wrong.

shane-tomlinson:
This work has been merged. Closing.

PR #3244, opened 2013-04-12
Collapse primary_user modules

arlolra:
Reduces duplicate code.
See Issue #3184


shane-tomlinson:
@arlolra - Thanks for the PR! I will review this week. The immediate concern I have is the upgrade of bcrypt, if that portion could be reverted and issued as a separate PR we can review the components individually.

arlolra:
Reverted. Sorry, `bcrypt v0.7.3` fails to install in `node v0.10.4`, at least for me.

lloyd:
@arlolra I'm working on the `nodejs-0.10.0` branch, feel free to collaborate there on the 0.10 port.  Seems there's just a bug or two left.

shane-tomlinson:
@arlolra - I have given this an initial review, the tests all pass which is good. A change I'd like to see on top of this, instead of creating two separate Modules using createModule, is have a single module that takes as an input parameter to start what type it is.  We could collapse the tests that way too. If you want to continue with this PR that is fantastic, if not, I can pick up where you finished up here. This is a huge step in the right direction, thank you!

arlolra:
@shane-tomlinson Updated as desired.

shane-tomlinson:
@arlolra - this is awesome. I have reviewed and things are looking good. This work has sent me down a rabbit hole of wondering whether things can be further simplified, but that'll be work for another PR, incremental improvements are always welcome. 

All the automated tests pass, but I am having issues manually testing. I want to make sure I understand why I am having these issues before merging.

Thanks again.

arlolra:
Out of curiosity, what issue are you running into when manually testing? Seems ok on my end.

shane-tomlinson:
@arlolra, thank you very much for this PR! I took it and made some additional simplifications, with a new PR created - #3458. Closing this one. Thanks again!

PR #3248, opened 2013-04-15
Remove no op code from email

ozten:
I don't believe the `_` function does anything here.

This removes this no op codepath.

### DIFF
--- a/lib/email.js
+++ b/lib/email.js
@@ -31,32 +31,29 @@ if (smtp_params && smtp_params.host) {
 
 const TEMPLATE_PATH = path.join(__dirname, "..", "resources", "email_templates");
 
-// the underbar decorator to allow getext to extract strings
-function _(str) { return str; }
-
 // a map of all the different emails we send
 const templates = {
   "new": {
     landing: 'verify_email_address',
-    subject: _("Confirm email address for Persona"),
+    subject: "Confirm email address for Persona",
     template: fs.readFileSync(path.join(TEMPLATE_PATH, 'new.ejs')),
     templateHTML: fs.readFileSync(path.join(TEMPLATE_PATH, 'new.html.ejs'))
   },
   "reset": {
     landing: 'reset_password',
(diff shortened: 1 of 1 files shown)


jaredhirsch:
@ozten I see the same noop function in lib/static/views. Is this a relic from how we used to do i18n? Are the two cases different?

jaredhirsch:
@ozten for more context, looks like the gettext underbar noop was added to lib/email in commit 5a28b8f, and to lib/static/views in e90869ce.

lloyd:
how will l10n code extract these strings?

ozten:
With gettext used in the .ejs template files.

lloyd:
@ozten - I don't understand, I'm sorry.  The only place I see the string 'Confirm email address for Persona' decorated in a way that l10n will extract it is in `lib/email.js`.  I grepped even!

So how will string extraction pick up this string if we remove the decoration?

ozten:
Great catch @lloyd!

Interesting. I forgot that `scripts/extract_po.sh` adds `_` as a keyword alias for gettext.

This is fine. The _only_ places we use the `_` alias is lib/email.js and lib/views.js (at least as far as I had poked around). This is potentially problematic.

Recommendations: 
- Use `_` more widely or 
- comment the link to Gettext wherever we declare `_` or 
- change `_` to  `___` or 
- use only `gettext` throughout email.js.

History: In many webapps we use `_` exclusively for Gettext. With Persona, we didn't want to use that as we used the underscore library liberally in the front and backend. So this support was baked into extract_po.sh but I completely forgot about it. If matjazz used our new i18n-abide based tools, they wouldn't have extracted these strings. Being emails, it probably would have taken a while for a localization bug to be filed.

Closing this pull request, as it isn't the right fix.

ozten:
Filed https://github.com/mozilla/i18n-abide/issues/26

lloyd:
+1 to s/_/gettext/ - the underbar is too sneaky.

PR #3255, opened 2013-04-16
Use node_modules lockdown

arlolra:
`node-gyp` kept failing for me because `process.env.HOME` was missing. This is fixed in `node-lockdown` in this commit [1302120a0b](https://github.com/mozilla/npm-lockdown/commit/1302120a0b9bcbd6344862413f6bf4c49d9ea686), however the changes aren't reflected in `browserid` because it uses `./scripts/lockdown`, which is a copy of `0.0.2`.

Instead, I changed the preinstall script to get lockdown and then run it from the `node_modules` folder.

If this is undesirable, I can just upgrade the version in `./scripts`


lloyd:
AHA!  I've been tracking this issue for weeks!  you're a genius!  I'll review and merge as soon as I can.

arlolra:
@lloyd Happy to help. Happier still if I were part of the Identity team. Feel free to [hire me](https://careers.mozilla.org/en-US/position/oDhfXfwQ) :D

package.json
```diff
@@ -53,7 +53,7 @@
         "node-inspector": "0.2.0beta3"
     },
     "scripts": {
-        "preinstall": "node ./scripts/lockdown",
+        "preinstall": "npm install lockdown; node ./node_modules/lockdown/lockdown.js",
```
seanmonstar: will this work on Windows?

arlolra: it does using `&&` instead of `;` to separate the commands. I'll update that

seanmonstar:
:+1: awesome! It's deeply bothered me that we were using a local (outdated) copy.

lloyd:
only issue here is we _want_ the lockdown script checked in.  idea is that if we npm install, then we can't trust our dependencies, cause anyone with control of the lockdown repo can fiddle em.  so checking it in was by design...

arlolra:
I see. It's a bit of chicken and egg problem. In that case, maybe just upgrade the version in `./scripts`?

Isn't there a loose dependency issue with `lockdown` anyways? If modules share a dependency (or sub dependency, say, `colors`),

```
"colors": {
  "0.3.0": "c247d64d34db0ca4dc8e43f3ecd6da98d0af94e7",
  "0.5.1": "7d0023eaeb154e8ee9fce75dcb923d0ed1667774",
  "0.6.0-1": "6dbb68ceb8bc60f2b313dcc5ce1599f06d19e67a"
},
```

they can inadvertently cause a compatible version to be installed, that's yet to be tested with a specific parent module. `npm shrinkwrap` gets this right, but doesn't provide checksums.

arlolra:
I see @fmarier has taken the liberty.

What I'm getting at may be mozilla/npm-lockdown#3

seanmonstar:
@lloyd what if we changed the package.json lockdown dependency to point at the git repo instead? Then we don't need to trust the npm repo, but still don't need a local copy (that grows stale).

seanmonstar:
@lloyd ping!

arlolra:
closing in favour of #3482

PR #3257, opened 2013-04-16
Issues 3196 3197 links

shane-tomlinson:
@callahad or @6a68 - could you review this one?

Two changes that @skinny97214 asked me to make - Add an "Account" link to the nav bar if the user is signed in and move "Develoers" to the footer.

### DIFF
--- a/resources/views/layout.ejs
+++ b/resources/views/layout.ejs
@@ -31,10 +31,10 @@
 
         <ul class="nav cf">
             <li><a href="/about"><%= gettext("How it works") %></a></li>
-            <li class="signOut"><a href="/"><%= gettext("Account") %></a></li>
+            <li class="ifAuthenticated"><a href="/"><%= gettext("Account") %></a></li>
 
-            <li class="signIn"><a class="signIn" href="/signin"><%= gettext("Sign In") %></a></li>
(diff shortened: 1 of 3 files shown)


callahad:
This looks pretty reasonable to me.

resources/views/layout.ejs
```diff
@@ -31,7 +31,7 @@
 
         <ul class="nav cf">
             <li><a href="/about"><%= gettext("How it works") %></a></li>
-            <li><a href="https://developer.mozilla.org/docs/persona" target="_blank"><%= gettext("Developers") %></a></li>
+            <li class="signOut"><a href="/"><%= gettext("Account") %></a></li>
```
callahad: Why the `signOut` class here?

shane-tomlinson: It's a mis-named class that means "show when the user is signed in". I was not sure whether I should expand the scope of the PR to rename it or not. I'd much rather re-name it if you agree.

callahad:
Shane is going to add a commit to rename the class, then this will be good to go.

shane-tomlinson:
@callahad  - could you re-review this?

callahad:
Looks trivially correct. Pulling and re-checking now...

PR #3259, opened 2013-04-16
Add ignore-tests option to run-all.js

edwindotcom:
This helps work around known browser webdriver issues, a heavy handed approach to #3202

### DIFF
--- a/automation-tests/lib/test-finder.js
+++ b/automation-tests/lib/test-finder.js
@@ -10,18 +10,18 @@ const path              = require('path'),
       tests_to_ignore   = require('../config/tests-to-ignore')(env),
       glob              = require('minimatch');
 
-exports.find = function(pattern, root, tests) {
+exports.find = function(pattern, root, tests, ignoreTests) {
   root = root || test_root_path;
   tests = tests || [];
   pattern = pattern || "*";
-
+  ignoreTests = ignoreTests || "";
   try {
     var files = fs.readdirSync(root);
     files.forEach(function(file) {
       var filePath = path.join(root, file);
       var stats = fs.statSync(filePath);
       if (stats.isFile() && /\.js$/.test(file)) {
-        if (tests_to_ignore.indexOf(file) === -1) {
+        if (tests_to_ignore.indexOf(file) === -1 && ignoreTests.split(',').indexOf(file.slice(0,-3)) === -1) {
           if (glob(file.replace(/\.js$/, ""), pattern)) {
             tests.push({
               name: file.replace('.js', ''),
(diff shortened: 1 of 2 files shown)


edwindotcom:
@6a68 review please?

jaredhirsch:
My concern with this patch is that we're now providing one implicit and one explicit way of excluding tests: the stuff in tests-to-ignore is implicitly included in the "ignores" set, so --ignore-tests as a command-line option won't do what it says it does.

We already have a [tests-to-ignore](https://github.com/mozilla/browserid/blob/dev/automation-tests/config/tests-to-ignore.js) file for this sort of thing--what would you think of extending that to handle platform-specific ignores? In that case, the tests_to_ignore var in test-finder.js could be the union of the tests-to-ignore for env and platform. It could be as simple as updating tests-to-ignore like this:

```
var testsToIgnore: {
  dev: [ ... ],
  stage: [ ... ],
  prod: [ ... ],
  safari: [ ... ]
}
```

And then updating [test-finder line 10](https://github.com/mozilla/browserid/blob/dev/automation-tests/lib/test-finder.js#L10) like this (but feel free to be less ugly/terse, this is just to illustrate the idea):

```
tests_to_ignore = require('../config/tests-to-ignore')(env).concat(require('../config/tests-to-ignore')(platform))
```

The tricky bit is that we currently have pretty divergent lists of platforms for local and sauce dev. We'd want to ensure that we ignore the tests in both cases. So this seems like an opportunity to unify and abstract how we label platforms for local and saucelabs development. I don't have specific ideas on how to do that at the moment.

edwindotcom:
Yeah - I considered updating the test-to-ignore file, here's my thinking:
pros: 
*tracked in github so nothing slips through the cracks
*could be more elegant to have one way to do this

cons:
*tests fail due to many reasons: out of sync trains, known issues, webdriver issues. Trying to solve that with a single dictionary is hard. We're talking about every browser/os config, dev/stage/prod , local, and multiple-firefox versions soon.  The data structure needed does not scale well and can be hard to edit.
*lots of pull requests bottle necks to process of getting stable test automation, slows down the process.  
*hard dependency on a checked-in file requires it be hot patched into prod/stage unless we wait 2 weeks for trains to go live.
*the work to build out better tests-to-ignore is non-trivial and our automation will continue peppered with known-issues as we wait for a full implementation.

Overall, tests-to-ignore is just more brittle and less flexible than this patch. The wait time for tests-to-ignore to percolate to prod/stage is just too long for Jenkins to be effective.

jaredhirsch:
i hear you. tallyho! :beers:

PR #3268, opened 2013-04-17
remove node-inspector as a dep - people can manually install this if they want it

lloyd:
npm install runs faster, and less software can possibly get into production (removing lines from lockdown.json == good), and node.js 0.10.0 port will be easier.


callahad:
r+

PR #3273, opened 2013-04-18
Prefetch dialog resources in the communication_iframe

fmarier:
Take advantage of the fact that we open a communication iframe to specify a list of dialog resources that could be prefetched by the browser.

Note that these prefetch lines are only suggestions only and the browser is free to ignore them (Chrome on Android has an "only prefetch on wifi" preference for example). However when prefetch happens, the dialog opens up instantly because all of its resources are already in the cache.

See http://feeding.cloud.geek.nz/posts/prefetching-resources-to-prime-browser/ for the background behind this patch.

### DIFF
--- a/resources/views/communication_iframe.ejs
+++ b/resources/views/communication_iframe.ejs
@@ -4,6 +4,15 @@
 <head><title>non-interactive iframe</title>
   <meta http-equiv="X-UA-Compatible" content="IE=Edge"/>
   <meta charset="utf-8">
+  <%- cachify(util.format('/production/%s/dialog.js', locale), {tag_format: '<link rel="prefetch" href="%s">'}) %>
+  <%- cachify('/production/dialog.css', {tag_format: '<link rel="prefetch" href="%s">'}) %>
+  <%- cachify('/common/i/grain.png', {tag_format: '<link rel="prefetch" href="%s">'}) %>
+  <%- cachify('/dialog/i/persona-logo-transparent.png', {tag_format: '<link rel="prefetch" href="%s">'}) %>
+  <%- cachify('/dialog/i/arrow_grey.png', {tag_format: '<link rel="prefetch" href="%s">'}) %>
+  <%- cachify('/common/i/button-loader.gif', {tag_format: '<link rel="prefetch" href="%s">'}) %>
+  <%- cachify('/common/i/button-arrow.png', {tag_format: '<link rel="prefetch" href="%s">'}) %>
+  <%- cachify('/common/i/button-arrow-hover.png', {tag_format: '<link rel="prefetch" href="%s">'}) %>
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@fmarier - THIS IS AWESOME!  Now that #3234 is merged, I am much more comfortable with this. One change I would like to see before merging, I worry this could introduce developer confusion when working in the dialog.

Since all dialog resources are loaded in the communication iframe, a developer who closes the dialog and immediately reloads it without first force loading the cache on the example RP will see the old code.

Can we disable the prefetch in dev mode, but have a flag to enable it for testing and then have it on by default in prod?

shane-tomlinson:
The initial performance results are mixed, especially on 3G.  I have to see why this is but I suspect it is because the Selenium test immediately opens the dialog as soon as the sign in button is ready, not giving the browser any time to believe it is idle.

```
Full Speed (cache empty), I:964, D:1068
Full Speed (cache full), I:300, D:508
Full Speed (cache empty), I:918, D:997
Full Speed (cache full), I:272, D:485
Full Speed (cache empty), I:893, D:1884
Full Speed (cache full), I:350, D:498
Full Speed (cache empty), I:873, D:1767
Full Speed (cache full), I:271, D:488
Full Speed (cache empty), I:907, D:2271
Full Speed (cache full), I:280, D:514
Full Speed (cache empty), I:880, D:1002
Full Speed (cache full), I:270, D:482
Full Speed (cache empty), I:784, D:907
Full Speed (cache full), I:284, D:496
Full Speed (cache empty), I:850, D:800
Full Speed (cache full), I:307, D:517
Full Speed (cache empty), I:840, D:776
Full Speed (cache full), I:266, D:490
Full Speed (cache empty), I:830, D:762
Full Speed (cache full), I:295, D:518
3G Average (cache empty), I:2417, D:5484
3G Average (cache full), I:918, D:1142
3G Average (cache empty), I:2426, D:5459
3G Average (cache full), I:927, D:1160
3G Average (cache empty), I:2534, D:5411
3G Average (cache full), I:919, D:1164
3G Average (cache empty), I:2580, D:5394
3G Average (cache full), I:926, D:1177
3G Average (cache empty), I:2622, D:5375
3G Average (cache full), I:922, D:1134
3G Average (cache empty), I:2408, D:5628
3G Average (cache full), I:933, D:1126
3G Average (cache empty), I:2626, D:5499
```

The full speed results are on average one second faster to load the dialog on cache empty than they are now, the 3G results are nearly identical to current.

fmarier:
@shane-tomlinson Prefetch now only happens when resources are minified.

shane-tomlinson:
@fmarier - I am getting an unusual number of sauce labs failures when running the selenium tests. I want to investigate to see if the problem is present in the current dev.

PR #3276, opened 2013-04-18
add cache headers to .well-known

lloyd:


lloyd:
click the button @shane-tomlinson - you know you want to.

PR #3288, opened 2013-04-20
Don't attempt to append to doc.head. Fixes #3287

### LINKED ISSUE #3287: in dev, primary login fails on IE8
async loading of bidbundle appends the script tag to document.head, which assumes document.head is accessible. Not sure why offhand, but IE8 tells me there's a document.body but no document.head.

At any rate, it's better to follow the advice in [3pjs](http://manning.com/vinegar) and append the dynamically-generated script tag after the 0th script tag on the page. Since there's at least one script tag (the one that loaded the running script), this is guaranteed to work.

Will open a PR momentarily.

jaredhirsch:
Attempting to append the dynamically-created script tag to document.head
    doesn't always work. In our case, it isn't working in IE8. Instead, we
    use an entry point that's guaranteed to exist, by finding the 0th script
    tag on the page. There must be at least one script tag, since the script
    itself is somehow running :-)

### DIFF
--- a/resources/static/common/js/crypto-loader.js
+++ b/resources/static/common/js/crypto-loader.js
@@ -18,7 +18,8 @@ BrowserID.CryptoLoader = (function() {
   function addScript(src) {
     var script = document.createElement("script");
     script.setAttribute("src", src);
-    document.head.appendChild(script);
+    var entryPoint = document.getElementsByTagName("script")[0];
+    entryPoint.parentNode.insertBefore(script, entry);
   }
 
   function waitUntilExists(checkFor, context, done) {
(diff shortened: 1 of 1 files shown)


jaredhirsch:
Ah, duh. document.head is html5, which is why IE8 sneers at it.

resources/static/common/js/crypto-loader.js
```diff
@@ -18,7 +18,8 @@ BrowserID.CryptoLoader = (function() {
   function addScript(src) {
     var script = document.createElement("script");
     script.setAttribute("src", src);
-    document.head.appendChild(script);
+    var entryPoint = document.getElementsByTagName("script")[0];
+    entryPoint.parentNode.insertBefore(script, entryPoint);
```
callahad, line 22: In your comments on #3287, you mention that the advice is:

> "append the dynamically-generated script tag **after** the 0th script tag."

Doesn't this insert the script tag **before** the existing one? Does that matter?

jaredhirsch, line 22: Derp, you're right. Doesn't matter so long as we have an entry point
guaranteed to exist.

PR #3304, opened 2013-04-24
renderError will publish an error_screen event

### LINKED ISSUE #3302: Publish error screens to KPI
It'd greatly help to know what percentage of users are "unsupported", and which browsers those are.

seanmonstar:
These events are stored in the event stream. This will allow us to see when a user drops out of the dialog because they ran into an error screen.

example: screen.error.relaySetup

fixes #3302

### DIFF
--- a/resources/static/common/js/lib/tinyscore.js
+++ b/resources/static/common/js/lib/tinyscore.js
@@ -58,6 +58,14 @@
       return -1;
     },
 
+    keyOf: function(obj, value) {
+      for (var key in obj) {
+        if (obj[key] === value) {
+          return key;
+        }
+      }
+    },
+
     size: function(obj) {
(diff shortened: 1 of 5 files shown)


seanmonstar:
I'd like to nominate this for merging into this train, as the feature is small, but would greatly increase our ability to understand how many of the % of users dropping off are doing so when hitting an error. Getting it in this train would let us know 2 weeks sooner. Waiting a train means we have 2 more weeks of less knowledge, not knowing which errors might be plaguing users. @kparlante do you agree?

@6a68 would you r? Shane is out the rest of the week.

kparlante:
Agreed, I think this information could be very helpful.

Cheers,
Katie

jaredhirsch:
r+ thanks for adding tests, all passing, code looks swell

PR #3307, opened 2013-04-25
B2g dev allow unverified

shane-tomlinson:
@jedp - allow unverified!

### DIFF
--- a/bin/verifier
+++ b/bin/verifier
@@ -7,6 +7,7 @@
 require('../lib/baseExceptions').addExceptionHandler();
 
 const
+_ = require('underscore'),
 util = require("util"),
 path = require('path'),
 url = require('url'),
@@ -146,13 +147,11 @@ function doVerification(req, resp, next) {
         rp: audience
       });
     } else {
-      resp.json({
+      resp.json(_.extend(r.success, {
         status : "okay",
-        email : r.success.email,
         audience : audience, // NOTE: we return the audience formatted as the RP provided it, not normalized in any way.
-        expires : new Date(r.success.expires).valueOf(),
-        issuer: r.success.issuer
-      });
+        expires : new Date(r.success.expires).valueOf()
+      }));
 
       metrics.report('verify', {
         result: 'success',
--- a/bin/verifier
+++ b/bin/verifier
@@ -19,6 +19,7 @@ logger = require('../lib/logging').logger,
 config = require('../lib/configuration'),
 shutdown = require('../lib/shutdown'),
 statsd = require('../lib/statsd'),
+booleanQuery = require('../lib/boolean-query'),
 toobusy = require('../lib/busy_middleware.js');
 
 logger.info("verifier server starting up");
@@ -60,6 +61,7 @@ if (statsd_config && statsd_config.enabled) {
 }
 
 app.use(express.bodyParser());
+app.use(booleanQuery);
 
 try {
   // explicitly relay VAR_PATH to children
@@ -93,6 +95,7 @@ function doVerification(req, resp, next) {
   } else if (req.body && req.body.forceIssuer) {
     forceIssuer =  req.body.forceIssuer;
   }
+  var allowUnverified = (req.query && req.query.allowUnverified) || req.body.allowUnverified;
 
   if (!(assertion && audience)) {
     // why couldn't we extract these guys?  Is it because the request parameters weren't encoded as we expect? GH-643
@@ -124,7 +127,8 @@ function doVerification(req, resp, next) {
   cc.enqueue({
     assertion: assertion,
     audience: audience,
-    forceIssuer: forceIssuer
+    forceIssuer: forceIssuer,
+    allowUnverified: !!allowUnverified
   }, function (err, r) {
     var reqTime = new Date - startTime;
     statsd.timing('assertion_verification_time', reqTime);
--- a/bin/verifier
+++ b/bin/verifier
@@ -86,17 +86,13 @@ try {
 }
 
 function doVerification(req, resp, next) {
+  req.query = req.query || {};
   req.body = req.body || {};
 
-  var assertion = (req.query && req.query.assertion) ? req.query.assertion : req. body.assertion;
-  var audience = (req.query && req.query.audience) ? req.query.audience : req.body.audience;
-  var forceIssuer = undefined;
-  if (req.query && req.query.forceIssuer) {
-    forceIssuer =  req.query.forceIssuer;
-  } else if (req.body && req.body.forceIssuer) {
-    forceIssuer =  req.body.forceIssuer;
-  }
-  var allowUnverified = (req.query && req.query.allowUnverified) || req.body.allowUnverified;
+  var assertion = req.query.assertion ? req.query.assertion : req.body.assertion;
+  var audience = req.query.audience ? req.query.audience : req.body.audience;
+  var forceIssuer = req.query.forceIssuer ? req.query.forceIssuer : req.body.forceIssuer;
+  var allowUnverified = req.query.allowUnverified ? req.query.allowUnverified : req.body.allowUnverified;
 
   if (!(assertion && audience)) {
     // why couldn't we extract these guys?  Is it because the request parameters weren't encoded as we expect? GH-643
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -316,13 +316,12 @@ exports.authForVerificationSecret = function(secret, cb) {
     if (!db.staged[secret]) return cb("no such secret");
 
     if (db.staged[secret].passwd) {
-      return cb(null, db.staged[secret].passwd, db.staged[secret].existing_user, true);
+      return cb(null, db.staged[secret].passwd, db.staged[secret].existing_user);
     }
 
     exports.checkAuth(db.staged[secret].existing_user, function (err, hash) {
-      // fourth parameter indicates that there was no
-      // password in the stage table
-      cb(err, hash, db.staged[secret].existing_user, false);
+      logger.debug("check auth " + err + " " + hash);
+      cb(err, hash, db.staged[secret].existing_user);
     });
   });
 };
@@ -474,6 +473,7 @@ exports.addPrimaryEmailToAccount = function(userID, emailToAdd, cb) {
 };
 
 exports.checkAuth = function(userID, cb) {
+  logger.debug("checkAuthh for userID " + userID);
   sync();
   var m;
   if (userID) {
@@ -613,6 +613,7 @@ exports.createUnverifiedUser = function(email, hash, cb) {
   var row = {
     id: getNextUserID(),
     lastPasswordReset: now(),
+    password: hash,
     emails: emailVal
   };
   db.users.push(row);
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -446,6 +446,8 @@ exports.completePasswordReset = function(secret, password, cb) {
         Object.keys(emails).forEach(function(email) {
           if (email !== o.email && emails[email].type === 'secondary') {
             emails[email].verified = false;
+          } else if (email === o.email) {
+            emails[email].verified = true;
           }
         });
         flush();
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -231,10 +231,19 @@ function addEmailToAccount(userID, email, type, cb) {
   });
 }
 
+function deleteOldSecrets(email) {
+  var secret = db.stagedEmails[email];
+  if (secret) {
+    delete db.stagedEmails[email];
+    delete db.staged[secret];
+  }
+}
+
 exports.stageUser = function(email, hash, cb) {
   secrets.generate(48, function(secret) {
     // overwrite previously staged users
     sync();
+    deleteOldSecrets(email);
     db.staged[secret] = {
       type: "add_account",
       email: email,
@@ -251,6 +260,7 @@ exports.stageEmail = function(existing_user, new_email, hash, cb) {
   secrets.generate(48, function(secret) {
     // overwrite previously staged users
     sync();
+    deleteOldSecrets(new_email);
     db.staged[secret] = {
       type: "add_email",
       existing_user: existing_user,
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -215,22 +215,14 @@ exports.emailToUID = function(email, cb) {
   });
 };
 
-exports.userOwnsEmail = function(uid, email, staged, cb) {
-   sync();
-  if (typeof staged === "function") {
-    cb = staged;
-    staged = false;
-  }
-   var m = jsel.match(":root > object:has(:root > .id:expr(x=" + ESC(uid) + ")):has(.emails > ." + ESC(email) + ")", db.users);
+exports.userOwnsEmail = function(uid, email, cb) {
+ sync();
+  var m = jsel.match(":root > object:has(:root > .id:expr(x=" + ESC(uid) + ")):has(.emails > ." + ESC(email) + ")", db.users);
   var owned = m && m.length === 1;
-  if (!owned && staged) {
-    m = jsel.match(":root > object:has(:root > .existing_user:expr(x=" + ESC(uid) +")):has(:root > .email:expr(x=" + ESC(email) + "))", db.staged);
-    owned = m && m.length;
-  }
-   process.nextTick(function() {
+  process.nextTick(function() {
     cb(null, owned);
-   });
- };
+  });
+};
  
 function addEmailToAccount(userID, email, type, cb) {
   // validate 'type' isn't bogus
@@ -492,14 +484,6 @@ exports.checkAuth = function(userID, cb) {
     if (m.length === 0) m = undefined;
     else m = m[0];
   }
-  process.nextTick(function() { cb(null, m); });
-};
-
-exports.checkUnverifiedAuth = function(email, cb) {
-  sync();
-  var m = jsel.match(":root > object:has(:root > .email:expr(x=" + ESC(email) +")) > .passwd", db.staged);
-  if (m.length === 0) m = undefined;
-  else m = m[0];
   process.nextTick(function() { cb(null, m); });
 };
 
--- a/lib/db/json.js
+++ b/lib/db/json.js
@@ -215,14 +215,23 @@ exports.emailToUID = function(email, cb) {
   });
 };
 
-exports.userOwnsEmail = function(uid, email, cb) {
-  sync();
-  var m = jsel.match(":root > object:has(:root > .id:expr(x=" + ESC(uid) + ")):has(.emails > ." + ESC(email) + ")", db.users);
-  process.nextTick(function() {
-    cb(null, m && m.length === 1);
-  });
-};
-
+exports.userOwnsEmail = function(uid, email, staged, cb) {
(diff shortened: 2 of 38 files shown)


jedp:
OMG

lib/db/json.js
```diff
@@ -199,13 +201,14 @@ exports.emailToUID = function(email, cb) {
 };
 
 exports.userOwnsEmail = function(uid, email, cb) {
-  sync();
+ sync();
   var m = jsel.match(":root > object:has(:root > .id:expr(x=" + ESC(uid) + ")):has(.emails > ." + ESC(email) + ")", db.users);
+  var owned = m && m.length === 1;
   process.nextTick(function() {
-    cb(null, m && m.length === 1);
+    cb(null, owned);
```
jedp, line 208: thank you - helpful clarification

lib/keysigner/ca.js
```diff
@@ -25,11 +25,17 @@ function parsePublicKey(serializedPK) {
...
-function certify(hostname, email, publicKey, expiration, cb) {
+function certify(hostname, email, publicKey, expiration, unverified, cb) {
   if (expiration == null)
     return cb("expiration cannot be null");
 
-  cert.sign({publicKey: publicKey, principal: {email: email}},
+  //TODO: use a different secret if unverified
+
+  var principal = {};
+  principal[ unverified ? 'unverified-email' : 'email' ] = email;
```
jedp, line 35: Some time in the near future (let's say after the merge is all done), I think we should rename `unverified-email` to `unverified_email`, so it's easier to access as a property of a javascript object.

resources/static/common/js/user.js
```diff
@@ -265,20 +268,31 @@ BrowserID.User = (function() {
...
-      pub: keypair.publicKey.toSimpleObject(),
-      priv: keypair.secretKey.toSimpleObject(),
-      cert: cert
-    });
-
-    storage.addEmail(email, email_obj);
-    if (onComplete) onComplete(true);
+    // XXX This needs to be looked at to make sure caching does not bite us.
+    // I would rather pass in the unverified flag so that caching does not
+    // cause problems here.
```
jedp: Good catch of what might be a subtle bug.  Let's file a follow-up issue to investigate after the merge?

resources/static/dialog/js/misc/state.js
```diff
@@ -228,10 +244,18 @@ BrowserID.State = (function() {
       else if (self.transitionNoPassword) {
         redirectToState("stage_transition_to_secondary", info);
       }
+      else if(self.newFxAccountEmail) {
+        startAction(false, "doStageUser", info);
+// TODO         startAction(false, "doStageResetPassword", info); ???
```
jedp, line 252: is this still a TODO?

resources/static/dialog/js/modules/dialog.js
```diff
@@ -122,6 +122,12 @@ BrowserID.Modules.Dialog = (function() {
     return returnTo;
   }
 
+  function fixupIssuer(url) {
+    // XXX do something awesome here.
```
jedp: Shall we file a follow-up issue?

jedp:
@ozten and @seanmonstar, this looks good to me.  But it's such a huge PR I'd be grateful to have another pair of eyes go over it before pushing the big green button.  Can one of you take a second look?

jedp:
Don't merge until forceIssuer PR lands!

lib/keysigner/ca.js
shane-tomlinson, line 35: I agree, as long as it follows existing conventions.

resources/static/dialog/js/misc/state.js
shane-tomlinson, line 252: Dunno, I made the same comment here: https://github.com/mozilla/browserid/pull/3240#discussion_r3895102

resources/static/dialog/js/modules/dialog.js
shane-tomlinson: It's already taken care of in the b2g_dev_force_issuer branch. Rebasing this and pushing.

shane-tomlinson:
@jedp, @seanmonstar - this should have a significantly smaller diff once the force-issuer branch has landed. 

In other news, this makes me happy (the instance is called force-issuer, but the code is allow-unverified):

```
➜  automation-tests git:(b2g_dev_allow_unverified) ✗ ./scripts/run-all.js -e force-issuer -p 5
adding add-primary-to-primary.js
adding add-primary-to-secondary.js
adding cancel-account.js
adding change-password-test.js
adding frontend-qunit-test.js
adding health-check-tests.js
adding broken-primary.js
adding primary-shuts-down-single-email.js
adding primary-starts-up-single-email.js
adding new-user-primary-test.js
adding new-user-secondary-test.js
ignoring public-terminals.js
adding remove-email.js
adding reset-password-test.js
adding returning-user.js
adding sign-in-test.js
Running 15 suite(s) on 1 platform(s), 5 at a time against https://force-issuer.personatest.org
>>>>>.........................<>.......<>.....<>.........................<>.......<>.................<>..........<>....................................<>......<>.......<>...........................<...............<......<.....<...<

216/216 tests passed
➜  automation-tests git:(b2g_dev_allow_unverified) ✗
```

shane-tomlinson:
when adding an email with "unverified allowed", I am currently forced to verify the email.

bin/verifier
```diff
@@ -135,13 +147,11 @@ function doVerification(req, resp, next) {
         rp: audience
       });
     } else {
-      resp.json({
+      resp.json(_.extend(r.success, {
         status : "okay",
-        email : r.success.email,
         audience : audience, // NOTE: we return the audience formatted as the RP provided it, not normalized in any way.
-        expires : new Date(r.success.expires).valueOf(),
-        issuer: r.success.issuer
```
shane-tomlinson, line 150: We could use a comment here about what comes back from the verifier. It took me a minute to grok why we removed email - I assume it is because the verifier could return either "email" or "unverified-email"

shane-tomlinson:
@jedp - It looks like there are many cases not handled by this PR.

If "allowUnverified" is set to true, I think the following flows need updated as well:
- add email
- transition_no_password
- transition_to_secondary

Also, do we still send primary users down the primary flow if allowUnverified is set to true? It makes sense to allow this.

seanmonstar:
@shane-tomlinson yes, this doesn't prevent primaries from authenticating, but if they do, they may require verification before giving back an assertion. We can't enforce allowUnverified on them. Which is why, to ensure a user doesn't need to verify, it must be used with forceIssuer as well.

lib/keysigner/ca.js
```diff
@@ -25,11 +25,17 @@ function parsePublicKey(serializedPK) {
   return jwcrypto.loadPublicKey(serializedPK);
 }
 
-function certify(hostname, email, publicKey, expiration, cb) {
+function certify(hostname, email, publicKey, expiration, unverified, cb) {
   if (expiration == null)
     return cb("expiration cannot be null");
 
-  cert.sign({publicKey: publicKey, principal: {email: email}},
+  //TODO: use a different secret if unverified
```
shane-tomlinson, line 32: Do we have to add a different key here?

bin/verifier
```diff
@@ -93,6 +96,7 @@ function doVerification(req, resp, next) {
   } else if (req.body && req.body.forceIssuer) {
     forceIssuer =  req.body.forceIssuer;
   }
+  var allowUnverified = (req.query && req.query.allowUnverified) || req.body.allowUnverified;
```
shane-tomlinson: Are we allowing allowUnverified as a GET parameter? If not, the req.query should be taken off.

shane-tomlinson: Do we allow GET parameters elsewhere?

lib/db/json.js
```diff
@@ -476,6 +473,7 @@ exports.addPrimaryEmailToAccount = function(userID, emailToAdd, cb) {
 };
 
 exports.checkAuth = function(userID, cb) {
+  logger.debug("checkAuthh for userID " + userID);
```
shane-tomlinson: Is this still needed?

lib/db/mysql.js
```diff
@@ -492,9 +492,14 @@ exports.completePasswordReset = function(secret, password, cb) {
         function(err) {
           if (err) return cb(err);
 
-          // update the password!
```
shane-tomlinson, line 495: In the JSON adapter, deleteOldSecrets has been added several times, but the same treatment has not been added here. Is this expected or does a deleteOldSecrets equivalent need to be added here?

lib/wsapi.js
```diff
@@ -22,6 +22,7 @@ config = require('./configuration'),
 logger = require('./logging.js').logger,
 httputils = require('./httputils.js'),
 forward = require('./http_forward.js').forward,
+booleanQuery = require('./boolean-query'),
```
shane-tomlinson: Do any wsapi calls need booleanQuery? booleanQuery only works with GET parameters.

lib/wsapi/complete_reset.js
```diff
@@ -70,7 +70,7 @@ exports.process = function(req, res) {
           });
         });
       } else {
-        return httputils.authRequired(res, 'Provide your password');
+        return httputils.authRequired(res, 'Provide your password- params were: ' + JSON.stringify(req.params) );
```
shane-tomlinson: Do we need the extra `- params were...`?

lib/wsapi/complete_reset.js
```diff
@@ -88,7 +88,7 @@ exports.process = function(req, res) {
           return postAuthentication(hash);
         });
       } else {
-        return httputils.authRequired(res, 'Provide your password');
+        return httputils.authRequired(res, 'Provide your password' + JSON.stringify(req.params));
```
shane-tomlinson: Likewise.

lib/wsapi/stage_user.js
```diff
@@ -58,30 +64,51 @@ exports.process = function(req, res) {
...
+        if (allowUnverified) {
+          // if its ok to be unverified, then create the user right now,
+          // and stage their email instead.
+          db.createUnverifiedUser(req.params.email, hash, function(err, uid, secret) {
+            if (err) {
+              cef_logger.alert("DB_FAILURE", "Cannot create unverified user; dbwriter failure", req, {msg: err});
+              return wsapi.databaseDown(res, err);
+            }
+
+            if (!req.session) req.session = {};
```
shane-tomlinson, line 76: From here to the else statement is duplicated below and should be collapsed. (the only difference is the unverified: true in res.json)

resources/static/common/js/user.js
```diff
@@ -489,8 +510,18 @@ BrowserID.User = (function() {
      */
     createSecondaryUser: function(email, password, onComplete, onFailure) {
       stageAddressVerification(email, password,
-        network.createUser.bind(network, email, password, origin),
-        onComplete, onFailure);
+        network.createUser.bind(network, email, password,
+            origin, allowUnverified), function(status) {
+              // update the cache if necessary
+              // XXX this needs a test.
```
shane-tomlinson: Add the test.

resources/static/dialog/js/misc/helpers.js
```diff
@@ -79,13 +79,20 @@
     var self=this;
     user.createSecondaryUser(email, password, function(status) {
       if (status.success) {
-        var info = { email: email, password: password };
-        self.publish("user_staged", info, info);
+        var msg = { email: email, password: password };
+        // XXX these need tests.
```
shane-tomlinson: Add the test.

resources/static/dialog/js/misc/helpers.js
```diff
@@ -79,13 +79,20 @@
...
+        if (status.unverified) {
+          msg.type = "secondary";
```
shane-tomlinson: Both of these "msg" settings are implicit in the message "unverified_created"

resources/static/dialog/js/misc/state.js
```diff
@@ -423,12 +431,18 @@ BrowserID.State = (function() {
...
       }
-      else if (info.state === 'unverified') {
+      else if (info.state === 'unverified' && !self.allowUnverified) {
         // user selected an unverified secondary email, kick them over to the
         // verify screen.
         redirectToState("stage_reverify_email", info);
       }
       else {
+        // make sure an unverified-certs are removed
+        if (record.unverified && info.state !== 'unverified') {
```
shane-tomlinson, line 443: This seems like it should be taken care of in user.js

resources/static/dialog/js/modules/actions.js
```diff
@@ -95,6 +95,13 @@ BrowserID.Modules.Actions = (function() {
       startService("required_email", info);
     },
 
+    doAuthenticateWithUnverifiedEmail: function(info) {
+      var self = this;
+      dialogHelpers.authenticateUser.call(this, info.email, info.password, function() {
+        self.publish("authenticated", info);
```
shane-tomlinson, line 102: From https://github.com/mozilla/browserid/pull/3307/files#r3992382 above (implicit), set the info.type and info.unverified here. here.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -274,6 +274,8 @@ BrowserID.Modules.Dialog = (function() {
         // forceAuthentication is used by the Marketplace to ensure that the
         // user knows the password to this account. We ignore any active session.
         if (paramsFromRP._experimental_forceAuthentication &&
+            // TODO this and _experimental_allowUnverified should throw an
```
shane-tomlinson: Make a better check like the TODO says.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -285,6 +287,14 @@ BrowserID.Modules.Dialog = (function() {
               fixupIssuer(paramsFromRP._experimental_forceIssuer);
         }
 
+        // allowUnverified means that the user doesn't need to have
+        // verified their email address in order to send an assertion.
+        // if the user *has* verified, it will be a verified assertion.
+        if (paramsFromRP._experimental_allowUnverified &&
+            true === paramsFromRP._experimental_allowUnverified) {
```
shane-tomlinson: This needs a better check as well.

resources/static/test/cases/dialog/js/modules/dialog.js
```diff
@@ -542,6 +542,15 @@
     );
   });
 
+  // TODO Add a test for forceIssuer
```
shane-tomlinson: Add a forceIssuer test if it doesn't already exist.

resources/static/dialog/js/misc/state.js
```diff
@@ -423,12 +431,18 @@ BrowserID.State = (function() {
...
+          storage.invalidateEmail(email);
```
shane-tomlinson: This only invalidates emails for the default issuer, should this invalidate emails for the currently forced issuer?

resources/static/dialog/js/misc/helpers.js
shane-tomlinson: I'm not convinced these are actually needed. Are they used somewhere?

shane-tomlinson:
@jedp, @seanmonstar, @benadida - what should happen, if for the same issuer, the user visits one RP and set up an account with an unverified email address and they visit another RP where allowUnverified is not set? Currently, there is a discrepancy if the user is authenticated to Persona vs if they are not.

If they _are_ signed in, they are forced to verify their email and sent to the verification screen.
If they _are not_ signed in, they are sent down the normal new user flow, including setting the account password.

This somewhat makes sense. In the case of a "stolen" account, the real owner of the address could come along and take the email address back from the original user.

This caused me to test another case where there is a verified problem:

User A, who is not the real owner of the address, visits a site that allows unverified emails
User B, who is the real owner of the address, visits the same RP.
User B is asked for the password User A typed. User B does not know the password. User B cannot sign in even though they own the address.

Even if User B opens the email address that was sent when User A registered the address, User B will still not know the password and cannot claim their own email address for themselves. 

wah wah.

shane-tomlinson:
So my last comment, the rightful owner of the address can still use password reset. What should happen if allowUnverified is set to true, should User B be forced to validate their email before they can snag it off of User A, or should they be signed straight in?

lib/keysigner/ca.js
jedp, line 32: I think this is a question we need @benadida to weigh in on

bin/verifier
jedp: It seems we do allow GET params, yes

lib/wsapi/complete_reset.js
jedp: kill the debugging cruft

lib/db/mysql.js
jedp, line 495: Good catch.  Hm.  I'm not sure.  It looks like the json database has `staged` and `stagedUser`, whereas the mysql has just `staged`, which it overwrites both on `stageUser` and `stageEmail`.  Looks like the json database is getting the same result by deleting both rows when either is updated?

lib/wsapi/stage_user.js
jedp, line 76: They have the same shape, but they're actually pretty different, I think: Different `db` functions, callback signatures, cef logging, `req.session` variables, and send-email functions.

lib/wsapi.js
jedp: Good catch. looks like we can remove this.

lib/db/mysql.js
jedp, line 495: Ja.  @seanmonstar confirms that this is ok.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -159,6 +159,13 @@ BrowserID.Modules.Dialog = (function() {
     return parsedTime;
   }
 
+  function validateBoolean(bool, name) {
+    if (typeof bool !== "boolean") {
+      throw new Error("invalid value for " + name + ": " + bool);
```
seanmonstar, line 164: a tangential conversation: should we be throwing these errors in the dialog, or right when they try to do it with `navigator.id.request({ _experimental_forceIssuer: "heck yes" });`

resources/static/common/js/network.js
```diff
@@ -15,6 +15,7 @@ BrowserID.Network = (function() {
       xhr = bid.XHR,
       post = xhr.post,
       get = xhr.get,
+      // XXX get this out of here!
```
shane-tomlinson, line 18: storage was not added for this PR, this is just a note to get it out.

resources/static/dialog/js/modules/dialog.js
shane-tomlinson, line 164: @seanmonstar - we throw some errors in the dialog, some from include.js. The advantage to the dialog, for request based parameters, is that checks only exist in one place. If we pushed the checks to a lower level, we would have to check in include.js as well as internal_api.js. I do think there are advantages to seeing the errors early though, as console messages - "why doesn't the browser pop up? oh, there's an error in the console."

shane-tomlinson:
@jedp - what more do we have to do to merge this PR?

jedp:
Oh frabjous day.

lib/keysigner/ca.js
benadida, line 32: longer term, we probably want a different issuer, but short term, this is okay.

PR #3311, opened 2013-04-26
Fix the failing serverTime test.

### LINKED ISSUE #3310: serverTime front end qunit test frequently fails
It is intermittent, but fails approximately 1/3 of the time when running the full test suite. I believe this is due to the asynchronous bidbundle load, but have to investigate further.

shane-tomlinson:
@seanmonstar - could you give this a quick review, if possible for today before the train is cut so we don't get false positives?

It was not giving enough leeway for the increased number of onContextChange handlers.

fixes #3310

### DIFF
--- a/resources/static/test/cases/common/js/network.js
+++ b/resources/static/test/cases/common/js/network.js
@@ -417,21 +417,18 @@
 
 
   asyncTest("serverTime", function() {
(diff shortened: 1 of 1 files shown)


seanmonstar:
That code itself looks fine, but I have no idea what it fixes, or what this test even exercises.

PR #3316, opened 2013-04-27
Mitigate accidental cache poisoning during deployment

lloyd:
@jrgm - if you merge this in the next hour, it goes into the train.  if you don't it doesn't.  your call.

Here's the delta in cachify: https://github.com/mozilla/connect-cachify/compare/9f879d...v0.0.14?w=1


lloyd:
ok, given:
1. @jrgm really wanted this
2. @seanmonstar and @ozten reviewed the actual 
3. it's trivial to revert in a hotpatch

I'm hitting the green button!

PR #3320, opened 2013-04-30
'eventStreams' is either undefined, null or an Array. So check for truth...

### LINKED ISSUE #3319: train-2013.04.26 - javascript error bringing up the dialog.
Right now train-2013.04.26 is on the old physical stage. I was just going to get it a quick look before I asked to push this out to AWS. And I hit an error on the main site:

STR

1) Change /etc/hosts to point to physical stage:

63.245.209.245 diresworb.org www.diresworb.org
63.245.209.240 anosrep.org login.anosrep.org www.anosrep.org
63.245.209.239 static.login.anosrep.org verifier.login.anosrep.org

2) Go to login.anosrep.org, and click signin to bring up the dialog.

Actual:

Chrome says "improper usage of API: TypeError: Cannot read property &#x27;length&#x27; of undefined"

If I'm right, I think the exception is here: https://github.com/mozilla/browserid/blob/dev/resources/static/common/js/modules/interaction_data.js#L348 and `eventStream` is undefined.

Which fits. This error is "most of the time", and stage is configured to sample kpi at 20%.

jrgm:
...iness

before trying to use it like an Array. Fixes #3319 for the branch. This is
probably not a fully correct fix.

@shane-tomlinson - if you have a different way you'd like to fix this, please do :-). This PR has a code smell. 

Note: this PR is for the train-2013.04.26 branch.

### DIFF
--- a/resources/static/common/js/modules/interaction_data.js
+++ b/resources/static/common/js/modules/interaction_data.js
@@ -332,8 +332,10 @@ BrowserID.Modules.InteractionData = (function() {
     if (data.duration) eventData.push(data.duration);
 
     var eventStream = self.getCurrentEventStream();
-    if (eventStream) eventStream.push(eventData);
-    setCurrentEventStream.call(self, eventStream);
+    if (eventStream) {
+      eventStream.push(eventData);
+      setCurrentEventStream.call(self, eventStream);
+    }
 
     return eventData;
   }
--- a/resources/static/common/js/modules/interaction_data.js
+++ b/resources/static/common/js/modules/interaction_data.js
@@ -300,7 +300,7 @@ BrowserID.Modules.InteractionData = (function() {
 
     // Base the offset of any event already on the event stream off of the new
     // startTime.
-    if (eventStream.length) {
+    if (eventStream && eventStream.length) {
       var delta = self.startTime - newStartTime;
 
       for (var i=0, event; event=eventStream[i]; ++i) {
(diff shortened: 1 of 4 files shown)


seanmonstar:
This solves the symptom, but the cause is that even though sampling isn't enabled, the event stream is being accessed. I wonder if just changing that check away from a strict equality with false will fix everything.

jrgm:
@seanmonstar Yeah, definitely, I'm just addressing the sympton here for the train. Absolutely, if there's a better approach that would be good.

seanmonstar:
I'd change all the instances of `if (self.samplingEnabled !== false) return` to `if (!self.samplingEnabled) return;`, but I don't know if there are a couple instances where @shane-tomlinson meant to differentiate between `false` and `undefined`.

jrgm:
@seanmonstar I actually tried what you say above, but still had errors, but didn't debug further.

jrgm:
Note: merging this commit will conflict with PR #3322 on the browserid.spec "Release:" value, but this one needs to go first.

resources/static/common/js/modules/interaction_data.js
```diff
@@ -332,7 +332,7 @@ BrowserID.Modules.InteractionData = (function() {
     if (data.duration) eventData.push(data.duration);
 
     var eventStream = self.getCurrentEventStream();
-    eventStream.push(eventData);
+    if (eventStream) eventStream.push(eventData);
```
shane-tomlinson: I would make this `if` statement cover both the eventStream.push and setCurrentEventStream.

shane-tomlinson:
@seanmonstar - `samplingEnabled === undefined` means the server has not yet told us whether we can collect results. So, collect them until the server says not to. As soon as the server says to not collect samples, set `samplingEnabled = false` and collect no more. This is so we can collect KPIs before session_context is called.

shane-tomlinson:
Aside from the one comment already made, this looks good. Nice debugging @jrgm. I have written a test that you can add on to this PR if you'd like - https://gist.github.com/shane-tomlinson/5488715

r+

Issue #3309, opened 2013-04-26
sporadic 503 error for https://login.anosrep.org/wsapi/auth_with_assertion


jrgm:
I've seen this _once_ in my own browser, and grepping on stage shows this has happened to 12% of requests to auth_with_assertion.

STR
1. Just go through a normal flow to sign up as a new primary user (eyedee.me, yahoo.com).

When the error occurs, the log messages in browserid.log are:

```
{"level":"info","message":"successfully created primary acct for 2y8qk101ir@eyedee.me (11609398)","timestamp":"2013-04-25T23:32:39.137Z"}
{"level":"warn","message":"database is down, cannot process request: Error: authenticateSession called with undefined lastPasswordReset","timestamp":"2013-04-25T23:32:39.139Z"}
{"level":"info","message":"127.0.0.1 - - [Thu, 25 Apr 2013 23:32:39 GMT] \"POST /wsapi/auth_with_assertion HTTP/1.1\" 503 41 \"https://login.anosrep.org/signin\" \"Mozilla/5.0 (Windows NT 6.1; WOW64; rv:16.0) Gecko/20100101 Firefox/16.0\"","timestamp":"2013-04-25T23:32:39.140Z"}
```

In former scl2 stage, since the beginning for March the URL was been called thousands of times with 100% success.

(Perhaps there is greater network "distance" between the db master and the slaves, and we are racing with replication between the creation of the user row by the dbwriter /wsapi/create_account_with_assertion, and the read from a slave with wsapi.authenticateSession. Unfortunately, ping is not available, and I'm not sure how I compute replication lag from pt-heartbeat, but I'll look into that).

karlht:
I have also replicated this (just once) in a normal manual browser login flow.

mostlygeek:
Here is what it looks like: 

![wgfK](https://f.cloud.github.com/assets/83972/433182/19e0c0be-aec1-11e2-8d18-015eb07d46ee.png)

jrgm:
One point of clarification. When I hit my error, I looked in the database slaves and the row was there. The error would have only occurred if the row does not exist.

lloyd:
We need someone to dig into code and enumerate the conditions under which this can occur.  This might lead to reliable steps to reproduce.

-- lloyd (thumb-typing)

lloyd:
We could also write some code to exercise this flow programmatically (against apis) to vet @jrgm 's timing theory - which sound incredibly plausible.

jrgm:
[...]

PR #3321, opened 2013-04-30
Fix sporadic 503s

lloyd:
And along the way, remove a consistency requirement that slaves are updated <5ms after write master is written. issue #3309


lloyd:
@jrgm where do you want this fix?  port to 2013.04.26 ?

PR #3325, opened 2013-04-30
WinChan detects if window gets closed

### LINKED ISSUE #1773: WinChan does not pick up dialog close while redirected to IdP
STR

1) Go to myfavoritebeer.org
2) Click "Sign in"
3) Sign in with an IdP that will cause redirect to IdP.
4) Close dialog
5) Sign in button is still greyed out.

seanmonstar:
if the dialog is redirected to an IdP to do auth, and the user dismisses
the dialog then, this will let winchan detect that the dialog has been
closed. it will trigger oncancel.

fixes #1773

**Related**: This should probably be reviewed on WinChan first. If changes need to be made there, I can modify here as well. See lloyd/winchan#28 and lloyd/winchan#29

### DIFF
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -34,7 +34,7 @@
       var userAgent = navigator.userAgent;
       return (userAgent.indexOf('Fennec/') != -1) ||  // XUL
              (userAgent.indexOf('Firefox/') != -1 && userAgent.indexOf('Android') != -1);   // Java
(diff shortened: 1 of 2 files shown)


seanmonstar:
nobody loves me!

jaredhirsch:
:beers: looking now

jaredhirsch:
DNS for ephemeral instances is out of commission at the moment, but it looks fine and has been r+'d into winchan upstream, so, r+

PR #3327, opened 2013-05-01
B2g session duration

shane-tomlinson:
@jedp - I got the tests running and added hooks on the front end. This seems like a sane approach.

We set session length on email verification too, we should send the suppress hint to the front end upon verification as well.

### DIFF
--- a/lib/wsapi/auth_with_assertion.js
+++ b/lib/wsapi/auth_with_assertion.js
@@ -58,7 +58,7 @@ exports.process = function(req, res) {
               res.json({ success: true,
                          userid: uid,
                          suppress_ask_if_users_computer:
-                             durationInfo.isFirefoxOSDuration
+                             durationInfo.suppressAskIsUsersComputer
               });
 
               // if this email was last used as a secondary, update the database
@@ -134,7 +134,7 @@ exports.process = function(req, res) {
             res.json({ success: true,
                        userid: r.userid,
                        suppress_ask_if_users_computer:
-                           durationInfo.isFirefoxOSDuration
+                           durationInfo.suppressAskIsUsersComputer
 
             });
           });
--- a/lib/wsapi/auth_with_assertion.js
+++ b/lib/wsapi/auth_with_assertion.js
@@ -40,25 +40,25 @@ exports.process = function(req, res) {
     // 2. if valid, does the user exist?
     db.emailType(email, function(err, type) {
       if (err) return wsapi.databaseDown(res, err);
+      var durationInfo = wsapiutils.getDurationInfo(req);
 
       // if this is a known email adddess, authenticate the user and we're done.
       if (type) {
         db.emailToUID(email, function(err, uid) {
           if (err) return wsapi.databaseDown(res, err);
           if (!uid) return res.json({ success: false, reason: "internal error" });
-          var sessionDuration = wsapiutils.getDurationMS(req);
           wsapi.authenticateSession(
             {
               session: req.session, uid: uid,
               level: 'assertion',
-              duration_ms: sessionDuration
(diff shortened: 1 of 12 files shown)


jedp:
@shane-tomlinson thanks for catching and fixing my `context` error; everything makes sense now :)

lib/wsapi/auth_with_assertion.js
```diff
@@ -45,17 +46,20 @@ exports.process = function(req, res) {
         db.emailToUID(email, function(err, uid) {
           if (err) return wsapi.databaseDown(res, err);
           if (!uid) return res.json({ success: false, reason: "internal error" });
+          var sessionDuration = wsapiutils.getDurationMS(req);
```
jedp: Cool. Yes, clearer to move this up here.

lib/wsapi/auth_with_assertion.js
```diff
@@ -45,17 +46,20 @@ exports.process = function(req, res) {
...
-                : config.get('authentication_duration_ms')
+              duration_ms: sessionDuration
             }, function(err) {
               if (err) return wsapi.databaseDown(res, err);
 
-              res.json({ success: true, userid: uid });
+              res.json({ success: true,
+                         userid: uid,
+                         suppress_ask_if_users_computer:
+                             wsapiutils.isFirefoxOSDuration(sessionDuration)
```
jedp: Looks good.  I wonder though whether we don't want to combine this function with `getDurationMS`, and have the new function return an object like `{duration_ms: xxx, is_firefox_os_duration: boolean}`, so that it's clear what the connection is between the two values?

lib/wsapi/auth_with_assertion.js
```diff
@@ -112,17 +116,21 @@ exports.process = function(req, res) {
...
-              config.get('ephemeral_session_duration_ms')
-              : config.get('authentication_duration_ms')
+            duration_ms: sessionDuration
           }, function (err) {
             if (err) return wsapi.databaseDown(res, err);
-            res.json({ success: true, userid: r.userid });
+            res.json({ success: true,
+                       userid: r.userid,
+                       suppress_ask_if_users_computer:
+                           wsapiutils.isFirefoxOSDuration(sessionDuration)
```
jedp: If we don't combine the functions (as proposed above), consider creating a variable to hold this value, as you did with `sessionDuration`?

shane-tomlinson:
@jedp - I was wondering if we should do something different, isFirefoxOSDuration seems like a hokey indicator for "suppress_ask_if_users_computer". Maybe instead, we replace isFirefoxOSDuration with a function shouldSuppressAskIfUsersComputer that looks at the session duration under the hood? This way the two are not necessarily tied together and can be changed at will without updating the semantic meaning of the function name.

shane-tomlinson:
@jedp - I am auditing the code for the complete_ wsapis.  For an unverified user, if the user already a session with a FFOS duration set, and then they complete the email verification in the same browser, it looks like the complete_\* wsapis reset the session duration to the "ephemeral_session_duration_ms". This negates our awesome session length. It looks like neither wsapi.js->authenticateSession nor node_modules/client-session/client-sessions.js->setDuration does not take this into account.

shane-tomlinson:
@jedp - I am also adding tests for verifying the address of a user staged with allowUnverified and am getting 401 errors that say the user must provide their password, not what I was expecting.

shane-tomlinson:
Link to #3336.

jedp:
+1 to your suggestion to create a function like `shouldSuppressAskIfUsersComputer`.

shane-tomlinson:
@jedp - I primarily wrote the tests, you did the heavy lifting. I give this an r+.

Do you want to hit the green button?

PR #3334, opened 2013-05-01
make verifier work with app:// urls on FirefoxOS

jedp:
The apps on FirefosOS desktop builds, and at least some of the apps on mobile, have an `app://` schema.  

This PR updates the verifier's url matcher so those apps can use persona.

### DIFF
--- a/tests/verifier-test.js
+++ b/tests/verifier-test.js
@@ -61,7 +61,9 @@ suite.addBatch({
     'http://fakesite.com:8100 and http://fakesite.com:80': matchesAudience(false),
     'https://fakesite.com:9100 and https://fakesite.com:443': matchesAudience(false),
     'http://fakesite.com:80 and http://fakesite.com:8000': matchesAudience(false),
-    'https://fakesite.com:443 and https://fakesite.com:9000': matchesAudience(false)
+    'https://fakesite.com:443 and https://fakesite.com:9000': matchesAudience(false),
+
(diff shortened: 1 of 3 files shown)


tests/verifier-test.js
```diff
@@ -61,7 +61,9 @@ suite.addBatch({
     'http://fakesite.com:8100 and http://fakesite.com:80': matchesAudience(false),
     'https://fakesite.com:9100 and https://fakesite.com:443': matchesAudience(false),
     'http://fakesite.com:80 and http://fakesite.com:8000': matchesAudience(false),
-    'https://fakesite.com:443 and https://fakesite.com:9000': matchesAudience(false)
+    'https://fakesite.com:443 and https://fakesite.com:9000': matchesAudience(false),
+
+    'app://browser.gaiamobile.org and app://browser.gaiamobile.org:80': matchesAudience(true)
```
jedp, line 66: Is this correct? or should we think of app:// as 443?  I don't know what to do about the port matching here

warner, line 66: I.. I'm not sure, but I think of apps as being outside the world of HTTP, which inclines me to make it ignore port number (which is does, so yay).  It depends upon what the platform does when it gets an app: URL (whether they're served internally from cached bundles or something). I think testing with 80 is fine, that makes sure the regexp isn't too restrictive, even if we don't intend to use a port number there.

warner:
As far as I can tell, this looks good. If you've gotten someone with more Signin context than me (I've been out of the loop for a while) to say that app: URLs are ok, then I say land it.

Issue #2656, opened 2012-10-29
lock account after X bad passwords


benadida:
After X bad passwords, the Persona account should be locked. Simplest agreed-upon behavior: after 3 bad passwords, treat account as unverified (e.g. password reset).

We might want to add a CAPTCHA to prevent denial-of-service, but in that case we want to add it to "forgot password," too, so that might not be as urgent.

zaach:
We discussed using a router monitor for handling brute force attacks, which is dependent on #2508.

seanmonstar:
Way too busy on bigtent and b2g, removing myself as assignee

ygjb:
Can we get an update on when this feature will be implemented?

benadida:
@lloyd @zaach can we prioritize this issue in the next 3-4 weeks?

benadida:
alright, I'm going to see if I can provide some assistance on this one. I'm thinking the router approach may not be the right one, as that would mean adding smarts in the router about which URLs do what... which seems like the wrong direction. @zaach please shout if you think I'm missing something here.

shane-tomlinson:
@benadida - is this still on the priority queue?

benadida:
Yes, very much so.

lloyd:
@fmarier yvan wants this so bad he can taste it....

fmarier:
@lloyd and I discussed this and we propose keeping this as simple as possible:
- new `failedpasswords` column in the user table
- that column is incremented by one every time a user submits the wrong password
- if that count reaches 10, then we clear the password in the DB and reset the failed count to 0
- next time the user tries to login, they get the "forgot password" flow and my do an email verification & set a new password

shane-tomlinson:
@fmarier - I am hesitant to wipe the user's password, this could lead alice, as somebody that doesn't like bob, to cause bob to have to reset his password over and over again.

What about when the password count reaches 10, lock the user out for several minutes and send an email with a link to optionally reset the password?

lloyd:
after spending time with the code, I'd like to take a slight tweak on @fmarier proposal:
- if that count reaches 10, then when they try to authenticate they see "your account is locked, please click forgot password to reset your password"

this is identical in spirit but something we can execute quickly.

@shane-tomlinson in the future we can add captcha or some nicer option, given the severity of this and how much the security team wants it, I think it is better to allow alice to be a jerk for the short term.

lloyd:
remaining work:
1. dbwriter api call
2. add configuration parameter
[...]

PR #3337, opened 2013-05-02
require a password reset when too many failed password attempts are made...

lloyd:
... - issue #2656 

Even though there's zero UI change, I claim this is ready for merge

### DIFF
--- a/lib/wsapi/authenticate_user.js
+++ b/lib/wsapi/authenticate_user.js
@@ -30,8 +30,9 @@ exports.args = {
 };
 
 exports.process = function(req, res) {
-  function fail(reason) {
+  function fail(reason, code) {
     var r = { success: false };
+    if (code) r.code = code;
     if (reason) r.reason = reason;
     cef_logger.warn("AUTH_FAILURE", "User authentication failed", req, {msg:reason});
     logger.debug('authentication fails for user: ' + req.params.email + (reason ? (' - ' + reason) : ""));
@@ -45,7 +46,7 @@ exports.process = function(req, res) {
       return fail('no such user');
     }
 
-    function compare(hash) {
+    function compare(hash, failedAuthTries) {
       logger.debug("hash is " + hash);
       if (err) return wsapi.databaseDown(res, err);
 
@@ -70,6 +71,7 @@ exports.process = function(req, res) {
           return fail("internal password check error");
         } else if (!success) {
           cef_logger.warn("AUTH_FAILURE", "Password mismatch", req);
+          incrementAuthCount(req, res, uid);
           return fail("password mismatch for user: " + req.params.email);
         } else {
           if (!req.session) req.session = {};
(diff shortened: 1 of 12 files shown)


lib/wsapi/authenticate_user.js
```diff
@@ -86,28 +88,94 @@ exports.process = function(req, res) {
...
 
 
+function incrementAuthCount(req, res, uid) {
+  var u = wsapi.forwardWritesTo;
+  var m = u.scheme === 'http' ? http : https;
+  var preq = m.request({
+    host: u.host,
+    port: u.port,
+    path: '/wsapi/increment_failed_auth_tries?' + querystring.stringify({userid: uid}),
+    method: "GET",
```
benadida, line 127: should this be a POST?

lib/wsapi/authenticate_user.js
```diff
@@ -86,28 +88,94 @@ exports.process = function(req, res) {
...
+
+function resetFailedAuthCount(req, res, uid) {
+  var u = wsapi.forwardWritesTo;
+  var m = u.scheme === 'http' ? http : https;
+  var preq = m.request({
+    host: u.host,
+    port: u.port,
+    path: '/wsapi/reset_failed_auth_tries',
+    method: "GET",
+    agent: false,
```
benadida, line 156: same here, should this be a POST?

lloyd, line 156: this is an internal only API call from browserid (with only DB read access) to dbwriter.  if it's a POST, we must forward cookies and CSRF tokens, which seems meaningless.  what do you think?

lib/wsapi/authenticate_user.js
lloyd, line 127: ditto: this is an internal only API call from browserid (with only DB read access) to dbwriter. if it's a POST, we must forward cookies and CSRF tokens, which seems meaningless. what do you think?

lib/wsapi/reset_failed_auth_tries.js
```diff
@@ -0,0 +1,22 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const
+db = require('../db.js'),
+wsapi = require('../wsapi.js');
+
+exports.method = 'get';
+exports.writes_db = true;
+exports.authed = true;
+exports.internal = true;
```
benadida, line 12: I'm assuming this guarantees that this call cannot be made from the user's browser. Might be good to test for this (I don't think I see any test that covers this.)

benadida:
to @lloyd re: discussion of GET v. POST.

Longer term, I suggest sticking to standard GET/POST behavior, maybe even disabling CSRF protection on internal calls if that makes things easier. But I can live with this as is given the urgency of the fix.

lib/wsapi/reset_failed_auth_tries.js
lloyd, line 12: Good point.  We had internal tests, but they did not excercise all internal apis.  update in commit 1a4572c.

lloyd:
@benadida 

re GET v. POST on internal APIs - I agree on both counts - we should change things and make these exempt from CSRF checks where appropriate, and it shouldn't block this PR.

re testing accessibility of internal APIs, our `internal-wsapi-test.js` is updated and does this (1a4572c)

re test stability - it's fixed (205bbe9)

at present, pending an r+ from @fmarier and @benadida, I believe this is ready to merge.

benadida:
r+ from me

fmarier:
r+ from me too

lloyd:
holding on this till the final FirefoxOS patches land.  @shane-tomlinson or @jedp ping me when everything needed to support FirefoxOS is done and I'll rebase and land this if I can do it safely.

PR #3344, opened 2013-05-03
Issue 3343 center delay

shane-tomlinson:
@jedp - mind reviewing? The easiest way to get to this is from the development menu


shane-tomlinson:
Woah. This is based off the wrong branch. Closing.

PR #3346, opened 2013-05-03
Hide the "loading" screen whenever the "check your email" screen is shown.

### LINKED ISSUE #3339: "check your email" screen not shown for unverified email when visiting a site that requires verified emails.
STR (using local dev because we have no test RPs with this functionality)

1) Open test RP
2) Select "Allow Unverified"
3) Open dialog
4) Create new account
5) On test RP, uncheck "Allow Unverified"
6) Open dialog
7) Select same email

Result:
"Loading" screen with spinner

Expected:
"Check your email" screen.

shane-tomlinson:
@jedp - I'm starting to feel bad about all of these requests, but can you do this one?

Fixes #3339

### DIFF
--- a/resources/static/dialog/js/modules/pick_email.js
+++ b/resources/static/dialog/js/modules/pick_email.js
@@ -65,9 +65,11 @@ BrowserID.Modules.PickEmail = (function() {
     if (!! record) {
       // Show the signing in screen as soon as the user presses the button so
       // that it does not seem like there is a huge delay while things being
-      // processed.
+      // processed. Show the load screen without a title (instead of "signing
+      // in") because the user may have to take some action after this - like
+      // verify their email address or answer yes/no to "is this your computer"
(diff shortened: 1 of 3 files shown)


resources/static/dialog/js/modules/pick_email.js
```diff
@@ -65,9 +65,11 @@ BrowserID.Modules.PickEmail = (function() {
     if (!! record) {
       // Show the signing in screen as soon as the user presses the button so
       // that it does not seem like there is a huge delay while things being
-      // processed.
+      // processed. Show the load screen without a title (instead of "signing
+      // in") because the user may have to take some action after this - like
+      // verify their email address or answer yes/no to "is this your computer"
       self.renderLoad("load", {
-        title: gettext("signing in")
+        title: ""
```
jedp, line 72: makes sense

PR #3358, opened 2013-05-03
Styling bag of goodies

shane-tomlinson:
@jedp - some styling fixes from the b2g merge. 4 commits, 3 issues.

### DIFF
--- a/resources/static/dialog/css/style.css
+++ b/resources/static/dialog/css/style.css
@@ -74,7 +74,7 @@ h2 {
   font-size: 18px;
 }
 
-.form_section h2.small {
+.form_section h2.deemphasize {
   font-size: 13px;
   font-weight: normal;
   line-height: 18px;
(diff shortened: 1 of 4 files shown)


resources/static/dialog/css/style.css
```diff
@@ -271,6 +277,16 @@ section > .contents {
     opacity: 1;
 }
 
+/**
+ * If the error or delay screens are shown, hide the load screen. If
+ * not hidden, the load screen causes a white top and bottom margin.
+ */
+.error #load, .delay #load {
+  opacity: 0;
+  -ms-filter:"progid:DXImageTransform.Microsoft.Alpha(Opacity=100)";
```
jedp, line 286: that looks exciting!

resources/static/dialog/css/style.css
```diff
@@ -71,6 +71,10 @@ h2 {
 }
 
 .form_section h2 {
+  font-size: 18px;
+}
+
+.form_section h2.small {
```
jedp: So `h2.small` gets used on desktop, and `h2` not `small` gets used on mobile; A small point, but those names seem counter-intuitive to me.  I think of mobile being "small".

jedp:
Thanks, Shane!

Issue #3369, opened 2013-05-07
params with leading underscore filtered out of b2g identity messages


jedp:
After all those conversations about `_experimental_*` properties, I failed to notice that underscore params are actually filtered out by the identity b2g message passer.

This may be regarded as a bug in the identity service, but it means that for now, we have to take our `_experimental_foo` properties and rename them to `experimental_foo`

jedp:
works  and merged.

PR #3370, opened 2013-05-07
Underscore props are filtered out of b2g identity messages, alas

jedp:
Fixes issue #3369


seanmonstar:
I imagine this comes from the 1-line patch for B2G that got a giant pile up in Bugzilla... but also, for some reason the lack of _ is scarier. Probably a stupid conception on my part, as the "experimental" prefix should scare away everyone else.

jaredhirsch:
wait, so does this mean I need to watch github during work week?! ;-)

shane-tomlinson:
@jedp - can you update the parameters list in example/rp/index.html as well?

shane-tomlinson:
@6a68 - Of course!

jedp:
@seanmonstar this is not actually that issue in b2g, but yeah, there is platform code that's already filtering underscore properties out of the dom caller's message before sending it through gecko.  I want to rework the way we do that, but it won't be in this run, unfortunately.

lloyd:
r+ given that @shane-tomlinson 's comments were addressed.

PR #3372, opened 2013-05-07
Disable Inputs

### LINKED ISSUE #3294: Inputs not disabled when adding another address or signing in for the first time
While the dialog is "Checking with your email provider," the text input is disabled, but the "next" button is still active. There's no spinner indicating progress / work.

Screen Shot 2013-04-22 at 10 31 35 AM

When adding an email address to an existing account, the dialog remains fully enabled and is interactive during provisioning.

Screen Shot 2013-04-22 at 10 34 12 AM

The text input and button should be disabled during these phases.

arlolra:
Fixes #3294

### DIFF
--- a/resources/static/dialog/js/modules/add_email.js
+++ b/resources/static/dialog/js/modules/add_email.js
@@ -13,7 +13,8 @@ BrowserID.Modules.AddEmail = (function() {
       complete = helpers.complete,
       tooltip = bid.Tooltip,
       hints = ["addressInfo"],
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@arlolra - I manually merged this in sha 80b30b3b9772f6e1cc2e5c1d04e4f534f55cad3c.

r+, tests continue to pass, it does what it says it does.

Thanks for the contribution!

PR #3381, opened 2013-05-09
Fix getSilentAssertion for B2G.

### LINKED ISSUE #3380: [b2g] Persona unable to perform automatic login after user signs in.
Affects b2g 1.0.1, this is a blocker.

Original Bugzilla bug is https://bugzilla.mozilla.org/show_bug.cgi?id=870412

shane-tomlinson:
Firefox OS 1.0.1 keeps one copy of the communication iframe around. When it loads, it fetches a copy of session context and then caches it. Whenever the user signs in to a tab, the communication iframe's session context cache is not updated. To avoid updates to B2G code, force a refresh of session context whenever calling user.getSilentAssertion.

This fixes #3380

### DIFF
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -226,6 +226,13 @@
 
     function checkAndEmit() {
       // this will re-certify the user if neccesary
+      // Firefox OS 1.0.1 keeps one copy of the communication iframe in memory
+      // all the time. The communication iframe gets a copy of session context
+      // when it first loads and caches it, it is never updated even if the user
+      // uses the dialog to sign in. To avoid manually updating the cache when the user
+      // signs in using the dialog, clear the cache every time a new tab
+      // requires internalWatch.
+      user.clearContext();
       user.getSilentAssertion(loggedInUser, function(email, assertion) {
         if (email) {
           // only send login events when the assertion is defined - when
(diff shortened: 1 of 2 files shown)


resources/static/dialog/js/misc/internal_api.js
```diff
@@ -226,6 +226,13 @@
 
     function checkAndEmit() {
       // this will re-certify the user if neccesary
+      // Firefox OS 1.0.1 keeps one copy of the communication iframe in memory
+      // all the time. The communication iframe gets a copy of session context
+      // when it first loads and caches it, it is never updated even if the user
+      // uses the dialog to sign in. To avoid manually updating the cache when the user
+      // signs in using the dialog, clear the cache every time a new tab
+      // requires internalWatch.
+      user.clearContext();
```
jedp, line 235: my god.  12 hours later.  so simple.

resources/static/test/cases/dialog/js/misc/internal_api.js
```diff
@@ -210,26 +210,28 @@
...
-        else {
-          ok(false, "unexpected method call: " + resp.method);
-        }
-      }, {
-        origin: ORIGIN,
-        loggedInUser: "testuser2@testuser.com"
-      }, console.log);
-    });
+    xhr.setContextInfo("auth_level", "password");
+    xhr.setContextInfo("userid", 1);
```
jedp, line 214: Because authenticate no longer means anything with internal.watch.  This makes sense.  And I don't imagine this pattern (authenticate around internal.*) would occur elsewhere.

jedp:
Fantastic work, Shane.  You saved our skins.  

Having testing with Edwin and you on platform, I feel very good about this.

I'm going to let Travis do his thing, but pending completion of the tests, this is r=me.

PR #3382, opened 2013-05-09
issuers for disabling federation on b2g

jedp:
I don't know if these are the right names, but we do need to settle on something, so here's kicking off the discussion.

/cc @lloyd, @benadida, @kumar303, @shane-tomlinson

### DIFF
--- a/config/aws.json
+++ b/config/aws.json
@@ -30,5 +30,9 @@
   "public_static_url": null,
 
   // enable the measurement of the dom loading
-  "measure_dom_loading": true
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@jedp - you discussed using login.persona.org on the list. Is that the agreed upon approach?

jedp:
I've updated the commit to allow either `firefoxos` or `login.persona.org` to give latitude to us and marketplace, looking ahead to the day when we swing `firefoxos.persona.org` to be a CNAME for `login`.

PR #3387, opened 2013-05-14
Fix typo which breaks ActiveX-based XHRs (fixes #3383)

### LINKED ISSUE #3383: Typo in micrajax breaks support for IE8 without native XMLHTTP support
There is a typo in [micrajax](https://github.com/mozilla/browserid/blob/2a638af0190162fc3480ed07458e66013d7ec775/resources/static/common/js/lib/micrajax.js#L27):

```
new ActiveXObject('Microsoft.XM/LHTTP');
```

should be:

```
new ActiveXObject('Microsoft.XMLHTTP');
```

It looks like the spurious slash was introduced in e00740316ccf61f60f6efc29905da6d039178fc8.

fmarier:
This can be tested on IE by disabling the "Enable native XMLHTTP support" setting in "Tools | Internet options | Advanced".

(I have not been able to verify it myself since I don't have an IE8 VM handy.)

### DIFF
--- a/resources/static/common/js/lib/micrajax.js
+++ b/resources/static/common/js/lib/micrajax.js
@@ -24,7 +24,7 @@ window.Micrajax = (function() {
     }
     else if (window.ActiveXObject) {
       // ...if not, try the ActiveX control
-      xhrObject = new ActiveXObject('Microsoft.XM/LHTTP');
+      xhrObject = new ActiveXObject('Microsoft.XMLHTTP');
     }
 
     return xhrObject;


callahad:
I haven't directly tested this, but r+ on the basis that it's logically identical to what [jquery 1.9.1 does](https://github.com/jquery/jquery/blob/1.9.1/src/ajax/xhr.js#L25-38).

shane-tomlinson:
@fmarier - can you submit a PR to https://github.com/stomlinson/micrajax as well?

fmarier:
@shane-tomlinson stomlinson/micrajax doesn't have that problem

shane-tomlinson:
@fmarier - Thanks for checking!

PR #3388, opened 2013-05-14
Stabilize selenium

shane-tomlinson:
@6a68 - could I get a review?

### DIFF
--- a/automation-tests/tests/add-primary-to-primary.js
+++ b/automation-tests/tests/add-primary-to-primary.js
@@ -18,44 +18,45 @@ user = require('../lib/user.js'),
 timeouts = require('../lib/timeouts.js');
 
 var browser,
+    testIdp,
     primaryEmail,
     secondPrimaryEmail;
 
 runner.run(module, {
   "setup all the things": function(done) {
-    testSetup.setup({ b:1, e:2 }, function(err, fix) {
+    testSetup.setup({ b:1, testidps:1 }, function(err, fix) {
       if (fix) {
         browser = fix.b[0];
-        primaryEmail = {
-          email: fix.e[0],
-          pass: fix.e[0].split('@')[0],
-        };
-        secondPrimaryEmail = {
-          email: fix.e[1],
-          pass: fix.e[1].split('@')[0],
-        };
+        testIdp = fix.testidps[0];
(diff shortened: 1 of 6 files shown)


jaredhirsch:
looking now

jaredhirsch:
@shane-tomlinson: It looks like the testSetup method has been changed, but those changes aren't included in this PR. I think maybe that file didn't get added? Otherwise I don't quite understand how testSetup is returning `fix.testidps`.

I have some other speculative feedback--not sure it'll be on the money until I see how testSetup changed:

As far as I can tell, the only reason we return testidp is to generate some fake emails, and then call `testidp.enableSupport`. Is there any reason we couldn't just ask testSetup to give us several fake emails, and then call `testidp.enableSupport` inside the setup function?

The advantages to doing this are making tests decoupled from testidp's existence, reducing boilerplate inside that testSetup callback (which itself is an ugly hack IMO), as well as making the testSetup call super explicit about the fixtures we are asking for: we don't want 1 testidp, we want 2 dummy emails.

Thoughts?

(edited to name-check shane to ensure he sees this comment :-) )

jaredhirsch:
oh also, @edmoz and @karlht and @jrgm might want to peep this change, since it impacts stability and changes our test infrastructure landscape. (We should really document all the services that underpin the selenium tests...)

shane-tomlinson:
@6a68 - Support for testidps was added a few months ago to help test BigTent and primary upgrade/downgrade paths, I made no changes there.

From my understanding, testidp primary support is not enabled by default so that upgrades can be tested. 

Scenario:
- User registers email address: testuser@xyz.com. xyz.com is not a primary.
- xyz.com adds primary support
- User tries to sign in to a site using testuser@xyz.com, at this point the user should see the "Your provider has been upgraded!" screen.

shane-tomlinson:
@6a68 - we could enable primary support by default, but we must surface testidp's existence so that we can force transition states. I am fine with doing so as long as we retrofit other tests that start off with the assumption that the testidp's primary support is disabled.

jaredhirsch:
@shane-tomlinson Ah, obviously I haven't kept a close enough eye on this codebase :-P

I think this is fine as it stands, given that extra context :beers:

Issue #3385, opened 2013-05-13
[ie8] train-2013.05.08 - frontend test failures


jrgm:
Works in other browsers, including IE9 and IE10, but not in IE8. These two
errors are shown below, and the test appears to stop executing shortly after.

https://jrgm-train0508.personatest.org/test/?filter=dialog%2Fjs%2Fmodules%2Fauthenticate%3A%20known%20secondary%20email%20declared%20in%20options%20-%20show%20password%20field

https://jrgm-train0508.personatest.org/test/?filter=dialog%2Fjs%2Fmodules%2Fauthenticate%3A%20allowUnverified%20with%20an%20unverified%20email%20declared%20in%20options%20-%20show%20password%20field

The message shown for the two test failures is:

```
1. Died on test #1: Can't move focus to the control because it is invisible, not enabled, or of a type that does 
not accept the focus. - { "name": "Error", "number": -2146826178, "description": "Can't move focus to the control
because it is invisible, not enabled, or of a type that does not accept the focus.", "message": "Can't move focus
to the control because it is invisible, not enabled, or of a type that does not accept the focus." }
```

Marking this as five star.

jaredhirsch:
looking now

jaredhirsch:
at [line 242 of dialog/js/modules/authenticate.js](https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/authenticate.js#L242) we give the PASSWORD_SELECTOR node focus via `dom.focus(PASSWORD_SELECTOR)`. If you set a breakpoint on that line and run the qunit tests, then ask IE8 if the node is hidden, IE8 will say that it is indeed hidden:

```
// assuming you've hit the breakpoint inside IE8 dev tools
>>x = $(PASSWORD_SELECTOR)
{...}
>>x.is(':hidden')
true
```

This explains the first error message.

Next, I'm going to git bisect recent history to see where this crept in. I suspect it's from an innocent-looking styling change.

jaredhirsch:
This goes back at least to the previous train. Not sure if it warrants five-star status.

Looking further back through history.

jrgm:
It's fundamentally different in this train-2013.05.08. I have passed this test in train-2013.04.26 (yes flaky). But now, we can't complete the tests at all. 

No, it may not be five-star, but needs to be worked on.

jrgm:
So I did a git bisect with an awsbox and it points to:

```
[...]

PR #3392, opened 2013-05-14
Issue 3385 ie8 failures

shane-tomlinson:
@jrgm, @6a68 - this fixes the IE8 ghettoness in the tests.

This depends on #3391 being merged for dev and may not be necessary for a merge into train-2013.05.08

### DIFF
--- a/resources/static/common/js/lib/dom-jquery.js
+++ b/resources/static/common/js/lib/dom-jquery.js
@@ -322,7 +322,17 @@ BrowserID.DOM = ( function() {
          * @param {selelector || element} elementToFocus
          */
         focus: function( elementToFocus ) {
-          jQuery( elementToFocus ).focus();
+          var el = jQuery( elementToFocus );
+
+          // IE8 blows up when trying to focus an invisible or disabled
+          // element. Keep that from happening. See issue #3385
+          if ( el.is( ':visible' ) && el.is( ':enabled' ) ) {
+            // IE8 is difficult. Sometimes a new element cannot be
+            // programatically focused if the old element is not first blurred.
+            jQuery( ':focus' ).blur();
+
+            el.focus();
+          }
         },
 
         /**
--- a/resources/static/test/cases/dialog/js/modules/set_password.js
+++ b/resources/static/test/cases/dialog/js/modules/set_password.js
@@ -57,6 +57,10 @@
 
   asyncTest("submit in password field with good password - skip to vpassword field", function() {
     $("#password").val("password");
+    $("#vpassword").val("");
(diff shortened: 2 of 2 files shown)


tests/cert-key-test.js
```diff
@@ -192,7 +192,7 @@ suite.addBatch({
         assert.isString(r);
       },
       "and adding this email": {
-        topic: function(assertion)  {
+        topic: function(err, assertion)  {
```
jaredhirsch, line 195: how is this related to the current bug?

resources/static/test/cases/dialog/js/modules/set_password.js
```diff
@@ -57,6 +57,10 @@
 
   asyncTest("submit in password field with good password - skip to vpassword field", function() {
     $("#password").val("password");
+    $("#vpassword").val("");
+    // IE8 is difficult. To programatically focus a new element, sometimes it
+    // is necessary to blur the old element.
+    $(":focus").blur();
```
jaredhirsch, line 63: it would really be nice to use our DOM library here. maybe I'll open a separate PR to fix.

resources/static/common/js/lib/dom-jquery.js
```diff
@@ -322,7 +322,17 @@ BrowserID.DOM = ( function() {
...
         focus: function( elementToFocus ) {
-          jQuery( elementToFocus ).focus();
+          var el = jQuery( elementToFocus );
+
+          // IE8 blows up when trying to focus an invisible or disabled
+          // element. Keep that from happening. See issue #3385
+          if ( el.is( ':visible' ) && el.is( ':enabled' ) ) {
+            // IE8 is difficult. Sometimes a new element cannot be
+            // programatically focused if the old element is not first blurred.
+            jQuery( ':focus' ).blur();
```
jaredhirsch, line 332: ah, so this is the problem? I thought it was that IE8 said that the PASSWORD_SELECTOR el was hidden and was trying to receive focus?

jaredhirsch, line 332: oh, duh. you only do this if the el is visible and enabled. should we wait a turn if it isn't?

jaredhirsch, line 332: just to clarify for 'the record', this patch fixes 2 bugs: first, don't try to focus an el unless it's visible and enabled. This fixes #3385, caused by trying to focus an el that is hidden. Second, if the target el is focusable, blur the currently-focused el (if any) before attempting to focus().

jaredhirsch:
:rainbow: :rainbow: what does it mean? R+, that's what

PR #3395, opened 2013-05-14
Fix the spinner (main site and it's wonkiness)

### LINKED ISSUE #3394: [main site] delay spinner does not spin

shane-tomlinson:
@jedp - could I get a quick review?

@jrgm - should we merge this into train-2013.05.08 as well?

fixes #3394


shane-tomlinson:
@jedp - An ephemeral instance is already set up for testing purposes - spinner.personatest.org.

jedp:
trying on a v1.0.1 unagi ...

jedp:
I like this a lot.  I like the spinner smaller, and it certainly looks centered on device.  r=me.

shane-tomlinson:
Jed just informed me that we can allow this to go into dev to make its way to prod in time. It doesn't look like we need to hotfix stage.

PR #3398, opened 2013-05-14
Fix for users who are unable to sign out.

### LINKED ISSUE #3386: Users cannot log out
We're seeing that calling `navigator.id.logout()` no longer logs the user out. You can see it for yourself:
1. Visit marketplace-dev.allizom.org
2. Sign in with Persona
3. Sign out (hover over the gear in the upper right)
4. Refresh the page

We're calling `navigator.id.watch()` with a blank string for the email and on sign out we're calling `navigator.id.logout()`, and yet Persona still magically fires `onlogin` with an assertion when the page loads.

shane-tomlinson:
@seanmonstar, @callahad, @jrgm - This is for a hotfix of prod. Review would be fantastic.

When upgrading localStorage from loggedIn to use the site namespace, I failed to remove the old loggedIn namespace, making it so users would sign out, refresh the page, and be signed in again.

fixes #3386

### DIFF
--- a/resources/static/common/js/storage.js
+++ b/resources/static/common/js/storage.js
@@ -44,6 +44,7 @@ BrowserID.Storage = (function() {
     for (var site in allInfo) {
       siteSet(site, "logged_in", allInfo[site]);
     }
+    storage.removeItem("loggedIn");
   }
   upgradeLoggedInInfo();
   // END TRANSITION CODE
@@ -595,5 +596,13 @@ BrowserID.Storage = (function() {
      * @method setDefaultValues
      */
     setDefaultValues: setDefaultValues
+    // BEGIN TRANSITION CODE
+    /**
+     * Upgrade the site->user logged in info from the loggedIn namespace to be
+     * under the site namespace.
+     */
+    ,
+    upgradeLoggedInInfo: upgradeLoggedInInfo
+    // END TRANSITION CODE
   };
 }());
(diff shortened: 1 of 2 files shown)


resources/static/common/js/storage.js
```diff
@@ -44,6 +44,7 @@ BrowserID.Storage = (function() {
     for (var site in allInfo) {
       siteSet(site, "logged_in", allInfo[site]);
     }
+    storage.removeItem("loggedIn");
```
shane-tomlinson, line 47: This is the key line - this removes the loggedIn namespace after upgrading.

resources/static/common/js/storage.js
```diff
@@ -595,5 +596,13 @@ BrowserID.Storage = (function() {
      * @method setDefaultValues
      */
     setDefaultValues: setDefaultValues
+    // BEGIN TRANSITION CODE
+    /**
+     * Upgrade the site->user logged in info from the loggedIn namespace to be
+     * under the site namespace.
+     */
+    ,
+    upgradeLoggedInInfo: upgradeLoggedInInfo
```
shane-tomlinson, line 605: I surfaced this function to be able to test it.

seanmonstar:
r+, waiting on travis to complete

jrgm:
@shane-tomlinson - Thanks for the fix! (p.s., we'll need this also on train-2013.05.08 and dev, of course).

PR #3400, opened 2013-05-14
make all unit tests executable

lloyd:
a test-only change for consistency and easy of invocation of specific tests.


callahad:
r+, just file mode changes in the tests/ dir. Seems reasonable

PR #3405, opened 2013-05-14
Last minute firefoxos fixes

lloyd:
1. (doesn't affect production code) aws / ephemeral specific change to let us support localization in aws instances
2. update verifier to add an `experimental_` prefix to `allowUnverified` and `forceIssuer`


jrgm:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=872336 to deploy this, and built the rpm as noted there. I could just update the rpm on the verifier, but that smells. 

Also, there is a hotfix on train-2013.04.26 to fix some users signout - https://github.com/mozilla/browserid/pull/3398 which I'd like to run through stage first. Ugh.

PR #3408, opened 2013-05-15
Issue 3407 mobile buttons

shane-tomlinson:
@jedp - could I get a review of this one?

Clean up buttons on mobile.

### DIFF
--- a/resources/static/common/css/style.css
+++ b/resources/static/common/css/style.css
@@ -402,6 +402,10 @@ a.secondary[disabled], .submit_disabled a.secondary, .submit_disabled a.secondar
   float: right;
 }
 
+.left {
+  float: left;
+}
+
 .center {
   text-align: center;
 }
--- a/resources/static/dialog/css/m.css
+++ b/resources/static/dialog/css/m.css
@@ -48,7 +48,7 @@
     height: 40px;
     min-width: 105px;
     font-size: 14px;
-    padding: 6px 25px 8px;
+    padding: 6px 0 8px;
     float: right;
     font-weight: 400;
     color: #414343;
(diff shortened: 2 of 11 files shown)


shane-tomlinson:
@jedp - I have an ephemeral instance set up already - "mobile-buttons.personatest.org" that you can use to test.

jedp:
Awesome that you set up an ephemeral instance to make this easy to test.  Checking it out now.

jedp:
Hmm.  Doesn't look right on device.  I presume you're testing on desktop and you're getting completely different results?  Great.



Also, the spinner began spinning the word 'loading' before the spinner appeared.  Maybe remove the alt text?

shane-tomlinson:
B2G uses Gecko 19, which uses the old flex-box. !@$!@^

jedp:
I think it might actually be gecko 18

jedp:
Buttons look great now.

There's a weird problem, though, with the keyboard flickering away and reappearing on email input.  This could be wrong, but it sort of seems like the dialog is trying to resize itself while text is being entered?

I'm trying the text input UI Test app on the phone, using the input type='email' field, and I can't get it to misbehave in the same way.

Yeah, and then sending the device to use firefoxos.persona.org, the problem is not evident.  So there must be something in this code that's causing the dialog to do something to dismiss and then recall the keyboard.  Hmm.

shane-tomlinson:
@jedp - that strange flickering I saw too and fixed in a different PR - #3419, commit 277c61d218c6255da0f58a7dbea033f111caa8bd. Want me to cherry-pick that fix over to this?

I'll rebase now because I merged stuff into dev after I pushed this PR.

jedp:
Sweet.  Testing now on mobile-buttons

jedp:
Looks beautiful on device.  Nice fix.

jedp:
Tests pass, looks good, r=me.  merging

PR #3415, opened 2013-05-16
Train 2013.05.08 - Fix for users who are unable to sign out.

jrgm:
This is the issue fixed in #3398, which is merged into train-2013.04.26 and dev branch.

This PR is to merge it into the pending stage train-2013.05.08.


lloyd:
r+

Issue #3342, opened 2013-05-03
Loading spinner does not spin on IE8/9, mobile webkit


shane-tomlinson:
Webkit is due to a Chrome for Android bug that causes the UI to be completely blank and unusable. IE is because it does not support CSS animations.

jaredhirsch:
PR has been merged, closing this bug accordingly

PR #3416, opened 2013-05-16
Enable the spinner on mobile webkit, it works now.

shane-tomlinson:
Since we changed the way the spinner is animated, it no longer gives Chrome on Android a problem.

issue #3342

### DIFF
--- a/resources/static/common/css/style.css
+++ b/resources/static/common/css/style.css
@@ -533,26 +533,26 @@ footer .help {
 
 @-webkit-keyframes spin {
     from {
-        -webkit-transform: rotate(0turn);
+        -webkit-transform: rotate(0deg);
     }
     to {
-        -webkit-transform: rotate(1turn);
+        -webkit-transform: rotate(365deg);
     }
 }
 @-moz-keyframes spin {
     from {
-        -moz-transform: rotate(0turn);
+        -moz-transform: rotate(0deg);
     }
     to {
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@6a68 - you have an Android 4.2 device, right?

If so, could you use Chrome to load up the dialog from spinner.123done.org and see if the spinner works for you?

jaredhirsch:
Hm, my 4.2.1 nexus handset has "browser", not chrome, as the default. Works fine in both embedded and chrome.

I have an old galaxy S on android 2.2, and it displays the spinner but doesn't seem to rotate it. Whatever.

I have a nexus 10 tablet, but it opens the popup in a new tab, rather than lightboxing and showing the spinner.

jaredhirsch:
@shane-tomlinson do you need me to do anything else here?

shane-tomlinson:
@6a68 - that is a fantastic summary, I didn't think to test my 2.2 device. Nothing else you need to do except click the green button if you are comfortable!

shane-tomlinson:
@6a68 - I changed the way the rotations are done and it works now on my 2.3.6 device, can you give it a go on your 2.2?

jaredhirsch:
@shane-tomlinson it spins on 2.2 :surfer:

@jedp @lloyd is this one worth merging to break the freeze? does marketplace need it?

PR #3419, opened 2013-05-16
Issue 3410 disable continue

shane-tomlinson:
@jedp - they are coming in fast and furious!

Disable "continue" button on mobile until the user has typed in a character.

```
* On mobile, rename the next button to continue
* Disable the "continue" button until the user types a char in the email field.
* Clean up the button hover states on desktop and mobile when the button is diabled.
```

This builds upon #3408, so that should be reviewed first.

The ephemeral instance to test on is disable-continue.personatest.org

### DIFF
--- a/resources/static/test/cases/dialog/js/modules/authenticate.js
+++ b/resources/static/test/cases/dialog/js/modules/authenticate.js
@@ -33,7 +33,8 @@
       TRANSITION_TO_SECONDARY_LABEL = "#authentication_form .label.transition_to_secondary",
       PASSWORD_LABEL = "#authentication_form .label.password_state",
       IDP_SELECTOR = "#authentication_form .authentication_idp_name",
-      AUTHENTICATION_CLASS = "authentication";
+      AUTHENTICATION_CLASS = "authentication",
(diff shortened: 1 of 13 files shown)


jedp:
you animal.  testing now

jedp:
Nice.  Looks great on device.  Tests good.  r=me.

resources/static/test/cases/dialog/js/modules/authenticate.js
```diff
@@ -464,5 +465,23 @@
...
+    // type a char
+    $(EMAIL_SELECTOR).val("t");
+    controller.emailChange();
+    ok( ! $(CONTINUE_BUTTON_SELECTOR).attr("disabled"));
+
+    // delete the char.
+    $(EMAIL_SELECTOR).val("");
+    controller.emailChange();
+    ok($(CONTINUE_BUTTON_SELECTOR).attr("disabled"));
+  });
```
jedp, line 484: You are an exceptionally thorough person.  Even under all this pressure, you do not relax your standards.  That's why this is all awesome.

PR #3420, opened 2013-05-16
Fix the set password input fields being too wide if the user enters an e...mail address > 31 characters on a mobile device.

### LINKED ISSUE #3409: [B2G] Verify password page is too wide for the frame
For the unverified user workflow at the step for create/verify password the input fields are too side. Also the footer text does not seem right, or at least to the precision standards of the Identity team.

The second screenshot shows panned to the right side; there is a grey border on the side.

shane-tomlinson:
@seanmonstar - could I get a quick review?

fixes #3409 

This can be tested on input-too-wide.personatest.org/input-too-wide.123done.org

### DIFF
--- a/resources/static/dialog/css/m.css
+++ b/resources/static/dialog/css/m.css
@@ -192,6 +192,12 @@
     visibility: hidden;
     display: block;
     height: 0;
+    /* hidden items still take up their natural width in the DOM, which can
+     * cause some funkiness if there is a really long email address. Long
+     * addresses cause the set password input boxes to be wider than the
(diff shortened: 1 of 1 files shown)


seanmonstar:
The input doesn't grow off screen, but the body does. Scrollbars show up briefly, to let the user know you can scroll.
screenshot_2013-05-16-12-22-25

shane-tomlinson:
Travis is happy, Selenium is happy. Sean is happy, I'm happy. Who wants to hit the button?

PR #3432, opened 2013-05-17
Make sure the email address is focused when the dialog starts up.

### LINKED ISSUE #3424: frontend tests failing on login.dev.anosrep.org with a real browser
There's at least a typo here - https://github.com/mozilla/browserid/blob/dev/resources/static/test/cases/dialog/js/misc/state.js#L578

But with that url quoted correctly locally, there is still a test failure - 

http://127.0.0.1:10002/test/?filter=dialog%2Fjs%2Fmodules%2Fauthenticate%3A%20authentication%20form%20initialized%20on%20startup%2C%20hidden%20on%20stop

It passes in the phantomjs tests though, perhaps because it is skipped (?).

shane-tomlinson:
@seanmonstar and @jrgm - could I get a review?
- The input element was not being focused because we prematurely said that the dialog was in the enterEmailState.
- Fix the failing QUnit test when it is run in isolation.

An ephemeral instance, focus-damnit.personatest.org is set up for testing.

fixes #3424

### DIFF
--- a/resources/static/common/js/lib/dom-jquery.js
+++ b/resources/static/common/js/lib/dom-jquery.js
@@ -329,7 +329,12 @@ BrowserID.DOM = ( function() {
           if ( el.is( ':visible' ) && el.is( ':enabled' ) ) {
             // IE8 is difficult. Sometimes a new element cannot be
             // programatically focused if the old element is not first blurred.
-            jQuery( ':focus' ).blur();
+            // IE8 is doubly difficult because the default element that is
+            // focused is the body. If you blur the body, it puts a dialog
+            // behind its parent window.
+            if (! jQuery( ':focus' ).is("body")) {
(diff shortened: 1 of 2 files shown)


resources/static/common/js/lib/dom-jquery.js
```diff
@@ -329,7 +329,12 @@ BrowserID.DOM = ( function() {
           if ( el.is( ':visible' ) && el.is( ':enabled' ) ) {
             // IE8 is difficult. Sometimes a new element cannot be
             // programatically focused if the old element is not first blurred.
-            jQuery( ':focus' ).blur();
+            // IE8 is doubly difficult because the default element that is
+            // focused is the body. If you blur the body, it puts a dialog
+            // behind its parent window.
```
seanmonstar, line 334: :crying_cat_face:

PR #3433, opened 2013-05-17
Fix submit buttons staying disabled after cancel from set-password and re-submit with same email address.

### LINKED ISSUE #3431: submit button stays disabled permanently after cancel from set password screen and re-submitting
STR

1) Open dialog, if signed in, sign out
2) Enter an unknown secondary email address
3) On the set password screen, click cancel
4) On authentication screen, click Next button without changing email address.
5) wah wah.

shane-tomlinson:
@seanmonstar - can you review this one too? This should probably be reviewed after #3432 because the same front end unit test fails here without the merge from there.
- Remove the "submit_disabled" class on the body when address info retreival is complete.
- fixes #3431


seanmonstar:
r+

says I can't automatically merge :crying_cat_face:

PR #3441, opened 2013-05-20
Use Feura Sans font for the mobile layout.

### LINKED ISSUE #3430: [b2g] B2G apps use Feura Sans font set, not Open Sans

shane-tomlinson:
@skinny97214, @mhanratty, I have updated the dialog available from http://all-together-now.123done.org to use the Feura Sans font set available from the https://github.com/mozilla-b2g/gaia/ repo. Can you see what you think?

fixes #3430

### DIFF
--- a/bin/static
+++ b/bin/static
@@ -14,7 +14,8 @@ http = require('http'),
 urlparse = require('urlparse'),
 express = require('express'),
 connect_fonts = require('connect-fonts'),
-connect_fonts_opensans = require('connect-fonts-opensans');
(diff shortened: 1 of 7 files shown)


jedp:
Holding off; @skinny97214 says no go on Feura

skinny97214:
Just to clarify, hold until Maureen says go.

shane-tomlinson:
Holding.

jedp:
In Issue #3430, @mhanratty confirms that we do want to use Feura Sans on gaia

PR #3444, opened 2013-05-21
Issue 3410 styling updates

shane-tomlinson:
@jedp - tag, you're it!
- Add a Persona watermark next to email addresses in the email picker
- Update the "submit" button style
- Give the buttons in "Is this your computer" more space
- Shrink the TOS/PP text by 1 pt.
- Update the :focus highlight on an input element
- Fix blue cancel button on Android 2.2/2.3 native browser.

This can be seen on all-together-now.


jedp:
It looks like there's a resize artifact on the right-hand side of those images.

I've tried again since Michael sent the high-res assets, but I'm still seeing the artifacts.

Apart from that, the flow and function seems perfect.

shane-tomlinson:
Wah wah, I'll have a look at what that artifact is. Thanks for the thorough review @jedp - I didn't notice that in my own testing until you pointed it out!

jedp:
The artifact on the right-hand side is said to be a rendering bug on gaia.  We'll follow that up in a bugzilla bug and link from here

PR #3445, opened 2013-05-21
Update the TOS/PP on mobile to follow the error screen styles with the OK button at the bottom....

shane-tomlinson:
@jedp - this can be seen on all-together-now

Should probably hold off until we get the final go ahead from @mhanratty or @skinny97214


shane-tomlinson:
terms of service

jedp:
@skinny97214 , @mhanratty: Please note that the horizontal and vertical scrollbars in the image above are _not_ visible on device.  They show up on the desktop emulator, but not when rendered on the phone.

jedp:
On email, Michael says "Perfect! It's a go."

PR #3452, opened 2013-05-24
Fix TOS/PP iframe scrolling.

### LINKED ISSUE #3442: [mobile][firefox][safari] cannot scroll Persona TOS/PP in dialog
https://discussions.apple.com/thread/4023861?start=0&tstart=0
https://bugzilla.mozilla.org/show_bug.cgi?id=703573

shane-tomlinson:
Sidestep the scrolling issue entirely. Only show the TOS/PP agreement in an embedded iframe IFF the browsing enviornment does not support window.open

fixes #3442


lloyd:
r+

Issue #2160, opened 2012-07-26
Support for IPv6


fmarier:
We don't appear to have a bug to track IPv6 support in Persona. We currently don't expose AAAA records on any of the Persona services.

I believe that this is a big deal in the developing world where IPv4 addresses are in short supply (or no longer available).

fmarier:
The new regexp that was introduced to fix #1981 will need to be extended to support IPv6 numerical addresses.

warner:
Other places that will need fixes (accepting ":", "[", and "]" in numeric hostnames):
- frontend validation of postMessage parameters: resources/status/common/js/user.js in the "setHostname" function (added in 780267-fixup-origin)
- the other frontend-validator in resources/static/dialog/js/modules/dialog.js (extractValidHostname)
- new tests to exercise this in resources/static/test/cases/dialog/js/modules/dialog.js (the tests only currently exercise names in URLs, not v4 or v6 addresses)

fmarier:
As far as ops go, the first step there is to build a case (e.g. stats and numbers on Internet users in various countries that don't have IPv4 access) and then send it to Mark Mayo.

fmarier:
http://www.ipv6.org.nz/2012/10/17/arin-scrapes-bottom-of-ipv4-barrel/

fmarier:
This issue is blocked until Amazon adds IPv6 to ELB.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

Issue #1981, opened 2012-07-11
improve validation of origin params to API calls


lloyd:
as @fmarier pointed out in issue #1979, we can do much better with our regex in validate.js - more tightly validating incoming origins.

Not extremely high priority, but why not?

for reference, http://blog.gnukai.com/2010/06/fqdn-regular-expression/

lloyd:
@zaach - you want?

zaach:
That's a wicked regex. I'm on it.

fmarier:
One thing that I'll be doing soon is refactoring the code that extract the domain from an email address (GH-1994). Once I do that, I'll validate the extracted domain, just in case.

So I would suggest that we create a domain/hostname validation function that the origin type checking code could use to check the hostname part of the origin. Or perhaps we should create a new 'hostname' type instead?

PR #3455, opened 2013-05-27
Add support of IPv6 in the hostname validation

quack1:
I added an IPv6 regular expression in the regular expression used to check the hostname in /lib/validate.js.

Currently, the regex only support IPv4 and "litteral" hostnames.

This PR could be part of the issue #2160 (derivated from #1981) submited by @fmarier.


shane-tomlinson:
@fmarier or @callahad - Can one of you review this? Are we ready for IPv6 hostnames?

shane-tomlinson:
Thanks @quack1 - We are not quite ready for IPv6 at this time.

Issue #3357, opened 2013-05-03
transitionToPrimary shows normal "<idp> makes it easy!" screen from sign in.


shane-tomlinson:
It doesn't show the awesome upgrade message.

shane-tomlinson:
This is from the sign in screen, I have not tried from the email picker.

shane-tomlinson:
Looking at the changes to modules/authentication.js, this does not appear to be a regression from the recent B2G merge.

PR #3458, opened 2013-05-29
Issue 3184 consolidate verify with primary

shane-tomlinson:
@ozten or @seanmonstar, could you guys review this one?

I worked off of the PR that @arlolra submitted to consolidate the "upgrade_primary_user" and "verify_primary_user" modules. These two modules were very similar to each other, so we collapsed them into one.

The selenium tests are still happy, so I am fairly confident that I did not change functionality. There is one user facing change, this is a fix for #3357. A user now sees the "transitionToPrimary" messaging if the user is not signed into Persona but their account is being upgraded.

I removed the "upgraded_primary_user"->"primary_user"->"verify_primary_user" sequence if the user is not signed in to their primary. Instead, I redirect directly to the primary, which picks up as it used to.

An ephemeral instance, "consolidation.personatest.org" is already set up.

And a final thanks goes to @arlolra for submitting the initial PR and getting this ball rolling. His changes made my work much easier.


automation-tests/pages/css.js
```diff
@@ -52,7 +52,7 @@ module.exports = {
     // the button you click on after typing and re-typing your password
     createUserButton: 'button#verify_user',
     returningUserButton: 'button.isReturning',
-    verifyWithPrimaryButton: 'button#verifyWithPrimary',
+    verifyWithPrimaryButton: 'button.isDesktop.verifyWithPrimary',
```
ozten, line 55: Nice, I like classes instead of ids.

resources/static/dialog/js/modules/verify_primary_user.js
```diff
@@ -49,47 +53,61 @@ BrowserID.Modules.VerifyPrimaryUser = (function() {
...
+
+      user.addressInfo(email, function(info) {
+        auth_url = info.auth;
+
+        if ("transition_to_primary" === info.state) {
+          options.transition_to_primary = true;
         }
-      }, self.getErrorDialog(errors.addressInfo));
 
+        if (showsPrimaryTransition(info.state)) {
```
ozten, line 101: knit pick: This could be an `else if` to make it more clear that these are mutally exclusive states.

shane-tomlinson, line 101: I dig, I was was hesitant to do it this way. Will update.

ozten:
Awesome work. Will merge once you update.

ozten:
Nice work Arlo, thanks for landing Shane!

resources/static/dialog/js/modules/upgrade_to_primary_user.js
```diff
@@ -1,72 +0,0 @@
...
-    ,
-    cancel: cancel
-    // END TESTING API
-  });
-
-  sc = Module.sc;
-
-  return Module;
-}());
-
```
seanmonstar, line 72: Glorious!

PR #3460, opened 2013-05-29
Fix new_user KPI for two scenarios:

### LINKED ISSUE #3359: KPI: cancel after mistyped email shouldn't count as new_account
- open a browser with empty local storage
- sign in to 123done.org (or wherever)
- enter an email address (non-bigtent) that doesn't exist on the fallback yet (mistyped or wrong email, from the user's perspective)
- hit cancel when it prompts you to create a password
- choose an email address that has been entered before properly
- enter the correct password and sign in

The blob that is generated has new_account = true, which it probably shouldn't.

The kpi-dashboard was confused by this case, as it hit the set_password screen and assertion_generated but not the email verification middle steps. The kpi-dashboard is not currently using new_account, fwiw, so this change isn't blocking anything. (I modified the data to treat any event stream with enter_password as not new).

### LINKED ISSUE #3459: [b2g] new_account KPI not reported for new FirefoxOS users

shane-tomlinson:
@kparlante - mind reviewing?
- FirefoxOS was not reporting new users
- If a user hits cancel from the "set password" screen.

fixes #3359
fixes #3459

### DIFF
--- a/resources/static/dialog/js/misc/state.js
+++ b/resources/static/dialog/js/misc/state.js
@@ -210,6 +210,10 @@ BrowserID.State = (function() {
     handleState("new_user", function(msg, info) {
       self.newUserEmail = info.email;
 
+      // Add new_account to the KPIs *before* the staging occurs allows us to
+      // know when we are losing users due to the email verification.
+      mediator.publish("kpi_data", { new_account: true });
+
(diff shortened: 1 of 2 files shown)


kparlante:
@shane-tomlinson I tested this out, works for me. Code looks right. r+

PR #3461, opened 2013-05-29
Fix primaries on FirefoxOS!

shane-tomlinson:
@jedp - this is a PR to fix the "relay frame cannot be found" error on FirefoxOS
- Save a marker in sessionStorage that indicates the return from a primary is in a native context. If so, do not attempt to set up the channel.
- Add the ability to redirect to a site from the dev menu. This is useful on FirefoxOS where the trusted UI has no URL bar.

### DIFF
--- a/resources/static/common/js/modules/development.js
+++ b/resources/static/common/js/modules/development.js
@@ -29,6 +29,7 @@ BrowserID.Modules.Development = (function() {
         this.click("#clearLocalStorage", clearLocalStorage);
         this.click("#clearEmailsForSites", clearEmailsForSites);
         this.click("#forceIsThisYourComputer", forceIsThisYourComputer);
+        this.click("#redirectTo", redirectTo);
         this.click("#closeDevelopment", close);
       }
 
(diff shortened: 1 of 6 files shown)


resources/static/common/js/modules/development.js
```diff
@@ -96,6 +97,15 @@ BrowserID.Modules.Development = (function() {
     storage.usersComputer.forceAsk(user.userid());
   }
 
+  function redirectTo() {
+    var href = dom.getInner("#siteToRedirectTo");
+
+    if (href) {
+      bid.module.stopAll();
+      document.location = href;
```
jedp, line 105: Wow, I can't believe b2g lets this happen.

In the logcat, gecko throws a `CheckSameOriginError`:

```
E/GeckoConsole( 2718): Security Error: Content at https://eyedee.me/ may not load data from https://all-together-now.personatest.org/sign_in#AUTH_RETURN.
```

but it goes through with it anyway!

As Ian has pointed out, this redirect bug may be involved: https://bugzilla.mozilla.org/show_bug.cgi?id=456957.

Nevertheless, you appear to have found a workaround that will at least get us through v1.0.1, and possibly v1.1!

jedp:
Let's doooo eeeeet.  We need something that works _now_, and this is that something.  Nice work, Shane!

jedp:
Note: not deleting the branch, so we can return to this if we choose.

PR #3474, opened 2013-05-30
Issue 3472 script to auto install node modules

quack1:
This small script is aimed to clone the browserid repository and install node 0.8.12 and all its necessary modules. 
Node 0.8.12 is installed through `nvm` which allows to install and use different versions of Node on the same system.

The script currently runs on Ubuntu 13.04 and Debian 6.0.x

Usage : 
- `./setup_node.sh install <directory_to_clone_browserid> <directory_to_install_nvm>


quack1:
Can someone review it ? (maybe @shane-tomlinson or @callahad )

callahad:
I need to set up a fresh VM for reviewing this, but I'll get to it later today.

quack1:
I just updated the script to install the last 0.10.x version available in the `nvm` repository (0.10.9).

jaredhirsch:
R- from me.

I'm not comfortable advocating people install nvm, I don't think it's worthwhile to commit to testing and maintaining multiple major node version compatibility. Given that we're now 0.10.x compatible (#3482) this patch doesn't seem necessary.

I also see an ipv6-related patch and two meaningless merge messages in the history, but those are beside the point.

jaredhirsch:
@quack1 appreciate you persistently trying to get involved :-) but in this particular case, a blog post or adding to the deb-flavored installation docs would be a better way to contribute. The docs are here https://github.com/mozilla/browserid/blob/dev/docs/SETUP_UBUNTU.md

quack1:
@6a68 I agree with you, I wrote this script before Persona was compatible with Node 0.10.x. I also thought about adding this to docs/SETUP_UBUNTU.md, but after a discussion on IRC, we came to the conclusion that a script could be better.

BTW, I actually also wrote a blog post, but it seems that I forgot to insert a link here :/ So, here is the post : http://quack1.me/persona_setup_ubuntu-en.html

And yes, I did some others commits related to IPv6, that I reverted "manually" to just add this script in the commit".

jaredhirsch:
@quack1 awesome! @callahad / @francois, do we collect developer-focused blog posts anywhere? I know we collect a list of libraries.

Issue #3471, opened 2013-05-30
[kpi] Change config/aws.json kpi to point at dev instead of staging


kparlante:
kpiggybank-staging has security group settings that mimic production settings, so most ephemeral servers are blocked from POSTing to it. https://kpiggybank-dev.personatest.org/ is what most ephemeral instances should point at. I'll make up a PR tomorrow when I'm not sleepy...

PR #3477, opened 2013-05-30
default ephemeral aws should use kpiggybank-dev

kparlante:
@lloyd: Fixes https://github.com/mozilla/browserid/issues/3471


lloyd:
r+

PR #3479, opened 2013-05-31
show the client time in the error.ejs dialog

### LINKED ISSUE #3478: show the client time in the error.ejs screen
In looking at another issue, it occurred to me that a useful bit of information in the 'See more info' part of the error screen would be the ISO8601 time for those cases where a user reports the text in that window back to us. It would help in correlating to backend logs.

Of course, the client clock may be **way off**,  so take that value with a grain of salt.

The dialog looks like this now:

show-toisostring-in-error-ejs

@shane-tomlinson: if you think this should have tests, pointers welcome. I looked around the qunit tests for the right place to put it, but wasn't sure where to put it.

jrgm:
Fixes #3478

### DIFF
--- a/resources/static/dialog/views/error.ejs
+++ b/resources/static/dialog/views/error.ejs
@@ -41,6 +41,8 @@
       <% if (typeof action !== "undefined") { %>
         <li>
           <strong id="action">Action: </strong><span id="errorTitle"><%= action.title %></span>
+          <br>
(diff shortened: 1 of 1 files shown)


jrgm:
Okay, typical "well that's just use this" from someone who doesn't live in the caniuse world. 

Closing this PR for now, since toISOString is a "new thing".

Issue #3465, opened 2013-05-29
errors shown when running backend tests (shutdown race?)


jrgm:
Thought I filed this a while ago, but I guess not.

I was running `WHAT_TESTS=back_mysql npm test` with the `dev` branch, locally on osx and noticed there were occasional chirps in stdout/stderr about `failed to clear authentication attempts`. The tests passed though.

Examples:

```
tests/primary-secondary-transition-test.js
  -> browserid (83630): error: failed to clear authentication attempts for 1: Error: socket hang up
tests/verify-in-different-browser-test.js
  -> browserid (84447): error: failed to clear authentication attempts for 1: Error: socket hang up
tests/registration-status-wsapi-test.js
  -> browserid (92998): error: failed to clear authentication attempts for 2: Error: connect ECONNREFUSED
```

I think this a race with shutdown in the tests, due to changes for PR-3337.

In https://github.com/mozilla/browserid/blob/dev/lib/wsapi/authenticate_user.js#L81, 
we try to do further work after telling the client that 'everything is
done'. 

If that test has nothing further to do (haven't traced all the way through each test), the test will begin shutdown of the test servers, and the race is on.

Anyways, this probably isn't a production issue (except maybe in a corner case of flapping servers). 

But perhaps for the local test, a little delay trick on mutual shutdown of local servers. (Might also help with other things).

shane-tomlinson:
The fix has been merged, closing.

PR #3480, opened 2013-05-31
slightly delay shutdown of run_locally.js daemons to avoid shutdown races

jrgm:
Fixes https://github.com/mozilla/browserid/issues/3465

### DIFF
--- a/tests/lib/start-stop.js
+++ b/tests/lib/start-stop.js
@@ -12,6 +12,10 @@ events = require('events'),
 config = require('../../lib/configuration.js'),
 db = require('../../lib/db.js');
 
+// at test completion, allow a moment for in-flight backend requests to finish
+// before shutting down the backend daemons. GH-3465.
+const DELAY_KILL_SIGINT_MS = 50;
+
 var proc = undefined;
 
 process.on('exit', function () {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@jrgm, do you know what the root cause of this is?

shane-tomlinson:
@jrgm - doh, you outlined the root cause in the original issue. This works for me locally, 

I am wondering if there is a way to guarantee that /wsapi/authenticate_user has completely finished before shutting down. For environments like Travis CI where no performance is very hit and miss, will 50ms always be enough? Can we set a bit on res from resetFailedAuthCount that is checked before attempting to shut down?

@lloyd, any input here or is this approach enough?

jrgm:
So the problem isn't that the webhead /wsapi/authenticate_user shuts down before it is finished. It's sort of the opposite. It's that authenticate_user responds to the test with OK, the test then tells the dbwriter and webhead (and router, etc.) to shutdown; if the dbwriter shuts down before the webhead, then the webhead's attempt to POST to 
/wsapi/reset_failed_auth_tries on the dbwriter gets a connection reset.

So there'd have to be some central coordination of "Is everybody ready to shutdown", but that doesn't seem to me like a direction to follow. Alternatively, I considered introduce a delay in each daemon (or actually in the shutdown.js that they all use to do cleanup at shutdown), but that also didn't seem to be the right thing. 

Is 50ms long enough? I can't be certain, but seems to work in my travis run. I really think the race is just a couple of milliseconds, but I could up the timeout to 100ms if you think I should.

jaredhirsch:
@jrgm in the interest of gettin' shit done, how about I merge this in, and if the problem persists, we up the delay some more?

jrgm:
@6a68 Sold!

PR #3486, opened 2013-06-03
Re-allow http TOS/PP agreements.

### LINKED ISSUE #3469: about changes for non-ssl RP and options {privacyPolicy,termsOfService}
So I see the commit for this change, https://github.com/mozilla/browserid/commit/5c81116a229d0776dfc6ddf2af8b5ecc5305048b, and I know this was done for b2g reasons.

At any rate, late as usual, but wondering about existing RP on desktop, where we are disabling them from using these features if they don't use https.

This may be a necessary regression, but at least the documentation should be updated if this is not a permanent change in the api:

```
https://developer.mozilla.org/en-US/docs/Web/API/navigator.id.request
https://developer.mozilla.org/en-US/docs/Web/API/navigator.id.get
```

Related: clicking on the tos and pp links in the dialog with desktop login.persona.org will now go to /embedded_tos and /embedded_privacy instead of the previous login.persona.org/{tos,privacy} url.

shane-tomlinson:
@jedp, @jrgm, this is the same fix as #3485, but against dev.

fixes #3469

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -324,12 +324,6 @@
       checkRenamed(options, "tosURL", "termsOfService");
       checkRenamed(options, "privacyURL", "privacyPolicy");
 
-      if (document.location.protocol !== "https:") {
-        warn("privacyPolicy and termsOfService are only allowed on https domains");
-        delete options.termsOfService;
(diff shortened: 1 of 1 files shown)


jedp:
As discussed with everyone.  Looks good!

PR #3487, opened 2013-06-03
Revert password failure lockout feature to avoid alter table.

jaredhirsch:
Long-running (10 hrs) alter table has led us to reconsider implementing a bad password count in mysql. Just for the moment, revert the relevant commits.


lloyd:
we want to revert before we try what resolved the long running alters last time?  https://id.etherpad.mozilla.org/alter-2012-08-17-wtaf

jaredhirsch:
@lloyd: just getting this branch ready while @jrgm @gene1wood and @Sheeri play with the DB. Parallelization ftw.

jrgm:
Priming the pump with mysqldump appears to bring this into an ~acceptable minutes-long window while the mysqldump and then alter table are run where signup, change password, add user, etc. write operations will lead to errors. 

Note: the mysqldump should be run on each downstream slave, and while it runs replication is delayed (leading to failures where we write the change and then try to read from a downstream slave). The alter cost will happen on each slave as the query moves down the replication tree. So if the alter takes N minutes and the maximum number of hops from the furthest slave to the master is M, then the window is N x M minutes. I do not know what our tree looks like these days; if it's not fairly flat, we can reduce impact by moving all read and write traffic into phx1 where the database master lives.

jrgm:
Also, two other things:
1. I'd like to get confirmation in production that this reproduces, say by taking query traffic off of a leaf slave node, do the mysqldump && alter table, and then drop the new column, and put traffic back in.
2. The row count in stage is ~7 times bigger than production tables. I don't know what the big-O of a mysql alter table works out to be, but we _may_ see quicker times in production.

gene1wood:
@jrgm The furthest leafs are 2 jumps away from the master. Example :
db1.iddb.phx1 <- db1.us-west-2 <- db2.us-west-2

jaredhirsch:
looks like it's safe to close this? @jrgm?

jrgm:
@6a68 yes. We sacrifice a chicken and then ...

lloyd:
sacrifice a larger animal, do the auto commit too...

PR #3491, opened 2013-06-04
Append FirefoxOS navigator.mozId properties on to the client created navigator.id object.

shane-tomlinson:
@jedp - one more for you! This can be tested on "kpi-this-session.personatest.org" and with eyedee.me, which now includes both include.js and authentication_api.js

This allows the shimmed primary APIs to work on FirefoxOS.

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -115,9 +115,15 @@
     // Is there a native implementation on this platform?
     // If so, hook navigator.id onto it.
     if (navigator.mozId) {
-      navigator.id = navigator.mozId;
-    } else {
-      navigator.id = {};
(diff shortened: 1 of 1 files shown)


resources/static/include_js/_include.js
```diff
@@ -115,9 +115,15 @@
     // Is there a native implementation on this platform?
     // If so, hook navigator.id onto it.
     if (navigator.mozId) {
-      navigator.id = navigator.mozId;
-    } else {
-      navigator.id = {};
+      for (var key in navigator.mozId) {
```
jedp, line 118: Is this safe?

shane-tomlinson:
Closing as "it doesn't work"

Issue #3306, opened 2013-04-25
KPI data is not sent for users who only open the Persona dialog once


jaredhirsch:
Currently, we don't fire KPI data at the time it's collected, we fire it on the subsequent trip through the dialog.

Since the dialog's localstorage is accessible to the communication_iframe, we could just have the iframe fire kpi data when it receives the dialog_complete event.

In particular, we could figure out how many users are bailing out of the dialog flow by saving some state in localstorage each time a dialog screen is displayed.

We could either just record the last screen viewed, or we could record the sequence of screens they viewed.

In either case, even if a user never uses persona again, we can figure out where in the flow they are bailing out.

seanmonstar:
It seems easiest and data-full to just allow the event stream to fill up as normal, and send the full interaction_data. That will show what screen they got to.

Also, we filed the same issue at the same time. #3305

lloyd:
Generally, i think we should try in multiple places.  Dialog and now also iframe.  Was that the intention?

-- lloyd (thumb-typing)

seanmonstar:
@lloyd I said as much in #3305, which I filed at the same time as Jared filed this one.

r0bl0rd:
Hey all - this is Rob Lord, incoming Mozillian. (More on that later.)

+1 for metrics at each stage of the funnel so we know our visible -> qualified -> best bet -> closed user counts and conversions. 

Rob
via iPhone

shane-tomlinson:
@6a68, @seanmonstar - We talked about doing it this way once and decided against it. Sending data from the iframe will not work for RPs that are using the .get API. So that we did not have to have some sort of lock mechanism to check who should send the data, we decided to do it only in the dialog. If we are to convert to sending data from the iframe, we have to make sure all of the flows are properly covered.

Changing the name of this issue from a proposed solution to a description of the problem - "KPI data is not sent for users who only open the Persona dialog once."

lloyd:
@seanmonstar word.

lloyd:
@shane-tomlinson good point about the locking mechanism.  I'm curious though if we can solve this with a creative algorithm that simply and elegantly addresses the race condition.
1. take the blob out of local storage
2. try the XHR
3. if it fails return the blob to local storage.

seanmonstar:
@lloyd :+1:

You'd have to very fast to open the dialog again before the communication frame finishes sending the small JSON blob back home.

lloyd:
[...]

PR #3495, opened 2013-06-05
Submit KPI data when the dialog closes either in the interaction_iframe OR in the internal_api.

shane-tomlinson:
@kparlante - this one is for you!

Only generate GUIDs for a KPI blob whenever needed:
- When the dialog is orphaned AND
- A staging event occurs without a confirmation event.

### DIFF
--- a/resources/static/common/js/models/interaction_data.js
+++ b/resources/static/common/js/models/interaction_data.js
@@ -10,6 +10,8 @@ BrowserID.Models.InteractionData = (function() {
       network = bid.Network,
       complete = bid.Helpers.complete,
       whitelistFilter = bid.Helpers.whitelistFilter,
+      SENT_ORPHAN_WITH_STAGE = "sent_orphan_with_stage",
+      STAGING_CONTINUATION = "staging_continuation",
       KPI_WHITELIST = [
         'event_stream',
         'lang',
@@ -20,7 +22,6 @@ BrowserID.Models.InteractionData = (function() {
         'number_sites_signed_in',
         'number_sites_remembered',
         'orphaned',
-        'linking_id',
         'new_account',
         'email_type',
         'rp_api'
@@ -50,14 +51,18 @@ BrowserID.Models.InteractionData = (function() {
 
     var interactionData = getInteractionData();
 
-    // If there is a linking_id, the user staged an email in the last dialog
-    // session but never verified the email. Associate this session and the
-    // previous session to see if the user completed verification. The back end
-    // will take care of any linking.
-    var guid = localStorage.getItem("linking_id");
-    if (guid) {
-      newData.linking_id = guid;
-      localStorage.removeItem("linking_id");
+    // If there is a sent_orphan_with_stage, the user staged an email in the
+    // last dialog session but never verified the email. Let the back end know
+    // so they can do rough calculations on how many of stagings were
+    // successfully completed.
+    if (localStorage.getItem(SENT_ORPHAN_WITH_STAGE)) {
+      localStorage.removeItem(SENT_ORPHAN_WITH_STAGE);
+
+      var events = newData.event_stream || [];
+      if (! hasEvent(events, STAGING_CONTINUATION)) {
+        events.unshift([STAGING_CONTINUATION, 0]);
+        newData.event_stream = events;
+      }
     }
 
     interactionData.current = newData;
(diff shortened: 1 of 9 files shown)


shane-tomlinson:
fixes #3306

resources/static/common/js/models/interaction_data.js
```diff
@@ -80,6 +80,26 @@ BrowserID.Models.InteractionData = (function() {
     }
   }
 
+  function publishCurrent(oncomplete) {
+    stageCurrent();
+    publishStaged(oncomplete);
+  }
+
+  function indexOfEvent(eventStream, eventName) {
```
shane-tomlinson, line 142: These two functions are no longer needed with the removal of the linking_id

kparlante:
@shane-tomlinson : I played with this one and did indeed see the blobs on the first go round (yay!). I tried with both myfavoritebeer and 123done, covering both get and watch apis.

While I don't I need a matching guid to link blobs when the user closes the dialog to wait for email verification, I do need some sort of flag for that next blob that succeeds, so I can tally a count for eventual new user successes.

FWIW, we can get rid of the orphaned flag if its only meaning is the existence of an assertion generated event.

shane-tomlinson:
@kparlante - What is the metric needs to be captured, % of signups that eventually complete? Something else? 

I'm shooting in the dark here, but I am trying to figure out how to handle the following scenarios:
1. User opens dialog, stages nothing, closes dialog.
2. User stages an address, closes the dialog, never does anything
3. User stages an address, verifies in a new tab while dialog is still open.
4. User stages an address, closes the dialog, and then verifies in same browser.
5. User stages an address, verifies in a different browser while the dialog is still open.
6. User stages an address, closes the dialog, verifies in a different browser.

#### Walking through cases

##### 1. User opens dialog, stages nothing, closes dialog.

KPI data sent when user closes dialog.

##### 2. User stages an address, closes the dialog, never does anything

KPI data sent when user closes dialog.

##### 3. User stages an address, verifies in a new tab while dialog is still open.

KPI data sent when dialog closes post-verification. KPI blob's event stream indicates that verification has completed and an assertion has been generated.

##### 4. User stages an address, closes the dialog, and then verifies in same browser.

KPI data is sent when user closes dialog, pre-verification. KPI blob's event stream indicates an address has been staged, but not that one has completed.

##### 5. User stages an address, verifies in a different browser while the dialog is still open.

KPI data is sent when dialog closes post-verification, like case 3. No KPI blob sent from second browser.

##### 6. User stages an address, closes the dialog, verifies in a different browser.

KPI data is sent when user closes dialog, pre-verification. No KPI blob sent from second browser.

Cases 3-6 are successful and need to indicate "success!"

In cases 3 and 5, we could use the "assertion_generated" event to indicate success, but for cases 4 & 6, since the user has already closed the dialog and they are not guaranteed to open the dialog a second time, we can't rely on using assertion_generated the next time the dialog is open to give us this data.

We currently do not log any events for wsapi calls on the main site, meaning no event is ever added to the event stream when an address is verified. A verification event seems like success criteria. 

If so, should we start logging events on email verification pages to capture successes?

kparlante:
@shane-tomlinson : We're trying to capture how many stages each "new user" makes it through when they come to the site. So, ideally, for a given day + segment, we want # "new" users @ screen.set_password, # "new" users @ user.user_staged, # "new" users @ user.user_confirmed, # "new" users @ assertion_generated. (We should probably add number of users who don't make it to screen.set_password as well, although we don't know for sure that they are "new" users).

I should add that its not important to be perfect, we just need the data to be good enough that we trust we have a pretty good picture of what's going on.

We can handle/count cases 1, 2, and 3 easily now, just by looking at the stages (especially with this fix where we actually get all of the blobs). We also easily handle case 5, assuming we see the assertion_generated in the event stream.

We were using "orphaned" to handle case 4, because the original blob would get modified to orphaned = false. I could then add counts for the last two stages if I saw orphaned = false + only the first two stages. This is the case that gets lost with the new change. I was thinking we could add a flag to the second blob that gets generated, something that means "this blob completes some previous blob, even if we're not telling you _which_ previous blob". I can then add counts to the last two stages for that segment/day. Basically a flag in place of a guid.

We're just losing case 6. I suppose we could handle it if they _ever_ come back to that original browser, I assume that ends up looking like case 4.

@lloyd and I talked about your idea of logging email verifications on the main site yesterday -- yes, I think this would be helpful to cross check the data we see in the KPIs. We could estimate case 6 for example.

shane-tomlinson:
@kparlante thanks for the explanation. I imagine any new flag would be cause for a data privacy review, so I am wondering if we can do this using the event stream. Instead of adding a new flag could we cover # 4 and 6 by logging verifications? I'm not exactly sure how it would look/work yet, I am trying to think up alternatives.

kparlante:
@shane-tomlinson Well, the first item on the event stream could be "removed-previous-orphaned-blob" or similar :-p .
Logging verifications will similarly require review (I suspect we'll want to do a review at the beginning of next quarter for a handful of things we want to do).

I think its useful to go ahead with this change even if we don't cover cases #4 and #6, so I do +r the change. If we go ahead without addressing it we should just log a new issue.

Another argument for not needing a review... isn't it logically the same as de-orphaning a blob?

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -178,6 +177,18 @@
     }, complete.curry(null));
   }
 
+  function setOrigin(origin) {
```
shane-tomlinson, line 180: This code is just moved from above for better proximity.

kparlante:
@shane-tomlinson : excellent! I tested this out for cases 4 and 6 (or at least the variant of 6 where the user eventually comes back to the original browser), works nicely. +r

Issue #3374, opened 2013-05-08
Errors thrown inside `navigator.id.watch` callbacks get swallowed


mattbasta:
If an error is thrown inside the `onlogin` or `onlogout` callbacks (and presumably `navigator.id.request`'s `oncancel`), it is not reported in the error console. This means that login seems to mysteriously stop working in these cases for no apparent reason. Tracing the code in Firefox's dev tools and Firebug produce no useful debugging information since they don't report exceptions as they happen.

seanmonstar:
Indeed, this seems true of `onlogin`: https://github.com/mozilla/browserid/blob/dev/resources/static/include_js/_include.js#L405

It seems like we didn't want errors in user provided scripts to stop our code from finishing up the event. Which makes sense... but it also sucks that you can't see the error.

Also worth noting that this behavior only happens when `onlogin` is called from the return value of the dialog. `onlogout`, `oncancel`, and `onmatch` aren't surrounded by try/catch, and the automatic login that would happen after redirecting once you completed email verification also isn't wrapped up.

mattbasta:
For what it's worth, you shouldn't need to wrap your code in a try/catch block anyway. If there's an internal problem with Persona, wouldn't you want that exposed so you can correct the issue? You'll get exactly zero feedback about broken internals if you swallow your own errors.

seanmonstar:
Ah, we don't swallow our own errors. That catch block is to protected the
internal code from errors written by the user in the `onlogin` callback. We
didn't want their errors stopping all our internals, methinks.

mattbasta:
Perhaps you can catch the error, assign it to a variable, and throw it once you're out of the critical code path?

deepak1556:
you can use domain feature of nodejs to solve the problem. put the code in the critical path to emit an event if error occurs and handle them in the listener.

seanmonstar:
@deepak1556 this is code that executes in the browser, when you call `navigator.id.watch()`

deepak1556:
a possible solution can be to them in a try-catch-final block without affecting code path and at same time throw the error to users.. 

``` javascript
  try {
   if (observers.login) observers.login(r.assertion);

 } catch(e if options.error = e ) {

            // client's observer threw an exception

 } finally { 

    console.log(options.error);
}
```

jedp:
@lloyd suggesting we catch it, console.log it, and re-throw it.  Sounds like a great feature.

lloyd:
[...]

PR #3504, opened 2013-06-09
Catch, log, and throw client errors

jedp:
As requested in issue #3374

When there's an error in client login code, log and then throw the error as a help for developers trying to integrate persona.

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -398,8 +398,11 @@
         if (!err && r && r.assertion) {
           try {
             if (observers.login) observers.login(r.assertion);
-          } catch(e) {
+          } catch(clientError) {
             // client's observer threw an exception
+            // help developers debug by logging the error
+            console.log(clientError);
+            throw clientError;
           }
         }
 


resources/static/include_js/_include.js
```diff
@@ -398,8 +398,11 @@
         if (!err && r && r.assertion) {
           try {
             if (observers.login) observers.login(r.assertion);
-          } catch(e) {
+          } catch(clientError) {
             // client's observer threw an exception
+            // help developers debug by logging the error
+            console.log(clientError);
+            throw clientError;
```
ozten, line 405: Awesome.

resources/static/include_js/_include.js
```diff
@@ -398,8 +398,11 @@
         if (!err && r && r.assertion) {
           try {
             if (observers.login) observers.login(r.assertion);
-          } catch(e) {
+          } catch(clientError) {
```
ozten, line 401: @shane-tomlinson or @6a68 - Any ideas on the best way to add a unit test?
- qunit - Doesn't currently call `navigator.id.request` and seems hairy to add there
- selenium - One would have to build a whole test site which had bad options to `request`

Merging w/o a test, as the patch is so straightforward.

shane-tomlinson, line 401: @ozten - we have no testing of include.js unfortunately. We have/had a bug open about it, but I cannot find it right now.

PR #3509, opened 2013-06-11
ToS: clarify that use in education and government is generally ok

fmarier:
This update to the terms of service comes from Jishnu Menon.

### DIFF
--- a/resources/views/tos.ejs
+++ b/resources/views/tos.ejs
@@ -62,7 +62,7 @@
   <ul>
     <li>Mozilla has the right to manage the Services to protect the rights and property of Mozilla and others and to facilitate the proper functioning of the Services, including disabling your account.</li>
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@jedp, @ozten - what repercussions will this have for the effort the translate the TOS/PP for FirefoxOS?

Otherwise, I am r+, this is a great change, thanks for pushing @fmarier

fmarier:
@shane-tomlinson Jishnu said that he should have the Polish and Spanish translations for us soon (with this change included).

shane-tomlinson:
Awesome. Merging.

callahad:
@fmarier Is there a bugzilla link or something from Jishnu that we can reference?

ozten:
Unknown, let's not block on that. Go for it.

ozten:
@callahad - The TOS/PP is a private legal bug. I've CC'd you.

Issue #3496, opened 2013-06-06
[b2g] tooltips can run off the screen.


zacc:
When using trusty ui in the modal dialog mode this error message is obscured.

I also need ot know what it says - something about logging in more than once?

![2013-06-06-06-22-50](https://f.cloud.github.com/assets/966053/618139/a7c0115e-ceac-11e2-9bbf-2bc3c7349bee.png)

jedp:
@zacc thanks for the report, and yes, this is horrible.  

We're already on this issue.  Please follow this bug: https://github.com/mozilla/browserid/issues/3154

Closing as duplicate.

zacc:
What does the message actually say though?

I was told to use an unverified user for this test but I am hitting this error message and I need to set up the device to handle the unverified user. Using a verified user in the meantime.

shane-tomlinson:
@zacc - an email is still sent to unverified users so that they can verify their ownership of the email address. This message is saying "hey, we just sent a message to that address!"

shane-tomlinson:
Re-opening to fix this before fully addressing #3154

shane-tomlinson:
This affects the current beta. @jrgm - is this worthy of a hotfix?

shane-tomlinson:
@jedp - the same comment goes to you - is this worthy of a hotfix?

shane-tomlinson:
@jedp and @jrgm - this bug should be limited to the "We just sent an email to that address! If you really want to send another, please wait a minute or two and try again." tooltip when Persona is accessed from Marketplace. When Persona is accessed from Marketplace, the RP info at the top of the screen is not displayed, which pushes the rest of the screen up. The code to anchor the tooltips did not check to ensure the tooltip boundaries were still on the screen.

The fix to this is an additional 10 lines added to tooltip.js to check the top and right boundaries of the tooltip. With tests, the changes come to 30 lines.

Issue #3154, opened 2013-03-28
convert the tooltips to inline indicators and highlights


skinny97214:
Kill them. Die. Die. Die.

We need to move to inline errors anyway for B2G. Let's make one visual style that works everywhere.

jaredhirsch:
Is there a wireframe or mock someplace that documents inline notification
styles? Are they already in the product and I'm just forgetting them?

skinny97214:
@6a68 They are coming from the b2g work to align our visual design more with Gaia.

![Screen Shot 2013-03-29 at 2 32 03 PM](https://f.cloud.github.com/assets/41135/319261/235ae242-98b8-11e2-8abd-8964c1399d3b.png)

callahad:
Assigning to @skinny97214 to clarify. Other comments and concerns coming up.

shane-tomlinson:
@skinny97214 - Can you elaborate more? My initial fears are that this is going to require a substantial amount of both work and testing to ensure our layout still functions correctly in letterific locales like German and Swedish. I want to find out more about the requirements before I go running for the hills.

skinny97214:
@shane-tomlinson What's your feeling on how soon we might be able to get to this?

jedp:
Update: @shane-tomlinson, @skinny97214, and @jedp meeting 6/10 to plan and schedule the work.

shane-tomlinson:
Link this to the corresponding bugzilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=879626

csuciu:
"Sorry, PIN doesn't match" is still displayed as tooltip

PR #3512, opened 2013-06-11
Constrain tooltips to be displayed on the screen.

shane-tomlinson:
@jedp and @jrgm - this is a hotfix to the current stage for #3496. An ephemeral instance exists to test, tooltips-off-screen-stage.personatest.org

### DIFF
--- a/resources/static/common/js/tooltip.js
+++ b/resources/static/common/js/tooltip.js
@@ -8,7 +8,8 @@ BrowserID.Tooltip = (function() {
   var ANIMATION_TIME = 250,
       TOOLTIP_MIN_DISPLAY = 2000,
       TOOLTIP_OFFSET_TOP_PX = 5,
-      TOOLTIP_OFFSET_LEFT_PX = 10,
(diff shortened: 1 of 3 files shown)


jaredhirsch:
@shane-tomlinson based on the screenshot in #3496, I wonder if tooltips will now obscure the fields they are supposed to point at?

shane-tomlinson:
@6a68 - one tooltip will, but having the tooltip fully readable is better than off the screen. For more context, read https://github.com/mozilla/browserid/issues/3496#issuecomment-19258637. I am OK with the tooltip obscuring the field for this single case, especially since tooltips should be going away soon now that initial work has started for #3154.

jaredhirsch:
@shane-tomlinson sounds good to me :-)

jedp:
Looks good to me, too.  Having the tooltip obscure that field seems fine.  It's still very usable (much more so than before), and the tooltips are a short-term fix anyway.

@jrgm I've merged the corresponding pr into dev

PR #3513, opened 2013-06-11
Prevent FirefoxOS showing a tooltip if the user entered a password but not a vpassword in the set password screen.

### LINKED ISSUE #3502: New account creation: Hitting 'enter' on first password field should focus the next
Currently, if you create a new account on b2g, when you are prompted to create a password, hitting 'enter' tries to submit the form, which fails because you haven't yet entered the password as second time for confirmation.

What should happen is that hitting 'enter' in the first field should focus on the second field, and then hitting 'enter' in the second field should try to submit.

shane-tomlinson:
@jedp - Can you test this?

ephemeral instance: enter-in-password.personatest.org

Does this deserve a hotfix to stage? @jrgm?

fixes #3502

### DIFF
--- a/resources/static/dialog/js/modules/set_password.js
+++ b/resources/static/dialog/js/modules/set_password.js
@@ -20,7 +20,13 @@ BrowserID.Modules.SetPassword = (function() {
         options = this.options,
         valid;
 
-    if (dom.is(PASSWORD_SELECTOR, ":focus") && pass && !vpass) {
+    // In FirefoxOS, when the user clicks on the keyboard, it causes the
+    // form fields to lose focus, meaning positive checks for focus will not
+    // match. If the user is not focused on the vpass field and there is
+    // a password but no vpassword, put them in the vpassword field without
+    // showing an error. See issue #3502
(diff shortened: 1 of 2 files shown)


resources/static/dialog/js/modules/set_password.js
```diff
@@ -20,7 +20,13 @@ BrowserID.Modules.SetPassword = (function() {
         options = this.options,
         valid;
 
-    if (dom.is(PASSWORD_SELECTOR, ":focus") && pass && !vpass) {
+    // In FirefoxOS, when the user clicks on the keyboard, it causes the
+    // form fields to lose focus, meaning positive checks for focus will not
+    // match. If the user is not focused on the vpass field and there is
+    // a password but no vpassword, put them in the vpassword field without
+    // showing an error. See issue #3502
```
jedp, line 27: very helpful comment

jedp:
So sweet.  This is honestly my favorite UX fix for FirefoxOS.  What does that say about me.  I don't know.  But this is beautiful.

jedp:
Restored the branch; will make a PR for stage.

PR #3516, opened 2013-06-12
Issue 2280 missing kpi fields

shane-tomlinson:
@kparlante, this builds on #3495 to ensure all the expected KPIs are there.

That should be reviewed before this, but I would like to get feedback if possible.


kparlante:
@shane-tomlinson : I tested #3495 without this PR and caused a blob to happen that excluded the KPIs (no window.unload event). I then tested with this PR and managed to again create a blob with no window.unload event, but did indeed see the KPIs show up. So this part looks good to me r+

PR #3517, opened 2013-06-12
Cleanup on the "email_chosen" state and reduce duplication of code.

shane-tomlinson:
@ozten, @seanmonstar or @jedp - would you guys mind giving this a review? I consolidated the code that decided how to handle transition states, I am hoping this is a little easier to follow. You guys had the most context from doing the original stuff.

No Selenium tests were changed putting this together, user facing functionality should be identical.

An ephemeral instance is open for testing: simplify-email-chosen.personatest.org
- Code that decided how to handle transition states and clear certs was sprinkled throughout the code, making it difficult to maintain.
- Move cert invalidation that happened in state.js into user.js.
- Rename user.js->checkEmailIssuer to checkForInvalidCerts
- Simplify checkForInvalidCerts
- Add more unit tests.


resources/static/dialog/js/misc/state.js
```diff
@@ -387,95 +410,89 @@ BrowserID.State = (function() {
...
+      user.resetCaches();
+      user.addressInfo(info.email, function(addressInfo) {
+        mediator.publish("kpi_data", { email_type: addressInfo.type });
+
+        // the cert could have been cleared by the call to addressInfo. re-load
+        // the record info.
+        record = user.getStoredEmailKeypair(email);
+
+        // XXX If the user has a cert, do we really care of the primary is
+        // offline? Shouldn't we just sign them in?
```
seanmonstar, line 437: If the primary is offline, the RP will likely be unable to verify the assertion...

PR #3523, opened 2013-06-13
Update coarse user agent parser for FirefoxOS

### LINKED ISSUE #3373: B2G Metrics
It would be great if we could be tracking B2G persona accesses when it goes to market.

As long as B2G isn't on its own isolated server, I believe it will show up in router-metrics, identifiable by the browser. I don't think we can currently identifiy verifier hits that are from B2G. With auto-login, we don't see a hit in router-metrics for every login.
- Do we want/need additional B2G metrics to count hits beyond calls to signin?
- Will B2G be on the main persona service before it goes to market?

If the answers are No and Yes, we can just close this bug.

shane-tomlinson:
us_matchers is now a table, the key is the OS, the value is the string to search for.

Add the current FirefoxOS string

fixes #3373

### DIFF
--- a/lib/coarse_user_agent_parser.js
+++ b/lib/coarse_user_agent_parser.js
@@ -19,13 +19,28 @@ exports.parse = function (ua_string) {
         browser: 'Unknown',
         version: 'Unknown'
       },
-      os_matchers = [
-        'iPod', 'iPad', 'iPhone',
-        'Android', 'BlackBerry', 'Linux', 'Macintosh',
+      os_matchers = {
+        'iPod': 'iPod',
+        'iPad': 'iPad',
+        'iPhone': 'iPhone',
+        'Android': 'Android',
+        'BlackBerry': 'BlackBerry',
+        'Linux': 'Linux',
+        'Macintosh': 'Macintosh',
+        // http://lawrencemandel.com/2012/07/27/decision-made-firefox-os-user-agent-string/
+        'FirefoxOS': 'Mozilla/5.0 (Mobile; rv:',
         // http://en.wikipedia.org/wiki/Microsoft_Windows#Timeline_of_releases
-        // Windows 8       Windows 7        Windows Vista     Windows XP        Windows 2000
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@kparlante, mind giving this a review?

lib/coarse_user_agent_parser.js
```diff
@@ -19,13 +19,28 @@ exports.parse = function (ua_string) {
...
+        'Linux': 'Linux',
+        'Macintosh': 'Macintosh',
+        // http://lawrencemandel.com/2012/07/27/decision-made-firefox-os-user-agent-string/
+        'FirefoxOS': 'Mozilla/5.0 (Mobile; rv:',
         // http://en.wikipedia.org/wiki/Microsoft_Windows#Timeline_of_releases
-        // Windows 8       Windows 7        Windows Vista     Windows XP        Windows 2000
-        'Windows NT 6.2', 'Windows NT 6.1', 'Windows NT 6.0', 'Windows NT 5.1', 'Windows NT 5.0'
-      ],
+        // Windows 8
+        'Windows NT 6.2': 'Windows NT 6.2',
```
shane-tomlinson, line 34: @kparlante - we could change the key for all of these windows versions to be their common name - ie Windows 8 instead of Windows NT 6.2

lib/coarse_user_agent_parser.js
```diff
@@ -19,13 +19,28 @@ exports.parse = function (ua_string) {
...
+      os_matchers = {
+        'iPod': 'iPod',
+        'iPad': 'iPad',
+        'iPhone': 'iPhone',
+        'Android': 'Android',
+        'BlackBerry': 'BlackBerry',
+        'Linux': 'Linux',
+        'Macintosh': 'Macintosh',
+        // http://lawrencemandel.com/2012/07/27/decision-made-firefox-os-user-agent-string/
+        'FirefoxOS': 'Mozilla/5.0 (Mobile; rv:',
```
ozten, line 31: Awesome, thank you.

ozten:
Looks good. @kparlante sorry to merge before your review. Just saw that Shane asked for you.

kparlante:
@ozten no worries, I'll have my fun with it as I add it to the segmentations!

PR #3534, opened 2013-06-13
'native' is reserved on Android 2.2, so access it array-style.

### LINKED ISSUE #3522: dev and beta sites not loading dialog on android 2.2
1. goto {dev,beta}.123done.org on a Android 2.2 stock browser, or Firefox
2. click the sign in button

actual: dialog opens and spins with loading icon

expected: it should load.  It works on Android 4.0 and desktop as expected.

jaredhirsch:
Fixes #3522. FML.

### DIFF
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -50,7 +50,8 @@ BrowserID.Modules.Dialog = (function() {
     if (win.sessionStorage.primaryVerificationFlow) {
       try {
         var info = JSON.parse(win.sessionStorage.primaryVerificationFlow);
-        if (info.native) return;
+        /*jshint sub: true */
+        if (info['native']) return;
       } catch(e) {
         self.renderError("error", {
           action: {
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@6a68, in test/cases/dialog/js/modules/verify_primary_user.js, there are two more instances of ".native". Can you update theses as well?

With that change, I am r+ on this, tests pass: selenium, frontend, backend.

jaredhirsch:
@shane-tomlinson nice catch, thanks!

I did a little digging. 'native' is an ES2/ES3 reserved word. IE7/8 (and IE9+ in compat mode presumably?) will barf on ES3 reserved words, as will Android up to 3 (not 100% sure on this, looking for a canonical ref on ES versions in embedded webkit).

It turns out there's a pull request opened just yesterday to fix a jshint bug where ES3 mode does not trigger errors when ES3 reserved words are used as object properties: https://github.com/jshint/jshint/pull/1139

Going to get this merged in for expediency's sake, then going to look at our jshint settings. If we can enable ES3 mode, and that patch lands, then we can theoretically avoid this craziness for other words, like, I dunno, 'volatile' or 'goto' or 'public' :grimacing:

shane-tomlinson:
Nice work on this @6a68!

PR #3545, opened 2013-06-17
Do not generate a silent assertion for an address in a transition state.

shane-tomlinson:
@lloyd, @ozten, @seanmonstar - need discussion and review.


jaredhirsch:
@shane-tomlinson holy :hankey: dude, that was quick! Glad I was on the right track, if going 20x slower than your analysis :-D

shane-tomlinson:
From the email I sent yesterday, emails converted from a primary-secondary were only marked as secondary when /wsapi/cert_key was generated. This meant if a user closed the original dialog before they verified their email address, when they re-opened the dialog the address would _still_ be in a transition state.

To avoid this, I made a change to the backend - as soon as a user verifies ownership of an email address with the fallback IdP, mark the address to a secondary. The user verified. They signaled their desire to convert.

I had to update one backend test, in address-info-test.js, line 373. The test checks the return status of /wsapi/address_info and is labelled "returns 'known' and not 'offline'". @ozten, you made the change on line 373 from "known" to "transition_to_secondary", I changed it back to "known".

I believe our behavior has been incorrect and could have led to confusion if users tried to convert their account from primary->secondary and closed the original dialog before verifying their ownership of the address. If users did this on an RP that used the .get API, the user would have continually been locked in the "transition_to_secondary" state because /wsapi/cert_key would never have been called.

What, if any, repercussions does my change have?

shane-tomlinson:
@ozten, @lloyd, @seanmonstar - can I get review/discussion on this?

lib/wsapi/complete_user_creation.js
```diff
@@ -95,7 +95,16 @@ exports.process = function(req, res) {
                                      duration_ms: durationInfo.durationMS
                                     }, function(err) {
             if (err) return wsapi.databaseDown(res, err);
-            res.json({ success: true });
+            // If the user created an account for an address whose primary
+            // is disabled, the database has the addresses type as "primary".
+            // Unless the type is changed to "secondary", /wsapi/address_info
+            // will return "transition_to_secondary" as the address state.
+            // The user has verified the address. The transition is complete.
+            // Get rid of the transition state.
```
ozten, line 103: This comment is long and confusing. Can we make it more concise? Write it to your future self.

resources/static/common/js/user.js
```diff
@@ -1178,17 +1188,12 @@ BrowserID.User = (function() {
      */
     checkForInvalidCerts: function(email, info, done) {
       function clearCert(email, idInfo) {
+        delete idInfo.priv;
```
ozten, line 1191: Nice.

resources/static/common/js/user.js
```diff
@@ -1464,24 +1476,86 @@ BrowserID.User = (function() {
...
-            }
-          } else {
-            onComplete(loggedInEmail, null);
+        var loggedInEmail = storage.site.get(origin, "logged_in");
+        // User is not signed in to Persona or not signed into the site.
+        if (!(authenticated && loggedInEmail))
+          return complete(onComplete, null, null);
+
+        User.resetCaches();
+        User.addressInfo(loggedInEmail, function(info) {
```
ozten, line 1536: This is expensive. I'm guessing you've tried hard to avoid this. Down the road we could 
- add an expiration to our cache, 
- Do a series of addressInfo calls while the dialog is idle
- ???
  or something to optimize this, or some other optimization to avoid the extra network hit.

tests/address-info-test.js
```diff
@@ -370,7 +370,7 @@ suite.addBatch({
           assert.isNull(e);
           var r = JSON.parse(r.body);
           assert.equal(r.type, "secondary");
-          assert.equal(r.state, "transition_to_secondary");
+          assert.equal(r.state, "known");
```
ozten, line 373: Sorry about that. I don't have much of this swapped into my head, but this looks good.

Issue #342, opened 2011-10-03
Safari on OSX errors trying to access relay frame.


shane-tomlinson:

shane-tomlinson:
This actually _crashes_ Safari 5.0.4 on Windows 7.

shane-tomlinson:
df6603eaafa45a9948c567e6e5baebd96cc51c94 fixes this.  Removed "autofocus" from the email element.  Stupidest crash bug ever.

PR #3554, opened 2013-06-19
compensate for focus issues on ie8 by clicking text fields before sending keystrokes to it

jrgm:
This is the wrong way to fix this, as it points out some focus issues on IE8 in the UI. So don't actually merge this PR if we can clean up where focus is wrong.

So, when https://github.com/lloyd/testidp.org/issues/10 is fixed, and the IE8 focus issues get resolved, the IE{8.9} situation goes from this:

pre

To this;
post

(Side note: Opera is confused about clicking on the button with an very long testidp.org email address that extends outside the view to the left; this seems to me a bug in OperaDriver targetting the top-left rather than the middle of the element. Some of the OSX failures for Safari are due to the lack of support for dealing with `alert`'s (already an issue on file)).

### DIFF
--- a/automation-tests/tests/add-primary-to-primary.js
+++ b/automation-tests/tests/add-primary-to-primary.js
@@ -65,6 +65,7 @@ runner.run(module, {
       .wclick(CSS['123done.org'].signInButton)
       .wwin(CSS['dialog'].windowName)
       .wclick(CSS['dialog'].useNewEmail)
+      .wclick(CSS['dialog'].newEmail) // Click this text field to help IE8.
       .wtype(CSS['dialog'].newEmail, secondPrimaryEmail)
       .wclick(CSS['dialog'].addNewEmailButton)
       .wclick(CSS['dialog'].verifyWithPrimaryButton)
--- a/automation-tests/tests/add-primary-to-secondary.js
+++ b/automation-tests/tests/add-primary-to-secondary.js
@@ -72,6 +72,7 @@ runner.run(module, {
       .wclick(CSS['123done.org'].signInButton)
       .wwin(CSS['dialog'].windowName)
       .wclick(CSS['dialog'].useNewEmail)
+      .wclick(CSS['dialog'].newEmail) // Click this text field to help IE8.
       .wtype(CSS['dialog'].newEmail, primaryEmail)
       .wclick(CSS['dialog'].addNewEmailButton)
       .wclick(CSS['dialog'].verifyWithPrimaryButton)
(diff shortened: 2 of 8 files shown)


shane-tomlinson:
@jrgm - did you add a wclick before _every_ wtype? If so, we may be better off extending wtype to first do a wclick before typing.

jaredhirsch:
@shane-tomlinson how about, if the UA is IE8, we wclick then type? ;-) :hack:

jrgm:
@shane-tomlinson I don't know if it was before _every_ wtype; it was for the cases where it was needed to get IE8 to work. 

Folding a wtype into wclick would look cleaner in the test code. But by "wrong way to fix this", I was thinking that this may be pointing at real places where we focus is not set correctly in IE8. For example, when you add another email to an account, on that screen, for IE8, the text field does not have focus.

jrgm:
And I probably should split out separate issues for screens where default focus is not set on IE8.

jrgm:
So, the lines where I added an extra click to set focus on a text field on IE8 are:

```
.wclick(CSS['dialog'].newEmail)                 add-primary-to-primary.js:68:
.wclick(CSS['dialog'].newEmail)                 add-primary-to-secondary.js:75:
.wclick(CSS['dialog'].newEmail)                 idp-transition/transition-to-secondary-forgot-password.js:86:
.wclick(CSS['dialog'].newEmail)                 idp-transition/transition-to-secondary.js:86:
.wclick(CSS['dialog'].newEmail)                 remove-email.js:112:
.wclick(CSS['dialog'].newEmail)                 remove-email.js:129:
.wclick(CSS['dialog'].newEmail)                 returning-user.js:66:
.wclick(CSS['dialog'].choosePassword)           idp-transition/primary-shuts-down-single-email.js:76:
.wclick(CSS['dialog'].choosePassword)           remove-email.js:132:
.wclick(CSS['dialog'].choosePassword)           returning-user.js:69:
.wclick(CSS['dialog'].choosePassword)           returning-user.js:71:
.wclick(CSS['dialog'].postVerificationPassword) reset-password-test.js:71:
```

The actual CSS for the above are:

```
newEmail: 'input#newEmail',
choosePassword: 'div#set_password input#password',
postVerificationPassword: 'input#password',
```

Fixing GH-3616 and GH-3617 on IE8 probably remove any need for this PR.

shane-tomlinson:
> But by "wrong way to fix this", I was thinking that this may be pointing at real places where we focus is not set correctly in IE8.

@jrgm - that's what I thought you meant, I had a PR [#3215] out to fix these but @6a68 and I decided "it wasn't ready". I am half tempted to resurrect that from the dead or use an approach similar to what @6a68 suggests in https://github.com/mozilla/browserid/pull/3215#issuecomment-16270216. I think instead of tabindex, I'd rather use a data-autofocus attribute. We saw how using autofocus with dynamically inserted elements can blow up certain versions of webkit in #342.

shane-tomlinson:
@jrgm - I am working on the "correct" way to do this, I'm gonna close this for now.

jrgm:
@shane-tomlinson yep, closing this makes good sense (I never meant for this to actually land).

shane-tomlinson:
@jrgm - BTW, thanks for the list of buttons, it makes it much easier to focus on specific problem areas.

PR #3565, opened 2013-06-20
When the user cancels from the "primary_offline" state, reset the caches so the user can try the address again.

### LINKED ISSUE #3155: If in a state in dialog where address lookup failed, should retry the lookup after cancel.
Enter an '@yahoo.com' email in the dialog.

If there is a network issue in looking up information from yahoo, in some cases the dialog will default to treat that address as a secondary; in other cases (not sure what the timing difference is), the dialog will transition to show the message `yahoo.com is not responding. Please wait a few minutes and try again.` with a cancel button.

The issue here is that if you click on `cancel` you go back to a screen where the email address you chose is shown; clicking `next` makes no attempt to retry to contact the network. It immediately goes back to the screen with 'not responding' and the cancel button. The user must quit the dialog to get out of this situation.

So, basically, clicking cancel should change state so that a retry is attempted.

We should get a fix for this on train-2013.03.28. Marking 5-star for now.

shane-tomlinson:
@ozten, you were the original module author, mind giving this a review?

An ephemeral instance is set up at https://address-lookup-faill.personatest.org

Fixes #3155

### DIFF
--- a/resources/static/dialog/js/misc/state.js
+++ b/resources/static/dialog/js/misc/state.js
@@ -421,8 +421,8 @@ BrowserID.State = (function() {
         // the record info.
         record = user.getStoredEmailKeypair(email);
 
-        // XXX If the user has a cert, do we really care of the primary is
-        // offline? Shouldn't we just sign them in?
+        // If a primary is offline, do not generate a cert for it because there
+        // is a good chance that the RP cannot fetch the IdP's public key.
         if ('offline' === addressInfo.state) {
           redirectToState("primary_offline", addressInfo);
         }
--- a/resources/static/dialog/js/modules/primary_offline.js
+++ b/resources/static/dialog/js/modules/primary_offline.js
@@ -5,10 +5,14 @@ BrowserID.Modules.PrimaryOffline = (function() {
   "use strict";
 
   var bid = BrowserID,
-      helpers = bid.Helpers;
+      helpers = bid.Helpers,
+      user = bid.User;
 
   function startDialogOver() {
     /*jshint validthis:true*/
+    // Reset the caches so that a user who cancels from this screen can go back
+    // and try the same address again in a few minutes.
+    user.resetCaches();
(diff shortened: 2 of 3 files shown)


resources/static/dialog/js/misc/state.js
```diff
@@ -421,8 +421,8 @@ BrowserID.State = (function() {
         // the record info.
         record = user.getStoredEmailKeypair(email);
 
-        // XXX If the user has a cert, do we really care of the primary is
-        // offline? Shouldn't we just sign them in?
+        // If a primary is offline, do not generate a cert for it because there
+        // is a good chance that the RP cannot fetch the IdP's public key.
```
ozten, line 425: Nice to have that question answered ;)

resources/static/dialog/js/modules/primary_offline.js
```diff
@@ -5,10 +5,14 @@ BrowserID.Modules.PrimaryOffline = (function() {
...
   var bid = BrowserID,
-      helpers = bid.Helpers;
+      helpers = bid.Helpers,
+      user = bid.User;
 
   function startDialogOver() {
     /*jshint validthis:true*/
+    // Reset the caches so that a user who cancels from this screen can go back
+    // and try the same address again in a few minutes.
+    user.resetCaches();
```
ozten, line 15: Nice and clean.

resources/static/test/cases/dialog/js/modules/primary_offline.js
```diff
@@ -53,5 +55,39 @@
...
+            email: "testuser@testuser.com",
+            type: "primary",
+            state: "known"
+          });
+          xhr.useResult("primary");
+
+          start();
+        }, testHelpers.unexpectedFailure);
+      }, testHelpers.unexpectedFailure);
+  });
```
ozten, line 90: Thanks for the test!

PR #3567, opened 2013-06-20
Fix primaries on FirefoxOS

shane-tomlinson:
@jedp - this is what I am thinking so far.

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -62,6 +62,11 @@ var conf = module.exports = convict({
     format: 'string = "https://login.persona.org"',
     env: 'PUBLIC_URL'
   },
+  firefoxos_url: {
+    doc: "The publically visible FirefoxOS URL of the deployment",
+    format: 'string',
+    env: 'FIREFOXOS_URL'
+  },
   public_static_url: {
     doc: "The publically visible URL from which static resources are served",
     format: 'string',
@@ -360,13 +365,17 @@ if (conf.get('env') === 'test_json') {
   conf.set('database.driver', 'mysql');
 }
 
-// if static and verifier urls are not explicitly set, default them to the
-// same as the public url (used in ephemeral and local deployments)
+// if static, verifier, and firefoxos urls are not explicitly set,
+// default them to the same as the public url (used in ephemeral
+// and local deployments)
 if (!conf.has('public_static_url') || conf.get('public_static_url') === null) {
   conf.set('public_static_url', conf.get('public_url'));
 }
 if (!conf.has('public_verifier_url')) {
   conf.set('public_verifier_url', conf.get('public_url'));
+}
+if (!conf.has('firefoxos_url') || conf.get('firefoxos_url') === null) {
+  conf.set('firefoxos_url', conf.get('public_url'));
 }
 
 // augment the cachify prefix with the public_static_url
@@ -387,12 +396,14 @@ conf.validate();
  */
 module.exports.performSubstitution = function(app) {
   if (conf.get('public_url') !== 'https://login.persona.org' ||
+      conf.get('firefoxos_url') !== 'https://firefoxos.persona.org' ||
       conf.get('public_static_url') !== 'https://static.login.persona.org' ||
       conf.get('public_verifier_url') !== 'https://verifier.login.persona.org')
   {
     app.use(postprocess(function(req, buffer) {
       return buffer.toString()
         .replace(new RegExp('https://login.persona.org', 'g'), conf.get('public_url'))
+        .replace(new RegExp('https://firefoxos.persona.org', 'g'), conf.get('firefoxos_url'))
         .replace(new RegExp('https://static.login.persona.org', 'g'), conf.get('public_static_url'))
         .replace(new RegExp('https://verifier.login.persona.org', 'g'), conf.get('public_verifier_url'));
     }));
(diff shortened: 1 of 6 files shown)


resources/static/authentication_api.js
```diff
@@ -33,16 +33,25 @@
       setTimeout(function() { cb(email); }, 0);
     };
 
+    var ipServer = 'https://login.persona.org';
+    if (navigator.userAgent ===
+          "Mozilla/5.0 (Mobile; rv:18.0) Gecko/18.0 Firefox/18.0") {
```
jedp: I guess we can just check that it matches "Mobile" and "Gecko"?

jedp:
That is pretty damn cool

resources/static/authentication_api.js
seanmonstar: Zoh my gawd, why? We can't feature detect FirefoxOS?

shane-tomlinson: @seanmonstar - proof of concept, don't sweat yet! Do you have a feature detection snippet for FFOS?

shane-tomlinson: I suppose we could check for navigator.mozId

jedp: maybe check for navigator.mozId and scan for Mobile in the UA string.  I don't know of a way to feature detect for fxos specifically

lib/configuration.js
```diff
@@ -393,6 +393,7 @@ module.exports.performSubstitution = function(app) {
     app.use(postprocess(function(req, buffer) {
       return buffer.toString()
         .replace(new RegExp('https://login.persona.org', 'g'), conf.get('public_url'))
+        .replace(new RegExp('https://firefoxos.persona.org', 'g'), conf.get('public_url'))
```
shane-tomlinson: the replacement should be pointing to the cname instead of public_url.

ozten:
Heroic!

lib/configuration.js
```diff
@@ -393,6 +398,7 @@ module.exports.performSubstitution = function(app) {
     app.use(postprocess(function(req, buffer) {
       return buffer.toString()
         .replace(new RegExp('https://login.persona.org', 'g'), conf.get('public_url'))
+        .replace(new RegExp('https://firefoxos.persona.org', 'g'), conf.get('firefoxos_url'))
```
jedp, line 406: How about something like `https://firefoxos\.(?:persona|(?:dev\.)?anosrep)\.org`, so we work with dev and stage, too?

shane-tomlinson, line 406: We always keep firefoxos.persona.org within the code, postprocess takes care of updating the firefoxos_url based on the environment.

jedp, line 406: Ah ok.  That makes sense.  Thanks for the comments in the code about this.

lib/configuration.js
```diff
@@ -360,14 +365,18 @@ if (conf.get('env') === 'test_json') {
...
 
-// if static and verifier urls are not explicitly set, default them to the
-// same as the public url (used in ephemeral and local deployments)
+// if static, verifier, and firefoxos urls are not explicitly set,
+// default them to the same as the public url (used in ephemeral
+// and local deployments)
 if (!conf.has('public_static_url') || conf.get('public_static_url') === null) {
   conf.set('public_static_url', conf.get('public_url'));
 }
 if (!conf.has('public_verifier_url')) {
```
jedp, line 374: Does this line also want the `|| conf.get('public_verifier_url') === null` test, like the others?

resources/static/authentication_api.js
```diff
@@ -33,16 +33,28 @@
       setTimeout(function() { cb(email); }, 0);
     };
 
+    var ipServer = 'https://login.persona.org';
+    // FirefoxOS UA string:
+    // "Mozilla/5.0 (Mobile; rv:18.0) Gecko/18.0 Firefox/18.0
+    // Android UA string:
+    // Mozilla/5.0 (Android; Mobile; rv:18.0) Gecko/18.0 Firefox/18.0
+    if (!!navigator.mozId && (navigator.userAgent.indexOf("(Mobile;") > -1)) {
```
jedp, line 41: Maybe we can query a property on `navigator.mozSettings`.  It's what the Settings app looks at to get the OS version etc.  I'm trying to find out whether there's a stable way to do that, with which vendors cannot interfere.

jedp, line 41: Oho! Or maybe UA sniffing is precisely what we want to do.  [Partners are not allowed to modify the FirefoxOS UA string](https://wiki.mozilla.org/B2G/User_Agent/Partner_Changes_Policy).

lib/configuration.js
shane-tomlinson, line 374: I don't believe so, public_verifier_url is not defined in production.js and does not need to be overridden. I left this code as is.

resources/static/authentication_api.js
shane-tomlinson, line 41: @jedp - do you know what we can get out of navigator.mozSettings?

jedp, line 41: All the device info you can see in the Settings app is taken from `navigator.mozSettings`.  I was surprised, though, when talking with someone on Deutsche Telekom's QA team to find that they don't have quite the same set of things we have our our dev builds.  I can look into this further to find out what the real story is.

jedp:
So, this is awesome.

PR #3571, opened 2013-06-21
Async snippet early AM draft [freaky friday fun]

jaredhirsch:
Not ready, want comments. @shane-tomlinson @seanmonstar @lloyd thoughts here?

I woke up thinking about the need for an async snippet this morning, maybe it's all the velocity conf tweets I've been reading, dunno ;-) Anyway, it's working locally for me.

This could go a number of different ways:
- it could grow into a self-hostable include.js
- it could just be a fancy async snippet with some stuff hidden behind the scenes + detailed docs, a la google analytics' [async snippet](https://developers.google.com/analytics/devguides/collection/gajs/)
- it could grow into an expert-mode async SDK (but that would change our API which is a huge decision, not to undertake lightly).

Thoughts/comments/heckling appreciated

### DIFF
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -47,6 +47,43 @@ pre {
 }
 
 </style>
+<script>
+navigator.id = navigator.id || {}; // important: does include.js bail if navigator.id, navigator.id.watch, or navigator.id.get is defined?
+navigator.id.q = [];
+
+function makeQueueable(funcName) {
+  var f = function() { navigator.id.q.push({func: funcName, args: arguments}) }
+  f.isQueueable = true; // so we can check it was overwritten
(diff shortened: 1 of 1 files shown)


example/rp/index.html
```diff
@@ -47,6 +47,19 @@
 }
 
 </style>
+<script>
+navigator.id = navigator.id || {}; // important: does include.js bail if navigator.id, navigator.id.watch, or navigator.id.get is defined?
```
jaredhirsch, line 51: actually, if navigator.id exists, makeQueueable isn't needed and we should just bail, I think.

jaredhirsch:
closing in favor of rebased version

PR #3580, opened 2013-06-25
Support localization of TOS and PP

ozten:
With this patch...
- tos.ejs and privacy.ejs are moved into `resources/views/<locale>/`
- On startup, localized TOS and PP files are noted
- For any supported locale, if files exists, links in dialog are updated


resources/views/signin.ejs
```diff
@@ -0,0 +1,117 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
```
jrgm: Did you really mean to add back this obsolete, previously removed view?

ozten: No, great catch! Thanks jrgm. Is it just a matter of removing `resources/views/signin.ejs`?

(I had a rather nasty merge as this work has been done for a while waiting on legal)

jrgm: Yeah, I /think/ it just needs this file removed (the routes and page-tests, etc., don't appear to have been added back).

resources/static/pages/js/start.js
```diff
@@ -158,6 +158,7 @@ $(function() {
         module = bid.about.create();
         module.start({});
       }
+      // TODO /en/tos
```
ozten, line 161: Do we need to do anything here? I don't think so and will remove TODO. Just wanted to call it out.

ozten:
All green except the last one is grey. Rerunning Travis build (3rd time). War on orange isn't just for bugzilla anymore ;)

lib/static/views.js
```diff
@@ -222,31 +254,45 @@ exports.setup = function(app) {
...
 
-  app.get("/tos", function(req, res) {
-    renderCachableView(req, res, 'tos.ejs', {
-      title: _('Terms of Service'),
-      fullpage: false
+  function pickLocale(supported, lang) {
+    var templateLocale = 'en',
+        locale = i18n.localeFrom(lang);
+
+    if (supported.indexOf(locale) !== -1) {
```
jedp, line 261: Should the logic be more nuanced here?  Say we have `pt` but not `pt-BR`; should we not fall back to `pt` instead of `en`?  (I honestly don't know.  Nationalist sentiments may prefer falling back to `en` in such a case.  I cannot speak for what should happen with `en-US` and `en-GB`, we two peoples "separated by a common language.")

resources/static/pages/js/start.js
jedp, line 161: I don't think so.  I believe this is actually your note, and this PR seems to provide what it was requesting?

I wonder, somewhat relatedly, whether the `helpers.log` message that follows in this file should print the path that was requested? (`helpers.log("Unknown path: " + path);`)  If there is any confusion in this url change, this may be useful?

jedp:
War on Orange.  That should be our band name.

lib/static/views.js
ozten, line 261: Exact match is best for political reasons.

We can add special logic on a case by case basis... but this doesn't happen much in `supported_languages` today.

resources/static/pages/js/start.js
ozten, line 161: Ya, my note. I was wondering if we needed to tweak the matches, but I found it works without it.

Thanks for double checking.

> (helpers.log("Unknown path: " + path);) 

Good call. Will do.

jedp:
I don't know what travis's beef is with this.  Tests look great locally, and manual testing of the urls and fallback cases is working perfectly.

jedp:
Thanks for re-pushing.  Travis is now all nice and green.

PR #3582, opened 2013-06-26
Ensure caches always check ETags on dynamic pages. Fixes #3547

### LINKED ISSUE #3547: investigate cache headers on /sign_in
There were reports of people getting 404 for content from static.login.persona.org, such as me, jedp and ednapirana. 

ednapirana's url was https://static.login.persona.org/v/f915b0aa15/production/en/dialog.css. 

That would be a correct url on the train-2013.04.26 release. (Production was updated to train-2013.05.22 at 2:30pm PDT today). 

```
curl -k -s https://persona-org-0525-1045101880.us-west-2.elb.amazonaws.com/v/f915b0aa15/production/en/dialog.css | md5sum
f915b0aa1558424be23575603f082c32  -
```

But the /sign_in dialog is served with `Cache-Control: public, max-age=0` and an ETag header, so a browser should request again each time the dialog comes up with 'If-None-Match', and that's what I see it doing with FF21. 

Anyways, is it a problem with the etags, or some weird browser cacheing, or weird DNS, or...

jaredhirsch:
@shane-tomlinson or @jrgm, mind having a look? I feel like this bug took me down a rabbit hole into an upside down world of eight-year-old blog posts/bugzilla bugs covering HTTP caching quirks in IE6 etc. Yuck!

Not sure the best way to test this. I considered actually changing the static endpoints and reloading the dialog, but it seems like there should be a way to do this via front-end unit test.

### DIFF
--- a/lib/static/views.js
+++ b/lib/static/views.js
@@ -39,7 +39,7 @@ function cacheAllTheThings(res) {
   if (config.get('env') !== 'local') {
     // allow caching, but require revalidation via ETag
     res.etagify();
-    res.setHeader('Cache-Control', 'public, max-age=0');
+    res.setHeader('Cache-Control', 'public, max-age=0, must-revalidate');
   } else {
     // disable all caching for local dev
     res.setHeader('Cache-Control', 'no-store');
--- a/tests/cache-header-tests.js
+++ b/tests/cache-header-tests.js
@@ -74,8 +74,8 @@ function hasProperCacheHeaders(path) {
       assert.strictEqual(r.statusCode, 200);
       // check X-Frame-Option headers
       hasProperFramingHeaders(r, path);
-      // ensure public, max-age=0
-      assert.strictEqual(r.headers['cache-control'], 'public, max-age=0');
+      // ensure public, max-age=0, must-revalidate
+      assert.strictEqual(r.headers['cache-control'], 'public, max-age=0, must-revalidate');
       // the behavior of combining a last-modified date and an etag is undefined by
       // rfc2616, so let's always use ETags, and ignore last modified date.
       assert.ok(r.headers['etag'])


lib/static/views.js
```diff
@@ -39,7 +39,7 @@ function cacheAllTheThings(res) {
   if (config.get('env') !== 'local') {
     // allow caching, but require revalidation via ETag
     res.etagify();
-    res.setHeader('Cache-Control', 'public, max-age=0');
+    res.setHeader('Cache-Control', 'public, max-age=0, must-revalidate');
   } else {
```
jrgm, line 43: @6a68 I was wondering about your suggestion to provide the 1.0 Expires header as well, for the possible benefit of proxy caches. But otherwise, since I suggested this, I'm r+, but would appreciate input from @shane-tomlinson, @lloyd and @fmarier.

shane-tomlinson, line 43: We considered this at one time - for more info: http://www.mnot.net/blog/2007/05/15/expires_max-age

shane-tomlinson:
Link to #3547

shane-tomlinson:
Interesting repercussions:

screen shot 2013-06-26 at 13 53 24
- This response allows all caches to store it.
- This response is already stale.
- This response cannot be served by a cache, because it is stale. 

I am most worried about the last one. Let's make sure individual browsers 304 in all cases we expect it to.

shane-tomlinson:
I did some very unscientific testing. I loaded https://caching.personatest.org/sign_in into a tab. I put my cursor into the URL bar and pressed enter to reload the page. CMD-R acts differently to pressing enter in the URL bar in that caching sometimes does not take effect with CMD-R.

Results:

Only Windows 7/IE9 304'd on the first reload. All other browsers that 304'd did so on the second reload.
- Firefox/OSX 304
- Chrome/OSX 304
- Safari/OSX _always_ re-downloads the entire page. It currently does this on login.persona.org/sign_in as well.
- Opera/OSX 304
- Windows XP/IE8 304
- Windows 7/IE8 304
- Windows 7/IE9 304 on FIRST reload
- Windows 8/IE10 304

jaredhirsch:
@shane-tomlinson - I find mnot's stuff kind of confusing--he isn't very careful about defining terms. For instance, does "This response cannot be served by a cache, because it is stale" apply to browser caches, or just proxy caches? I suspect the latter, but because the terms aren't carefully defined, one of us will have to go digging around in RFC 2616 to try to figure it out.

Thanks for testing against an awsbox, mang! I suspect that your win7/ie9 has a dirty cache. There's a great article on how hard this is to clean from webpagetest's maintainer, I'll dig up the ref shortly & try to repro separately.

lib/static/views.js
jaredhirsch, line 43: I dug a bit on whether HTTP 1.0 is even still in use, found this from 2k9 on an [erlang mailing list](https://groups.google.com/forum/#!msg/erlang-programming/9jG_SHjvs84/dO9WUS239ggJ):

> Talking to the NginX developers, I have clarified a couple of things:
> 
> 1) HTTP/1.0 is in use by proxies, some mobile clients, and IE when configured to use a proxy. So 1.0 appears to still account for a non-trivial % of traffic on the web overall.

I don't know which IE version. I'm not sure if it's worth adding the extra header.

jrgm:
Just pointing out that redbot has additional explanation in tooltips. Here's what it says in full for current production vs. the awsbox with this patch about the last point:

What it says for current production:

http://redbot.org/?uri=https%3A%2F%2Flogin.persona.org%2Fsign_in
  This response might be served by a cache, even though it is stale.

  HTTP allows stale responses to be served under some circumstances; for
  example, if the origin server can't be contacted, a stale response can be
  used (even if it doesn't have explicit freshness information). This
  behaviour can be prevented by using the Cache-Control: must-revalidate
  response directive.

What it says after adding `must-revalidate`:

http://redbot.org/?uri=https%3A%2F%2Fcaching.personatest.org%2Fsign_in
  This response cannot be served by a cache, because it is stale.

  The Cache-Control: must-revalidate directive forbids caches from using stale
  responses to satisfy requests.  For example, caches often use stale responses
  when they cannot connect to the origin server; when this directive is present,
  they will return an error rather than a stale response.

jaredhirsch:
@jrgm thanks for pointing that out! I guess mnot's tooltips confirm that we've made the change we wanted to make?

jaredhirsch:
@lloyd it turns out that #3547 is caused by a browser bug, not this header change. Thoughts on whether we should still make this change?

shane-tomlinson:
@6a68, @jrgm - I am tempted to add the cache headers anyways to ensure that browsers and proxies are not serving up stale versions under any "normal" circumstances - DNS cache problems not included.

jaredhirsch:
/me does it!!! @shane-tomlinson thanks for the review

PR #3585, opened 2013-06-26
Give useful information to the KPIs when reporting a custom error.

### LINKED ISSUE #3584: KPIs report "screen.error." if module.renderError is called with custom error.
screen.error. contains no useful developer facing information. 

Link to #2407

shane-tomlinson:
- If reporting a custom error type, use the action.title field as the reported error.

fixes #3584

### DIFF
--- a/resources/static/common/js/modules/interaction_data.js
+++ b/resources/static/common/js/modules/interaction_data.js
@@ -47,12 +47,16 @@ BrowserID.Modules.InteractionData = (function() {
     var parts = [];
 
     if (data.action) {
-      parts.push(_.keyOf(errors, data.action));
+      var errorType = _.keyOf(errors, data.action)
+                          || data.action.title || "unknown";
+      parts.push(errorType);
     }
 
     if (data.network && data.network.status > 399) {
       parts.push(data.network.status);
     }
+
+    if (!parts.length) parts[0] = "unknown";
 
     return 'screen.error.' + parts.join('.');
   }
--- a/resources/static/test/cases/common/js/modules/interaction_data.js
+++ b/resources/static/test/cases/common/js/modules/interaction_data.js
@@ -382,19 +382,37 @@
     start();
   });
 
-  test("error_screen formats an error object", function() {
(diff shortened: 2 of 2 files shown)


shane-tomlinson:
@kparlante - could you review this one?

seanmonstar:
Isn't the title localized?

shane-tomlinson:
@seanmonstar - generally not since these are catastrophic errors. The hope is users can send us an email or tweet us with the error.

@kparlante - an example of these errors is when an input parameter to navigator.id is invalid, we toss up an error screen with the error info.

seanmonstar:
So looking through error-messages.js, I see a few that are wrapped in
gettext.

shane-tomlinson:
@seanmonstar - the messages that you point out can reasonably be displayed to the user as part of the normal flow, the rest are supposed to be developer facing messages that are shown very rarely, in the case of a catastrophic error. 

cannotConfirm - is displayed when there is a problem with the signup link and the user cannot confirm their ownershipo of an address.
cannotComplete - is displayed if the link is alright, but for some reason the verification failed.
cookiesDisabled - is displayed if the user's cookies are disabled.
xhrError - generic.

In these cases, the table key (cannotConfirm, cannotComplete, cookiesDisabled, xhrError) is logged in the KPI data instead of the translated strings.

When we started localizing strings, we decided to only localize the subset of strings that users would see as a result of an action that they themselves took - ie, clicking a verification link twice, disabling cookies. Perhaps these strings should be moved out of this table to reduce confusion?

I'm down for a different solution if needed, my intent is to increase the usefulness of the blank custom errors that are being reported in the KPIs. These blank error messages provide us with no useful information that we can use to track down errors. :(

seanmonstar:
Ah right, I forgot that before the title, it will try the key from Errors
first. My bad :+1:

kparlante:
@shane-tomlinson : I have no issues with the code and don't want to block this. I haven't been able to create a blob that triggers this, though. FWIW, I tried using persona with cookies disabled (on mac Chrome and Safari) and no kpi blob gets sent.

Issue #3466, opened 2013-05-29
error seen once when signup new user at main site


jrgm:
I hit this once, but on retrying this ~10 more times, I haven't hit it again. This was with FF21 using an ephemeral instance with train-2013.05.22 - https://jrgmtrain0522.personatest.org.

STR
1. Load the main site and click signin.
2. In the dialog, enter a new email address and the password twice. Dialog transitions to polling for user_completion_status.
3. Load that email in a new window/tab of the same browser.

Expected: user completes creation in the new window, the dialog dismisses and the first window transitions to signed in state.
Actual: user completes creation in the new window, but I got this error below, and the dialog stops.

Timestamp: 5/29/13 4:29:51 PM
Error: Error: module not started for check_registration
Source File: https://jrgmtrain0522.personatest.org/v/6f499d473e/production/en/dialog.js
Line: 4

jrgm:
I was running selenium with IE8 (after repairing the tests for IE8) and had some failures. Saw the error icon, so ran the tests again and got lucky is seeing it in a live test with this message. The occassionally failing test was new-user-secondary-test and the error happens after the username and passwords have been submitted and the dialog transitions to the polling state.

jrgm:
The tests were running against stage which is now the train-2013.06.05

jrgm:
To clarify, by "this message" I mean "module not started for check_registration".

jrgm:
I saw this again in GH-3560. It's seems this may have become easier to hit. I'm raising this to 5-star for investigation.

jaredhirsch:
looking now

jaredhirsch:
Well, been at it for 90 minutes or so, here's what I've traced thus far.

The failure screen is triggered when the function call [in check_registration#startCheck](https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/check_registration.js#L47-L52) fires its errback:

```
# inside check_registration#startCheck
user[self.verifier](self.email, function(status) { ... }, self.getErrorDialog(...))
```

That's a dynamic function call, it's a bit much to unpack. But note that the basic deal is, we're calling a dynamically-selected function on the user object, `user[self.verifier]`, and the last argument is the errback, which displays the error screen observed when the bug attacks, `self.getErrorDialog(errors.registration, oncomplete)`.

[...]

PR #3587, opened 2013-06-26
Better sync dialog service state

jaredhirsch:
Needs comments. @shane-tomlinson mind taking a look?

When investigating #3466, I found out that there are some edge cases in which BrowserID.Modules.Actions doesn't cleanly sync the currently-running service state held by BrowserID.module.

In this diff, I've added some try/catches to ensure that the `runningService` state in Actions is properly unset and reset.

However, in making this change, the dialog/js/modules/actions unit tests fail--because they expect uncaught Errors to be thrown, and assert on the contents of the error message. This doesn't seem to match up with the actual use of the `start` and `stop` methods in the codebase, though: `start` and `stop` are called by [`startService`](https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/actions.js#L19-L33), which in turn is called by methods like [`doRPInfo`](https://github.com/mozilla/browserid/blob/dev/resources/static/dialog/js/modules/actions.js#L191-L193). No try/catches anywhere.

I can see two ways to go: either we modify the unit tests, or maybe we merge bid.Modules.Actions and bid.module, so that there's no need to juggle state across the two. Making the latter change is currently is a bit outside what I can pull off, I'd like to get input before delving in.

I guess a third option would be to let the Errors be uncaught and just be a little more careful about when we unset and reset `runningService` without adding all the try/catch/finally stuff.

### DIFF
--- a/resources/static/dialog/js/modules/actions.js
+++ b/resources/static/dialog/js/modules/actions.js
@@ -24,19 +24,16 @@ BrowserID.Modules.Actions = (function() {
       try {
         serviceManager.stop(runningService);
       } catch (e) {
-        // stop() throws if module wasn't running, no action needed,
-        // but IE8 can't do try/finally without a dummy catch. see eg
-        // http://webbugtrack.blogspot.com/2007/11/bug-184-catch-to-try-catch-finally-in.html
-      } finally {
-        runningService = null;
+        // stop() throws if runningService wasn't running, no action needed.
       }
+      runningService = null;
     }
 
     var module;
     try {
       module = serviceManager.start(name, options);
     } catch (e) {
-      // module failed to start or was already running.
+      // service failed to start or was already running.
       // we must parse the error message to determine which.
       if (e.message.indexOf('already running') > -1) {
         runningService = name;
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@6a68 - I didn't get a chance to look at this today, I will tomorrow. Thanks for digging in!

shane-tomlinson:
@6a68 - I am uncomfortable covering up these errors and would like to find out what the root cause is. 

I definitely agree that the majority of the code in actions and the stuff that makes sure only one action is running should be separated. Moving this code into lib/module.js may not be the right place because several modules are usually running, it's just that only one module started from actions.js can be running at a time.

resources/static/dialog/js/modules/actions.js
```diff
@@ -21,10 +21,25 @@ BrowserID.Modules.Actions = (function() {
 
     // Only one service outside of the main dialog allowed.
     if(runningService) {
-      serviceManager.stop(runningService);
+      try {
+        serviceManager.stop(runningService);
+      } catch (e) {
+        // stop() throws if runningService wasn't running, no action needed.
```
shane-tomlinson, line 27: maybe we should log the error into the KPIs?

shane-tomlinson:
@6a68 - I am going to close this for now until we can do a deep dive here. Let's collaborate on a branch on either your or my repo until we are ready for final review!

PR #3588, opened 2013-06-27
Fixes #3573 : make sure the '.buttonrow > .right' buttons line up on the right hand side

### LINKED ISSUE #3573: Next button on authentication page is not lined up with the right hand side of the input box.
screen shot 2013-06-24 at 12 51 58

chilts:
Patch for #3573.

Hoping that this doesn't affect anything else.

### DIFF
--- a/resources/views/dialog.ejs
+++ b/resources/views/dialog.ejs
@@ -95,7 +95,7 @@
 
                     <p class="submit cf buttonrow">
 
-                      <button class="isDesktop isStart isAddressInfo right">
+                      <button class="isDesktop isStart isAddressInfo">
                         <%= gettext('next') %>
                       </button>
 


shane-tomlinson:
@chilts - I didn't get a chance to review this today, I'll have a look at it tomorrow!

shane-tomlinson:
@chilts - unfortunately, that removes the margin when there is a button on the right and an anchor on the left (see set password or add email screens for an example). I forget the repercussions of just removing the right class from that button. Perhaps the "right" class on that button is just a relic of a past time and can safely be removed?

chilts:
@shane-tomlinson ok, I'll re-look into this. I tried a number of different pages but must have missed that one. Thanks. :)

shane-tomlinson:
@chilts - I'm gonna close this until we have a better solution. Want to pow wow on this? Thanks man!

chilts:
@shane-tomlinson thanks for reviewing and pulling :)

PR #3589, opened 2013-06-27
RP Branding!

### LINKED ISSUE #3286: Stronger relying party (RP) brand presence for sign-in and account creation UI elements
The Webmaker team is rebuilding its toolchain and experience, and has decided to make Persona an integral part of that solution, providing SSO and outsourcing identity authentication so that we're not responsible for yet another password store.

One issue we've run into is that from a user experience perspective, it is very difficult to communicate to a new user that they need to create a new Persona in order to obtain a Webmaker account. There is no way to "white label" account creation so that we can just show a new (non-Persona equipped) user an "email / password" field and have them sign up for a Persona.

We'd love the opportunity to discuss requirements and a potential solution here, most likely at the API level. I suspect we're one of the first consumers who is trying to go completely Persona-based, as opposed to Persona _or_ our own username/password system.

EDIT: morphed title of bug as per callahad's request; this isn't compete "white label", but rather just providing a stronger emphasis on the RP brand elements throughout the UI.

shane-tomlinson:
fixes #3286


callahad:
:sparkles: :exclamation:

resources/static/dialog/js/modules/dialog.js
```diff
@@ -293,6 +318,11 @@ BrowserID.Modules.Dialog = (function() {
           }
         }
 
+        if (paramsFromRP.backgroundColor) {
```
callahad, line 321: Do we want to hide this behind `experimental_backgroundColor`?

shane-tomlinson, line 321: @callahad this is probably a good way to go. I was going to merge it as is to get maximum exposure/testing before next Wednesday and let us debate until then.

jaredhirsch:
creating an account + signing in and out works fine for me on android 2.2, will look more later if it's not yet merged

jaredhirsch:
not yet merged, looking more now.

resources/static/dialog/js/modules/dialog.js
ozten, line 321: I think the experimental prefix should be weighed against how confident we are that people want it and how soon we want people to use it in it's final form.

The other properties we've put behind experimental, such as `forceIssuer` have had deep skepticism for inclusion and we don't want anyone using it who wouldn't be willing to update their code to an un-prefixed version.

I like experimental, as it is conservative... but I think we're pretty confident that we need this feature. We don't want partners to have to change their code in a 6 weeks or whatever to remove the experimental part, right?

jaredhirsch:
derp, out of time, didn't find anything to complain about :-)

resources/static/dialog/js/modules/dialog.js
```diff
@@ -293,6 +318,11 @@ BrowserID.Modules.Dialog = (function() {
...
+          var backgroundColor = validateBackgroundColor(paramsFromRP.backgroundColor);
```
jaredhirsch, line 322: Wait, maybe I did find a nit to pick ;-)

If the RP specifies an invalid background color, validateBackgroundColor will throw here. What's catching that error? What happens to the underlying page if we bail on execution right here--should we actually return a falsy backgroundColor if validation fails, and console.log the error instead?

resources/static/include_js/_include.js
```diff
@@ -461,6 +461,7 @@
         opts.tosURL = passedOptions.tosURL || undefined;
         opts.siteName = passedOptions.siteName || undefined;
         opts.siteLogo = passedOptions.siteLogo || undefined;
+        opts.backgroundColor = passedOptions.backgroundColor || undefined;
```
jaredhirsch, line 464: I'm not sure why we'd need to explicitly pass undefined if passedOptions.backgroundColor doesn't exist here. Is this just to please JSHint or something?

resources/static/dialog/js/modules/dialog.js
jaredhirsch, line 321: +1 to not prefixing this. it's an optional config parameter, not a new method, I'd just expose it for those who want to use it.

jaredhirsch:
@shane-tomlinson spoke too soon, wound up getting sucked into reading the code more closely :-)

Couple nits above, only one that really should be addressed is `validateBackgroundColor` throwing an uncaught Error if the color doesn't validate.

:beers: nice work shane and @djc

resources/static/dialog/js/modules/dialog.js
djc, line 321: FWIW, I also don't see much value in prefixing this given the response we've had to this on the mailing list. It also seems fairly innocuous; not that much complexity, very small API surface.

resources/static/include_js/_include.js
djc, line 464: I just followed along with all the other options... What you're saying really applies to most of them, doesn't it?

resources/static/dialog/js/modules/dialog.js
djc, line 322: I just did some more testing on this. In the current case, you get the error version of the dialog ("Please close this window and try again."). When you click "more info", you get to see "improper usage of API: Error: invalid backgroundColor: floozy". So actually, I think that's pretty good behavior.

We could go with console.logging() a warning message about the backgroundColor and proceeding by ignoring the color passed, but I don't know that that's better. It seems like pretty much all RPs will only pick a color once, while they're setting up RP functionality, and for that case it would be better to fail as obnoxiously as possible. Once the RP is setup and in actual use, invalid background colors should be very rare.

shane-tomlinson:
@6a68 - Thanks for the thorough review! validateBackgroundColor is within the massive try/catch block that does all the input parameter validation. 

The ongoing standard has been to throw an error in the validation function that is caught on line 379, causing a user facing error screen. The error screen displays "Improper usage of API: " + error. 

@lloyd and I discussed how to surface RP provided input validation errors last year and decided on an approach that would be immediately noticed by the RP developer. If there is an input validation error, stop the dialog from processing and surface the API error.

resources/static/include_js/_include.js
shane-tomlinson, line 464: @6a68 - that is a very good question, and we should immediately remove them. An RP could redefine undefined and break this code. Man, undefined is used like this all over the place!

shane-tomlinson, line 464: @6a68, @djc - https://github.com/mozilla/browserid/issues/3598

jaredhirsch:
Ah, ok. Thanks for the feedback, @djc and @shane-tomlinson. R+ :rainbow: :rainbow:

resources/static/include_js/_include.js
jaredhirsch, line 464: @shane-tomlinson I completely didn't catch that `undefined` could be explicitly reassigned here, it just looked funny to explicitly assign to it. Good catch!

PR #3590, opened 2013-06-27
Async snippet 2

jaredhirsch:
@shane-tomlinson mind taking a look? I could use some feedback.
#### Summary

This is a more well-rounded successor to the async snippet prototype I started last week (#3571). Seeking feedback.
#### Next Steps

If this looks good, I think the next step would be to create some RP integration docs (blog post and/or MDN page), and _maybe_ create a separate local testing example RP. Does this seem reasonable? Thoughts?
#### Background

We're in a bit of an awkward position relative to your typical third-party provider because we are shimming browser APIs, so our async snippet really does want to set global state. So, an async Persona snippet needs two parts: (1) something in the head that stubs out nav.id.\* before the script has loaded and sets up queues to collect function calls, and (2) the usual async snippet which builds a script tag, appends to the page, then unqueues waiting function calls when the script is ready.

This two-part thing seems complicated for a typical RP to have to integrate. I'm not sure.

The alternative I can think of is to put both parts of the snippet in the head, and wrap part (2) in a window.onload handler. This would also give additional flexibility, since the listener could be configured to listen for custom events--if the RP wanted to load some of their content scripts, and then fire nav.id.watch, for instance.

### DIFF
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -85,7 +85,7 @@ function onScriptLoaded(s) {
         // if navigator.id's reference has been replaced by the shim,
         // _qd will no longer be available.
         if (nav.id._qd) {
-          // bail till ready, back off slowly with no timeout because EDGE/2G
+          // bail till ready, back off slowly with no timeout
           return setTimeout(onParsed, parseWait+=50);
         }
         while (next = idq.shift()) {
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -47,6 +47,43 @@ pre {
 }
 
 </style>
+<script>
+navigator.id = navigator.id || {}; // important: does include.js bail if navigator.id, navigator.id.watch, or navigator.id.get is defined?
+navigator.id.q = [];
+
+function makeQueueable(funcName) {
+  var f = function() { navigator.id.q.push({func: funcName, args: arguments}) }
+  f.isQueueable = true; // so we can check it was overwritten
+  return f;
+}
+navigator.id.watch = makeQueueable('watch');
+navigator.id.get = makeQueueable('get');
+
+var scr = document.createElement('script');
+scr.async = true;
+scr.src = 'https://login.persona.org/include.js';
+scr.onload = scr.onreadystatechange = function() {
+  var rdyState = scr.readyState;
+  if (!rdyState || /loaded|complete/.test(rdyState)) {
+    var pollCount = 0;
+    function onParsed() {
+      if (navigator.id.watch.isQueueable || navigator.id.get.isQueueable) {
+        pollCount++;
+        if (pollCount > 200) { throw new Error('include.js failed to load after 10 sec') }
+        return setTimeout(onParsed, 50); // bail till ready, todo verify it's acutally needed to wait for parse time.
+      }
+      while (navigator.id.q.length) {
+        var next = navigator.id.q.shift();
+        navigator.id[next.func].apply(null, next.args);
+      }
+      scr.onload = null;
+      scr.onreadystatechange = null;
+    }
+  }
+};
+var entry = document.getElementsByTagName('script')[0];
+entry.parentNode.insertBefore(scr, entry);
+</script>
 </head>
 <body>
 <div class="title">
@@ -125,7 +162,6 @@ pre {
 </body>
 
 <script src="jquery-min.js"></script>
-<script src="https://login.persona.org/include.js"></script>
 <script>
 
 try {
--- a/example/rp/index.html
+++ b/example/rp/index.html
@@ -49,41 +49,17 @@ pre {
 </style>
 <script>
 navigator.id = navigator.id || {}; // important: does include.js bail if navigator.id, navigator.id.watch, or navigator.id.get is defined?
-navigator.id.q = [];
+navigator.idq = [];
 
 function makeQueueable(funcName) {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Dude, this is fun.

After _a lot_ of hand tweaking, I was able to get it down to 740 bytes after being run through uglifyjs

https://gist.github.com/shane-tomlinson/5884391

shane-tomlinson:
The code in that gist works in IE8 and FF. Haven't tested others yet.

shane-tomlinson:
@6a68 - My opinion is that, if this were to be released, both halves should be combined into a single snippet that can be included. RPs would include the snippet whenever they are ready to start using navigator.id functions. We want to make it as simple as possible for RPs to be able to get the performance wins out of an async load.

Does this get us any closer to the goal of allowing RPs to self host include.js? 

I believe it does, I am trying to imagine how a message proxy would work using a pre-defined API like what is occurring here.

jaredhirsch:
@shane-tomlinson Thanks for having a look! Nicely done with that gist, great stuff :-D

I was thinking the same thing--splitting the snippet is a lot of complexity to ask RPs to manage. The problem is that, in order to not ask RPs to use a different async API, we have to stub out nav.id.\* globals with a queue really early in the page load, so we do need to be in the head--on the other hand, we could put the embed code in a tiny locally-hosted file, which takes the sting out a bit. We can defer the script tag loading until onload, to absolutely, definitely not affect time-to-content. I can't think of a simpler way...

On the plus side, waiting to load include.js until 'onload' fires means we could also wait until some RP-specified custom event fires, like 'onContentLoaded' or something. We can make this easy-to-specify by having room in the snippet for a custom event variable, something like what disqus does:

``` html
<!-- disqus snippet as example of specifying config vars -->
<script type="text/javascript">
    /* * * CONFIGURATION VARIABLES: EDIT BEFORE PASTING INTO YOUR WEBPAGE * * */
    var disqus_shortname = '<example>'; // Required - Replace example with your forum shortname

    /* * * DON'T EDIT BELOW THIS LINE * * */
    (function() {
        var dsq = document.createElement('script'); dsq.type = 'text/javascript'; dsq.async = true;
```

jaredhirsch:
@shane-tomlinson About self-hosting: arguably the embed snippet is the self-hostable part. Especially if we allow async loading (maybe another config option they can set), I think it's reasonable to tell RPs not to self-host, maybe?

Again, it's odd doing third-party stuff in a browser shim context, because, someday, the APIs will just be in the page, and delayed-loading semantics won't make a ton of sense. But, I think this would help adoption for now.

seanmonstar:
It'd definitely be awesome to not load the entire include.js if the async
shim detects native support already exists.

jaredhirsch:
@shane-tomlinson @seanmonstar Updated. Now we put the whole embed blob in the doc head, wait until DOMContentLoaded (or IE's doc.onreadystatechange == 'complete'), and then do the async loading-and-dequeueing routine.

This also includes shane's fantastic changes from that gist he linked above, awesomesauce

example/rp/index.html
```diff
@@ -47,6 +47,73 @@
...
+function onScriptLoaded(s) {
+  return function() {
+    rdyState = s.readyState;
+    if (!rdyState || /loaded|complete/.test(rdyState)) {
+      (function onParsed() {
+        // if navigator.id's reference has been replaced by the shim,
+        // _qd will no longer be available.
+        if (nav.id._qd) {
+          // bail till ready, back off slowly with no timeout because EDGE/2G
+          return setTimeout(onParsed, parseWait+=50);
```
jaredhirsch, line 89: I also changed the polling logic to slowly back off forever, instead of giving up after waiting 10 seconds for parsing.

I'm hoping this will avoid a really spooky class of bugs, like the browser backgrounded the tab but it still runs setTimeouts periodically, or batches a thousand of them and runs them when the tab is foregrounded, etc.

jaredhirsch, line 89: I'll fix that comment to not refer to network slowness, derp.

jaredhirsch:
just screwing around, I found that assigning reused strings like 'DOMContentLoaded' to variables reduced the size a bit.

@shane-tomlinson how do you test this in IE8 locally? maybe I need to finally get VM networking ironed out on my lappy.

chilts:
Not sure if this is helpful, but I have seen the odd site (https://github.com/altryne/browserID-jQuery#how-to-use) firstly do a check to see if navigator.id already exists:

```
<script>navigator.id || document.write('<script src="https://login.persona.org/include.js"><\/script>')</script>
```

@6a68 Could we do a similar check at the very top of your script here and quit early if it's true. After that, we don't bother with detecting onload nor loading include.js. Does this look feasible or even ok to do it?

seanmonstar:
@chilts its there, at line 54.

chilts:
@seanmonstar ah sorry, thanks. :)

jaredhirsch:
@chilts this is pretty compact front-end JS, probably not the easiest place to jump in. (Also, doc.write is **_evil**_ ;-)

If you're interested in what's going on here, I'd suggest having a look at the Manning book Third-party JavaScript, which covers a lot of basic techniques for async loading.

jaredhirsch:
I'm not sure where this goes from here. Closing the PR because it's just lingering at this point. Might turn this into a self-updating self-hostable snippet some Friday, but going to leave it for now.

PR #3592, opened 2013-06-27
Fix returning from IdP verification on FirefoxOS

### LINKED ISSUE #3572: [b2g] intermittent 'improper usage of API: json parse' when signing in with primaries
I saw this 2 times on different sign in email but it doesn't repro the 2nd time I execute it.  Here are the vague steps:
1. set pref on phone to use login.anosrep.org
2. sign into beta.123done.org with a user@yahoo.com, logout
3. sign into beta.123done.org with a user@test-moz-idp.org, logout
4. sign into beta.123done.org with a user@gmail.com
   start these steps again over an over a few times.

actual: intermittently you'll see this:

shane-tomlinson:
@jedp, @jrgm - here is the other half of the FirefoxOS series of PRs!

Background: sessionStorage can be wiped on FirefoxOS when authenticating with the IdP. We need to store a momento with dialog bootstrapping info for when the user returns from the IdP.
- store primary IdP verification flow info in localStorage with a nonce
- Store the nonce in window.name for FirefoxOS, sessionStorage for IE8.
- When returning from the IdP, look for nonce, load info from localStorage.

fixes #3572

### DIFF
--- a/resources/static/common/js/storage.js
+++ b/resources/static/common/js/storage.js
@@ -28,7 +28,9 @@ BrowserID.Storage = (function() {
   "use strict";
 
   var ONE_DAY_IN_MS = (1000 * 60 * 60 * 24),
-      storage = BrowserID.getStorage();
+      IDP_INFO_LIFESPAN_MS = 1000 * 60 * 60,
+      storage = BrowserID.getStorage(),
+      win = window;
 
   // Set default values immediately so that IE8 localStorage synchronization
   // issues do not become a factor. See issue #2206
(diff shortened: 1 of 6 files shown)


resources/static/common/js/storage.js
```diff
@@ -442,6 +444,118 @@ BrowserID.Storage = (function() {
     storage.emailToUserID = JSON.stringify(allInfo);
   }
 
+  // generate a random string for use as a per-window nonce.
+  function generateRandomString(length) {
```
jedp: Would there be anything wrong with just using `String(Math.random())`?

shane-tomlinson: @jedp - nope.

resources/static/common/js/storage.js
```diff
@@ -442,6 +444,118 @@ BrowserID.Storage = (function() {
...
+    if (!info) {
+      info = nonce;
+      nonce = generateRandomString(8);
+    }
+
+    // The nonce is a bit tricky. Since sessionStorage is not reliable in
+    // FirefoxOS, we need an alternate method. window.name survives across page
+    // redirects in FirefoxOS, but not in IE8. In IE8, when the user redirects
+    // to their primary, window.name is reset to "__persona_dialog". In IE8,
+    // sessionStorage survives redirects to the Idp. So, we save to both.
```
jedp, line 500: So as a side effect of this, IE8 users would not be able to conduct two concurrent flows with a Primary.  (Which seems like a pretty unlikely occurrence.)

shane-tomlinson, line 500: @jedp - they still should be able to, the default is to look for the nonce in sessionStorage. Even FirefoxOS will first check for the nonce in sessionStorage, and if that is not there it will go to window.name.

So I believe concurrent windows should be fine.

jedp, line 500: (The good @shane-tomlinson related to me in IRC that he has tested this and it works - concurrent logins on IE8.  Awesome.)

jedp:
This tests brilliantly.  Using the 1.0.1 device and my latest 1.1 build, I can use `sendmypin` and `eyedee.me` to sign in to:
- navigator.mozId tests (1.0.1 and 1.1)
- navigator.mozId.get (1.1 - test suite not installed on 1.0.1)
- `primaries.123done.org`

So this is sure looking like @shane-tomlinson has done it again.  Beautiful work, Shane.

jedp:
I'm loving this.  Merging.

(Future generations, note that this PR is the other half of #3567.)

PR #3594, opened 2013-06-27
Fixes #3467 - Redirect old /signin, /signup, /forgot to the homepage

### LINKED ISSUE #3467: retired links /signin, /forgot, and /signup are now 404
Sorry, I missed this earlier in train-2013.04.26.

However, in the move to use the dialog on the main site, the routes for
/signin, /forgot (and the legacy redirect of /signup were removed).

The problem is that those links have either been bookmarked or indexed by
search engines, and I did see a few 404 for /signup when reviewing data during
the aws move. Better user experience would be to make these 301 to '/'.

You can choose to wontfix this if you want.

chilts:
I've modified the REDIRECTS and redirect() function so that we can
also specify whether the redirect is a 301 or a 302. Prior to this
we were only able to do the default 302 redirect.


jaredhirsch:
R+, redirects working on ephemeral instance and tests all passing locally. Ignoring scumbag travis, merging.

PR #3597, opened 2013-06-28
'native' is a reserved word.

### LINKED ISSUE #3595: android 2.3.6 default browser/dev - "Sign in with Persona" button never shown on dev.123done.org
When loading the dev RP locally, the dialog never gets past "Connecting to Persona..."

shane-tomlinson:
@6a68 - this is another instance of #3534. Could you give this a review?

Android 2.3.6 pukes when it used as the key to an object. When it is wrapped in quotes, everything is fine.

fixes #3595

### DIFF
--- a/resources/static/dialog/js/modules/verify_primary_user.js
+++ b/resources/static/dialog/js/modules/verify_primary_user.js
@@ -33,7 +33,7 @@ BrowserID.Modules.VerifyPrimaryUser = (function() {
       email: email,
       // native is used when the user returns from the primary to prevent
       // WinChan from establishing the postMessage channel.
-      native: win.document.location.hash === "#NATIVE"
+      'native': win.document.location.hash === "#NATIVE"
     });
 
     var url = helpers.toURL(auth_url, { email: email });


jaredhirsch:
@shane-tomlinson sure, gotta walk a dog then I'll work on reproing with my 2.2 handset. Android just jumped over IE8 to the top of my hit list :-P

resources/static/dialog/js/modules/verify_primary_user.js
```diff
@@ -33,7 +33,7 @@ BrowserID.Modules.VerifyPrimaryUser = (function() {
       email: email,
       // native is used when the user returns from the primary to prevent
       // WinChan from establishing the postMessage channel.
-      native: win.document.location.hash === "#NATIVE"
+      'native': win.document.location.hash === "#NATIVE"
```
jedp, line 36: I've also double-checked, and I don't see any other instances in `resources/` where `native` occurs as a bareword.  This does look like the only one that needs fixing.

shane-tomlinson:
@6a68 - how are you feeling about this one?

jaredhirsch:
@shane-tomlinson sorry for the delay! dev is down and there's no telling when it'll be back up. the code looks good, merging in.

PR #3604, opened 2013-06-29
Bugzilla 873241 spanish tos pp

ozten:
Spanish versions of TOS and PP.


ozten:
Green after triggering Travis rebuilds twice.

jaredhirsch:
This looks good to me (I thought Polish was included in this PR as well, I guess not). Merging in. @ozten if my test steps in #3605 missed anything, I'll reverify against dev.

PR #3608, opened 2013-07-01
Update tooltip copy with @ryanfeeley's suggestions.

### LINKED ISSUE #3607: Tooltip copy is too verbose
From https://id.etherpad.mozilla.org/frontend-tooltips

@ryanfeeley and I are pow-wowing to clean up the tooltip copy.

Link to #3154 
- Incorrect old password, password not updated => Old password incorrect
- New password is required. => New password required
- Old and new passwords are the same. => Passwords must be different
- Old password is required. => Old password required
- Password is required. => Password field required
- Password must be between 8 and 80 characters long. => Passwords must be 8–80 characters long
- Please choose an email address. => Select an email address
- The account cannot be verified with this username and password. => Password incorrect
- The email field is required. => Email field required
- The password field is required. => Password field required
- That address is already added to your account! => Already added!
- These passwords don't match! => Passwords don't match!
- This email address and password do not match. => Password incorrect
- This field must be an email address. => Must be an email address
- Verification password is required. => Verification password required
- We just sent an email to that address! If you really want to send another, wait a minute or two and try again. => Initial confirmation sent – wait a few minutes before requesting another

shane-tomlinson:
@ryanfeeley - this is what I have so far for the tooltip updates. How are these looking?

fixes #3607

### DIFF
--- a/resources/static/dialog/views/add_email.ejs
+++ b/resources/static/dialog/views/add_email.ejs
@@ -12,19 +12,19 @@
               <input id="newEmail" name="newEmail" type="email" autocapitalize="off" autocorrect="off" spellcheck="false" maxlength="254" <% if (typeof email !== "undefined") { %> value="<%= email %>" <% } %> placeholder="<%= gettext("email address") %>"/>
 
               <div id="email_format" class="tooltip" for="newEmail">
-                <%= gettext('This field must be an email address.') %>
+                <%= gettext('Must be an email address') %>
               </div>
 
               <div id="email_required" class="tooltip" for="newEmail">
-                <%= gettext('The email field is required.') %>
+                <%= gettext('Email field required') %>
               </div>
 
               <div id="could_not_add" class="tooltip" for="newEmail">
-                <%= gettext('We just sent an email to that address! If you really want to send another, wait a minute or two and try again.') %>
+                <%= gettext('Initial confirmation sent - wait a few minutes before requesting another') %>
               </div>
 
               <div id="already_own_address" class="tooltip" for="newEmail">
-                <%= gettext('That address is already added to your account!') %>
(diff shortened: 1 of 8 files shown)


resources/static/dialog/views/add_email.ejs
```diff
@@ -12,19 +12,19 @@
...
               </div>
 
               <div id="email_required" class="tooltip" for="newEmail">
-                <%= gettext('The email field is required.') %>
+                <%= gettext('Email field required') %>
               </div>
 
               <div id="could_not_add" class="tooltip" for="newEmail">
-                <%= gettext('We just sent an email to that address! If you really want to send another, wait a minute or two and try again.') %>
+                <%= gettext('Initial confirmation sent - wait a few minutes before requesting another') %>
```
shane-tomlinson: ndash this!

jaredhirsch:
@mathjazz Looks like we're wanting to make lots of our strings a bit more terse. Do we need to give l10n more lead time than usual when this lands?

_edit_: probably easier to read #3607 to see the list of affected strings

shane-tomlinson:
@6a68 - @djc had the foresight to mention in the channel that we should inform @mathjazz, so he is already aware. I am holding off on the merge until early in the next train so that we can get the new strings merged, extracted a bit earlier than usual, and off to the l10n crew.

mathjazz:
Yeah,

Thanks for the heads up.

It's very important that we give 2 weekends for localizers, so branching
the train today with the l10n deadline on 07/16 we should be ok:
http://personatra.in/#2013.07.03

-Matjaž

jaredhirsch:
@stomlinson thanks for the info, I was wondering if we wanted to get this merged before I branched.

PR #3613, opened 2013-07-02
Define `undefined` to explicitly be undefined in case an RP or IdP has accidentally redefined undefined.

### LINKED ISSUE #3598: include.js sets variables = undefined. undefined can be redefined to no longer be undefined.
See https://github.com/mozilla/browserid/pull/3589#discussion_r4931463. Nice catch @6a68.

shane-tomlinson:
@seanmonstar and @callahad - mind reviewing?

fixes #3598

### DIFF
--- a/resources/static/provisioning_api.js
+++ b/resources/static/provisioning_api.js
@@ -9,6 +9,9 @@
 (function() {
   "use strict";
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
looks good to me

Issue #3610, opened 2013-07-01
KPIs data does not contain whether an XHR response was non-200


shane-tomlinson:
This makes it difficult to tell if there are responses that are regularly failing and to track down why certain error screens may be shown.

Link to #3466

seanmonstar:
These lines will try to collect the response code if an error screen is shown from a failed request: https://github.com/mozilla/browserid/blob/dev/resources/static/common/js/modules/interaction_data.js#L55-L57

shane-tomlinson:
Thanks @seanmonstar - those only collect status > 399, why aren't we collecting the others? The context of that magic number is forgotten.

seanmonstar:
Thinking back, that function is called when an error screen is shown. It would include any network request error data if it was an request error. Less than 400 means the error happened client-side.

Ignore this, we should record include every response code as a separate piece of data, regardless of future error screens.

shane-tomlinson:
As @seanmonstar notes in https://github.com/mozilla/browserid/pull/3620/files#r5071514, we should also record aborted XHR requests to make sure they only happen after window.unload.

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #3620, opened 2013-07-03
Fix responses of aborted xhr requests when user closes dialog or browses away from RP.

### LINKED ISSUE #2423: Chrome / Safari call onlogout when onlogin navigates away
Our application uses the `watch()` API, and submits a POST form from `onlogin` and `onlogout`. This works fine on Firefox, but causes strange behavior on Chrome and Safari.

The call to `watch()` looks roughly as follows: (This is done in a `jQuery(function() { ... });`)

```
navigator.id.watch({
    loggedInEmail: email,
    onlogin: function(assertion) {
        var form = $('<form/>', {
            method: 'post',
            action: '/login'
        });
        $('<input/>', {
            type: 'hidden',
            name: 'assertion',
            value: assertion
        }).appendTo(form);
        form.appendTo(document.body).submit();
    },
    onlogout: function() {
        var form = $('<form/>', {
            method: 'post',
            action: '/logout'
        });
        form.appendTo(document.body).submit();
    }
});
```

The application has page links directly hooked up to `request()` and `logout()`.

In Chrome, when logging in, the popup opens normally, and I walk through the login process. After the popup closes, `onlogin` is called as usual. However, there's also a call to `onlogout` immediately after, before the page unloads. This does not happen when `form.submit()` is commented.

The net effect is that `onlogout` submits another form, cancelling the first form from `onlogin`. The browser now loads the `/logout` action, which simply redirects back to our login form. When the login form again calls `watch()`, it immediately gets the `onlogin` notification, which now completes successfully. (Ie., no further `onlogout` call in this situation, the `/login` action is called, and our app redirects to the dashboard.)

### LINKED ISSUE #3618: Errors can be logged if user closes dialog while XHR request is in flight.
Breaking this out from #3466. See https://id.etherpad.mozilla.org/kpi-issue-3466-data.

See also: #2423, #3610, #2560.

Copying in comment https://github.com/mozilla/browserid/issues/3466#issuecomment-20289563:
# 

There are a couple of interesting things going on.
- All but one data point from the sample data set is Chrome.
- screen.error.registration is on the event stream after window.unload even though it occurs before window.unload in time.
- Screen.error.registration should only occur if /wsapi/user_creation_status returns a non-200 error code.
- In the set that completes, /wsapi/authenticate_user is called immedaitely after /wsapi/user_creation_status.

I wonder if for a small subset of these XHR requests, the user closes the dialog to verify their email address while /wsapi/user_creation_status is in flight. If so, the XHR request would be aborted, causing Chrome to complete the with a status of 0 [see #2423  and http://stackoverflow.com/questions/11666421/jquery-ajax-request-abort-in-chrome]

This could be caused by:
1) /wsapi/user_creation_status is in flight when user triggers dialog close
2) /wsapi/user_creation_status is started after the user triggers dialog close
#1 could happen because XHR requests are not aborted on dialog close.
#2 could happen because the email verification poll is on a repeating setTimeout that does not take dialog close into consideration.

log with error screen:

```
[["xhr_complete.GET/wsapi/session_context", 4095, 525], ["window.channel_established", 4849], 
["screen.rp_info", 4853], ["screen.authenticate", 4858], ["user.can_interact", 4922], 
["xhr_complete.GET/wsapi/address_info", 20146, 2774], ["screen.set_password", 23639], 
["xhr_complete.POST/wsapi/stage_user", 42254, 1210], ["user.user_staged", 43467], 
["screen.check_registration", 43468], 
["xhr_complete.GET/wsapi/user_creation_status", 43503, 238, 3], 
["window.unload", 50001], ["screen.error.registration", 49986, 13]]
```

log of new user flow with successful completion:

```
"event_stream": [["window.dom_loading", 382], 
["xhr_complete.GET/wsapi/session_context", 1931, 63, 2], 
["window.channel_established", 2006], ["screen.rp_info", 2007], 
["screen.authenticate", 2008],
["user.can_interact", 2028], ["xhr_complete.GET/wsapi/address_info", 4112, 1768], 
["screen.set_password", 5953],
["xhr_complete.POST/wsapi/stage_user", 9480, 411], ["user.user_staged", 9893], 
["screen.check_registration", 9894], 
["xhr_complete.GET/wsapi/user_creation_status", 9911, 15, 5], 
["xhr_complete.POST/wsapi/authenticate_user", 22004, 404], 
["xhr_complete.GET/wsapi/list_emails", 22410, 13, 2], 
["user.user_confirmed", 22437], ["generate_assertion", 22437], 
["screen.generate_assertion", 22437], 
["xhr_complete.POST/wsapi/cert_key", 22462, 785], ["assertion_generated", 23258], 
["window.unload", 25044]],
```

### LINKED ISSUE #3619: onlogout not called if there is an error when calling /wsapi/session_context from the communication_iframe
This was found from a manual inspection of the code:

no failure callback is passed to network.cookiesEnabled in https://github.com/mozilla/browserid/blob/49eb5e6f0e6c2965a204558c334d1e5fec37a8cb/resources/static/communication_iframe/start.js#L48-L80

network.cookiesEnabled calls /wsapi/session_context. If /wsapi/session_context fails, the error is dropped on the ground.

Link to #2423, #2560, #3618

shane-tomlinson:
@seanmonstar and @6a68 - yo dudes, I could use some help/review on this!
- Add an abort method to the XHR object.
- Keep track of all outstanding XHR requests. In the window.onbeforeunload handler, abort all outstanding requests.
- errors no longer logged on the KPIs when the user closes the dialog
- onlogout no longer called in the communication_iframe for aborted XHR requests
- onlogout called in the communication_iframe for all other XHR errors.

changes to micrajax were pushed to origin repo - https://github.com/stomlinson/micrajax, https://github.com/stomlinson/micrajax/commit/2364453e387b914d556c53de8ab01188e067146f

fixes #2423
fixes #3618
fixes #3619

### DIFF
--- a/resources/static/common/js/xhr.js
+++ b/resources/static/common/js/xhr.js
@@ -10,7 +10,16 @@ BrowserID.XHR = (function() {
       BROWSERID_VERSION = BrowserID.CODE_VERSION,
       context,
       csrf_token,
-      time_until_delay;
+      time_until_delay,
+      outstandingRequests = {};
+
+  /**
+   * Abort any outstanding XHR requests if the user browses away or reloads.
+   * This prevents the onlogout message from being sent if XHR requests fail
+   * because the user browses away from the current page.
+   * See issue #2423
+   */
+  window.onbeforeunload = abortAll;
 
   function clearContext() {
     csrf_token = context = undefined;
@@ -47,6 +56,9 @@ BrowserID.XHR = (function() {
   }
 
   function xhrComplete(reqInfo) {
+    outstandingRequests[reqInfo.eventTime] = null;
+    delete outstandingRequests[reqInfo.eventTime];
+
     reqInfo.duration = new Date() - reqInfo.eventTime;
     mediator.publish("xhr_complete", reqInfo);
   }
@@ -71,6 +83,7 @@ BrowserID.XHR = (function() {
             delayTimeout = null;
           }
 
+          reqInfo.resp = resp;
           xhrComplete(reqInfo);
           if(options.defer_success) {
             _.defer(successCB.curry(resp, jqXHR, textResponse));
@@ -85,7 +98,25 @@ BrowserID.XHR = (function() {
             delayTimeout = null;
           }
 
+          reqInfo.resp = resp;
           xhrComplete(reqInfo);
+
+          /**
+           * XHR request was aborted. Don't do anything. A request can be
+           * aborted when:
+           * 1. user shuts the dialog while an XHR request is in flight
+           * 2. user browses away from an RP while an XHR request in the
+           *        communication_iframe is in flight.
+           * We cannot prevent these from happening, we can only abort XHR
+           * requests when it does happen and clean up afterwards.
+           * See issues #3618, #2423, #2560.
+           *
+           * We differentiate aborted requests from other errors because
+           * a status code of 0 can be returned for other XHR issues like
+           * illegal cross domain requests or when the user has no
+           * connectivity.
+           */
+          if (resp.statusText === "aborted") return;
           _.defer(xhrError.curry(errorCB, reqInfo, resp, jqXHR, textResponse));
         };
 
@@ -102,7 +133,9 @@ BrowserID.XHR = (function() {
     }
 
     mediator.publish("xhr_start", reqInfo);
-    transport.ajax(req);
+    var xhrObj = transport.ajax(req);
+    outstandingRequests[reqInfo.eventTime] = xhrObj;
+    return xhrObj;
   }
 
   function get(options) {
@@ -147,6 +180,11 @@ BrowserID.XHR = (function() {
     }, options.error);
   }
 
+  function abortAll() {
+    for (var eventTime in outstandingRequests) {
+      outstandingRequests[eventTime].abort();
+    }
+  }
 
   return {
     /**
@@ -191,7 +229,13 @@ BrowserID.XHR = (function() {
      * Clear the current context
      * @method clearContext
      */
-    clearContext: clearContext
+    clearContext: clearContext,
+
+    /**
+     * Abort all outstanding XHR requests
+     * @method abortAll
+     */
+    abortAll: abortAll
   };
 }());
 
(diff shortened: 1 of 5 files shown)


resources/static/common/js/xhr.js
```diff
@@ -10,7 +10,16 @@ BrowserID.XHR = (function() {
...
+      time_until_delay,
+      outstandingRequests = {};
+
+  /**
+   * Abort any outstanding XHR requests if the user browses away or reloads.
+   * This prevents the onlogout message from being sent if XHR requests fail
+   * because the user browses away from the current page.
+   * See issue #2423
+   */
+  window.onbeforeunload = abortAll;
```
seanmonstar: In case other places also wanted to listen, I'd use an actual event listener here. `window.addEventListener` and `window.attachEvent`.

resources/static/common/js/xhr.js
```diff
@@ -85,7 +98,25 @@ BrowserID.XHR = (function() {
             delayTimeout = null;
           }
 
+          reqInfo.resp = resp;
           xhrComplete(reqInfo);
+
+          /**
+           * XHR request was aborted. Don't do anything. A request can be
```
seanmonstar, line 113: Are we certain? We don't want to keep track of this in some way, perhaps in the event stream? I'd hate to have a case where requests can abort and we don't know about it.

shane-tomlinson, line 113: @seanmonstar - We can log them, but we don't really want to show an XHR error. 

A status code of 0 is returned for other XHR errors like illegal cross domain requests or when the user has no connectivity, for which we do show an XHR error.

resources/static/common/js/xhr.js
shane-tomlinson: Noted.

resources/static/common/js/xhr.js
shane-tomlinson, line 113: @seanmonstar - I was concerned about this too, we actually log that the request completes (see the first non-empty line above the comment), but we do not log the status of the return. I was worried about this last week and opened #3610, but did not want to expand this PR to tackle that too. Do you think I should?

seanmonstar, line 113: ah, i see, this still calls `xhrComplete` which publishes an event. So then sure, in #3610, it'd be nice to include that these requests were aborted. At least so we can be sure that aborted requests only happen after unload was triggered.

PR #3630, opened 2013-07-08
Refactor eliminate required email

### LINKED ISSUE #1873: remove requiredEmail legacy code.

shane-tomlinson:
@seanmonstar, @6a68, anybody else that wants to review. I don't mind. I want the remnants of this PITA feature out of the code. 907 total LOC removed. 

An ephemeral instance, https://no-required-email.personatest.org is available to test.

Note, no Selenium tests were changed in this PR, so I am somewhat confident that user facing functionality is the same. I did change a couple of selectors, which will require me to notify @stephend and co.

fixes #1873


callahad:
\o/!

jaredhirsch:
@shane-tomlinson I had to cherry-pick 9398590dc393f2cfd5661d749ebfb066b4b6e20c into this branch to fix the abide lockdown error.

resources/static/dialog/css/style.css
```diff
@@ -472,7 +472,7 @@ a.emphasize:active {
 /**
  * Start off hiding all of the state specific elements
  */
-.isReturning, .isStart, .isAddressInfo, .isMobile {
+.isReturning, .isStart, .isAddressInfo, .isMobile, .isRequiredEmail {
```
jaredhirsch: I'm confused by the changes in this file. It seems like you might have removed then re-added .isRequiredEmail, .requiredEmail, .isNotRequiredEmail classes?

jaredhirsch:
Grepping, I see some leftover requiredEmail references in the authenticate module. Should I look more closely, do those belong there?

resources/static/dialog/views/required_email.ejs
```diff
@@ -1,67 +0,0 @@
-<!-- This Source Code Form is subject to the terms of the Mozilla Public
-   - License, v. 2.0. If a copy of the MPL was not distributed with this
-   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-
-  <strong>
-    <% if(secondary_auth) { %>
-      <%= gettext("Sign in using") %>
-    <% } else { %>
-      <%= gettext("The site requested you sign in using") %>
```
jaredhirsch, line 9: Do we need to do anything to remove now-unused strings from the list generated by the l10n code?

Most of these strings seem reusable, but I'm guessing this one (for example) is no longer needed.

resources/static/dialog/views/required_email.ejs
```diff
@@ -1,67 +0,0 @@
...
-              <label for="email"><%= gettext("Email") %></label>
-              <input id="required_email" type="email" value="<%= email %>" disabled />
-              <div id="could_not_add" class="tooltip" for="required_email">
-                <%= gettext('Initial confirmation sent &ndash; wait a few minutes before requesting another') %>
-              </div>
-          </li>
-
-          <% if (verify && primary) { %>
-              <li>
-                  <%= gettext("You must sign in with your email provider to verify ownership of this address. This window will be redirected to") %>
```
jaredhirsch, line 24: Probably this string as well

resources/static/test/cases/dialog/js/misc/state.js
```diff
@@ -485,16 +485,16 @@
...
 
 
-  asyncTest("email_chosen with verified secondary email, user must authenticate - call doAuthenticateWithRequiredEmail", function() {
+  asyncTest("email_chosen with verified secondary email, user must authenticate - call doAuthenticate", function() {
     storage.addEmail(TEST_EMAIL);
 
     xhr.setContextInfo("auth_level", "assertion");
 
-    mediator.publish("start", { privacyPolicy: "priv.html", termsOfService: "tos.html" });
+    mediator.publish("start");
```
jaredhirsch, line 493: I'm not sure I understand why this change was made.

resources/views/dialog.ejs
```diff
@@ -19,14 +19,19 @@
                      DOMContentLoaded or else the Firefox password manager does
                      not fill in passwords -->
                 <div id="authentication_form" class="form_section">
+
                     <div class="isDesktop">
-                      <h2 class="isStart persona_intro deemphasize">
+                      <h2 class="isRequiredEmail">
```
jaredhirsch: I have to admit I'm finding the isRequiredEmail and isNotRequiredEmail classnames confusing. Maybe it's just me?

jaredhirsch:
I'm not sure if this code path needs to be changed, but the comment is definitely out of date: [user.js#L226-229](https://github.com/shane-tomlinson/browserid/blob/36f7fb339b4d95596bdf349218540152d3b8f2ca/resources/static/common/js/user.js#L226-L229)

jaredhirsch:
@shane-tomlinson All tests passing, and I've read through the changes, but I don't think I quite understand what's going on, as shown by my comments above. I don't know how much of that is inexperience on my part--take my feedback with a grain of salt.

resources/static/dialog/views/required_email.ejs
shane-tomlinson, line 9: I asked @mathjazz about what happens to these unused strings a few months back, he said they just sit peacefully at the bottom of the extracted files, are marked as "unused" and nothing really needs to happen.

resources/static/test/cases/dialog/js/misc/state.js
shane-tomlinson, line 493: The RP TOS/PP used to be displayed by the authenticate module. Since we moved this behavior to the rp_info module, the extra parameters to start were not actually exercising any code paths. I figured it would be a good time to clean it up.

resources/views/dialog.ejs
shane-tomlinson: Not at all, this is the part I am most hesitant about as well. I could use some help here.

The reason why I am doing all this juggling with class names is for Firefox's password manager to work, the password field must be in the DOM before DOMContentLoaded. The password field cannot be removed and re-added. So, instead of using a dynamically written template, I always have the password field in the DOM and use classes to show/hide various elements.

"requiredEmail" is used if a user is currently authenticated to the "assertion" level, then they chose an email address that is vouched for by the fallback identity provider. At this point, the user must enter in their Persona password. I wasn't quite sure what to call this.

So, isRequiredEmail is shown to users who are following the above path, isNotRequiredEmail is shown to all others.

When the user is viewing the "isRequiredEmail" flow, the "forgot your password?" text is shown above the password input box instead of next to the "sign in" button. Next to the sign in button, a "cancel" button is shown which allows the user to go back to the email picker.

requiredEmail is a bad name. Is there a better name? Is there a better approach? I'd like the intention to be crystal clear before merging, I don't want to remove 900 LOC with the intention of clarifying things only to have them still be super confusing.

shane-tomlinson:
Thanks for the feedback @6a68. Your reviews are always super useful, your points are valid, and I could use some help. Can you read my responses and we can come up with a path forward that makes the intentions much more obvious? No sense in prematurely merging if the result is still not clear.

resources/static/test/cases/dialog/js/misc/state.js
jaredhirsch, line 493: makes sense :-)

resources/views/dialog.ejs
jaredhirsch: Hmm. Maybe "isRequiredEmail" could be "passwordRequired", "needsPassword", "notAuthed", something like that. "isNotRequiredEmail" is a bit tougher--because it seems to stand for all the other cases.

Maybe, "passwordNeeded" and "passwordNotNeeded"? This is a tough one!

resources/static/dialog/views/required_email.ejs
jaredhirsch, line 9: Awesome.

jaredhirsch:
@shane-tomlinson I think this is a solid PR, my uncertainty is more about not fully understanding the client-side codebase.

I'm R+ if/when you're satisfied with it. (I took a swing at tweaking the class names, let me know if you have other specific feedback you're after.)

PR #3640, opened 2013-07-10
Fix the lockdown error with i18n-abide

### LINKED ISSUE #3639: npm lockdown fails on i18n-abide
package.json and lockdown.json are looking for i18n-abide@0.0.8beta4, `npm search i18n-abide` lists the version as 0.0.8-beta4

```
i18n-abide@0.0.8beta4 node_modules/i18n-abide
....
....
....
LOCKDOWN ERRORS:
    package version i18n-abide@0.0.8-beta4 not in lockdown.json!
```

shane-tomlinson:
@6a68, @seanmonstar, @callahad, @lloyd, @ozten - can someone review this?

fixes #3639

### DIFF
--- a/lockdown.json
+++ b/lockdown.json
@@ -165,7 +165,7 @@
     "0.1.1": "d9d82735a5f85f950761ac3909ba9485ec0af4f1"
   },
   "i18n-abide": {
-    "0.0.8beta4": "b106bdaa156f90652e04b75ef6a1079c271a1b45"
+    "0.0.8-beta4": "b106bdaa156f90652e04b75ef6a1079c271a1b45"
   },
   "iconv-lite": {
     "0.2.7": "45be2390d27af4b7613aac4ee4d957e3f4cbdb54"
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "etagify": "0.0.2",
     "express": "2.5.0",
     "gobbledygook": "0.0.3",
-    "i18n-abide": "0.0.8beta4",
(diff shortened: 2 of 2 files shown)


callahad:
Huh. Odd, but the hash doesn't change. Testing locally real quick, but expect r+

callahad:
From my reading of [the docs](https://npmjs.org/doc/json.html#version), this seems like an upstream bug:

> A numeric item separated from the main three-number version by a hyphen will be interpreted as a "build" number, and will increase the version. But, if the tag is not a number separated by a hyphen, then it's treated as a pre-release tag, and is less than the version without a tag. So, `0.1.2-7` > `0.1.2-7-beta` > `0.1.2-6` > `0.1.2` > `0.1.2beta`

This also breaks the gmail bridge (and presumably anything that's relying on i18n-abide and lockdown, right now)

callahad:
Looks good to me. Tests pass. r+. I'll leave the green button up to Shane.

It looks like the hyphen is here to stay, as it was standardized in semver 2.0, which npm just updated to. Perhaps the best thing to do right now is the thing that would work on both sides of that divide: release i18n-abide 0.0.8 without a beta label ;)

seanmonstar:
@callahad isn't it being sub 1.0 enough to indicate its beta?

callahad:
That's really ozten's call :)

ozten:
I've published 0.0.8.

callahad:
Thanks!

PR #3641, opened 2013-07-10
Hide the "checking with your email provider" message whenever the check is complete.

### LINKED ISSUE #3638: "checking with email provider" state from "add email" never goes away
STR

1) Open dialog
2) Create an account with a secondary address
3) Re-open dialog
4) Add new secondary address to same account
5) DO NOT VERIFY, instead, re-open dialog and try to re-add address from #4.
6) Tooltip is displayed, "checking with your email provider" never goes away.

Expected: "checking with your email provider" goes away.

screen shot 2013-07-10 at 15 24 51

shane-tomlinson:
@6a68 or @seanmonstar - HALP PLEAZE!

fixes #3638

### DIFF
--- a/resources/static/dialog/js/modules/add_email.js
+++ b/resources/static/dialog/js/modules/add_email.js
@@ -43,11 +43,16 @@ BrowserID.Modules.AddEmail = (function() {
 
     if (email) {
       showHint("addressInfo");
+
       dom.setAttr(EMAIL_SELECTOR, 'disabled', 'disabled');
       dom.addClass(BODY_SELECTOR, SUBMIT_DISABLED_CLASS);
+
       dialogHelpers.addEmail.call(self, email, function removeHint(status) {
+        hideHint("addressInfo");
+
+        // !status means there was a problem adding the email. Let the user
+        // modify the email address and try to re-add.
         if (!status) {
-          hideHint("addressInfo");
           dom.removeAttr(EMAIL_SELECTOR, 'disabled');
           dom.removeClass(BODY_SELECTOR, SUBMIT_DISABLED_CLASS);
         }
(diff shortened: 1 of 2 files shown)


seanmonstar:
I'd r+, but looking at travis, omgwtfbbq, did the abortAll blow something up?

shane-tomlinson:
@seanmonstar - I am trying to figure that out, I saw the issue manifest locally yesterday. I restarted my server and was unable to reproduce again.

jaredhirsch:
looking now.

jaredhirsch:
@shane-tomlinson The front-end xhr test error is really weird. I repro'd twice ([failure msg](http://6a68.pastebin.mozilla.org/2624283)), then actually started the server, verified the test passed in real browsers, and couldn't repro using phantom after that.

Tried wiping out node_modules, reinstalling, re-running tests, couldn't repro.
Tried uninstalling/reinstalling phantomjs, still couldn't repro.
Tried installing a different version of nodejs over existing, switched from 0.8.24 to 0.10.13, deleted node_modules, reinstalled, re-ran tests. Repro'd -- but only on the first two runs.
Went back from 0.10.13 to 0.8.24, reinstalled node_modules, etc., couldn't repro.

I don't think it's a known issue with phantom, as [according to its changelog](https://github.com/ariya/phantomjs/blob/master/ChangeLog#L36), xhr abort support was added in 1.9.0, which I've used across all tests (that's what homebrew installs).

Anyway, aside from this test weirdness, the code looks fine to me. If the test errors recur, we can address those in a separate bug. Merging in.

shane-tomlinson:
Thanks for the testing @6a68. The XHR failure is ...  yeah... I have taken care of it in #3655. The problem is a subtle timing bug due to global state being kept across test runs causing the two tests to interfere with each other.

jaredhirsch:
Ah, nice. Refactoring pays off right away, then :-)

PR #3654, opened 2013-07-11
Extend internal API to embed arbitrary payload data in assertion (WebRTC)

ryanseys:
The internal API is extended to allow the signing of arbitrary data (`payload`) to be included in the assertion. This is done to support the signing of a WebRTC PeerConnection **fingerprint** (the `payload`), used in the PeerConnection Identity module of WebRTC as specified at: http://dev.w3.org/2011/webrtc/editor/webrtc.html#identity

The assertions given from internal API will be included in the `offer` and `answer` SDP objects returned from `RTCPeerConnection.createOffer()` and `RTCPeerConnection.createAnswer()`

These changes are necessary to ensure BrowserID supports the WebRTC Identity specification to be landed in future web browsers.

/cc @warner @shane-tomlinson

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -1343,9 +1343,11 @@ BrowserID.User = (function() {
      * @param {string} audience - Audience to use for the assertion.
      * @param {function} [onComplete] - Called with assertion, null otw.
      * @param {function} [onFailure] - Called on error.
+     * @param {object} [payload] - optional extra payload for assertion
      */
-    getAssertion: function(email, audience, onComplete, onFailure) {
+    getAssertion: function(email, audience, onComplete, onFailure, payload) {
       var storedID = storage.getEmail(email, issuer),
+          payload = payload || {},
           assertion,
           self=this;
 
@@ -1365,7 +1367,8 @@ BrowserID.User = (function() {
             // raise "script has become unresponsive" errors.
             setTimeout(function() {
               jwcrypto.assertion.sign(
-                {}, {audience: audience, expiresAt: expirationDate},
+                payload,
+                {audience: audience, expiresAt: expirationDate},
                 sk,
                 function(err, signedAssertion) {
                   assertion = jwcrypto.cert.bundle([idInfo.cert], signedAssertion);
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -118,9 +118,13 @@
       // first, check the required email field, if that is not specified, go
       // check if an email is associated with this site. If that is not
       // available, there is not enough information to continue.
-      var requiredEmail = options.requiredEmail || storage.site.get(origin, "email");
+      // if options.origin is available, check login from that url instead of origin for assertion
+      var login_origin = options.origin || origin;
+      var requiredEmail = options.requiredEmail
+                       || storage.site.get(login_origin, "logged_in"); //get the logged in user (required for silent get)
       if(requiredEmail) {
-        getSilent(origin, requiredEmail, complete);
+        // always get assertion for origin passed, never options.origin
+        getSilent(origin, requiredEmail, complete, options.payload);
       }
       else {
         complete();
@@ -156,7 +160,7 @@
   /*
    * Get an assertion without user interaction - internal use
    */
-  function getSilent(origin, email, callback) {
+  function getSilent(origin, email, callback, payload) {
     function complete(assertion) {
       assertion = assertionObjectToString(assertion);
       callback && callback(assertion || null);
@@ -169,7 +173,7 @@
       if(authenticated) {
         user.getAssertion(email, user.getOrigin(), function(assertion) {
           complete(assertion || null);
-        }, complete.curry(null));
+        }, complete.curry(null), payload);
       }
       else {
         complete(null);
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -51,7 +51,7 @@ var types = {
   origin: function(x) {
     /* origin regex
     /^                          // beginning
-    (?:https?|app):\/\/         // starts with http://, https://, or app:// (b2g desktop)
+    (?:https?|app|rtcweb):\/\/  // starts with http://, https://, or app:// (b2g desktop), or rtcweb:// (RTC auth)
     (?=.{1,254}(?::|$))         // hostname must be within 1-254 characters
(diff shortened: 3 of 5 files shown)


resources/static/common/js/user.js
```diff
@@ -1343,9 +1343,11 @@ BrowserID.User = (function() {
      * @param {string} audience - Audience to use for the assertion.
      * @param {function} [onComplete] - Called with assertion, null otw.
      * @param {function} [onFailure] - Called on error.
+     * @param {object} [payload] - optional extra payload for assertion
      */
-    getAssertion: function(email, audience, onComplete, onFailure) {
+    getAssertion: function(email, audience, onComplete, onFailure, payload) {
```
jedp, line 1348: Yes, we know this is grody; but we can't change the sequence of the arguments in this signature.  FirefoxOS and other things depend on the status quo.

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -118,9 +118,13 @@
       // first, check the required email field, if that is not specified, go
       // check if an email is associated with this site. If that is not
       // available, there is not enough information to continue.
-      var requiredEmail = options.requiredEmail || storage.site.get(origin, "email");
+      // if options.origin is available, check login from that url instead of origin for assertion
+      var login_origin = options.origin || origin;
```
jedp, line 122: Here we check if the user is logged into the current origin.  If so, we will get a silent assertion below for the _preferred_ origin specified in `options`.  In the case of rtc, this will be `rtcweb://peerconnection`.  The thinking here is that if Alice has visited ekrpokr.com, a site that facilitates the establishment of rtc video chats, then Alice wants an rtc peer connection verified by the same identity she used to sign into ekrpokr.com.  Therefore the user agent switches the audience in the assertion.  Additionally, the one-time unique peer-connection key fingerprint is embedded in the assertion (`payload.extra`), which prevents replay attacks.)

resources/static/common/js/user.js
seanmonstar, line 1348: You can adjust the order to be nicer. This is inside `user.js`, which isn't called directly by FxOS. You'd just need to find every call, which is quick grep away.

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -156,7 +160,7 @@
   /*
    * Get an assertion without user interaction - internal use
    */
-  function getSilent(origin, email, callback) {
+  function getSilent(origin, email, callback, payload) {
```
seanmonstar, line 163: This looks to be the function that is called by FxOS, but the order can still be adjusted:

``` js
function getSilent(origin, email, payload, callback) {
  //payload is optional
  if (typeof payload === 'function') {
    callback = payload;
    payload = undefined;
  }
  // ...
}
```

shane-tomlinson, line 163: +1 to non-grodyness!
or:

```
function getSilent(origin, email, payload, callback) {
  //payload is optional
  if (typeof payload === 'function') {
    callback = payload;
    var undefined;    // just in case someone accidentally redefines undefined.
    payload = undefined;
  }
```

resources/static/dialog/js/misc/internal_api.js
shane-tomlinson, line 122: @jedp - couldn't you just pass options.origin as origin? I'm not following this logic fully.

FWIW, in #3630, I am ripping out requiredEmail. See https://github.com/mozilla/browserid/pull/3630/files#L3L121.

resources/static/common/js/user.js
shane-tomlinson, line 1348: @jedp - I'd rather do what @seanmonstar suggests here.

These are the places where it matters:

```
./misc/helpers.js:    user.getAssertion(email, user.getOrigin(), function(assert) {
./misc/internal_api.js:        user.getAssertion(email, user.getOrigin(), function(assertion) {
./user.js:              User.getAssertion(email, "https://login.persona.org", function(assertion) {
./user.js:                  User.getAssertion(email, audience, onComplete, onFailure);
./user.js:              User.getAssertion(email, audience, onComplete, onFailure);
./user.js:          User.getAssertion(loggedInEmail, origin,
```

Or, easier yet, you could follow the approach in:
https://github.com/mozilla/browserid/pull/3654/files#r5158662

resources/static/common/js/user.js
```diff
@@ -1365,7 +1367,8 @@ BrowserID.User = (function() {
             // raise "script has become unresponsive" errors.
             setTimeout(function() {
               jwcrypto.assertion.sign(
-                {}, {audience: audience, expiresAt: expirationDate},
+                payload,
```
shane-tomlinson, line 1370: Is there a way to test this change?

shane-tomlinson:
It looks like this introduced some JSHint errors. https://travis-ci.org/mozilla/browserid/jobs/8987942

resources/static/dialog/js/misc/internal_api.js
jedp, line 163: yes let's do that. thanks.

jedp:
Thanks @shane-tomlinson and @seanmonstar for your helpful coments.  We'll revise.

resources/static/dialog/js/misc/internal_api.js
ryanseys, line 122: in our special case the request is made with the following param:
`origin` = `"rtcweb://peerconnection"`
`options.origin` = `"http://example.com"` where `example.com` is the site we are logged into.

This change allows us to verify identity with origin site, while generating an assertion for the made up origin `rtcweb://peerconnection`, a url that is specifically crafted to not represent something that looks like a real website, making it unspoofable.

jaredhirsch:
closing. please open a new PR when the revised code is ready. :beers:

jaredhirsch:
I believe #3952 unblocked this

ryanseys:
Unblocked or solved this? Can we now embed signed arbitrary data in the assertion or am I misunderstanding what #3952 accomplishes?

jaredhirsch:
Well, the original description refers to embedding the extra spinach inside SDP objects. If that work's part of this bug, then, 'unblocked'. If it's just getting the spinach in there, then, 'solved'. :beers: either way

ryanseys:
solved! :wink:

ryanseys:
Correction: unblocked.

PR #3655, opened 2013-07-12
Issue 3647 async requests

### LINKED ISSUE #3647: intermittent front end test failure for /slow_request
```
common/js/xhr: abortAll aborts outstanding requests, triggers xhr_complete (2, 3, 5)
   - failed
Expected:
"/slow_request"
Result:
"/wsapi/session_context"
Diff:
"/slow_request" "/wsapi/session_context"
Source:
    at http://127.0.0.1:10002/test/cases/common/js/xhr.js:183

   - Died on test #5: Unmatched start: 1 - 
{ "message": "Unmatched start: 1", "line": 406, "sourceId": 197014816, 
"sourceURL": "http://127.0.0.1:10002/test/qunit/qunit.js", 
"stack": "Error: Unmatched start: 1 
at http://127.0.0.1:10002/test/qunit/qunit.js:406
at http://127.0.0.1:10002/test/cases/common/js/xhr.js:185 
at fire (http://127.0.0.1:10002/common/js/lib/hub.js:50) 
at http://127.0.0.1:10002/common/js/javascript-extensions.js:19 
at xhrComplete (http://127.0.0.1:10002/common/js/xhr.js:71) 
at http://127.0.0.1:10002/common/js/xhr.js:110 
at http://127.0.0.1:10002/test/mocks/xhr.js:306 
at abortAll (http://127.0.0.1:10002/common/js/xhr.js:193) 
at http://127.0.0.1:10002/test/cases/common/js/xhr.js:194 
at http://127.0.0.1:10002/test/qunit/qunit.js:102 
at http://127.0.0.1:10002/test/qunit/qunit.js:232 
at process (http://127.0.0.1:10002/test/qunit/qunit.js:875) 
at http://127.0.0.1:10002/test/qunit/qunit.js:418",
```

shane-tomlinson:
@callahad - Yo dude, this is one step to "getting there"! Mind reviewing?

So, basically, xhr.js was a heap of global state. We know, we don't like global state. So, I made it a real true blue module with no global state. This allowed me to simplify a lot of things.

An ephemeral instance is set up to run the selenium tests: https://async-requests.personatest.org

fixes #3647


jaredhirsch:
stealing, looking now.

jaredhirsch:
Unit tests passing for me locally.

resources/static/common/js/modules/xhr.js
```diff
@@ -0,0 +1,319 @@
...
+
+  function init(config) {
+    /*jshint validthis: true*/
+    config = config || {};
+
+    var self = this;
+
+    self.outstandingRequests = {};
+    self.transport = config.transport || XHRTransport;
+    self.time_until_delay = config.time_until_delay || 0;
```
jaredhirsch: Later on (line 149) we do a falsy check of time_until_delay, so I don't see why we bother setting it to zero here.

resources/static/common/js/modules/xhr.js
```diff
@@ -0,0 +1,319 @@
...
+    /*jshint validthis: true*/
+    var outstandingRequests = this.outstandingRequests;
+    for (var eventTime in outstandingRequests) {
+      outstandingRequests[eventTime].xhr.abort();
+      outstandingRequests[eventTime] = null;
+      delete outstandingRequests[eventTime];
+    }
+  }
+
+  function getRequest(options) {
```
jaredhirsch: This is a confusing name--it seems like it might mean, "perform an xhr GET request". What it really means is, `getRequestInfo` or something, the old version called this `reqInfo`.

As far as I can tell, this function is only called once, at line 138 of this file.

How about we just inline it, then we can rename the awkward `makeRequest` to `request` or `XHRRequest` or something.

resources/static/common/js/modules/xhr.js
```diff
@@ -0,0 +1,319 @@
...
+     */
+    post: post,
+
+    /**
+     * Get the session context
+     * @method getContext
+     * @param {function} complete
+     * @param {function} error
+     */
+    getContext: getContext,
```
jaredhirsch, line 68: I really like this name change. Now it's clearer that xhr gets the context (XHR#getContext), while network/user let you do things with the context (Network#withContext, User#withContext).

resources/static/common/js/modules/xhr.js
```diff
@@ -0,0 +1,319 @@
...
+    // We defer the responses because otherwise the transport eats any
+    // exceptions that are thrown in the response handlers and it
+    // becomes very difficult to debug.
+    if (request.defer_success) {
+      setTimeout(function() {
+        complete(request.success, resp, xhrObj, textResponse);
+      }, 0);
+    }
+    else {
+      request.success(resp, xhrObj, textResponse);
```
jaredhirsch, line 262: Why do we call complete() inside the setTimeout (line 258) instead of calling req.success directly? This differs from old xhr.js, I suspect I'm missing something fairly obvious and should take a break.

jaredhirsch:
Taking a break. So far, XHR module looks good, modulo a few minor nits.

Going to look at collaborator classes, try to run Selenium tests again (grrr timeouts), and very likely R+ by end of day.

resources/static/common/js/network.js
```diff
@@ -44,12 +51,12 @@ BrowserID.Network = (function() {
         // If cookies are disabled, some browsers throw an exception. Ignore
         // this, the backend will see that cookies are disabled.
       }
-      xhr.withContext(cb, onFailure);
+      xhr.getContext(cb, onFailure);
```
jaredhirsch, line 54: Here's a question. Why does XHR hold the context? That bit of state seems like it should probably belong to Network or User.

I know that xhr is a transport, but should it be a stateful transport? Does it need to be? I'm not sure on the answer.

resources/static/common/js/network.js
```diff
@@ -95,8 +102,17 @@ BrowserID.Network = (function() {
...
+      config = config || {};
+
       // Any time the context info changes, we want to know about it.
       mediator.subscribe('context_info', onContextChange);
+      if (config.xhr) {
+        xhr = config.xhr;
+      } else {
+        xhr = XHR.create();
+        // A request that takes more than 10 seconds is considered delayed.
+        xhr.init({ time_until_delay: 10 * 1000 });
```
jaredhirsch: Is there some way we can pull out magic numbers like this? Maybe hide them inside `BrowserID.config` or somethin?

resources/static/dialog/js/start.js
```diff
@@ -10,11 +10,8 @@
       modules = bid.Modules,
       user = bid.User,
       network = bid.Network,
-      mediator = bid.Mediator,
-      xhr = bid.XHR;
+      mediator = bid.Mediator;
 
-  // A request that takes more than 10 seconds is considered delayed.
-  xhr.init({ time_until_delay: 10 * 1000 });
```
jaredhirsch, line 17: :rainbow: Really nice that we're hiding the xhr transport inside Network. I dig it.

resources/views/test.ejs
```diff
@@ -115,7 +115,7 @@
     <script src="/common/js/javascript-extensions.js"></script>
     <script src="/common/js/lib/bidbundle.js"></script>
     <script src="/build/code_version.js"></script>
-    <script src="https://testmob.org/include.js"></script>
+    <!--script src="https://testmob.org/include.js"></script-->
```
jaredhirsch: Was this intentional? Seems like it might have been a stray debugging commit.

jaredhirsch:
R+

Some questions and minor complaints, but looking good overall.

Feel free to merge after addressing my comments above.

:ship: it :-)

resources/static/common/js/modules/xhr.js
shane-tomlinson: I think that is left over from a bygone era. Will remove.

resources/static/common/js/network.js
shane-tomlinson, line 54: That's an excellent question, one that I completely forgot about. I want session to only be held in user.js. Thanks for pointing that out! I forgot about my own intentions. I'll see how much effort it involves and we can decide whether it makes sense to pile on here or whether it is better in its own PR.

resources/static/common/js/modules/xhr.js
shane-tomlinson: good suggestion. I'll rename to `getRequestInfo` and turn `makeRequest` into `request`

shane-tomlinson: I take that back, I inlined it.

resources/static/common/js/modules/xhr.js
shane-tomlinson, line 262: The old version used _.defer on https://github.com/mozilla/browserid/pull/3655/files#L3L97

resources/views/test.ejs
shane-tomlinson: This was not intentional, excellent catch.

resources/static/common/js/network.js
shane-tomlinson, line 54: Ahha, the context about context has come back to me. The reason I store the context here is so that we can ensure that we call post with a CSRF token without requiring a higher level caller to worry about calling session_context and passing the token from user.js->network.js->xhr.js. 

This seems like a heavy handed way of going about things that leads to too much context lying around. 
Perhaps a better way is to have network.js take care of automatically fetching session_context before any posts, then set the CSRF token in xhr.js? If a post is attempted without a CSRF token, xhr throws an exception? I agree, context should not be here.

resources/static/common/js/network.js
shane-tomlinson: We have some generic config in browserid.js, I'll add a new constant there.

shane-tomlinson:
@6a68 - Thanks for your great feedback. I have updated the PR with your suggestions, and am working on a new PR that moves getContext out of xhr.js and into network.js. The work started to grow in size larger than I was comfortable with for this PR.

PR #3656, opened 2013-07-12
Inline those tooltips!

### LINKED ISSUE #3154: convert the tooltips to inline indicators and highlights

shane-tomlinson:
I don't really care who reviews!

@ryanfeeley has given the tentative OK, wants to do a bit more testing.

A test instance, inline-tooltips.personatest.org can be used to test.

fixes #3154

### DIFF
--- a/resources/static/common/js/tooltip.js
+++ b/resources/static/common/js/tooltip.js
@@ -6,15 +6,11 @@ BrowserID.Tooltip = (function() {
   "use strict";
 
   var ANIMATION_TIME = 250,
-      TOOLTIP_MIN_DISPLAY = 2000,
-      TOOLTIP_OFFSET_TOP_PX = 5,
-      TOOLTIP_OFFSET_LEFT_PX = 0,
-      TOOLTIP_MARGIN_TO_SCREEN_EDGE_PX = 20,
-      READ_WPM = 200,
+      animationTime = ANIMATION_TIME,
       bid = BrowserID,
       dom = bid.DOM,
       renderer = bid.Renderer,
-      onlyAttachedTooltip;
+      visibleTooltip;
 
   // This file is made up of two parts, the first part is the Tooltip generic
   // object type, the second half is the BrowserID.Tooltip singleton logic.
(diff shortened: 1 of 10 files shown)


jaredhirsch:
reviewing now.

resources/static/common/js/tooltip.js
```diff
@@ -26,147 +22,141 @@ BrowserID.Tooltip = (function() {
...
 
-      return displayTimeMS;
-    }
-  });
+      dom.stopAnimations(tooltipEl);
+      if (animationTime) {
+        // If tooltipEl does not exist in the DOM, jQuery never calls the
+        // slideDown done function. Avoid the blowup by using a setTimeout.
+        dom.slideDown(tooltipEl, animationTime);
+        if (done) setTimeout(done, animationTime);
```
jaredhirsch: I think we should abstract out this jQuery quirk inside dom-jquery, here and in slideUp as well.

Aside from this case, it would actually be super helpful to have a little DOM helper to check if an element is in the DOM.

I can't think of a good name, but basically, `dom.isInDOM = function(el) { !!jQuery(el).closest('html').length }`. [This jsperf test](http://jsperf.com/jquery-element-in-dom/2) shows it's a performant way of checking whether a node is in the doc or not, found via [this SO thread](http://stackoverflow.com/questions/3086068/how-do-i-check-whether-a-jquery-element-is-in-the-dom).

resources/static/common/js/tooltip.js
```diff
@@ -26,147 +22,141 @@ BrowserID.Tooltip = (function() {
...
+      }
+      else {
+        dom.show(tooltipEl);
+        if (done) done();
```
jaredhirsch: It'd be cool to pass `done` as an optional second parameter to `dom.show` as well, then this whole stanza could just be:

```
animationTime ? dom.slideDown(tooltipEl, animationTime, done) : dom.show(tooltipEl, done)
```

Alternatively, if animationTime is zero, maybe `slideDown` should just toggle show. I dunno if jQuery does that for you automatically.

resources/static/common/js/tooltip.js
```diff
@@ -26,147 +22,141 @@ BrowserID.Tooltip = (function() {
...
-    }
+    // Only one tooltip can be shown at a time, see issue #1615
+    removeVisibleTooltip();
 
-    return {
-      text: tooltipText,
-      anchor: tooltipAnchor,
-      done: complete
-    };
+    visibleTooltip = new Tooltip();
```
jaredhirsch: I'm a bit confused by this. `showTooltip` creates a new Tooltip, but `removeVisibleTooltip` never actually deletes the old one, it just calls `stop`, which in turn hides it.

Why are we creating a new Tooltip here, and not inside a hidden/protected constructor function? Is this some other way of doing singletons that I've forgotten/don't know about? Or is the code convention just to only call showTooltip once per page?

resources/static/common/js/tooltip.js
shane-tomlinson: Both of those are fantastic ideas, I feel like what I am doing here is super gross.

resources/static/common/js/tooltip.js
shane-tomlinson: The major difference is slideDown is async, even with a period of 0, whereas show is sync. I'll go with the first option.

resources/static/common/js/tooltip.js
shane-tomlinson: showTooltip can be called at any time, with the intention is that only one tooltip is ever displayed at any given instant. So yeah, BrowserID.Tooltip is a public facing singleton, with the showTooltip method that shows a Tooltip instance. 

The differentiation between BrowserID.Tooltip (public facing singleton) and Tooltip, the instantiated object is historical. All of it used to be taken care of by the singleton, but that became too complex for my brain to handle.

showTooltip will remove any of the old tooltips from the DOM, and then overwrites the visibleTooltip element with a reference to a new Tooltip. In removeVisibleTooltip, I probably should delete the reference upon completion.

I think my own historical context is preventing me from fully seeing an alternative, cleaner, solution. Any help that allows me to make changes to tooltip.js only is very much appreciated!

resources/static/common/js/tooltip.js
shane-tomlinson: I think the dom.isInDOM function would also be very useful in the unit tests.

jaredhirsch:
R+, feel free to merge once you've addressed my concerns above.

Unit tests look good, works smoothly on IE8 and Android 2.2.

Great work, mang. :beers:

jaredhirsch:
R+, brimming with win.

PR #3657, opened 2013-07-12
Issues 3616 3617 focus

shane-tomlinson:
@6a68 - yo man, at one point you reviewed a PR that I did to help focus the correct element, esp with IE8. I have taken another stab at this, fixing problems on several fronts, but want to get your feedback!

### DIFF
--- a/automation-tests/lib/wd-extensions.js
+++ b/automation-tests/lib/wd-extensions.js
@@ -253,7 +253,8 @@ wd.prototype.wgetAttribute = function(opts, attribute, cb) {
 // wait for an element to be displayed, then type in it.
 wd.prototype.wtype = function(opts, text, cb) {
   var self = this;
-  self.waitForDisplayed(opts, function(err, el) {
+  // IE8 can't type into an element unless it is already focused.
+  self.wclick(opts, function(err, el) {
     if (err) return cb(err);
     self.type(el, text, cb);
   });
--- a/resources/static/common/js/modules/page_module.js
+++ b/resources/static/common/js/modules/page_module.js
@@ -56,18 +56,46 @@ BrowserID.Modules.PageModule = (function() {
       self.click(CANCEL_DIALOG_SELECTOR, cancelDialog);
     },
 
-    renderForm: function(template, data) {
+    renderForm: function(template, data, done) {
       var self=this;
 
       dom.removeClass("body", "rptospp");
 
       screens.form.show(template, data);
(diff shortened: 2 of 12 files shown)


automation-tests/lib/wd-extensions.js
```diff
@@ -253,7 +253,8 @@ wd.prototype.wgetAttribute = function(opts, attribute, cb) {
 // wait for an element to be displayed, then type in it.
 wd.prototype.wtype = function(opts, text, cb) {
   var self = this;
-  self.waitForDisplayed(opts, function(err, el) {
+  // IE8 can't type into an element unless it is already focused.
+  self.wclick(opts, function(err, el) {
```
shane-tomlinson, line 257: Fixes the Selenium tests to a large extent for when you have to click into an element before typing. This is primarily needed for when the user sets their password. I could update the tests to do that, or I could make it generic, wasn't sure which way to go.

resources/static/common/js/modules/page_module.js
```diff
@@ -56,18 +56,46 @@ BrowserID.Modules.PageModule = (function() {
...
+     * Pre-defined element search order:
+     *   1) visible element with data-autofocus=true
+     *   2) visible, enabled input element
+     *   3) visible button
+     * @method focus
+     * @param {string || element} [el] - element to focus
+     * @param {function} [done] - called with boolean, true if element
+     *   was found to focus, false otw.
+     */
+    focus: function(el, done) {
```
shane-tomlinson, line 89: Here is the crux of the change. First, look for an element with the data-autofocus attribute. If it is true, try to focus. To avoid breaking backwards compatibility, look for the first enabled, visible input element. Fall back to the first visible button.

resources/static/dialog/js/modules/pick_email.js
```diff
@@ -20,15 +20,7 @@ BrowserID.Modules.PickEmail = (function() {
 
   function pickEmailState(event) {
     /*jshint validthis: true*/
-    var self=this,
-        // focus the first radio button by default.
-        focusSelector = "input[type=radio]:eq(0)";
-
-    // unless a radio button is checked, then focus it.
-    if (dom.getElements("input[type=radio]:checked").length) {
```
shane-tomlinson, line 28: I add the "data-autofocus" attribute to the correct email address to focus.

shane-tomlinson:
Closing, this was supposed to be opened against my repo for feedback.

PR #3665, opened 2013-07-16
Aws enable locale

ozten:
A new script to be used from inside an awsbox deployment, to enable certain locales.

Example:

```
ssh app@foobar.personatest.org
$ emacs t.json
```

Contents of t.json:

```
{
    "supported_languages": ["es", "pl", "sq"]
}

./scripts/aws_enable_locales.js t.json
```

Script takes a few minutes, but will do the following:
- Writes new locales to `/home/app/config.json`
- Re-runs `compress`
- Does `forever restartall`

### DIFF
--- a/scripts/aws_enable_locales.js
+++ b/scripts/aws_enable_locales.js
@@ -1,0 +1,117 @@
+#!/usr/bin/env node
+
+const exec = require('child_process').exec,
+fs = require('fs'),
+optimist = require('optimist'),
+path = require('path');
+
+
+const argv = optimist
+  .usage('Usage: $0 path/to/anyfile_with_locales.json\n\n' +
+    ' -a, -c, -u are for development of this script outside of aws and ' +
+    'can be ignored.\n\n' +
+    'During development, create a sample config.json and ' +
+    'include var_path in it.')
+  .options('a', {
+    alias: 'aws-config',
+    "default": '/home/app/code/config/aws.json',
+    describe: 'Path to aws.json config file'
+  })
+  .options('c', {
+    alias: 'compress',
+    "default": '/home/app/code/scripts/compress',
+    describe: 'Path to compress script'
+  })
+  .options('u', {
+    alias: 'update-config',
+    "default": '/home/app/config.json',
+    describe: 'The config file we should update. ' +
+      'We will overrite any existing locale setting'
+  })
+  .argv;
+
+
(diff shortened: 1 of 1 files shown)


ozten:
@jedp I'm always forgetting the steps, so I put together this script. What do you think? Too special purpose for awsboxen?

shane-tomlinson:
@ozten - I for one have wanted something like this for a while, esp when debugging locale specific problems.

shane-tomlinson:
@ozten - can you rebase this against the current dev? I keep seeing test failures due to a bug that we fixed yesterday.

ozten:
Re bah bah bah bah base!

jaredhirsch:
@ozten @jedp how can I help get this over the finish line?

ozten:
How that the train is cut... Feel free to merge. Not much risk, as it is a dev tool to be executed from within an awsbox VM.

scripts/aws_enable_locales.js
```diff
@@ -0,0 +1,117 @@
...
+fs = require('fs'),
+optimist = require('optimist'),
+path = require('path');
+
+
+const argv = optimist
+  .usage('Usage: $0 path/to/anyfile_with_locales.json\n\n' +
+    ' -a, -c, -u are for development of this script outside of aws and ' +
+    'can be ignored.\n\n' +
+    'During development, create a sample config.json and ' +
```
shane-tomlinson, line 13: This comment could be clarified, perhaps with an example. I'm not sure what it means.

shane-tomlinson:
@ozten - what about a simplification where you can just say which locales you want to enable, from the command line, without adding an extra .json file?

ozten:
Ya, that works.

Using a JSON file is nice when you want to enable all locales. You can just use `config/l10n-prod.json`. But if you want to enable 1 or 2 locales, then the locals as a list would be nicer.

shane-tomlinson:
@ozten - for sure, I do sometimes enable all locales at once for testing. More often than not though, I enable one or two to see if things are working as I expect.

jaredhirsch:
@ozten Hmm, this seems to have stalled. Since it's just a dev tool, are we good to merge, or do you want to add a commit to address @shane-tomlinson's feature request?

jaredhirsch:
chatted with @shane-tomlinson in IRC, he's fine with merging as-is to get this over the finish line. So...

Issue #3666, opened 2013-07-16
context has seeped into xhr.js, too many mixed concerns.


shane-tomlinson:

shane-tomlinson:
Closed wiht #3667

jaredhirsch:
:beers:

PR #3667, opened 2013-07-16
Issue 3666 context out of xhr

shane-tomlinson:
@6a68 - yo dude, I am trying to pull the context stuff out of xhr and move it into network.js. It has gone well, but I could use some collaboration time.

Specifically, https://github.com/shane-tomlinson/browserid/compare/issue_3666_context_out_of_xhr?expand=1#L1R53.  I want to change this to:

`get({`

but doing so has repercussions that I did not foresee. The old code requested session_context and returned results would be returned immediately instead of being deferred. This is the _only_ request that acted like this. I want to make this request consistent with all other requests and have its completion deferred. 

Some tests depend on this behavior, the tests are in `user.js` and `state.js`. You can see in the first commit the changes I had to make to the tests. I was a bit uncomfortable doing this without some input to see if the scope of the PR is expanding too far.

This is a continuation of mozilla/browserid#3655 to address mozilla/browserid#3666.

### DIFF
--- a/resources/static/common/js/modules/xhr.js
+++ b/resources/static/common/js/modules/xhr.js
@@ -36,6 +36,13 @@ BrowserID.Modules.XHR = (function() {
     /**
      * Low level request
      * @method request
+     * @param {object} config
+     *   {string} config.url
+     *   {string} [config.type]
+     *   {object} [config.data]
+     *   {function} [config.success]
+     *   {function} [config.error]
+     * @returns {object} xhr object
      */
     request: makeRequest,
 
@@ -44,8 +51,10 @@ BrowserID.Modules.XHR = (function() {
      * @method get
      * @param {object} config
      *   {string} config.url
+     *   {object} [config.data]
      *   {function} [config.success]
      *   {function} [config.error]
+     * @returns {object} xhr object
      */
     get: get,
 
@@ -54,8 +63,11 @@ BrowserID.Modules.XHR = (function() {
      * @method post
      * @param {object} config
      *   {string} config.url
+     *   {object} config.data
+     *   {string} config.data.csrf
      *   {function} [config.success]
      *   {function} [config.error]
+     * @returns {object} xhr object
      */
     post: post,
 
@@ -163,7 +175,7 @@ BrowserID.Modules.XHR = (function() {
       type: "GET",
       defer_success: true
     });
-    this.request(options);
+    return this.request(options);
   }
 
   function post(options) {
@@ -185,7 +197,7 @@ BrowserID.Modules.XHR = (function() {
                   "missing csrf token from POST request");
     }
 
-    this.request(options);
+    return this.request(options);
   }
 
   function abortAll() {
--- a/resources/static/common/js/modules/xhr.js
+++ b/resources/static/common/js/modules/xhr.js
@@ -77,6 +77,7 @@ BrowserID.Modules.XHR = (function() {
     var self = this;
 
     self.outstandingRequests = {};
+    self.outstandingTimers = [];
     self.transport = config.transport || XHRTransport;
     self.time_until_delay = config.time_until_delay;
 
@@ -102,7 +103,12 @@ BrowserID.Modules.XHR = (function() {
 
   function stop() {
     /*jshint validthis: true*/
-    var abortAllListener = this.abortAllListener;
+    var self = this;
+
+    // abort all requests
+    self.abortAll();
+
+    var abortAllListener = self.abortAllListener;
     if (window.removeEventListener) {
       window.removeEventListener("beforeunload", abortAllListener, false);
     } else if (window.detachEvent) {
@@ -129,9 +135,11 @@ BrowserID.Modules.XHR = (function() {
      * the user informing them of the slowness.
      */
     if (self.time_until_delay) {
-      request.slowRequestTimeout = setTimeout(function() {
+      var timer = request.slowRequestTimeout = setTimeout(function() {
+        removeTimer.call(self, timer);
         onXHRResponseDelayed.call(self, request);
       }, self.time_until_delay);
+      addTimer.call(self, timer);
     }
 
     self.publish("xhr_start", request);
@@ -182,11 +190,18 @@ BrowserID.Modules.XHR = (function() {
 
   function abortAll() {
     /*jshint validthis: true*/
-    var outstandingRequests = this.outstandingRequests;
+    var self = this;
+    var outstandingRequests = self.outstandingRequests;
     for (var eventTime in outstandingRequests) {
       outstandingRequests[eventTime].xhr.abort();
       outstandingRequests[eventTime] = null;
       delete outstandingRequests[eventTime];
+    }
+
+    // abort any outstanding response timers
+    var timer;
+    while (timer = self.outstandingTimers.pop()) {
+      clearTimeout(timer);
     }
   }
 
@@ -221,9 +236,11 @@ BrowserID.Modules.XHR = (function() {
     // exceptions that are thrown in the response handlers and it
     // becomes very difficult to debug.
     if (request.defer_success) {
-      setTimeout(function() {
+      var timer = setTimeout(function() {
+        removeTimer.call(self, timer);
         complete(request.success, resp, xhrObj, textResponse);
       }, 0);
+      addTimer.call(self, timer);
     }
     else {
       request.success(resp, xhrObj, textResponse);
@@ -253,11 +270,14 @@ BrowserID.Modules.XHR = (function() {
     if (xhrObj && xhrObj.statusText === "aborted") return;
 
     // See note in success about why we defer responses
-    setTimeout(function() {
+    var timer = setTimeout(function() {
+      removeTimer.call(self, timer);
       var errorInfo = getErrorInfo(request, textStatus, errorThrown);
       self.publish("xhr_error", errorInfo);
       complete(request.error, errorInfo);
     }, 0);
+
+    addTimer.call(self, timer);
   }
 
 
@@ -272,13 +292,30 @@ BrowserID.Modules.XHR = (function() {
     outstandingRequests[request.eventTime] = null;
     delete outstandingRequests[request.eventTime];
 
-    if (request.slowRequestTimeout) {
-      clearTimeout(request.slowRequestTimeout);
+    var timer = request.slowRequestTimeout;
+    if (timer) {
+      removeTimer.call(this, timer);
       request.slowRequestTimeout = null;
     }
 
     request.duration = new Date().getTime() - request.eventTime;
     this.publish("xhr_complete", request);
+  }
+
+  // Timers keep track of outstanding setTimeouts that must be cleared on
+  // abortAll or when the module stops to prevent interference between tests.
+  function removeTimer(timer) {
+    /*jshint validthis: true*/
+    var self=this;
+    var index = _.indexOf(self.outstandingTimers, timer);
+    if (index > -1) {
+      self.outstandingTimers.splice(index, 1);
+    }
+  }
+
+  function addTimer(timer) {
+    /*jshint validthis: true*/
+    this.outstandingTimers.push(timer);
   }
 
 
--- a/resources/static/common/js/modules/xhr.js
+++ b/resources/static/common/js/modules/xhr.js
@@ -60,20 +60,6 @@ BrowserID.Modules.XHR = (function() {
     post: post,
 
     /**
-     * Get the session context
-     * @method getContext
-     * @param {function} complete
-     * @param {function} error
-     */
-    getContext: getContext,
-
-    /**
-     * Clear the current context
-     * @method clearContext
-     */
-    clearContext: clearContext,
-
-    /**
      * Abort all outstanding XHR requests
      * @method abortAll
      */
@@ -83,11 +69,6 @@ BrowserID.Modules.XHR = (function() {
   sc = XHR.sc;
 
   return XHR;
-
-  function clearContext() {
-    /*jshint validthis: true*/
-    this.context = undefined;
-  }
 
   function init(config) {
     /*jshint validthis: true*/
@@ -136,6 +117,7 @@ BrowserID.Modules.XHR = (function() {
     var self = this;
 
     var request = getRequestInfo(options);
+
     // The request obj must be added to list of outstanding requests in
     // case request is synchronous. This makes sure all housekeeping is kept in
     // order.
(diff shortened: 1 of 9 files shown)


resources/static/common/js/network.js
```diff
@@ -8,57 +8,70 @@ BrowserID.Network = (function() {
   var bid = BrowserID,
       complete = bid.Helpers.complete,
       context,
-      server_time,
```
shane-tomlinson, line 11: These three fields are also stored in the context variable. too many sources of truth hurt my brain, so I removed the duplicates.

resources/static/common/js/network.js
```diff
@@ -8,57 +8,70 @@ BrowserID.Network = (function() {
...
       XHR = bid.Modules.XHR,
       xhr,
       // XXX get this out of here!
       storage = bid.Storage;
 
-  function post() {
-    xhr.post.apply(xhr, [].slice.call(arguments, 0));
-  }
+  function post(options) {
+    if (!context) {
```
shane-tomlinson, line 19: Do the context check here in network.js instead of xhr.js, this makes more sense to me.

resources/static/common/js/network.js
```diff
@@ -8,57 +8,70 @@ BrowserID.Network = (function() {
...
+      document.cookie = "can_set_cookies=1";
+    } catch(e) {
+      // If cookies are disabled, some browsers throw an exception. Ignore
+      // this, the backend will see that cookies are disabled.
     }
+
+    // use the low level request because get defers success which jacks up
+    // a bunch of tests that expect this to return immediately. Who coded this?
+    // (me, I shrug embarrassed).
+    xhr.request({
```
shane-tomlinson: This is the line in question, I want to change this to be just `get`, but tests erroneously expect this call to immediately return when using a mock low level XHR object.

resources/static/common/js/network.js
```diff
@@ -8,57 +8,70 @@ BrowserID.Network = (function() {
...
+      local_time: new Date().getTime()
+    });
+
+    mediator.publish("context_info", newContext);
   }
 
   function clearContext() {
-    if (xhr) xhr.clearContext();
     var undef;
-    context = server_time = undef;
```
shane-tomlinson, line 61: We were incorrectly not clearing the multiple sources of truth, and that hurt my head and prompted me to only use context.

resources/static/common/js/network.js
```diff
@@ -168,7 +179,13 @@ BrowserID.Network = (function() {
     withContext: withContext,
 
     setContext: function(field, value) {
```
shane-tomlinson, line 178: setContext is used in testing, when used in testing, the entire context object can be set at once.

resources/static/test/cases/common/js/network.js
```diff
@@ -553,18 +553,22 @@
   asyncTest("cookiesEnabled with onComplete exception thrown - should not call onComplete a second time", function() {
     // Since we are manually throwing an exception, it must be caught
     // below.
-    try {
-      network.cookiesEnabled(function(status) {
-        // if there is a problem, this callback will be called a second time
-        // with a false status.
-        equal(status, true, "cookies are enabled, correct status");
-        start();
+    network.withContext(function() {
```
shane-tomlinson, line 556: This test was just plain out not testing what it needed to test. You can't check for whether an exception was thrown from within the catch block, it might never be called.

resources/static/common/js/network.js
```diff
@@ -8,57 +8,70 @@ BrowserID.Network = (function() {
...
+      error: onFailure
+    });
+  }
+
+  function setContext(newContext) {
+    context = _.extend({}, newContext, {
+      local_time: new Date().getTime()
+    });
+
+    mediator.publish("context_info", newContext);
```
jaredhirsch: should mediator publish context here, rather than newContext? do we want local_time in the published signal?

resources/static/common/js/network.js
jaredhirsch, line 178: I'm trying to imagine if this could accidentally bite us, say if you didn't pass a second argument. Is it possible to check if we're in test mode before allowing this? /me squirms

resources/static/test/cases/common/js/network.js
jaredhirsch, line 556: yup, good catch

resources/static/common/js/network.js
shane-tomlinson, line 178: mmm.. What if instead of surfacing a testing variable, if we check if arguments.length === 1 and if the first argument is an object?

resources/static/common/js/modules/xhr.js
```diff
@@ -147,9 +135,11 @@ BrowserID.Modules.XHR = (function() {
      * the user informing them of the slowness.
      */
     if (self.time_until_delay) {
-      request.slowRequestTimeout = setTimeout(function() {
+      var timer = request.slowRequestTimeout = setTimeout(function() {
```
shane-tomlinson, line 150: All setTimeout timers are kept track of and aborted when the module stops to prevent cross-test contamination. This became a problem when the call to /wsapi/session_context started using "defer_success" like all the other requests.

resources/static/common/js/network.js
```diff
@@ -6,59 +6,69 @@ BrowserID.Network = (function() {
...
-      // to ensure the cookie is set for that single call.
-      try {
-        document.cookie = "can_set_cookies=1";
-      } catch(e) {
-        // If cookies are disabled, some browsers throw an exception. Ignore
-        // this, the backend will see that cookies are disabled.
-      }
-      xhr.getContext(cb, onFailure);
+  function withContext(done, onFailure) {
+    if (typeof context !== "undefined") return complete(done, context);
```
jaredhirsch: inside `post` we do the falsy check, `if (!context)`. Frankly, I prefer a looser check unless you explicitly want to distinguish between falsy types, but our jshint settings seem to imply that we prefer super-strict checks under all circumstances. So either let's loosen this one or tighten the check in `post`.

resources/static/common/js/network.js
```diff
@@ -6,59 +6,69 @@ BrowserID.Network = (function() {
...
+    });
+
+    mediator.publish("context_info", context);
   }
 
   function clearContext() {
-    if (xhr) xhr.clearContext();
     var undef;
-    context = server_time = undef;
+    context = undef;
```
jaredhirsch, line 71: I like this undef thing (unset things by assigning to an uninitialized variable instead of 'undefined' itself), but it's at odds with straight-up setting to `undefined` elsewhere in this PR. Probably we should be consistent about how we unset things--I prefer this approach because it's more minifiable, but your call on which to use.

resources/static/common/js/network.js
```diff
@@ -168,7 +176,15 @@ BrowserID.Network = (function() {
...
     setContext: function(field, value) {
-      if (context) context[field] = value;
+      if (arguments.length === 1) {
+        if (!helpers.isObject(field)) throw new Error("invalid context");
+
+        // an object was passed in for the context. Used for testing.
+        setContext(field);
+      }
+      else {
+        if (context) context[field] = value;
```
jaredhirsch, line 186: again--here's a truthy check. if we want to be uniformly strict, I suppose `isObject(context)` or `typeof context !== 'undefined'` should be used instead.

resources/static/common/js/network.js
```diff
@@ -579,13 +595,14 @@ BrowserID.Network = (function() {
...
+        var server_time = context.server_time;
         try {
           if (!server_time) throw new Error("can't get server time!");
-          var offset = (new Date()).getTime() - server_time.local;
-          complete(onComplete, new Date(offset + server_time.remote));
+          var offset = (new Date()).getTime() - context.local_time;
+          complete(onComplete, new Date(offset + server_time));
         } catch(e) {
-          complete(onFailure, e.toString());
+          complete(onFailure, String(e));
```
jaredhirsch, line 605: I'm not sure I follow this change. Why's this a better way? I'm not familiar with it.

resources/static/common/js/network.js
jaredhirsch, line 178: sure, let's do it.

resources/static/common/js/network.js
```diff
@@ -600,12 +617,13 @@ BrowserID.Network = (function() {
...
     domainKeyCreationTime: function(onComplete, onFailure) {
-      withContext(function() {
+      withContext(function(context) {
+        var domain_key_creation_time = context.domain_key_creation_time;
         try {
           if (!domain_key_creation_time) throw new Error("can't get domain key creation time!");
           complete(onComplete, new Date(domain_key_creation_time));
         } catch(e) {
-          complete(onFailure, e.toString());
+          complete(onFailure, String(e));
```
jaredhirsch, line 626: ditto

resources/static/dialog/js/misc/state.js
```diff
@@ -261,6 +261,7 @@ BrowserID.State = (function() {
         startAction(false, "doStageUser", info);
 // TODO         startAction(false, "doStageResetPassword", info); ???
```
jaredhirsch: too much scope creep to address this TODO while we're here? your call, I'm not sure.

resources/static/dialog/js/misc/state.js
```diff
@@ -261,6 +261,7 @@ BrowserID.State = (function() {
...
       }
+      complete(info.complete);
```
jaredhirsch, line 263: This seems significant. Can you explain a bit why this has been added?

resources/static/test/cases/common/js/helpers.js
```diff
@@ -162,4 +162,35 @@
...
+    equal(helpers.isFunction([]), false);
+  });
+
+  test("isArray", function() {
+    var someFunc = function() {};
+    equal(helpers.isArray([]), true);
+    equal(helpers.isArray(new Array()), true);
+    equal(helpers.isArray({}), false);
+    equal(helpers.isArray(""), false);
+  });
```
jaredhirsch, line 196: one of the beauties of our `isArray` implementation is that it correctly distinguishes arrays from non-Array things with length, like a DOM NodeList or an `arguments` object. Maybe add tests for those here? `arguments` should be easy, the DOM bit might be more trouble than it's worth.

_edit: derp._

resources/static/test/cases/dialog/js/misc/state.js
```diff
@@ -55,12 +55,19 @@
...
+        testActionStarted("doSetPassword");
+        mediator.publish("password_set", {
+          complete: function() {
+            equal(actions.info[stagingAction].email, TEST_EMAIL,
+                      "correct email sent to " + stagingAction);
+            start();
+          }
+        });
+      }
+    });
```
jaredhirsch, line 70: hmm, this seems connected to the mysterious `complete` call inside `state.js`. Is this another thing that accidentally got pushed, or just something you've fixed while you were in there?

resources/static/common/js/modules/xhr.js
```diff
@@ -169,59 +159,50 @@ BrowserID.Modules.XHR = (function() {
 
   function get(options) {
     /*jshint validthis: true*/
-    var req = _.extend(options, {
+    options = _.extend(options, {
       type: "GET",
       defer_success: true
     });
-    this.request(req);
+    this.request(options);
```
jaredhirsch: `makeRequest` returns the request object, but `get` and `post` swallow it and return undefined.

Would it be useful to return the request upward to callers? Or is it a leaky abstraction asking for trouble? I'm not sure. Either way, we probably should be consistent, as `makeRequest` is part of the exported module API along with `get` and `post`. Your call.

Actually, it looks like we publish the request in the `xhr_start`, `xhr_complete`, and `xhr_delay` signals, so I think `get` and `post` should return `this.request()`? Yay uniformity.

resources/static/common/js/modules/xhr.js
jaredhirsch, line 150: <3

jaredhirsch:
@shane-tomlinson Welp, I've had quite a nit-picking fiesta at your expense. I'm kinda fried at the moment, can't provide good high-level input. But xhr.js is looking really sleek at this point, :beers: all around.

Tests passing for me locally, and Travis seems happy too.

R+ modulo assorted minor complaints. Feel free to merge in after addressing (or dismissing, as the case may be)

shane-tomlinson:
Thanks for the awesome review @6a68, time for some updates!

resources/static/common/js/network.js
shane-tomlinson, line 605: @seanmonstar convinced me that String(e) was the way to go when printing exceptions for the reasons outlined in http://stackoverflow.com/questions/3945202/whats-the-difference-between-stringvalue-vs-value-tostring - mainly for when we start passing exceptions around. Even though we are guaranteed to have `e` here, I changed these to be consistent with what we have done in other areas.

resources/static/common/js/modules/xhr.js
shane-tomlinson: Good point. I originally did not return the XHR object because `post` was taking care of fetching the CSRF token, which resulted in 2 requests. That is now taken care of in network, meaning `post` can safely return the object.

resources/static/common/js/network.js
shane-tomlinson: Updating & testing.

resources/static/dialog/js/misc/state.js
shane-tomlinson: I'm not sure _what_ that line is supposed to do. I think it can now safely be removed, as we aren't seeing any issues with new FFOS accounts.

resources/static/test/cases/dialog/js/misc/state.js
shane-tomlinson, line 70: Dude, you made me dig deep to figure out why I made this change.

Here's what happened:

The mocked call to /wsapi/session_context used to return synchronously, we made the change to return asynchronously like all the other calls. This caused this test, which previously depended on the synchronous behavior in the `mediator.publish(stagingMessage..` to return immediately. This is now no longer the case, and and our assumption was broken. So, now, depend on the asynchronous behavior and the fact that the complete message on R60 will be called when the staging has completed. The password_set bit was changed to asynchronous behavior just to keep consistency, it is not strictly necessary, but I like consistent behavior so I don't have to remember special cases.

Phew. I can relax again.

resources/static/test/cases/common/js/helpers.js
shane-tomlinson, line 196: Added a test for arguments.

PR #3670, opened 2013-07-16
Remove transition code added in #3398 (877281c).

jaredhirsch:
@shane-tomlinson, to be merged after the train is branched.

Poking at storage-related stuff for #1500, I noticed some transition code more than a month old, should be ripe for plucking out of the codebase.

All tests passing. Let me know if I've missed anything. This is essentially a revert of 877281c5.


jaredhirsch:
@shane-tomlinson thoughts on this one?

PR #3685, opened 2013-07-19
Remove reference to locale/compile-json.sh.

ozten:
Conservative fix. See Issue#3683 for another suggestion.

Fixes Issue#3683

### DIFF
--- a/scripts/browserid.spec
+++ b/scripts/browserid.spec
@@ -23,7 +23,7 @@ persona server & web home for persona.org
 %build
 npm install
 export PATH=$PWD/node_modules/.bin:$PATH
-./locale/compile-json.sh locale/ resources/static/i18n/
+./node_modules/.bin/compile-json locale/ resources/static/i18n/
 echo "$GIT_REVISION" > resources/static/ver.txt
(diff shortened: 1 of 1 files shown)


callahad:
I had to `mkdir resources/static/i18n`, but after that things appeared to compile fine. The resulting RPM includes compiled json/js files in that directory.

So, r+

ozten:
Thanks for the review!

The old script did ensure this directory exists, I'll add that to the rpm spec

jrgm:
I have built the rpm successfully with this change. I'll note that there are two subtle differences 1) all the metadata is now squashed into a single string keyed by `Project-Id-Version`, and 2) newlines in text strings are no longer escaped. e.g., 

```
-      "Removing the last address will cancel your Persona account.\\nAre you sure you want to continue?": [                
+      "Removing the last address will cancel your Persona account.\nAre you sure you want to continue?": [   
```

ozten:
Great catches! We'll see in stage if newlines are significant. I didn't see any issues in http://translate.123done.org/

jrgm:
Yeah, the one place where there is a newline looks correct on translate.personatest.org (an alert on main site).

PR #3690, opened 2013-07-22
Check for failures in provisioning after returning from an IdP.

### LINKED ISSUE #3520: with 3rd party cookies disabled, primary IdP flow fails inexplicably
The failure mode is confusing.  We need to better message this to users, especially with gmail coming.  labeling 4 star to spark discussion.  It seems like there are a lot of issues around 3rd party cookies opened and closed, but none seem to address the primary idp case and get us a satisfying user facing error message.

shane-tomlinson:
- This is a standin for checking for disabled 3rd party cookies. If we expect
  provisioning to succeed, but it fails, this is probably due to disabled 3rd party cookies.

fixes #3520

### DIFF
--- a/resources/static/dialog/views/primary_user_not_verified.ejs
+++ b/resources/static/dialog/views/primary_user_not_verified.ejs
@@ -12,7 +12,7 @@
 
 <p>
   <%-
-  format(gettext("This may be due to disabled 3rd party cookies. Please ensure 3rd party cookies are enabled or allow 3rd party cookies from %(idpName)s"), {
+  format(gettext("This may be due to your browser blocking third-party cookies.  Please ensure your browser\'s privacy preferences are set to accept third-party cookies for %(idpName)s and try again."), {
     idpName: idpName
   })
   %>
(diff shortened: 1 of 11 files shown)


resources/static/dialog/views/primary_user_not_verified.ejs
```diff
@@ -0,0 +1,20 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<h2 id="primary_user_not_verified">
+  <%-
+  format(gettext("Cannot verify %(email)s"), {
+    email: _.escape(email)
```
seanmonstar: ejs has systems to escape. You can have the whole string escaped by using an equal sign: `<%= your text %>`, or a helper `escape` is always available in ejs context.

shane-tomlinson: @seanmonstar - I wasn't sure how to handle this one, because we do the `escape(email)` in other places. Doh, that is without the _.

seanmonstar: `escape` in other places is because we're adding html into the string also, and dont want the html escaped,

PR #3692, opened 2013-07-22
Issue 3691 primary user verified

### LINKED ISSUE #3691: Copy when returning from IdP is confusing and verbose.
@ryanfeeley agrees. He suggests changing it to look like:

screen shot 2013-07-22 at 19 41 01

shane-tomlinson:
An ephemeral instance, https://primary-verified.personatest.org is set up for testing.

fixes #3691

### DIFF
--- a/resources/static/dialog/views/primary_user_verified.ejs
+++ b/resources/static/dialog/views/primary_user_verified.ejs
@@ -3,17 +3,9 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 
   <div class="form_section" id="primary_user_verified">
-      <h2><%= gettext("Address Verified!") %></h2>
+      <%- format(gettext("%(email)s verified!"), {
+        email: "<strong>" + email + "</strong>"
+      }) %>
 
-      <ul class="inputs">
-
-          <li>
-              <label for="email" class="isDesktop"><%= gettext("Email") %></label>
-              <input id="email" type="email" disabled value="<%= email %>" />
-          </li>
-
-          <li>
-              <%= gettext("Signing you in&hellip;") %>
-          </li>
-      </ul>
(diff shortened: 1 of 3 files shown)


resources/static/dialog/views/primary_user_verified.ejs
```diff
@@ -3,17 +3,9 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 
   <div class="form_section" id="primary_user_verified">
-      <h2><%= gettext("Address Verified!") %></h2>
+      <%- format(gettext("%(email)s verified!"), {
```
seanmonstar, line 6: The `=` was switched to a `-`, which won't escape the email address in this output.

jaredhirsch, line 6: I assume that was to support the strong tags inserted on either side of the email on line 7.

Maybe a better solution is to put the escaped string inside strong tags:

```
<strong><%= gettext("%(email)s verified!"), {email: email}) %></strong>
```

shane-tomlinson, line 6: crap, after all of that, I failed to escape the email address.

shane-tomlinson, line 6: And this is why code reviews are a must. Thanks @seanmonstar and @6a68

shane-tomlinson, line 6: I was also thinking, what do you think of possibly exposing helper functions, something like "strong" that could be used to wrap text in a strong element?

shane-tomlinson:
@seanmonstar - Could you give another pass at this?  Does the 500 ms delay feel too short?

seanmonstar:
nah, it feels way better. the 3seconds makes people start to wonder if they need to do something.

PR #3696, opened 2013-07-23
Fix the xhr_delay being called on every request.

### LINKED ISSUE #3695: xhr_delay called for all requests

shane-tomlinson:
@6a68 - TAG!

fixes #3695


shane-tomlinson:
an ephemeral instance, xhr-delay.personatest.org is set up.

jaredhirsch:
R+ ftwz

PR #3697, opened 2013-07-23
Remove unneccesary(?) var app = undefined;

ryanseys:
Perhaps someone can explain why there is a need for the additional `var app = undefined;` in these modules?


jrgm:
r+

PR #3703, opened 2013-07-24
Issue 3689 edit email field

### LINKED ISSUE #3689: unable to change email after cancelling from "set password" screen for new user
This is due to the "remove required email" PR. DOH.

shane-tomlinson:
@6a68 - When ripping out requiredEmail, I made some assumptions in the authenticate module that were incorrect. I assumed that if the email was specified when the module started, it meant that the email address was immutable. This is not the case. A new user who enters an email address, goes to the set password screen, and then cancels and goes back to the authenticate screen should still be able to edit the email address.

Users who must authenticate post-email-verification, on assertion->password credential upgrade, or on transition_to_secondary must enter the password, but in this case the email is immutable.

So, I pass in whether the email is mutable or not from the state machine.

An ephemeral instance (with the fix contained in #3702) has already been pushed to run selenium tests against is set up - https://cancel-set-password.personatest.org

fixes #3689

### DIFF
--- a/automation-tests/tests/new-user/new-user-secondary-test.js
+++ b/automation-tests/tests/new-user/new-user-secondary-test.js
@@ -42,6 +42,13 @@ var new_secondary_123done_two_browsers = {
   "switch to the persona dialog": function(done) {
     browser.wwin(CSS['persona.org'].windowName, done);
   },
+  "enter incorrect email, cancel set password screen": function(done) {
+    browser.chain({onError: done})
+      .wtype(CSS['dialog'].emailInput, "incorrect_email@restmail.net")
+      .wclick(CSS['dialog'].newEmailNextButton)
+      .wclick(CSS['dialog'].submitCancelButton)
+      .wclear(CSS['dialog'].emailInput, done);
+  },
   "go through signup flow": function(done) {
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
fixes #3661

automation-tests/tests/new-user/new-user-secondary-test.js
```diff
@@ -42,6 +42,13 @@ var new_secondary_123done_two_browsers = {
   "switch to the persona dialog": function(done) {
     browser.wwin(CSS['persona.org'].windowName, done);
   },
+  "enter incorrect email, cancel set password screen": function(done) {
```
jaredhirsch, line 45: Initially, the idea was to only cover the happy paths with Selenium tests, on the assumption they were brittle.

They're pretty darn solid at this point--maybe we should work out a way to build lots and lots of little edge case-exercising tests or subtests?

Not something to change in this particular test, but something to think about.

jaredhirsch:
Unit tests passing for me locally. Code looks good, but I want to go through it once more before merging in.

jaredhirsch:
R+ I think

PR #3710, opened 2013-07-25
Change the default rpBackground color to match the footer.

### LINKED ISSUE #3677: Make the default RP backgroundColor ligher
Per Jon of Voost:

> right now, the background involves color and an image. I'd hate to lose the grey grainy image as I think it is pretty nice right now.

If the site doesn't specify any backgroundColor preference, maybe we could bring back the grain and a slight tint to keep the defaults more aesthetically pleasing / interesting?

shane-tomlinson:
@callahad - I give you the review.

The note from Shorelander is:

>  Hi Shane,
> 
>  I don't think we should add the noise back. As far as default colors, if the darker grey is default then I would
>  suggest lightening it up. Still darker than the left, but lighter than it is now.
> 
>  Thanks
> 
> ```
>  Stephen
> ```

fixes #3677

### DIFF
--- a/resources/static/dialog/css/style.css
+++ b/resources/static/dialog/css/style.css
@@ -295,7 +295,7 @@ section > .contents {
 
 
 .rpBackground {
-  background-color: #ccc;
+  background-color: #ebeced;
 }
 
 #signIn .table {


jaredhirsch:
@shane-tomlinson I'm not sure about the color change, but I'm not not sure. Merging.

jaredhirsch:
er I mean, I'm not sure it's wrong. oh monday.

PR #3713, opened 2013-07-25
3705: allow siteLogo parameter to include domain if served via https

SamPenrose:
This delta:
- continues to insist on the https scheme (explicitly or via scheme-relative or domain-absolute URIs)
- continues to forbid relative URIs
- allows specification of any domain

callahad says next step is to think carefully about security corner cases. Along those lines:
- we are relying on the browser to correctly sandbox the response as an image
- what information could leak to an arbitrary domain that doesn't leak from the authorization requests to the IdP's domain?

### DIFF
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -97,6 +97,48 @@ BrowserID.Modules.Dialog = (function() {
     this.publish("window_unload");
   }
 
+  function publishKpis(rpAPI) {
+    /*jshint validthis: true*/
+
+    // By default, a dialog is an orphan. It is only not an orphan if an
+    // assertion is generated. When an assertion is generated, orphaned will
+    // be set to false (currently in state.js).
+    var kpis = {
+      orphaned: true,
+      rp_api: rpAPI || "unknown"
+    };
+
+    // only publish the kpi's in aggregate.
+    this.publish("kpi_data", kpis);
+  }
+
+  function validateRpParams(origin_url, paramsFromRP) {
+    /*jshint validthis: true*/
+    var self=this;
+
+    startValidator.call(self);
+
+    var params;
+    try {
+      paramsFromRP.origin_url = origin_url;
+      params = self.validator.validate(paramsFromRP);
+    } catch(e) {
+      // note: renderError accepts HTML and cheerfully injects it into a
+      // frame with a powerful origin. So convert 'e' first.
+      self.renderError("error", {
+        action: {
+          title: "error in " + _.escape(origin_url),
+          message: "improper usage of API: " + _.escape(e)
+        }
+      });
+
+      throw e;
+    }
+
+    return params;
+  }
+
+
   var Dialog = bid.Modules.PageModule.extend({
     start: function(options) {
       var self=this;
@@ -144,35 +186,16 @@ BrowserID.Modules.Dialog = (function() {
 
       user.setOrigin(origin_url);
 
-      // By default, a dialog is an orphan. It is only not an orphan if an
-      // assertion is generated. When an assertion is generated, orphaned will
-      // be set to false (currently in state.js).
-      var kpis = {
-        orphaned: true
-      };
-
       if (self.startExternalDependencies) {
         startActions.call(self, success, error);
         startStateMachine.call(self);
       }
 
-      // The validator is always started.
-      startValidator.call(self);
-
       var params;
       try {
-        paramsFromRP.origin_url = origin_url;
-        params = self.validator.validate(paramsFromRP);
+        params = validateRpParams.call(self, origin_url, paramsFromRP);
       } catch(e) {
-        // note: renderError accepts HTML and cheerfully injects it into a
-        // frame with a powerful origin. So convert 'e' first.
-        self.renderError("error", {
-          action: {
-            title: "error in " + _.escape(origin_url),
-            message: "improper usage of API: " + _.escape(e)
-          }
-        });
-
+        // input parameter validation failure. Stop.
         return e;
       }
 
@@ -184,14 +207,10 @@ BrowserID.Modules.Dialog = (function() {
       if (params.startTime)
         self.publish("start_time", params.startTime);
 
-      if (params.rpAPI)
-        kpis.rp_api = params.rpAPI;
+      publishKpis.call(self, params.rpAPI);
 
       if (params.returnTo)
         user.setReturnTo(params.returnTo);
-
-      // only publish the kpi's in aggregate.
-      self.publish("kpi_data", kpis);
 
 
       // XXX Perhaps put this into the state machine.
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -249,7 +249,7 @@ BrowserID.Modules.Dialog = (function() {
       user.setOrigin(origin_url);
 
       // By default, a dialog is an orphan. It is only not an orphan if an
-      // assertion is generated. When an asseriton is generated, orphaned will
+      // assertion is generated. When an assertion is generated, orphaned will
       // be set to false (currently in state.js).
       var kpis = {
         orphaned: true
--- a/resources/static/dialog/js/modules/dialog.js
+++ b/resources/static/dialog/js/modules/dialog.js
@@ -304,14 +304,16 @@ BrowserID.Modules.Dialog = (function() {
         }
 
         if (paramsFromRP.siteLogo) {
-          // Until we have our head around the dangers of data uris and images
-          // that come from other domains, only allow absolute paths from the
-          // origin.
-          params.siteLogo = fixupAbsolutePath(origin_url, paramsFromRP.siteLogo);
+          // Regularize URL; throws error if input is relative.
+          params.siteLogo = fixupURL(origin_url, paramsFromRP.siteLogo);
           // To avoid mixed content errors, only allow siteLogos to be served
           // from https RPs
           /*jshint newcap:false*/
           if (URLParse(origin_url).scheme !== "https") {
+            throw new Error("only https sites can specify a siteLogo");
+          }
+          // XXX eliminate previous? This test has some overlap with it.
+          if (URLParse(params.siteLogo).scheme !== "https") {
             throw new Error("only https sites can specify a siteLogo");
           }
         }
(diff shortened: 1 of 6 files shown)


resources/static/dialog/js/modules/dialog.js
```diff
@@ -304,16 +304,18 @@ BrowserID.Modules.Dialog = (function() {
...
+          params.siteLogo = fixupURL(origin_url, paramsFromRP.siteLogo);
           // To avoid mixed content errors, only allow siteLogos to be served
           // from https RPs
           /*jshint newcap:false*/
           if (URLParse(origin_url).scheme !== "https") {
             throw new Error("only https sites can specify a siteLogo");
           }
+          // XXX eliminate previous? This test has some overlap with it.
+          if (URLParse(params.siteLogo).scheme !== "https") {
+            throw new Error("only https sites can specify a siteLogo");
```
shane-tomlinson: We are really after siteLogo coming from an https domain to avoid mixed content messages. What if instead of "only https sites can specify a siteLogo" it says "siteLogo must be on an https domain". After that, we should be able to get rid of the "only https sites can specify a siteLogo" check - an http site that specifies an https siteLogo will not cause the problem.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -249,7 +249,7 @@ BrowserID.Modules.Dialog = (function() {
       user.setOrigin(origin_url);
 
       // By default, a dialog is an orphan. It is only not an orphan if an
-      // assertion is generated. When an asseriton is generated, orphaned will
+      // assertion is generated. When an assertion is generated, orphaned will
```
shane-tomlinson: +1 to spelling fixes!

callahad:
The only thing I can think of is that we're allowing a hypothetical attacker to specify the domain we place in the `src` attribute. Until now, that domain was required to be the domain of the RP.

Does this change anything meaningful?

The only corner case I can think of is that they could point siteLogo at some endpoint on persona.org itself, but I can't think of any meaningful way in which that would be different from the status quo.

resources/static/dialog/js/modules/dialog.js
SamPenrose: I was wondering about that case. Here's the next edit, which I think is closer to what you want (the data: acceptance is not in yet, this is just bypassing the original_url test and clearer error message):

```
var validSchemes = {'https': 1, 'data': 1}; // 'app'?
    if (paramsFromRP.siteLogo) {
      // Regularize URL; throws error if input is relative.
      params.siteLogo = fixupURL(origin_url, paramsFromRP.siteLogo);
      /*jshint newcap:false*/
      if (!(URLParse(params.siteLogo).scheme in validSchemes)) {
        throw new Error("siteLogos can only be served from " + _.keys(validSchemes) + " schemes.");
      }
    }
```

SamPenrose:
Some people, when confronted with a problem, think "I know, I'll use /^(?:([^:\/?#]+):)?(?:\/\/((?:(([^:@]_)(?::([^:@]_))?)?@)?([^:\/?#]_)(?::(\d_))?))?((((?:[^?#\/]_\/)_)([^?#]_))(?:\?([^#]_))?(?:#(.*))?)/".

urlparse.js does not recognize data URIs.

resources/static/dialog/js/modules/dialog.js
SamPenrose: ... and now it looks different; see https://github.com/SamPenrose/browserid/commit/a9e40d02ee95413af90b42d6de8556424f88dd10

resources/static/dialog/js/modules/dialog.js
```diff
@@ -303,16 +303,32 @@ BrowserID.Modules.Dialog = (function() {
...
-          // origin.
-          params.siteLogo = fixupAbsolutePath(origin_url, paramsFromRP.siteLogo);
-          // To avoid mixed content errors, only allow siteLogos to be served
-          // from https RPs
-          /*jshint newcap:false*/
-          if (URLParse(origin_url).scheme !== "https") {
-            throw new Error("only https sites can specify a siteLogo");
+          dataMatches = paramsFromRP.siteLogo.match(dataUriRegex);
+          if (dataMatches) {
+	    if ((dataMatches[1].toLowerCase() === 'image')
```
shane-tomlinson: Can you set your editor's config to use spaces instead of tabs? 2 spaces per indent.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -303,16 +303,32 @@ BrowserID.Modules.Dialog = (function() {
           params.privacyPolicy = fixupURL(origin_url, paramsFromRP.privacyPolicy);
         }
 
+        var validLogoSchemes = {"https": 1, 'data': 1};
+        // 'data:image/png;base64,iV...' -> ['data:image/png;base64,iV...', 'image', 'png', ...]
+        // ... therefore mimetype -> [1]/[2]
+        var dataUriRegex = /^data:(.+)\/(.+);base64,(.*)$/;
+        var dataMatches = null;
+        // who needs a shared mimetype parsing library?
+        var imageMimeTypes = {'png': 1, 'gif': 1, 'jpg': 1, 'jpeg':1, 'svg': 1}
```
shane-tomlinson: We have been using arrays with _.indexOf(array, itemToFind) > -1 for this type of check - see validateRPAPI in the same file.

Don't forget about the semi-colons.

resources/static/dialog/js/modules/dialog.js
```diff
@@ -303,16 +303,32 @@ BrowserID.Modules.Dialog = (function() {
...
+          dataMatches = paramsFromRP.siteLogo.match(dataUriRegex);
+          if (dataMatches) {
+	    if ((dataMatches[1].toLowerCase() === 'image')
+                 &&
+                (dataMatches[2].toLowerCase() in imageMimeTypes)) {
+                ; // Good to go.
+            } else {
+              throw new Error("bad data URI for siteLogo: " + paramsFromRP.siteLogo.slice(0, 15) + " ...");
+            }
+	  } else {
```
shane-tomlinson: Same here.

resources/static/dialog/js/modules/dialog.js
shane-tomlinson: Can these checks be extracted into a "validate" method for clarity?

resources/static/test/cases/dialog/js/modules/dialog.js
```diff
@@ -443,32 +443,78 @@
...
-    var URL = HTTPS_TEST_DOMAIN + "://logo.png";
-    testMustBeAbsolutePath({ siteLogo: URL }, URL);
+  asyncTest("get with local https: siteLogo - allowed", function() {    
+    createController({
+      ready: function() {
+        var siteLogo = HTTPS_TEST_DOMAIN + "://logo.png";
+        var retval = controller.get(HTTPS_TEST_DOMAIN, {
+          siteLogo: siteLogo
+        });
+        equal(typeof retval, "undefined", "no error expected");
```
shane-tomlinson: You can also use testHelpers.testUndefined for a higher level test that does the same thing.

shane-tomlinson:
I'm excited to see that you are digging in and getting things done @SamPenrose. While you are digging in there, don't be afraid of refactoring bits of code if it helps you get your job done. Perhaps do the refactorings first in one commit, and your updates in another commit so we can follow along. Sometimes I do it in the reverse order, but I generally try to keep them in separate commits.

Also keep track of where you have difficulties understanding conceptually how pieces fit together. If your question isn't already answered in https://github.com/mozilla/browserid/wiki/Front-End-Development, feel free to modify it. If you have suggestions on where to make things clearer, note them down and send ideas to @seanmonstar, @6a68 and I.

resources/static/dialog/js/modules/dialog.js
SamPenrose: Editor/tabs -- yeah, sorry. I actually invested 45min in this yesterday.

resources/static/dialog/js/modules/dialog.js
SamPenrose: Semi-colons: sorry, per previous, wrestling with the dissatisfying state of JS modes in emacs as a side effect of switching from CentOS 5 to Mountain Lion. I'll get my act together shortly.

indexOf: the format specifies where the image extension should be. So I could loop over all formats and break if indexOf(matches, format) === 2, but that seems less obvious to me. Did you mean you want me to avoid the:

  testValue in whitelistObject

idiom?

resources/static/dialog/js/modules/dialog.js
SamPenrose: Is this underscore_versus camelCase inconsistency something we're stuck with due to public-facing API?

```
get: function(origin_url, paramsFromRP, success, error)
```

seanmonstar: Nope, that's just an oversight. Kill it with fire!

resources/static/dialog/js/modules/dialog.js
shane-tomlinson: Instead of declaring imageMimeTypes as an object with a `1` value for each field, and then checking on line [318](https://github.com/mozilla/browserid/pull/3713/files#L0R318) with `in`, we have been using an array, then checking whether something is in the array using indexOf:

```
var imageMimeTypes = ['png', 'gif', 'jpg', 'jpeg', 'svg'];
...
(imageMimeTypes.indexOf(dataMatches[2].toLowerCase()) > -1) { 
```

resources/static/dialog/js/modules/dialog.js
shane-tomlinson: I agree with @seanmonstar. Kill.

shane-tomlinson:
When submitting PRs, create a feature branch from dev with a new name, for this something like `issue_3705_data-uri_site_logo`. It makes it easier for you when developing multiple features, and it makes it easier for reviewers to understand the purpose of branches are when doing a `git branch -ra`

Our general approach to things is outlined in http://lloyd.io/applying-gitflow

jaredhirsch:
looking now.

jaredhirsch:
@SamPenrose OK, I see what happened. You had extracted the siteLogo-related validation stuff into a separate method within the Dialog module (in d4f50bf), but @shane-tomlinson went all out with a beautiful-yet-sprawling extraction of all the validation code out of the dialog module and into its own module in his PR.

I'll be honest, the merge commit (79eb6f3) is totally incomprehensible to me. I'm going to play around with doing the merge myself, or maybe cherry-picking these commits on top of current dev in different orders. I'll update this ticket in a few with my results.

SamPenrose:
@6a68 Would it be faster / better for me just to create a new branch from dev, merge Shane's fork, and then re-type my edits on top? Don't want to stick you with cleaning the stables when I have a shovel too ...

jaredhirsch:
@SamPenrose closing, feel free to reopen with a new pull request.

Issue #3718, opened 2013-07-26
upgrade 3rd-Eden/useragent when it gets support for IE11


jrgm:
@6a68 asked about this in https://github.com/3rd-Eden/useragent/issues/27, and the answer was basically "when IE11 is released". This bug is a reminder that we should update our version of useragent at that time.

PR #3719, opened 2013-07-26
Add P3p header test

jrgm:
This is not really about second-guessing 3rd-Eden/useragent; it's more about exercising lib/p3p.js middleware in the server for regression tests.

Also, 3rd-Eden/useragent will need to be upgraded when IE11 is live. https://github.com/mozilla/browserid/issues/3718


jrgm:
Hmm, just noticed in https://github.com/3rd-Eden/useragent/issues/27 that it mentions
https://github.com/3rd-Eden/useragent/commit/55aff614f5ff012a3a9fcafe70fab63c5dd9d749 which says it adds support for IE11, but I tried updating to useragent@2.0.6 and it still doesn't detect the user-agent string I have for IE11 preview - 'Mozilla/5.0 (Windows NT 6.3; WOW64; Trident/7.0; Touch; rv:11.0) like Gecko'

PR #3721, opened 2013-07-26
Show a "Signing in" message after the user selects an email from the email picker or successfully authenticates

### LINKED ISSUE #3708: no message provided during arrow animation when a user successfully authenticates.

shane-tomlinson:
@ryanfeeley already approves these changes from a UX perspective.
- Give the same signin notice for both primary and secondary sign ins
- Replace primary_user_verified.ejs with signing_in.ejs

An ephemeral instance is re-used - primary-verified.personatest.org

fixes #3708


resources/static/dialog/js/modules/generate_assertion.js
```diff
@@ -6,25 +6,36 @@ BrowserID.Modules.GenerateAssertion = (function() {
 
   var bid = BrowserID,
       dialogHelpers = bid.Helpers.Dialog,
-      sc,
-      user = bid.User,
-      storage = bid.Storage;
+      DELAY_BEFORE_ASSERTION_GENERATION = 750,
```
jaredhirsch: it's not clear to me why we need this delay

shane-tomlinson: cause it looks funny without some sort of delay.

resources/static/dialog/js/modules/primary_user_provisioned.js
```diff
@@ -17,22 +16,16 @@ BrowserID.Modules.PrimaryUserProvisioned = (function() {
       var self = this,
           email = options.email,
           assertion = options.assertion,
-          addEmailToCurrentUser = !!options.add,
           complete = function(status) {
             options.ready && options.ready(status || false);
-          },
-          delay = options.complete_delay || 500;
```
shane-tomlinson, line 24: We used to have a delay here too. 500 seemed a bit too short. 3000 was way too long.

resources/static/dialog/js/modules/generate_assertion.js
jaredhirsch: answer from IRC:

```
stomlinson _6a68: it is a bring over from the other module. Without the delay, the animation starts
immediately and it is a bit offputting. 
```

jaredhirsch: I already said this in IRC, but if this is just a UI delay, we should call it something like SIGNIN_ANIMATION_DELAY.

It'd be good to split out the UI and assertion generation bits more cleanly--maybe we can call dialogHelpers.getAssertion and then throw an arrow animation function call inside the setTimeout.

I guess dialogHelpers.getAssertion gets the assertion, then fires the 'assertion_generated' signal, which leads to the doAssertionGenerated call firing, which then immediately calls dialogHelpers.animateClose.

I'm not sure what the cleanest way is to separate all this. 90% of dialog/js/misc/state.js is a one-liner passthru, it's funky to be making UI calls in there for the doAssertionGenerated case.

Maybe we could immediately generate the assertion, and when it's ready, make the dialogHelpers.animateClose call in a setTimeout?

I'm sure I don't see the full picture here, but hopefully this helps.

jaredhirsch: Shane and I went around for a while talking about how to tweak this, see http://irclog.gr/#show/irc.mozilla.org/identity/303278 if interested. I'm sure another PR will materialize any moment.

resources/static/dialog/js/misc/helpers.js
```diff
@@ -13,6 +13,7 @@
       dom = bid.DOM,
       // This is a magic number, it is the same width as the arrow. See
       // resources/static/dialog/css/style.css #signIn for the arrow width.
+      DELAY_BEFORE_ANIMATION_BEGINS = 750,
       ARROW_WIDTH = 136;
 
   function animateClose(callback) {
```
shane-tomlinson: @6a68 - should we move this into the new module as well since there is nothing else that calls it?

resources/static/dialog/js/modules/actions.js
```diff
@@ -13,8 +13,7 @@ BrowserID.Modules.Actions = (function() {
       mediator = bid.Mediator,
       dialogHelpers = bid.Helpers.Dialog,
       runningService,
-      onsuccess,
-      onerror;
```
shane-tomlinson, line 17: onerror was never called. I took it out.

resources/static/dialog/js/modules/actions.js
```diff
@@ -127,15 +125,11 @@ BrowserID.Modules.Actions = (function() {
...
-      // Clear onerror before the call to onsuccess - the code to onsuccess
-      // calls window.close, which would trigger the onerror callback if we
-      // tried this afterwards.
-      this.hideWait();
-      dialogHelpers.animateClose(function() {
-        onerror = null;
-        if(onsuccess) onsuccess(info);
-      });
+    doCompleteSignIn: function(info) {
+      info.ready = function() {
```
shane-tomlinson, line 129: What is better, coming back here where onsuccess is already defined, or passing onsuccess into complete_sign_in?

jaredhirsch:
I got sidetracked looking at jank in the dialog, will have to review on Monday morning.

jaredhirsch:
ehh, Travis. Tests passing for me locally with 0.8/mysql running, merging in.

PR #3733, opened 2013-07-30
add a test for wsapi/email_addition_status (which was not being called previously)

jrgm:
I was looking at code coverage for browserid, and we were not ever calling /wsapi/email_addition_status (except for utter failure in stalled-mysql-test). 
1. I suppose I could have retrofitted some existing tests to add this call, but I don't mind having a distinct test for this wsapi.
2. There is another code path in email_addition_status not covered by this test, but reaching it would require changes to the lib/db code to expose a call, which I've done, but I'll put that in a separate pull requests after this one lands (or is rejected ;-) ).

@shane-tomlinson please note if I'm not using this wsapi the right way. Thanks.

### DIFF
--- a/tests/email-addition-status-test.js
+++ b/tests/email-addition-status-test.js
@@ -73,6 +73,23 @@ suite.addBatch({
   }
 });
 
+
+suite.addBatch({
+  "Checking email_addition_status for address that has not yet been added": {
+    topic: wsapi.get('/wsapi/email_addition_status', {
+      email: TEST_EMAIL_ADDED,
+    }),
+    "shows that 'status' is now 'failed'": function (err, r) {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@jrgm - this looks great to me. I am added a PR to your PR to add a test that checks the status of an email address that has not yet been staged. https://github.com/jrgm/browserid/pull/2

One thing that concerns me is that none of the  `*_status` wsapi tests check for the `mustAuth` state. That is a separate issue and can be resolved separately.

If you like my addition to your addition, merge away!

If you don't like my addition to your addition, merge away anyways. r+.

jrgm:
Okay, merged in @shane-tomlinson's addition. Thanks!

Merging in this test.

PR #3736, opened 2013-07-30
Add arrow overlap to avoid animation flicker. Fixes #3668.

### LINKED ISSUE #3668: Dialog jank during signin animation on Chrome only
When the chevron ( div.arrow or div.arrowContainer ) animates rightward after a successful login, you can see a flashing dark vertical line move across the dialog, too.

I can't repro on Safari or FF, but it's bad on Chrome/Mac. I'm not having luck getting a screen capture of the animation in progress.

jaredhirsch:
@seanmonstar mind looking?

I'm not sure if I should reposition `.arrow` or `.arrowContainer`, but this works, meh.

Yay webkit CSS subpixel rounding errors during transition of adjacent els.


seanmonstar:
lolwut ok

PR #3739, opened 2013-07-30
fix assertions with uppercase domains + proxy_idps

seanmonstar:
and a test!


jaredhirsch:
ok, I don't even understand what's wrong with travis. tests are passing for me, though. in the cavalier late afternoon spirit, merging in.

Issue #3728, opened 2013-07-29
persona user ID receives error message "Gmail.com is not responding"


jedp:
ed at emrsoftware.com reported in [Bug 898851](https://bugzilla.mozilla.org/show_bug.cgi?id=898851):

User Agent: Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.1; WOW64; Trident/6.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; Media Center PC 6.0; InfoPath.2; .NET4.0C; .NET4.0E)

Steps to reproduce:

A current User tries to login to my website (golfgrudge.com) using Persona User Id  email address = "ARJ412@Gmail.com".

Actual results:

After hitting next button, the user immediately gets the response "Gmail.com is not responding. Please wait a few minutes and try again. "  This happens every time, and happens using IE, Chrome, FireFox, (every thing I've tried it on.)  This user has been logging in successfully for months.  I assume this has something to do with the Persona / GMAIL interaction that has been recently added to Persona.

Expected results:

Persona should have the user verify the email by typing in the password, and then successfully authenticating the user.

/cc @callahad

jedp:
Ed additionally comments:

> I've tested logging in to my website with other gmail accounts, and have not had any problem.  So far just the gmail email address listed above has this issue.

callahad:
I suspect DNS issues. Can they resolve gmail.login.persona.org? Can they hit ver.txt on the same?

seanmonstar:
Nah, testing that email address myself gives me the same error.

seanmonstar:
I get the feeling this is related to his email having an uppercase G.

Trying with the g in lowercase, I get that the state is transition, and I see it load the provisioning page just fine. It's of course not proven yet, and then in the dialog, something uses the address_info again, only it uses the `normalizedEmail`. Since that includes the capital G, it hits the network again (cache key missed), and we get back that the state is offline.

```
https://login.persona.org/wsapi/address_info?email=ARJ412%40gmail.com&issuer=default
{
  "auth":"https://gmail.login.persona.org/authenticate",
  "prov":"https://gmail.login.persona.org/provision",
  "type":"primary",
  "issuer":"gmail.login.persona.org",
  "normalizedEmail":"ARJ412@Gmail.com",
  "state":"transition_to_primary"
}
```

```
[...]

PR #3741, opened 2013-07-31
fix verifier to find proxy_idps with uppercase domains

seanmonstar:


lib/verifier/certassertion.js
```diff
@@ -164,7 +164,7 @@ function verify(assertion, audience, forceIssuer, allowUnverified, successCB, er
         return errorCB("missing email");
       }
 
-      var domainFromEmail = email.replace(/^.*@/, '');
+      var domainFromEmail = primary.domainFromEmail(email);
```
fmarier, line 167: yay for centralizing that check into a separate function!

jrgm:
Okay, in IRC, @fmarier gave this an r+ too. So merging this to dev.

jrgm:
link to https://github.com/mozilla/browserid/issues/3728

PR #3743, opened 2013-07-31
Issue 3705 site logo ur is

SamPenrose:
1) Supports https: URIs from any domain while excluding http: and relative URIs.
2) Supports data:image/ URIs.
3) Does NOT restrict the latter by size.
4) Does NOT support app: URIs.
5) RP parameter validation factored into modules/validate_rp_params by @shane-tomlinson .

### DIFF
--- a/resources/static/dialog/js/modules/validate_rp_params.js
+++ b/resources/static/dialog/js/modules/validate_rp_params.js
@@ -1,0 +1,279 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+BrowserID.Modules.ValidateRpParams = (function() {
+  "use strict";
+
+  /**
+   * This module validates parameters that come from the RP.
+   * the validate method is called with paramsFromRP and returns params, a list
+   * of cleaned parameters than can be assumed to be safe.
+   */
+
+  var bid = BrowserID,
+      user = bid.User,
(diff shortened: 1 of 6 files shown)


resources/static/dialog/js/modules/validate_rp_params.js
```diff
@@ -0,0 +1,295 @@
...
+    // into an array that looks like:
+    //   ['data:image/png;base64,iV...', 'image', 'png', ...]
+    // which means that mimetype proper is represented as-> [1]/[2]
+    var dataUriRegex = /^data:(.+)\/(.+);base64,(.*)$/;
+
+    dataMatches = inputLogoUri.match(dataUriRegex);
+    if (dataMatches) {
+      if ((dataMatches[1].toLowerCase() === 'image')
+           &&
+          (imageMimeTypes.indexOf(dataMatches[2].toLowerCase()) > -1)) {
```
jaredhirsch: use _.indexOf for ie8 compat

SamPenrose:
indexOf fix pushed

jaredhirsch:
@SamPenrose thanks for the quick IE8 fix! I think this looks good, but the refactor gives it serious size. I want to get it on dev and poke at it, seems like a lot of changes to merge in just before branching.

I'll branch the train tonight, re-read this once more in the morning, and merge after that, modulo any final nits.

shane-tomlinson:
@SamPenrose - I'm a bit late to the party, have you checked large (>32KB) data URIs on IE8?

SamPenrose:
Nope. Perhaps foolishly took silence for consent on not restricting size.

PR #3744, opened 2013-07-31
hotfix assertions and verification with uppercase domains

seanmonstar:
@jrgm r?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,7 @@ train-2013.07.17:
   * On each Travis-CI run, record linux, mysql, and phantomjs versions for future reference: #3626
   * (hotfix 2013.07.19) fix l10n rpm building: #3683, #3685
   * (hotfix 2013.07.29) Hotfix primary emails with uppercase: #3728, #3732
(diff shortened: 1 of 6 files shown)


jrgm:
So this, in general, works for me without sideshow. bigtent for an email with a capital in the domain is till broken, but @seanmonstar has a PR to fix (so bigtent is the same as it was for those users). In particular, for the case where this began: a user like foo@Gmail.com, with type secondary, cannot login, that case is fixed and the user can be upgraded to primary.

I did run into an odd situation with multiple email addresses like foo+1@gmail.com, foo+2@gmail.com, where things got confused and it wouldn't allow me to proceed with the log (saying on a repeat visit, for an email that was already allowed that 'you are logging in as foo+1@gmail.com but are currently logged in as foo@gmail.com). However, this is probably unrelated to the changes, and I'll try to find STR in production.

So r+ jrgm, and merging. I'll build the rpm and update the ticket for this train to be deployed to stage (but I am not going to ship this this week, since another matter may need to be dealt with first).

PR #3748, opened 2013-08-02
support csv list for --platform, and update readme with current output of --help

jrgm:
A convenience. A quoted glob like "--platform `'{linux_foo,win7_bar}'`" would (and still) works, but who wants to type that?


shane-tomlinson:
@jrgm - this slipped past me until now - this is a very useful improvement!

PR #3750, opened 2013-08-04
fixes #2728; content-type header is not guaranteed to exist

### LINKED ISSUE #2728: Missing error check; lib/primary.js crashes if no content-type in response
Just happened to notice this, but this should check that there is a content-type before using it.

```
res.headers['content-type'].indexOf('application/json') !== 0
```

https://github.com/mozilla/browserid/blob/dev/lib/primary.js#L167

jrgm:
Fixes #2728

### DIFF
--- a/lib/http_forward.js
+++ b/lib/http_forward.js
@@ -98,7 +98,8 @@ exports.forward = function(dest, req, res, cb) {
   // if the body has already been parsed, we'll write it
   if (req.body) {
     var data;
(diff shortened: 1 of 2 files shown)


jaredhirsch:
@jrgm The code looks good, but seems like we should add tests for these edge cases.

jrgm:
Hmm, I don't like saying this, but to test those would require rewriting some of that to make it more testable. I don't think this fix warrants that work.

shane-tomlinson:
Testing now.

PR #3751, opened 2013-08-05
Clean up jshint warnings in ./automation-tests

jrgm:
I like jshint, but opinions may vary. Anyways, this cleans up jshint warnings in ./automation-tests.
- it grandfathers use of 'foo["bar"]' since we use a lot of it
- cleans up a bunch of whiny stuff about semicolons
- removes some dead code jshint pointed out
- fix a couple of leaks to global of variables.
- prefers strict equality tests

I don't see any significant difference in test pass/fail rates on chrome, ff{win,linux}, ie9 and ie10.

### DIFF
--- a/automation-tests/tests/public-terminals.js
+++ b/automation-tests/tests/public-terminals.js
@@ -4,6 +4,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/*jshint sub: true */
+
 const
 path = require('path'),
 assert = require('../lib/asserts.js'),
--- a/automation-tests/tests/public-terminals.js
+++ b/automation-tests/tests/public-terminals.js
@@ -68,6 +68,8 @@ runner.run(module, {
       .wfind(CSS['persona.org'].accountManagerHeader, done);
   },
   'hack localStorage to simulate 60 seconds of inactivity': function(done) {
+    /*jshint evil:true */
+
     // JSON.parse to get user id, JSON.parse usersCOmputer, rewind time,
     // stringify and finally set as local storage.
     var rewindOneMinute = '(function() { ' +
@@ -84,7 +86,7 @@ runner.run(module, {
     browser.chain({onError: done})
       .wclick(CSS['persona.org'].header.signIn)
       .wfind(CSS['persona.org'].accountManagerHeader) // make sure we're logged in
-      .eval(rewindOneMinute, function(err, out) {
+      .eval(rewindOneMinute, function(err) {
         // you can echo out eval's return value for debugging
         // console.error(out);
         done(err);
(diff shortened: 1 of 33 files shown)


automation-tests/tests/public-terminals.js
```diff
@@ -66,6 +68,8 @@ runner.run(module, {
       .wfind(CSS['persona.org'].accountManagerHeader, done);
   },
   'hack localStorage to simulate 60 seconds of inactivity': function(done) {
+    /*jshint evil:true */
```
shane-tomlinson, line 71: E.V.I.L.

automation-tests/tests/sign-in-test.js
```diff
@@ -4,6 +4,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/*jshint sub: true */
```
shane-tomlinson, line 7: Since there are so many of these, we should put it into the .jshintrc file. I thought there was a .jshintrc specifically for the automation-tests directory, but I do not see it. :/

shane-tomlinson:
Testing now.

shane-tomlinson:
before:

```
➜  automation-tests git:(dev) ✗ find . -name *.js | grep -v "node_modules" | xargs -n 1 jshint
./lib/asserts.js: line 3, col 68, Missing semicolon.
./lib/asserts.js: line 5, col 2, Missing semicolon.
./lib/asserts.js: line 9, col 11, Expected '!==' and instead saw '!='.
./lib/asserts.js: line 10, col 2, Missing semicolon.

4 errors
./lib/personatestuser.js: line 14, col 24, Expected '===' and instead saw '=='.
./lib/personatestuser.js: line 19, col 36, ['PERSONA_ENV'] is better written in dot notation.
./lib/personatestuser.js: line 23, col 39, Expected '===' and instead saw '=='.
./lib/personatestuser.js: line 24, col 80, Missing semicolon.
./lib/personatestuser.js: line 19, col 10, 'env' is defined but never used.

5 errors
...
```

after:

```
➜  automation-tests git:(autotest-jshint) ✗ find . -name *.js | grep -v "node_modules" | xargs -n 1 jshint
➜  automation-tests git:(autotest-jshint) ✗
```

automation-tests/tests/sign-in-test.js
jrgm, line 7: Yeah, there isn't .jshintrc in automation-tests, and I thought of adding one, but chose to grandfather existing so new stuff had to either declare or use the prevailing rule. But I could go with an automation-tests/.jshintrc (but I personally prefer foo.bar to foo['bar'] where the latter is not required).

shane-tomlinson, line 7: @jrgm - nice avatar. w.r.t. "prefer foo.bar to foo['bar'] where the latter is not required" - I'm there with you.

PR #3754, opened 2013-08-06
Fix l10n rpm train 2013.07.31

jrgm:


ozten:
Sorry about that, thanks!

PR #3760, opened 2013-08-08
Train 2013.07.31+hotfix 868967

fmarier:
This was hotfixed on train-2013.07.17 after train-2013.07.31 was branched. Therefore we need to hotfix this train too.


jrgm:
r=jrgm, merging to train-2013.07.31 and will rebuild the rpm.

PR #3761, opened 2013-08-08
Explicitly set rejectUnauthorized to false when using https.request()

fmarier:
This is the default in node 0.10 but not in 0.8. That older version
needs both `agent: false` and `rejectUnauthorized: true` before it
will reject invalid TLS certificates.

http://nodejs.org/api/https.html#https_https_request_options_callback

In production, this is normally not needed because our squid proxy
is also set to reject invalid certificates, but setting it like
this means less surprises when we upgrade to node 0.10 and protects
anybody that might be running our code without a proxy on the old
version of node.

https://bugzilla.mozilla.org/show_bug.cgi?id=901650

### DIFF
--- a/lib/http_forward.js
+++ b/lib/http_forward.js
@@ -42,6 +42,7 @@ exports.forward = function(dest, req, res, cb) {
     path: u.path,
     method: req.method,
     headers: req.headers,
+    rejectUnauthorized: true,
     agent: false
   }, function(pres) {
 
--- a/lib/primary.js
+++ b/lib/primary.js
@@ -150,6 +150,7 @@ var fetchWellKnown = function (currentDomain, principalDomain, clientCB) {
     req = https.get({
       host: currentDomain,
       path: WELL_KNOWN_URL + "?domain=" + principalDomain,
+      rejectUnauthorized: true,
       agent: false
     }, handleResponse);
   }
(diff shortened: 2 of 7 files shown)


shane-tomlinson:
@fmarier - when doing a quick `grep -r "agent: false"` in the `lib` directory, there are 12 locations. rejectUnauthorized was only added in 10 places. Is this intentional?

shane-tomlinson:
Are TLS failures logged somewhere? If not, should they be?

fmarier:
I think these are the occurrences that @shane-tomlinson is talking about:
1. https://github.com/mozilla/browserid/blob/b0607a255a66c7804e6f09063639ca1be7157035/lib/primary.js#L144
2. https://github.com/mozilla/browserid/blob/b0607a255a66c7804e6f09063639ca1be7157035/lib/wsapi/interaction_data.js#L57
3. https://github.com/mozilla/browserid/blob/b0607a255a66c7804e6f09063639ca1be7157035/lib/heartbeat.js#L103

The first two are for HTTP connections so that option doesn't apply.

The third one is for the heartbeat. I didn't change it because I didn't want to mess with our monitoring stuff and because it's not security-critical, there's probably no harm in doing it though. I'm happy to update this pull request if you think it's worth it.

fmarier:
I don't think we log TLS failures. As for whether or not it's worth it, I don't know. I suppose they will be mostly mis-configuration on the part of IdPs (as opposed to people attacking IdPs), but it might still be interesting to see them at a low debug level. @shane-tomlinson Are you suggesting CEF-logging them?

shane-tomlinson:
@fmarier - that I am, but that can be for another PR.

shane-tomlinson:
Testing now.

PR #3774, opened 2013-08-10
update inspect_localstorage to work with B2G (and fwiw opera blink)

jrgm:
Make this tool work with Opera Blink, and with B2G (by adb-pulling the webappsstore.sqlite database to local disk)
Also add extended help with `--help`


scripts/inspect_localstorage.js
```diff
@@ -88,36 +109,106 @@ const OPTIONS = {
...
   v: {
-    describe: 'show the name of the database file',
+    describe: 'show a bit more detail about program operation',
     'default': false
   },
 };
 
+function tmpdir() {
+  // from nodejs 0.10.15
+  var isWindows = (process.platform === 'win32');
```
shane-tomlinson: would [tmp](https://npmjs.org/package/tmp) work here as well?

scripts/inspect_localstorage.js
```diff
@@ -88,36 +109,106 @@ const OPTIONS = {
...
+      process.env.TEMP ||
+      '/tmp';
+  }
+}
+
+function adbPullWebappsstore(dbfile, callerCb) {
+  var ADB_EOL = '\r\n';
+
+  function findSqlitePath(callback) {
+    var glob = './data/b2g/mozilla/*/' + WEBAPPSSTORE_SQLITE;
```
shane-tomlinson, line 122: This is HOT.

scripts/inspect_localstorage.js
```diff
@@ -88,36 +109,106 @@ const OPTIONS = {
...
+    var cmd = "adb shell 'ls " + glob + "'";
```
shane-tomlinson, line 123: Is the `adb` dep checked for or is it assumed to be installed?

jrgm, line 123: It's not checked. If it's not in the path, this message is printed, which I think is enough for anyone working with b2g and interested in localStorage to figure out what is wrong:

```
*** ERROR: Pulling from b2g:  Command failed: /bin/sh: adb: command not found
```

jrgm:
@shane-tomlinson For some reason, I was thinking that 0.8 didn't have os.tmpDir() (but it's 0.6 that didn't  have that). Fixed now.

scripts/inspect_localstorage.js
shane-tomlinson, line 123: That's plenty, thanks @jrgm.

PR #3776, opened 2013-08-12
Redirect flow

### LINKED ISSUE #2034: Support environments without popups (phonegap, etc.)
Several folks on dev-identity are asking about getting BrowserID to work with things like PhoneGap. Apparently one of the primary issues is opening a popup: it opens in a browser outside of the PhoneGap application, breaking the connection between Persona and the app.

seanmonstar:
Adds a redirect flow for browsers that don't support the dialog flow. Most notably: Chrome for iOS and Windows Phone IE. The RP has no control over whether this can happen, to keep the experience consistent for all users.

This flow only works with the `watch()` API. Without it, after being redirected, the user would still not be logged in. An `unsupported_dialog_without_watch` screen will be shown on these browsers if `navigator.id.get()` is used. The wording could probably use some help.

fixes #2034 :beers:

### DIFF
--- a/lib/static/views.js
+++ b/lib/static/views.js
@@ -229,6 +229,17 @@ exports.setup = function(app) {
     });
   });
 
+  app.get("/unsupported_dialog_without_watch", function(req,res) {
+    renderCachableView(req, res, 'unsupported_dialog_without_watch.ejs', {
+      title: _('Unsupported Browser without Watch'),
+      layout: 'dialog_layout.ejs',
+      useJavascript: false,
+      // without the javascript bundle, there is no point in measuring the
+      // window opened time.
+      measureDomLoading: false
+    });
+  });
+
   app.get("/cookies_disabled", function(req,res) {
     renderCachableView(req, res, 'cookies_disabled.ejs', {
       title: _('Cookies Are Disabled'),
--- a/lib/static/views.js
+++ b/lib/static/views.js
@@ -236,7 +236,8 @@ exports.setup = function(app) {
       useJavascript: false,
       // without the javascript bundle, there is no point in measuring the
       // window opened time.
-      measureDomLoading: false
+      measureDomLoading: false,
+      showLoading: true
     });
   });
 
--- a/lib/static/views.js
+++ b/lib/static/views.js
@@ -237,7 +237,7 @@ exports.setup = function(app) {
       // without the javascript bundle, there is no point in measuring the
       // window opened time.
       measureDomLoading: false,
-      showLoading: true
+      showLoading: false
     });
   });
 
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -133,6 +133,8 @@
       (isFennec ? undefined :
        "menubar=0,location=1,resizable=1,scrollbars=1,status=0,width=700,height=375");
 
+    var needsPopupFix = true;
+
     var w;
 
     // table of registered observers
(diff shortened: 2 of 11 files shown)


resources/static/include_js/_include.js
```diff
@@ -351,17 +361,30 @@
...
+        }
+      }
+      
+      if (!isSupported()) {
+        var reason = noSupportReason();
+        var url = "unsupported_dialog";
 
         if(reason === "LOCALSTORAGE_DISABLED") {
           url = "cookies_disabled";
+        } else if (reason === REQUIRES_WATCH) {
```
shane-tomlinson, line 381: I dig the use of a constant here, should we do it to the others?

resources/views/unsupported_dialog_without_watch.ejs
```diff
@@ -0,0 +1,17 @@
...
+
+  <section id="error" class="cookies_disabled">
+      <div class="table">
+        <div class="vertical contents">
+          <h2 id="reason">
+            <%= gettext("Persona requires the watch() API in this browser") %>
+          </h2>
+
+          <p>
+            <%- format(gettext("If you're a developer, look at using <a %s>navigator.id.watch()</a>. If not, tell the website owner."), [" target='_blank' href='https://developer.mozilla.org/en-US/docs/Web/API/navigator.id.watch'"]) %>
```
shane-tomlinson, line 13: Much better than my message!

shane-tomlinson:
I'm SO excited for this! I'll set up an ephemeral instance and show @ryanfeeley today to see if we need any text updates.

lib/static/views.js
```diff
@@ -227,6 +227,17 @@ exports.setup = function(app) {
     });
   });
 
+  app.get("/unsupported_dialog_without_watch", function(req,res) {
```
shane-tomlinson, line 232: We should add some tests for this.

resources/static/include_js/_include.js
```diff
@@ -351,17 +361,30 @@
...
+          url = "unsupported_dialog_without_watch";
```
shane-tomlinson, line 382: Much cleaner than the approach that I took.

shane-tomlinson:
The text currently looks like:

screen shot 2013-08-13 at 11 50 52

@ryanfeeley - how can we improve this?

shane-tomlinson:
@seanmonstar - I added a PR to this PR to add tests for the new screen.

shane-tomlinson:
@seanmonstar - I do not receive the error message when using Chrome for iOS on my ipod touch when visiting redirect-flow.myfavoritebeer.org.

shane-tomlinson:
@seanmonstar - w.r.t. my last comment - I get the unsupported dialog, but I also get the "connecting to persona" screen that never goes away.

resources/static/include_js/_include.js
seanmonstar, line 381: yea, when we can, constants ftw. I'd probably just fix them as I touched that section. _shrug_

seanmonstar:
rebased against dev, since it had fallen out of sync.

resources/static/test/cases/common/js/storage.js
```diff
@@ -237,7 +237,23 @@
 
     storage.idpVerification.clear();
     var info = storage.idpVerification.get("expired");
-    equal(typeof info, "undefined");
+    testHelpers.testUndefined(info);
+  });
+
+  test("rpRequest functions, set, get, clear", function() {
+    storage.rpRequest.set({
+      origin: "https://testuser.com"
```
seanmonstar, line 245: missing comma!

shane-tomlinson, line 245: Wow. How'd I not see that error?

shane-tomlinson:
@seanmonstar - I am closing and re-opening this PR against a branch on origin for final review. I updated the broken test and have added @ryanfeeley's text suggestions.

PR #3780, opened 2013-08-13
record frontend test details in a temp file

jrgm:
I've been doing this locally with trains for a while, so would like to have this always available. Yeah, there might be more elegant ways to do this such as report back to a server, but anyways.

### DIFF
--- a/automation-tests/tests/frontend-qunit-test.js
+++ b/automation-tests/tests/frontend-qunit-test.js
@@ -12,23 +12,9 @@ utils = require('../lib/utils.js'),
 runner = require('../lib/runner.js'),
 persona_urls = require('../lib/urls.js'),
 fs = require('fs'),
+os = require('os'),
 path = require('path'),
 testSetup = require('../lib/test-setup.js');
-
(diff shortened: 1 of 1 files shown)


automation-tests/tests/frontend-qunit-test.js
```diff
@@ -4,19 +4,37 @@
...
 const
 assert = require('../lib/asserts.js'),
 utils = require('../lib/utils.js'),
 runner = require('../lib/runner.js'),
 persona_urls = require('../lib/urls.js'),
+fs = require('fs'),
 path = require('path'),
 testSetup = require('../lib/test-setup.js');
 
+function tmpdir() {
```
shane-tomlinson: @jrgm - can a library like [tmp](https://npmjs.org/package/tmp) be used for this?

shane-tomlinson:
Aside from my comment about using a library like tmp, r+. I dig it. Merge when ready.

jrgm:
@shane-tomlinson yeah, same brain cramp over node api versions. Fixed now.

Issue #3212, opened 2013-04-09
/include.orig.js is 404 on train-2013.03.29 on stage


jrgm:
1. Go to https://login.anosrep.org/include.orig.js => 404

lib/static/views.js is looking for it in /build/include.js, but there is nothing there. Nor do I see a file `include.orig.js` anywhere in the unpacked rpm.

shane-tomlinson:
This is the revenge of #3056 I believe. @callahad - wanna look at this?

shane-tomlinson:
@jrgm - how is the server process started on staging/prod, is run_locally.js called at all?

callahad:
sure thing

jrgm:
So npm install is run on a build box and then an rpm is packaged with scripts/rpmbuild.sh and scripts/browserid.spec. That rpm is installed on stage/prod and is started with a shell script that sets CONFIG_FILES and does `bin/browserid`.

So, no, no run_locally.js.

shane-tomlinson:
@jrgm - are any resources that we place into the "production" subdirectory included in the rpm package? I am thinking if we still do the initial build step to create the uncompressed include.js into build, then copy that as include.orig.js into production, it should be copied into the rpm and available to people that request it.

jrgm:
Sorry, I got behind on all the email. So, see https://github.com/mozilla/browserid/blob/dev/scripts/browserid.spec#L35

So, resources/static/production is packaged into the rpm, so you're idea would work.

swh:
Someone should get on this - it should be a simple fix, and it looks really scruffy!

Worst case just remove the comment from the top of the compressed .js file.

edwindotcom:
+1 - still 404'ing

lbredeso:
I just ran into this while trying to find the JavaScript that handles "returnTo" in the Observer API:  https://developer.mozilla.org/en-US/docs/Web/API/navigator.id.request

https://login.persona.org/include.js still references https://login.persona.org/include.orig.js, which still returns a 404.

callahad:
<-- Jerk who didn't fix include.orig.js

chilts:
I'll take a look at this. Assigning myself.

seanmonstar:
Here's the reasons why: The uncompressed file is cleaned up by the rpmbuild process.

https://github.com/mozilla/browserid/blob/dev/lib/static/views.js#L145
https://github.com/mozilla/browserid/blob/dev/scripts/browserid.spec#L30

[...]

PR #3795, opened 2013-08-15
make /include.orig.js available again. Fixes GH-3212.

jrgm:
I took @djc's change to browserid.spec and added the necessary adjustment in views to show this (but check my math). 

This whole area could use some overall rationalization, but for now let's just fix the 404 for /include.orig.js

I've built an rpm and installed it in a pseudo production environment and seems good to me.

@shane-tomlinson, @6a68 please consider adding this before you branch train-2013.08.14 tomorrow. 

 Fixes GH-3212.


shane-tomlinson:
Testing now.

PR #3801, opened 2013-08-15
Redirect flow

shane-tomlinson:
@seanmonstar - here is the replacement for #3776. This is pushed to redirect-ios.personatest.org.

I'm happy with it if you are happy with it!

### DIFF
--- a/resources/static/test/cases/common/js/storage.js
+++ b/resources/static/test/cases/common/js/storage.js
@@ -219,7 +219,7 @@
 
     storage.idpVerification.clear();
     info = storage.idpVerification.get();
-    equal(typeof info, "undefined");
+    testHelpers.testUndefined(info);
   });
 
   test("idpVerification - clear old info", function() {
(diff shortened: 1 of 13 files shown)


resources/static/test/cases/common/js/storage.js
```diff
@@ -237,7 +237,23 @@
 
     storage.idpVerification.clear();
     var info = storage.idpVerification.get("expired");
-    equal(typeof info, "undefined");
+    testHelpers.testUndefined(info);
+  });
+
+  test("rpRequest functions, set, get, clear", function() {
+    storage.rpRequest.set({
+      origin: "https://testuser.com"
```
shane-tomlinson: Are you kidding me? Where did the fix for this go?

shane-tomlinson:
@seanmonstar - I'm not sure _why_ those last two commits are showing up here, when rebasing this from dev, I see "Current branch redirect-flow is up to date." When I push to origin, I see "Everything up-to-date". Closing and re-opening.

Issue #3755, opened 2013-08-07
metrics output should really be JSON


jaredhirsch:
the metrics files, like `router-metrics.json`, aren't valid JSON. They are a bunch of stringified JSON objects, but they aren't collected inside a top-level object, and they aren't separated by commas.

it would be way easier to process those files if they were valid JSON, so, do that.

jaredhirsch:
link to mozilla/identity-ops#102 ("Update the cron job processing the metrics file to send metrics data to a kpiggybank collector")

callahad:
If this is a streaming logfile, I'm not sure how we could sanely ensure that it's valid JSON at all times.

Every time we wanted to append, wouldn't we need to truncate the file to kill the final `}`, then write the next line with with a leading `,` and a trailing `}`?

Seems complex :(

jaredhirsch:
@callahad ah, that's true. i'm no longer attempting to parse metrics files, so, closing as "meh."

PR #3809, opened 2013-08-19
Issue 3529 metrics to piggybank

jaredhirsch:
@shane-tomlinson Could I get your feedback on this? Not sure it's ready, but maybe it is (might also need to wait for privacy review before merging, I'm not sure).

This is a fix for mozilla/identity-ops#102 that sends some additional RP metrics to kpiggybank along with the existing kpi data. We do this by extracting the `wsapi/interaction_data` kpi code into a separate lib, and dropping a few additional logger calls into `bin/verifier` and `bin/router`.

There is a separate privacy bug open for collecting the RP metrics, and there's also a ton of related conversation, inside mozilla/identity-ops#102.

Open questions:
- what format should I use for the new data? what should I call the keys, how should I namespace them?
- do we need to wait until privacy review is done to merge in? 
- are there other metrics I should fold into this PR? I'm thinking particularly of #3755.

### DIFF
--- a/bin/router
+++ b/bin/router
@@ -23,6 +23,7 @@ logger = require('../lib/logging.js').logger,
 forward = require('../lib/http_forward').forward,
 addP3PHeader = require('../lib/p3p.js'),
 shutdown = require('../lib/shutdown'),
+store_kpi = require('../lib/kpi'),
 toobusy = require('../lib/busy_middleware.js');
 
 var app = express.createServer();
@@ -179,10 +180,16 @@ app.use(function(req, res, next) {
       // IP address (this probably needs to be replaced with the X-forwarded-for value
       ip: ipAddress
     });
+    store_kpi({
+      type: 'signin',
+      browser: req.headers['user-agent'],
+      rp: referer,
+      // IP address (this probably needs to be replaced with the X-forwarded-for value
+      ip: ipAddress
+    });
   };
 
   // log metrics
-  // TODO also log to kpiggybank here
   if (req.url === '/sign_in') userEntry(req);
 
   forward(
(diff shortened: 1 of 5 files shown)


jaredhirsch:
ah, looks like I need to update some tests

bin/router
```diff
@@ -160,8 +161,35 @@ wsapi.routeSetup(app, {
 //catch-all
 app.use(function(req, res, next) {
 
+  // utility function to log a bunch of stuff at user entry point
+  function userEntry(req) {
+    var ipAddress = req.connection.remoteAddress;
+    if (req.headers['x-real-ip']) ipAddress = req.headers['x-real-ip'];
```
shane-tomlinson, line 167: This is probably a big nono as it is definitely fingerprintable information.

bin/router
```diff
@@ -160,8 +161,35 @@ wsapi.routeSetup(app, {
...
+
+    var referer = null;
+    try {
+      // don't log more than we need
+      referer = urlparse(req.headers.referer).originOnly().toString();
+    } catch(e) {
+      // ignore malformed referrers.  just log null
+    }
+
+    metrics.report('signin', {
```
shane-tomlinson, line 177: What is the difference between metrics and store_kpi?

shane-tomlinson:
I am comfortable with this approach as a whole; I think collecting KPIs from the router directly neatly avoids writing new software to transfer and parse log files. It would be nice if there were a way to consolidate the myriad loggers into one.

I believe sending the IP address will not be allowed, do we need the IP address or just a count of hits? Are we looking for unique signins?

Before merge, I want to make sure we are not running afoul of any data safety rules/policies or else we'll have to rip it out of production.

jaredhirsch:
this has been lingering for a while. needs to be rebased, tests fixed, and is blocked by privacy reivew.

i'll push to an issue_3529_metrics_to_piggybank branch on the mozilla repo, in case of handoff to @shane-tomlinson  and close this out for now.

PR #3817, opened 2013-08-21
Train 2013.08.15 hotfix 3811

### LINKED ISSUE #3811: allowUnverified accounts can't login to non-allowUnverified RPs
@stephendonner found this bug in production, and @shane-tomlinson figured out STR locally.

STR:
1. create a new secondary with an RP that allows unverified accounts, like an FXOS phone
2. then, without clicking the link in your email, try to use that unverified account to log into an RP that does not allow unverified, like persona.org
3. when you enter the email into the dialog and hit 'next', the button will spin, but nothing will happen--you cannot log in.

We observed this today with the webqa test account not being able to log into production persona.org.

STR locally (thanks to shane):
1. load up the dev RP
2. click 'allow unverified'
3. create a new secondary acct
4. visit the dev login.persona.org and sign out
5. re-load up the dev RP
6. un-select 'allow unverified' if it is selected
7. load the dialog
8. enter the email address from step 3

The underlying bug is rather subtle and funky:
- [the authenticate module](https://github.com/mozilla/browserid/blob/6a8b8e/resources/static/dialog/js/modules/authenticate.js#L129-L137) fires 'email_chosen'
- the state machine's ['email_chosen' handler](https://github.com/mozilla/browserid/blob/6a8b8e/resources/static/dialog/js/misc/state.js#L460-L464) fires 'stage_reverify_email' if the address is unverified and the RP doesn't allowUnverified
- the state machine's ['stage_reverify_email' handler](https://github.com/mozilla/browserid/blob/6a8b8e/resources/static/dialog/js/misc/state.js#L493) fires a corresponding [action](https://github.com/mozilla/browserid/blob/6a8b8e/resources/static/dialog/js/modules/actions.js#L110) which calls into a [dialogHelper](https://github.com/mozilla/browserid/blob/6a8b8e/resources/static/dialog/js/misc/helpers.js#L76-L87) which, finally,
- calls [user.js.requestEmailReverify](https://github.com/mozilla/browserid/blob/6a8b8e/resources/static/common/js/user.js#L850-L861), which doesn't find anything in storage (since this is a different browser, I think?) and gives up.

This is an obscure but ugly edge case. :beers: to webqa for flushing it out

shane-tomlinson:
@lloyd, @6a68, @jrgm - here you go guys!

There is an ephemeral instance, https://unverified-signin.personatest.org

fixes #3811

### DIFF
--- a/resources/static/dialog/js/modules/authenticate.js
+++ b/resources/static/dialog/js/modules/authenticate.js
@@ -50,17 +50,29 @@ BrowserID.Modules.Authenticate = (function() {
     /*jshint validthis:true*/
     var self = this;
     /*
-     * If this is is a required email, we are making the assumption
-     * that it is a secondary address and has a password. The only two ways to
-     * get here with emailSpecified are post-reset-password where the email
-     * verification occurs in a second browser and when signed in to the
-     * assertion level and then choose a secondary backed address.
+     * If this is is a required email (emailSpecified && !emailMutable),
+     * we make the assumption that the address is a secondary with a password.
+     * emailSpecified can occur when (OR):
+     * 1) post-reset-password where the email verification occurs in a
+     *    second browser.
+     * 2) When signed in to the assertion level and then a fallback IdP backed
+     *    address is chosen.
+     *
+     * An unverified email address can occur when (OR):
+     * 1) User registers address on a site that uses allowUnverified and never
+     *    verifies the address.
+     * 2) User has an account with multiple addresses backed by the fallback
+     *    IdP, they perform a password reset, verify the first address,
+     *    then try to sign in using one of the other addresses in a browser
+     *    without a session.
+     * In either case, the user must enter the account password. If the address
+     * must be verified, verification will be taken care of by the state machine.
      */
     return (self.emailSpecified && !self.emailMutable) ||
               (info && info.email && info.type === "secondary" &&
                   (info.state === "known" ||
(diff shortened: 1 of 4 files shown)


resources/static/dialog/js/modules/authenticate.js
```diff
@@ -50,17 +50,29 @@ BrowserID.Modules.Authenticate = (function() {
...
+     *    then try to sign in using one of the other addresses in a browser
+     *    without a session.
+     * In either case, the user must enter the account password. If the address
+     * must be verified, verification will be taken care of by the state machine.
      */
     return (self.emailSpecified && !self.emailMutable) ||
               (info && info.email && info.type === "secondary" &&
                   (info.state === "known" ||
                    info.state === "transition_to_secondary" ||
-                   info.state === "unverified" && self.allowUnverified));
```
shane-tomlinson, line 63: This is the crux of the fix! We were being _too strict_ with this check. If an RP does not call .request with `allowUnverified: true` and the user types an unverified email address that has a password into the auth screen, the user should be forced to type the password and then sent to the "verify your address!" screen. If the RP calls .request with `allowUnverified: true`, the user should type in their password and then an assertion should be generated.

After the user types their password and authenticates with the backend, the "authenticated" signal will be triggered, and the state machine will take the correct action based on the "allowUnverified" flag from the RP.

jrgm:
Sorry @shane-tomlinson. I forgot to say thank you!

PR #3820, opened 2013-08-22
if no emails after sync, log user out

### LINKED ISSUE #1472: What should /sign_in dialog do if /wsapi/list_emails returns an empty list
We talked about this in the triage meeting, and noted that GH-759 is the covering issue that needs to be addressed. However, I'm filing this as a separate issue about how the dialog should behave if it ever gets back an empty email list (for whatever reason).

Steps to reproduce this empty list:
1. Initial condition: two browsers that have both already authenticated for the same email address. This account has only that one single email address associated with it.
2. With browser-1, click signin for beta.myfavoritebeer.org. In the /sign_in dialog, click 'this is not me'.
3. Enter the same email as before, then click 'forgot your password' and then 'reset password'. Load the link in the email, enter a "new" password, etc., and browser-1 is now re-authed for that email address, and is logged in to beta.myfavoritebeer.org.
4. With browser-2, click signin. 

Expected: 

Either I should have a choice to signin with the email I'd used previously with this browser, or I should be given the textbox to enter an email address and then password.

Actual:

I get a dialog box with no email list and a 'sign in' button. Clicking 'sign in' puts up an alert with 'The selected email is invalid or has been deleted'.

img

When the dialog comes up, the call to /wsapi/session_context returns 200 OK, with payload '"authenticated":true,"auth_level":"password"'.
The call to /wsapi/list_emails is also 200 OK, and returns payload '{}'. 

Workaround:

In browser-2, click 'this is not me', and then enter the email address and (now-not-forgotten) password.

seanmonstar:
fixes #1472

not actually a hotfix, my mistake. but it does fix a now 5-star.

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -1070,7 +1070,16 @@ BrowserID.User = (function() {
       User.checkAuthentication(function(authenticated) {
         if (authenticated) {
           User.syncEmails(function() {
-            complete(onComplete, authenticated);
+            // no emails means they must have been removed from this
+            // account. at any rate, the user can't do anything without
+            // any emails, so log them out.
+            if (storage.getEmailCount() === 0) {
+              User.logoutUser(function() {
+                complete(onComplete, false);
+              }, onFailure);
+            } else {
+              complete(onComplete, authenticated);
+            }
           }, onFailure);
         }
         else {
--- a/resources/static/test/cases/common/js/user.js
+++ b/resources/static/test/cases/common/js/user.js
@@ -845,6 +845,17 @@
     failureCheck(lib.checkAuthenticationAndSync);
   });
 
+  asyncTest("checkAuthenticationAndSync with no emails", function() {
+    xhr.setContextInfo("auth_level", "assertion");
+    xhr.useResult("no_identities");
+    storage.clear();
+
+    lib.checkAuthenticationAndSync(function(authed) {
+      equal(authed, false, "we are not authenticated");
+      start();
+    }, testHelpers.unexpectedXHRFailure);
+  });
+
   asyncTest("isEmailRegistered with registered email", function() {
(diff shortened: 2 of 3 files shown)


jrgm:
Thanks @seanmonstar. Works for me in testing locally; I'll let @shane-tomlinson or @6a68 look and merge. 

Sorry, for abusing 5-star (sort of sorry). This is something that people do hit, and it had been hanging around for too long.

shane-tomlinson:
@jrgm - do you want this to also be made into a hotfix against the current stage?

shane-tomlinson:
@seanmonstar, @jrgm - I am unable to reproduce using @jrgm's original steps, the email addresses are not being re-synced after clicking the "sign in" button. See comment https://github.com/mozilla/browserid/issues/1472#issuecomment-23082064.

For a full fix, "checkAuthenticationAndSync" should be called after clicking on "sign in". I'm not sure when this was removed.

seanmonstar:
This was the new STR I reported in the issue a month ago:
1. I logged in with a brand new email. 
2. in another browser, I logged into my normal account
3.  "Add another email", and added the same email I used previously in step 1.
4. Try to sign in with browser and email from step 1.
   
   It "stole" the email from the brand new account. Trying to use it, I was still logged in, but the list was empty.

jrgm:
In reply to:

```
@jrgm - do you want this to also be made into a hotfix against the current stage?
```

No. I regret bumping this to 5-star, as that was abusing that category. It's just that this had been hanging around for so long, and users kept occasionally bumping into it, leaving some users perhaps baffled by how to fix this for themselves. But it's been around for a while, so we can just pick it up in the next 08.28 train.

jaredhirsch:
@seanmonstar's STR worked for me to repro the bug and the fix locally. :beers:, merging in

jaredhirsch:
derp.

Issue #3779, opened 2013-08-13
[train-2013.07.31 and dev] consistent FE test failures with IE8 and IE9


jrgm:
There was a flaky test in train-2013.07.17, producing the message

```
"common/js/modules/interaction_data: addEvent with eventTime and duration - eventTime used as basis to calculate offset, duration is third item in event (1, 3, 4)"
```

in 2 of 5 runs against an ephemeral instance with IE8.

But in train-2013.07.31 and dev against respective ephemerals, I'm seeing consistent failures, summarized as:

```
{
  "IE9": {
    "6/10": "common/js/network: serverTime (1, 0, 1)",
    "10/10": "common/js/modules/interaction_data: start_time message sets the startTime to calculate event time offset (1, 0, 1)",
  },
  "IE8": {
    "10/10": "common/js/network: serverTime (1, 0, 1)",
    "10/10": "common/js/modules/xhr: xhr_delay not called for completed request (1, 1, 2)",
    "10/10": "common/js/modules/interaction_data: start_time message sets the startTime to calculate event time offset (1, 0, 1)"
  }
}
```

The js/network and js/modules/interaction_data are timing related on slower JS engines. 

But js/modules/xhr looks like a basic 'IE8 JS/DOM is not like the other ones' failure:

```
modules/xhr: xhr_delay not called for completed request (1, 1, 2)
Expected:  "undefined" 
Result:  "object" 
Diff:  "undefined" "object"
```

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #3836, opened 2013-08-27
Update to qunit 1.12.0 (fixes #3797, but some open issues for tests that never assert)

### LINKED ISSUE #3797: compile failures in ./resources/static/test are not detected by QUnit
This came up when I ran the redirect-flow branch on an ephemeral with IE8.

Stderr does spit out a SyntaxError message, but... 

... Despite a syntax error in resources/static/test/cases/common/js/storage.js, the Qunit (phantomjs or IE8) test reported "all passed".

The problem in this case was that the syntax error meant none of those tests for the storage case were executed. 

So, is there a setting of QUnit to fix this? Otherwise, the other defence would be jshint, but there's a lot of tuned-to-many-browsers changes that would need to happen. (I suppose another approach is to keep better track of how many tests should exist, but that's also failure prone).

jrgm:
(Obviously not ready for merging since it has failing tests.)

So, I was looking into setting up a `window.onerror` handler to deal with compile failures in QUnit tests (GH-3797). But in looking at QUnit source code, I realized that this already had been added some time ago. So by upgrading to latest QUnit, we fix GH-3797 for free. 

However, along the way, QUnit also added the requirement that tests must have at least one assertion (which is sensible). There were about 13 tests that did not meet this criterion, but most of them are cases where they only want to assert failure (e.g., this callback must never be called after xhr module shutdown). This PR fixes most of those, but I'm left with some cases that I don't grok and at least #2 and #3 look like bugs in the tests.

To ease review of these changes, I've put a `git diff -U12` with more context at http://people.mozilla.org/~jmorrison/issue-3797-diff-U12.txt. I've also setup https://jrgmnewqunit.personatest.org which has this PR, and https://jrgmoldqunit.personatest.org which is the same changes to the tests minus upgrading QUnit version (and that instance passes (modulo known failures GH-3779 on IE9 and IE8)).

The remaining failures:
- https://github.com/mozilla/browserid/blob/dev/resources/static/test/cases/dialog/js/misc/state.js#L389-L397

```
dialog/js/misc/state: primary_user_authenticating stops all modules (1, 0, 1) Rerun 4 ms
  Expected at least one assertion, but none were run - call expect(0) to accept zero assertions.
  Source:  
  @https://jrgmnewqunit.personatest.org/test/cases/dialog/js/misc/state.js:389
  @https://jrgmnewqunit.personatest.org/test/cases/dialog/js/misc/state.js:4
```

This one is flaky. The code when run in isolation from any other tests it always passes (i.e., no exception is thrown). But when run with all the other tests, most of the time an exception **is** thrown, but occasionally passes (calls `equal()`). Given the comment "ignore exception, it tries shutting down all the modules", perhaps this is acceptable, in which case we could just add `equal(true, "This is ok too")` in the `catch` clause. @shane-tomlinson: you wrote this test. Comments?
- https://github.com/mozilla/browserid/blob/dev/resources/static/test/cases/pages/js/verify_secondary_address.js#L169-L195

```
pages/js/verify_secondary_address: redirect: message shows with correct timeout (1, 0, 1) Rerun 45 ms
  Expected at least one assertion, but none were run - call expect(0) to accept zero assertions.
  Source:  
  QUnit.asyncTest@https://jrgmnewqunit.personatest.org/test/qunit/qunit.js:401
```

In this case, the mocked out replaceFormWithNotice callback is never called. @seanmonstar - Comments?
- https://github.com/mozilla/browserid/blob/dev/resources/static/test/cases/dialog/js/modules/primary_user_provisioned.js#L117-L131

```
dialog/js/modules/primary_user_provisioned: start controller with `add: true` adds email to user's list (1, 0, 1)Rerun30 ms
  Expected at least one assertion, but none were run - call expect(0) to accept zero assertions.
  Source:  
  QUnit.asyncTest@https://jrgmnewqunit.personatest.org/test/qunit/qunit.js:401
```

In this case, the `ready` callback is never called. @shane-tomlinson?

### DIFF
--- a/resources/static/test/cases/common/js/modules/interaction_data.js
+++ b/resources/static/test/cases/common/js/modules/interaction_data.js
@@ -330,6 +330,7 @@
     var startTime = new Date().getTime() - 1000;
     try {
       controller.addEvent("start_time", startTime);
+      ok(true); // assert affirmatively that the previous line did not throw.
     } catch(e) {
       ok(false);
     }
--- a/resources/static/test/cases/common/js/modules/xhr.js
+++ b/resources/static/test/cases/common/js/modules/xhr.js
@@ -238,6 +238,8 @@
   });
 
   asyncTest("success responses not called after module stops", function() {
+    expect(0); // neither the error nor success callbacks below should fire.
+
     xhr.get({
       url: "/wsapi/session_context",
       error: testHelpers.unexpectedXHRFailure,
@@ -251,7 +253,10 @@
   });
 
   asyncTest("error responses not called after module stops", function() {
+    expect(0); // neither the error nor success callbacks below should fire.
+
     transport.useResult("contextAjaxError");
+
     xhr.get({
       url: "/wsapi/session_context",
       error: testHelpers.unexpectedXHRFailure,
--- a/resources/static/test/cases/dialog/js/misc/internal_api.js
+++ b/resources/static/test/cases/dialog/js/misc/internal_api.js
@@ -135,6 +135,8 @@
   });
 
   test(".get with invalid string passed for options - assertion is not generated", function() {
+    expect(0); // we only assert when called unexpectedly
+
     dialogModule.get_success_value = "simulated_assertion";
 
     internal.get(ORIGIN, function(assertion) {
--- a/resources/static/test/cases/dialog/js/modules/authenticate.js
+++ b/resources/static/test/cases/dialog/js/modules/authenticate.js
@@ -87,6 +87,8 @@
   }
 
   function testInvalidPassword(password) {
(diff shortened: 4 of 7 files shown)


jrgm:
There's also another flaky test that shows up when run via selenium/sauce, and occasionally shows up manually for FF and maybe some others  

submit in password field with good password - skip to vpassword field                                                                       

https://github.com/mozilla/browserid/blob/dev/resources/static/test/cases/dialog/js/modules/set_password.js#L59-L71

shane-tomlinson:
@jrgm - This is looking really promising, though am  getting a couple of test failures. :/

shane-tomlinson:
Closing in favor of #3839

PR #3839, opened 2013-08-28
Qunit update

shane-tomlinson:
Reissue of #3836 fixing all the things.

@jrgm - wanna run through this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ notifications:
     use_notice: false
     skip_join: false
   email:
-    - lloyd@mozilla.com
     - jrgm@mozilla.com
 
 env:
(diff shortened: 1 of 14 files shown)


shane-tomlinson:
Now, why is Phantom failing when Firefox isn't.

jrgm:
Oh, haha. The new QUnit updates #qunit-testresult with the currently running test name. Our phantom runner was just polling for the string 'completed' in that innerText. So any test name with 'completed' would stop the test, and then we'd fail the test since it had no '\d+ failed' string ;-)

I looked at using the official runner addon from qunit, but it wasn't working, so I'll just make the pattern match a bit more specific on what stops the test (for now).

jrgm:
Okay, this passes on travis/phantom now, as well as with FF 23, Chrome 28, IE{8,9,10}, Opera 12, Safari 6 run via Saucelabs. (modulo known test failures in IE9 and IE8 -- GH-3779 -- which are independent of this update).

shane-tomlinson:
@jrgm - Man, I spent an hour and a half looking in the wrong direction yesterday w.r.t. the last commit to search for test failures.

PR #3840, opened 2013-08-28
If the user has not yet clicked on "Yes, this is my computer", remove the cert after generating an assertion.

shane-tomlinson:
@fmarier - This goes along with what we talked about on the front end. The odd thing is that when re-loading a page that uses .watch, if the user is authenticated to the IdP, a new cert is generated automatically. Is that the desired behavior?

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -1383,6 +1383,13 @@ BrowserID.User = (function() {
                   // issuer is used for B2G to get silent assertions to get
                   // assertions backed by certs from a special issuer.
                   storage.site.set(audience, "issuer", issuer);
+
+                  /**
+                   * If a user who signs with a primary address is not authenticated to Persona,
+                   * an assertion is first generated with the audience login.persona.org to sign the user
+                   * into Persona, another assertion is generated to sign the user into the RP. The cert should
+                   * be removed after generating the assertion for the RP.
+                   */
                   if (audience !== PERSONA_ORG_AUDIENCE && !storage.usersComputer.confirmed(email)) {
                     // If the user has not confirmed that this is their
                     // computer, immediately invalidate the cert so that nobody
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -26,7 +26,8 @@ BrowserID.User = (function() {
       userid,
       auth_status,
       issuer = "default",
-      allowUnverified = false;
+      allowUnverified = false,
+      PERSONA_ORG_AUDIENCE = "https://login.persona.org";
 
   var TRANSITION_STATES = [
     "transition_to_secondary",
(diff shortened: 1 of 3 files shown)


fmarier:
I think that's the right behaviour. IdPs should also have some kind of policy for ephemeral v. long-term sessions. They should be able to set that policy.

shane-tomlinson:
Uh oh, consistent Selenium test failures:

``` bash
➜  automation-tests git:(remove-certs-after-assertion-generation) ✗ ./scripts/run-all.js -p 5 -e onetime-cert -t remove-email
ignoring public-terminals.js
adding remove-email.js
Running 1 suite(s) on 1 platform(s), 1 at a time against https://onetime-cert.personatest.org
>........!.<

10/11 tests passed, here are your failures:
remove-email - win7_firefox_22 (1 failure(s), 0 warnings):
  9: log in to 123done using secondPrimaryEmail, errors:
    *   timeout hit
  >> https://saucelabs.com/tests/256f3622d5e64e2389d8fcd7df1d0bc5
```

shane-tomlinson:
@fmarier - I am going to make the wild claim that this is ready for review/merge if you are comfortable with it.

resources/static/common/js/user.js
```diff
@@ -1382,6 +1383,12 @@ BrowserID.User = (function() {
                   // issuer is used for B2G to get silent assertions to get
                   // assertions backed by certs from a special issuer.
                   storage.site.set(audience, "issuer", issuer);
+                  if (audience !== PERSONA_ORG_AUDIENCE && !storage.usersComputer.confirmed(email)) {
```
fmarier, line 1393: Why is `login.persona.org` a special audience in this case? (I'm not saying it's wrong, I just don't understand it...)

fmarier:
@shane-tomlinson I think it's ready as well. I've just added one small question inline.

jrgm:
Yeah, however, @shane-tomlinson's comment https://github.com/mozilla/browserid/pull/3840#issuecomment-23927803 still stands.

fmarier:
Oh, I thought the second commit he pushed was to fix that...

jrgm:
I tried it earlier this afternoon PDT with https://onetime-cert.personatest.org and it failed. But perhaps, that fix was not pushed to there.

resources/static/common/js/user.js
shane-tomlinson, line 1393: @fmarier - this is a strangeness in Persona. If a user who signs with a primary address is not authenticated to Persona, an assertion is first generated with the audience login.persona.org to sign the user into Persona, and once that occurs, another assertion is generated to sign the user into the RP.

shane-tomlinson:
@jrgm - the change to fix the Selenium test failure was contained within the tests, the ephemeral instance shouldn't need the code. Updating anyways, might as well test the proposed code in its entirety.

PR #3843, opened 2013-08-29
Email hint

shane-tomlinson:
@lloyd - BAM!

### DIFF
--- a/resources/static/dialog/js/modules/pick_email.js
+++ b/resources/static/dialog/js/modules/pick_email.js
@@ -121,6 +121,17 @@ BrowserID.Modules.PickEmail = (function() {
     this.publish("notme");
   }
 
+  function getPreselectedEmail(options) {
+    var emailHint = options.emailHint;
+    // Only use the email hint as the preselected email if the user owns the
+    // address, otherwise get the last used email for this site.
+    if (emailHint && user.getStoredEmailKeypair(emailHint)) {
+      return emailHint;
+    }
+
+    return user.getOriginEmail();
+  }
+
   var Module = bid.Modules.PageModule.extend({
     start: function(options) {
       var origin = user.getOrigin(),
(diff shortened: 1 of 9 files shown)


resources/static/dialog/js/modules/pick_email.js
```diff
@@ -134,7 +145,7 @@ BrowserID.Modules.PickEmail = (function() {
 
       self.renderForm("pick_email", {
         identities: identities,
-        siteEmail: user.getOriginEmail(),
+        preselectedEmail: getPreselectedEmail(options),
```
lloyd, line 148: integration into the picker makes sense, preferring email_hint over last used makes good sense.

resources/static/test/cases/dialog/js/misc/state.js
```diff
@@ -792,6 +819,18 @@
...
+    mediator.publish("start", {
+      emailHint: "testuser@testuser.com"
+    });
+
+    mediator.publish("pick_email");
+
+    testActionStarted("doPickEmail", {
+      emailHint: "testuser@testuser.com"
+    });
+  });
```
lloyd, line 833: w00t for tests.  <3

resources/static/dialog/js/modules/validate_rp_params.js
```diff
@@ -111,6 +111,15 @@ BrowserID.Modules.ValidateRpParams = (function() {
             "experimental_allowUnverified");
       }
 
+      // emailHint allows a site to provide a hint to the persona dialog
+      // of the desired email to verify.  It allows a site who knows an
+      // email but must verify it to offer a streamlined user experience
+      if (paramsFromRP.experimental_emailHint) {
+        params.emailHint = validateEmail(
+            paramsFromRP.experimental_emailHint,
+            "experimental_emailHint");
```
lloyd, line 120: tested - when the calling site passes in an invalid email hint we display an error in dialog.  This seems a reasonable approach.  Alternatively we could simply ignore the hint.  

But for an experimental feature, the first choice seems fine.

shane-tomlinson, line 120: I think what I did here seems heavy handed because it can be due to an input error by the user, in which case I feel like we should just ignore the hint?

lloyd:
r+ - this is great use of the experimental api to quickly land a feature requested by the community.  I see no reason to delay, hitting the green button!

PR #3854, opened 2013-09-04
Realms

### LINKED ISSUE #2555: Single Sign On
I'm working on Persona integration and looking for a few answers. I have a site that allows a user to login at http://www.example.com, but they can also login at https://www.example.com. Currently Persona considers these two domains to be completely different when determining if the user is logged in, I understand that they are considered cross-domain due to differing protocols.

However in practice we consider both of these to be our same site. The same would also be true if the user attempts to login at http://www.example.com and attempts to access http://othersubdomain.example.com. In practice we would consider the user to be logged into both, with all other forms of login currently used with our site.

Is it possible to allow a user to login with Persona for an entire hostname?
Such as considering the user logged in for *.example.com.

I understand there are possible security issues if this behavior was the default. However it would be nice to specify an option to handle an entire hostname.

If my understanding of how a user is logged in is completely off base, can you please explain how I can have a user logged into multiple domains technically considered cross-domain with one login action and to have the ability to log them out with one action.

seanmonstar:
This adds realm support to Persona, specifically using the watch() api.

Realms allow a group of sites to state they are connected as part of a group, and if a user logs into 1 site of the group, or realm, the user should be logged into all the sites in realm.

This can be done by passing an extra parameter to watch():

```
navigator.id.watch({
  realm: 'http://realm.123done.org'
});
```

The second part of the puzzle is that the watched realm must publish a well-known file describing everyone that is part of the realm.

```
GET http://realm.123done.org/.well-known/browserid-realm
{
  realm: ['http://foo.123done.org', 'http://bar.123done.org']
}
```

If the site's origin is found in the list of the realms found at the well-known file, then Persona will believe it is part of the realm. If a user had logged in to `foo.123done.org`, and then visited `bar.123done.org`, and that site also watches the same realm, the user will be automatically logged in upon page load.

fixes #2555
### Blockers to merging
- [ ] B2G testing with @jedp 
- [ ] QA? perhaps QA is involved once merged to dev?
- [ ] Plan for automated testing, how can we avoid regression?  Do we need a new (set) of RPs so we can confirm we don't regress on this feature?
- [ ] security testing with Yvan
- [ ] legal bug with @lloyd @seanmonstar and Jishnu
- [ ] identity-ops deploying new proxy https://github.com/mozilla/identity-ops/commit/e6e293149747a18278d79fe4d3b5fa24bbbd0e69


bin/proxy
```diff
@@ -20,11 +20,12 @@ var addy = config.has('bind_to.host') ? config.get('bind_to.host') : "127.0.0.1"
 forward.setTimeout(config.get('declaration_of_support_timeout_ms'));
 
 const allowed = /^https:\/\/[a-zA-Z0-9\.\-_]+\/\.well-known\/browserid(\?.*)?$/;
+const allowed_realm = /^https?:\/\/[a-zA-Z0-9\.\-_:]+\/\.well-known\/browserid-realm(\?.*)?$/;
 const allowed_kpi = /^https:\/\/[a-zA-Z0-9\.\-_]+\/wsapi\/interaction_data(\?.*)?$/;
 
 var server = http.createServer(function (req, res) {
   var url = req.url;
-  if (!allowed.test(url) && !allowed_kpi.test(url) ) {
+  if (!allowed.test(url) && !allowed_realm.test(url) && !allowed_kpi.test(url) ) {
```
seanmonstar, line 28: will this also require changes to our production proxy?

lib/request.js
```diff
@@ -0,0 +1,29 @@
+const request = require('request');
```
seanmonstar: i dearly wanted to fix up `lib/primary.js` to use `request` also, but it's off-topic of this feature.

resources/static/test/cases/dialog/js/misc/internal_api.js
```diff
@@ -101,12 +101,20 @@
...
 
         internal.get(ORIGIN, function(assertion) {
-          ok(assertion, "assertion generated using stored email address for site.");
+          //ok(assertion, "assertion generated using stored email address for site.");
+          ok(true);
+          // TODO: what the heck does this even test?
+          // It seems to be able to generate an assertion silently for a
+          // site if the user has logged into it before, even if the
+          // user is not currently signed in. Is that really what is
+          // wanted? That sounds broken to my limited understanding.
```
seanmonstar, line 117: B2G brokeness?

resources/static/include_js/_include.js
```diff
@@ -354,6 +357,8 @@
 
       options.start_time = (new Date()).getTime();
 
+      options.realm = realm;
```
seanmonstar, line 360: This realm value is set from `watch()`. My guess is that the client code in **B2G** would need to be adjusted to do the same thing. Is that true? Does that suck horribly?

example/rp/index.html
```diff
@@ -179,6 +179,7 @@
 
 navigator.id.watch({
   loggedInUser: (storage.loggedInUser === 'null') ? null : storage.loggedInUser,
+  realm: 'http://localhost:10001',
```
shane-tomlinson: _must_ a site specify a realm? If not, can we make this optional or create a new example page for it?

lib/realm.js
```diff
@@ -0,0 +1,43 @@
...
+      err = new Error('missing realm json');
+    } else if (!util.isArray(body.realm)) {
+      err = new Error('realm property is not an array');
+    } else {
+      body.realm = body.realm.filter(function(value) {
+        try {
+          validate.is.origin(value);
+          return true;
+        } catch(err) {
+          return false;
```
shane-tomlinson: Should there be some sort of error surfaced here?

lib/request.js
shane-tomlinson: Add an issue for it!

lib/realm.js
```diff
@@ -0,0 +1,43 @@
...
+
+const WELL_KNOWN_PATH = '/.well-known/browserid-realm';
+
+function isJsonType(res) {
+  var contentType = res.headers['content-type'];
+  return contentType && contentType.indexOf('application/json') === 0;
+}
+
+exports.checkSupport = function realmSupport(realm, callback) {
+  request(realm + WELL_KNOWN_PATH, { json: true }, function onRequest(err, res, body) {
```
shane-tomlinson: Must realms be defined at an https endpoint like /.well-known/browserid?

lib/request.js
```diff
@@ -0,0 +1,29 @@
...
+      host: HTTP_PROXY.host,
+      port: HTTP_PROXY.port
+    };
+  }
+
+  if (!('timeout' in options)) {
+    options.timeout = REQUEST_TIMEOUT;
+  }
+
+  options.strictSSL = true;
```
shane-tomlinson: Ah yes, it looks like declaration of support must be SSL.

lockdown.json
```diff
@@ -33,6 +42,9 @@
   "bindings": {
     "1.0.0": "c3ccde60e9de6807c6f1aa4ef4843af29191c828"
   },
+  "boom": {
```
shane-tomlinson: I like boom, can we get rid of our homegrown error code library and use it instead?

resources/static/common/js/storage.js
```diff
@@ -184,38 +184,15 @@ BrowserID.Storage = (function() {
...
-
-    storage.siteInfo = JSON.stringify(allSiteInfo);
-  }
-
-  function siteGet(site, key) {
-    var allSiteInfo = JSON.parse(storage.siteInfo || "{}");
-    var siteInfo = allSiteInfo[site];
-
-    return siteInfo && siteInfo[key];
+    generic3KeySet('siteInfo', site, key, value);
```
shane-tomlinson, line 193: I dig. I was also looking to replace all this cruft with https://github.com/shane-tomlinson/easierStorage

resources/static/common/js/storage.js
```diff
@@ -271,6 +268,11 @@ BrowserID.Storage = (function() {
       delete allSiteInfo[site].logged_in;
     }
     storage.siteInfo = JSON.stringify(allSiteInfo);
+    var allRealmInfo = JSON.parse(storage.realmInfo || "{}");
+    for (var realm in allRealmInfo) {
+      delete allRealmInfo[realm].logged_in;
+    }
+    storage.realmInfo = JSON.stringify(allRealmInfo);
```
shane-tomlinson, line 287: realmInfo needs to be added to the list of localStorage items initialized on startup or else IE8 won't know how to sync between windows. Does realmInfo make sense in its own namespace or can this info be jammed in under siteInfo?

resources/static/common/js/user.js
```diff
@@ -1527,35 +1538,61 @@ BrowserID.User = (function() {
...
-          if (isTransitioning(info.state))
-            return complete(onComplete, null, null);
-
-          // If there has not been an issuer change and Persona's view of the
-          // world agrees with the sites, then skip assertion generation.
-          if (( ! info.issuerChange)
-                  && (loggedInEmail === siteSpecifiedEmail)) {
-            return complete(onComplete, loggedInEmail, null);
+        User.isRealmValid(function(realmAgrees) {
+          var loggedInEmail = User.getOriginLoggedIn();
```
shane-tomlinson, line 1546: can User.isRealmValid be called after fetching getOriginLoggedIn to reduce the number of network calls in the case that there is an email saved for the domain?

resources/static/common/js/user.js
```diff
@@ -1527,35 +1538,61 @@ BrowserID.User = (function() {
...
+            if (isTransitioning(info.state))
+              return complete(onComplete, null, null);
+
+            // If there has not been an issuer change and Persona's view of the
+            // world agrees with the sites, then skip assertion generation.
+            if (( ! info.issuerChange)
+                    // could be different if logged into realm but not
+                    // site, so we need to generate a site-specific
+                    // assertion
+                    && (loggedInEmail === User.getOriginLoggedIn())
```
shane-tomlinson, line 1571: loggedInEmail was set to User.getOriginLoggedIn() in line 1546, can it be modified between then and here? If not, this check does not seem necessary.

resources/static/common/js/user.js
```diff
@@ -1527,35 +1538,61 @@ BrowserID.User = (function() {
...
+                    && (loggedInEmail === siteSpecifiedEmail)) {
+              if (realmAgrees) {
+                User.setRealmLoggedIn(loggedInEmail);
```
shane-tomlinson, line 1574: Is this necessary from getSilentAssertion or should calling User.setRealmLoggedIn be done only if the user interacts with the dialog?

resources/static/common/js/user.js
```diff
@@ -1570,6 +1607,26 @@ BrowserID.User = (function() {
...
+              // don't remove user's realm association if site isn't
+              // reall part of realm.
+              //
+              // Example: 
+              // 1. foo.com claims { realm: 'bar.com' }
+              // 2. bar.com's browserid-realm does not include 'foo.com'
+              // 3. logging out of foo.com should not log out user from
+              //    bar.com realm, because bar.com does not agree.
+              if (realmsAgree) {
+                storage.realm.remove(realm, "logged_in");
```
shane-tomlinson, line 1624: If the realms DO agree, does signing out of foo.com sign the user out from all sites specified by bar.com?

resources/static/common/js/user.js
```diff
@@ -1645,6 +1702,52 @@ BrowserID.User = (function() {
...
+      storage.site.set(origin, "logged_in", email);
+    },
+    
+    getOriginLoggedIn: function getOriginLoggedIn() {
+      return storage.site.get(origin, "logged_in");
+    },
+
+    setRealmLoggedIn: function setRealmLoggedIn(email) {
+      if (realm) {
+        storage.realm.set(realm, "logged_in", email);
```
shane-tomlinson, line 1727: Do you think it makes sense to have a "realm" model to put a lot of this code? I am in the process of separating out a lot of the RP related info into its own model to simplify things and realm information might be a good candidate also.

shane-tomlinson, line 1727: For more info: https://github.com/shane-tomlinson/browserid/blob/fdc5ddd920b72ac7d584470e5d51955318772bad/resources/static/common/js/models/rp_info.js and the overall approach in https://github.com/shane-tomlinson/browserid/pull/12/files to reduce the number of places where RP info is stored.

resources/static/include_js/_include.js
shane-tomlinson, line 360: @jedp - what would we need to do to surface realms on FirefoxOS? @seanmonstar - at a minimum, we'll have to surface the new option in the internal api (if it hasn't been done already)

resources/static/dialog/js/misc/internal_api.js
```diff
@@ -159,19 +159,10 @@
...
-      // User must be authenticated to get an assertion.
-      if(authenticated) {
-        user.getAssertion(email, user.getOrigin(), function(assertion) {
-          complete(assertion || null);
-        }, complete.curry(null));
-      }
-      else {
-        complete(null);
-      }
+    user.getSilentAssertion(email, function(_email, assertion) {
```
shane-tomlinson, line 162: Does getSilentAssertion follow the exact same semantics?

resources/static/dialog/js/misc/state.js
```diff
@@ -556,15 +556,28 @@ BrowserID.State = (function() {
...
           assertion: info.assertion,
           email: self.email
         });
+
+        complete(info.complete);
+      }
+
+      if (info.assertion !== null) {
+        self.success = true;
+        user.setOriginLoggedIn(self.email);
```
shane-tomlinson, line 572: I feel like this should be done in modules/generate_assertion.js, it doesn't seem to make much sense here.

resources/static/dialog/js/misc/state.js
```diff
@@ -556,15 +556,28 @@ BrowserID.State = (function() {
...
+
+        // at this point, we don't want to blow up at the very end if a
+        // realmInfo request fails, do we?
+        user.isRealmValid(function(isValid) {
```
shane-tomlinson, line 576: Perhaps all of the realm checking/setting should be done in modules/generate_assertion.js instead of here.

resources/static/dialog/js/modules/validate_rp_params.js
```diff
@@ -201,6 +205,18 @@ BrowserID.Modules.ValidateRpParams = (function() {
     return issuer;
   }
 
+  function fixupRealm(realm) {
+    // A realm should only be scheme + host (+ port).
+    /*jshint newcap:false*/
+    var u = URLParse(realm);
+    if (u.scheme + '://' + u.authority !== realm) {
```
shane-tomlinson: My tiny brain has a hard time with the order of operations, can you add a ( ) around the `u.scheme...u.authority`?

tests/address-info-test.js
```diff
@@ -15,10 +15,9 @@ vows = require('vows'),
 start_stop = require('./lib/start-stop.js'),
 wsapi = require('./lib/wsapi.js'),
 db = require('../lib/db.js'),
-primary = require('./lib/primary.js');
+primary = require('./lib/primary.js'),
 config = require('../lib/configuration.js'),
 bcrypt = require('bcrypt'),
-primary = require('./lib/primary.js'),
```
shane-tomlinson, line 21: nice cleanup.

example/rp/index.html
seanmonstar: A site doesn't have to. I was trying to figure out how to make this optional. Perhaps if the page is loaded with `?realm=1` ?

lib/realm.js
seanmonstar: Nope, http and https are both valid. The protocol is required in the listed realm argument (wsapi/realm_info validates its a proper `origin`).

lib/realm.js
seanmonstar: This is a filter on the array provided by a support document. If a value in the array isn't a proper origin, it gets dumped.

So, `{ realm: ['foo.com', 'http://bar.com', 'baz'] }` becomes `{ realm: ['http://bar.com'] }`. Our client side doesn't care if there's errors or if it's malformed, simply if the declared realm is in the support document.

lib/request.js
seanmonstar: No, it doesn't have to be. This option is only used for HTTPS requests. But considering we pretty much always use an outgoing proxy, it's likely never actually needed. Still, doesn't hurt, and is safer.

resources/static/common/js/user.js
seanmonstar, line 1546: I had that originally, but found a case where it's pretty much always needed if a realm is declared (if no realm is declared, not request is actually made).

If you've logged into https://webmaker.org, and they later setup their realm, when you visit webmaker.org again, it should notice the realm, check its validity, and then set the logged_in flag on the realm as well. Even though there was already a logged_in flag on the site. Yea?

resources/static/common/js/user.js
seanmonstar, line 1571: It can be at L1552. If there's on logged_in email from site, but there is from a valid realm, then I set `loggedInEmail` to that instead. Then, this check is to see if it was the same.

resources/static/common/js/user.js
seanmonstar, line 1574: This is necessary for the example I describe with webmaker.org, and turning on realm support after you've already logged in. It should upgrade you to being logged into the realm, methinks.

resources/static/common/js/user.js
seanmonstar, line 1624: Oh dang, I forgot that case. It probably should, huh? That means I'll need to track all the domains that are related in realmInfo.

resources/static/common/js/storage.js
seanmonstar, line 287: good catch.

I created realmInfo namespace because I wasn't sure if I was going to need more fields besides just `logged_in`. It looks like I'll need more to keep track of all current sites related to a realm.

resources/static/common/js/user.js
seanmonstar, line 1727: Hmm, interesting. If it seems like `siteInfo` should do the same, then I should do that with `realmInfo`. Should they both?

resources/static/dialog/js/misc/internal_api.js
seanmonstar, line 162: No, it does not. Previously, this would create an assertion if the user didn't have `logged_in` set. I don't know if that was desired. I don't know when get({ silent: true }) is called from B2G. Hence my cry for help.

resources/static/dialog/js/misc/state.js
seanmonstar, line 572: True. I was just updating L561 to use `user.setOriginLoggedIn` instead of accessing `storage` directly. I can move both to generate_assertion, I think.

lib/realm.js
shane-tomlinson: I admit, this makes me uncomfortable - is there any way that defining a realm file over http can be used to hijack a user? @fmarier or @warner, can we get your opinions here?

bin/proxy
shane-tomlinson, line 28: Possibly, @gene1wood or @jrgm?

lib/realm.js
shane-tomlinson: ok, that makes sense.

lib/wsapi/realm_info.js
```diff
@@ -0,0 +1,28 @@
...
+exports.writes_db = false;
+exports.authed = false;
+exports.args = {
+  realm: 'origin'
+};
+exports.i18n = false;
+
+exports.process = function realm_info(req, res) {
+  var domain = req.params.realm.toLowerCase();
+  realm.checkSupport(domain, function onRealm(err, body) {
```
shane-tomlinson, line 17: I'm not sure whether this comment is better here or in `../ream` - it doesn't look like this information is cached, which could cause a heavy load on the realm `host`. Perhaps it is not needed at this point, but we should give it some thought.

lockdown.json
```diff
@@ -280,6 +322,9 @@
   "nub": {
     "0.0.0": "b369bd32bdde66af59605c3b0520bc219dccc04f"
   },
+  "oauth-sign": {
```
shane-tomlinson: eh?

lockdown.json
```diff
@@ -352,8 +398,10 @@
   "slide": {
     "1.1.4": "2b23f1949b369ed61d22bd6570ff0320302fc8df"
   },
+  "sntp": {
```
shane-tomlinson: eheh?

lockdown.json
```diff
@@ -371,6 +419,9 @@
   "traverse": {
     "0.6.3": "a053ffa1b6179b9240ea16d74bfd604bd6b6e41b"
   },
+  "tunnel-agent": {
```
shane-tomlinson: and finally, eheheh? Are these all required by `request`?

resources/static/common/js/user.js
```diff
@@ -17,6 +17,7 @@ BrowserID.User = (function() {
       provisioning = bid.Provisioning,
       addressCache = {},
```
shane-tomlinson, line 18: Is it time we consider a simple caching module?

bin/proxy
gene1wood, line 28: @shane-tomlinson Will this change result in persona initiating outbound connections out to the internet to new destinations (destinations other than sites' well-known/.browserid files and identity bridge endpoints [yahoo, gmail, etc])?

bin/proxy
```diff
@@ -20,11 +20,12 @@ var addy = config.has('bind_to.host') ? config.get('bind_to.host') : "127.0.0.1"
 forward.setTimeout(config.get('declaration_of_support_timeout_ms'));
 
 const allowed = /^https:\/\/[a-zA-Z0-9\.\-_]+\/\.well-known\/browserid(\?.*)?$/;
+const allowed_realm = /^https?:\/\/[a-zA-Z0-9\.\-_:]+\/\.well-known\/browserid-realm(\?.*)?$/;
```
shane-tomlinson: do we need to allow the extra .\* on the end? Does it give us anything?

bin/proxy
shane-tomlinson, line 28: @gene1wood - indeed, the new outbound request will search for realm files at <rp_location>/.well-known/browserid-realm

gene1wood, line 28: Proxy server support added here https://github.com/mozilla/identity-ops/commit/e6e293149747a18278d79fe4d3b5fa24bbbd0e69

When QA wants to see this code in stage they'll need to request that new proxy server AMIs be built with this new code in the process. (cc @jrgm)

shane-tomlinson, line 28: Excellent, thanks @gene1wood

lib/realm.js
seanmonstar: So, I could have missed something, but here's my reasoning as to why we should allow http as well.

First, we allow RPs to use Persona for login even over http. Their sites don't need to be secured with HTTPS. Of course, there's risk for the sites, that an attacker could adjust DNS to make the user log into the wrong site, or an attack could capture the assertion as the site POSTs it to their own server for verification, and login as the user. Still, we let the sites accept those risks, instead of demanding everyone who wants to use Persona needs to setup HTTPS.

Likewise, the realm is simply a group of sites that a user is logging in to. An attacker could hijack DNS to an http site, and declare a different realm than normal. Or they could serve up their own browserid-realm file, which could include a different list of sites. The user isn't automatically logged into any of those sites at that time. The user then needs to visit one of those sites, and that site (whether normally or by the attacker) would need to declare being part of the hijacked realm.

In the end, the worst that could happen is a user is logged into a realm or site that isn't really part of the realm.

lib/wsapi/realm_info.js
seanmonstar, line 17: Ideally, we could let squid handle caching this file, just as we do when requesting the IdP support document. possible @gene1wood ?

lockdown.json
seanmonstar: yea they are... i can switch away from `request`, but its the most popular and tested library.

resources/static/common/js/user.js
seanmonstar, line 18: eh, seems like more trouble than it's worth. maybe if we wanted LRUCache, but otherwise, a simple object is simple.

lib/realm.js
gene1wood: @seanmonstar If we want https and http for the /.well-known/browserid-realm URL I'll need to tweak my proxy config change (I'd assumed this was https only like /.well-known/browserid

lib/wsapi/realm_info.js
gene1wood, line 17: @seanmonstar The squid proxy server will respect whatever caching directives it receives from the target site in the HTTP headers that come with the /.well-known/browserid-realm document. If the target asserts long cache times, the squid proxy will respect that and not make an outbound request.

fmarier:
Noting that a security review (by Yvan and others on the security assurance team) of this feature is tentatively scheduled for Wednesday 2013-09-11 at 13:00.

fmarier:
Regarding the requirement (or lack of) for HTTPS when fetching the realm document, I would vote for requiring SSL.

Sean is right, we don't require it for RPs. That's a compromise we made because so many RPs are served over HTTP and requiring HTTPS would hurt our adoption. So it's quite a reasonable compromise. However, let's be honest: these RPs are totally insecure. Anybody sniffing the network traffic [could steal the session cookies and take over accounts](http://codebutler.github.io/firesheep/).

Now, we're adding SSO functionality for sites that want to cooperate with each other. I think it's perfectly acceptable to raise the bar here. Requiring SSL would mean that, at the very least, anybody using this would need to get a cert for `realm.theircompany.com` and then serve the JSON file from there over HTTPS. That's similar to the HTTPS requirement we have for RP logo. I think it's pretty easy to satisfy. Maybe there's a problem I'm not seeing?

seanmonstar:
Thinking about it more, I guess it's not too bad to require HTTPS for browserid-realm file. It reduces the amount of attack space. I still think that sites that want to participate in the realm shouldn't be required to use HTTPS, but I was conflating the 2. I figured the `realm` value in watch() and in the well-known needed to have the same format (including a scheme).

Since we'll assume HTTPS for the well-known, that cleans up the realm property to this:

``` js
navigator.id.watch({
  realm: 'foo.com' // not https://foo.com
});
```

seanmonstar:
Sadly, this means testing on the local example rp won't work, since it now requires https, and I'll have to fudge with the tests also. https is mean.

ozten:
Can you build a `SHIMMED_PRIMARIES` like mechanism which subverts `https`?

seanmonstar:
@ozten I can, but I dislike it. _hacks away_

jrgm:
You can use something like stunnel[1] to get https working with the local environment.
[1] https://www.stunnel.org/

ozten:
Word. I also dislike SHIMMED_PRIMARIES ( I think a config file or something would be nicer), but something like it.

seanmonstar:
Yea, I've just now make a `SHIMMED_REALMS` dealio, and it works. But I had to remove 2 tests that were specific to http requests (status code and content-type), which is unfortunate.

seanmonstar:
New commit requires that the browserid-realm file be served over HTTPS.

lib/realm.js
```diff
@@ -0,0 +1,87 @@
...
+const util = require('util');
+const validate = require('./validate');
+
+const WELL_KNOWN_PATH = '/.well-known/browserid-realm';
+
+const SHIMMED_REALMS = {};
+
+// Support for "shimmed realms" for local development.
+// CSV values:
+//  <realm>|<browserid-realm filepath>,
```
shane-tomlinson, line 18: This took me a while to figure out how to do from bash command line:

``` bash
SHIMMED_REALMS=127.0.0.1\|example/rp/.well-known/browserid-realm npm start
```

Can you add a comment about how to make it so? This info should probably go into the [front end wiki](https://github.com/mozilla/browserid/wiki/Front-End-Development) as well.

lib/realm.js
```diff
@@ -0,0 +1,87 @@
...
+function isJsonType(res) {
+  var contentType = res.headers['content-type'];
+  return contentType && contentType.indexOf('application/json') === 0;
+}
+
+function onRealmInfo(err, body, callback) {
+  if (err) {
+    // just pass err
+  } else if (!(body && body.realm)) {
+    err = new Error('missing realm json');
```
shane-tomlinson: Would it make the code clearer to use these as guard clauses and immediately return:

``` js
if (err) return callback(err);
if (!(body && body.realm)) return callback(new Error('missing realm json'));
...
```

lib/validate.js
```diff
@@ -18,6 +18,7 @@ httputils = require('./httputils.js'),
 check = require('validator').check;
 
 var hostnameRegex = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$|^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$/;
+var hostRegex = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$|^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])(:\d{1,5})?$/;
```
shane-tomlinson: Head explosion. What are the differences between this and hostnameRegex? Both allow IP addresses, both allow domain names. It looks like hostRegex allows a port, if that is so, can you add a comment?

shane-tomlinson: That reminds me, is an RP able to specify the port the realm's support document is located on?

lib/realm.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+exports.checkSupport = function realmSupport(realm, callback) {
+  if (SHIMMED_REALMS[realm]) {
+    loadShimmedRealm(realm, callback);
+  } else {
+    request('https://' + realm + WELL_KNOWN_PATH, { json: true }, function onRequest(err, res, body) {
+      if (err) {
+        logger.debug(err);
+      } else if (!isJsonType(res)) {
```
shane-tomlinson: Same comment about guard clauses here. Secondly, if onRealmInfo just calls `callback` when there is error, can we just call `callback` directly from here?

lib/wsapi/realm_info.js
```diff
@@ -0,0 +1,28 @@
...
+    var json = {};
+    if (err) {
+      logger.info('"' + domain + '" realm support is misconfigured: ' + err);
+      json.realm = [];
```
shane-tomlinson, line 24: IIUC, the intent here is to not block users from signing in to a site if the support document is misconfigured?

resources/static/common/js/network.js
```diff
@@ -526,6 +526,25 @@ BrowserID.Network = (function() {
...
     /**
+     * Get infornation about a browserid-realm.
+     * @method realmInfo
+     * @param {string} realm - Hostname of realm to check.
+     * @param {function} onComplete - Called with an object on success,
+     *   contained these properties:
+     *     realm: array of hostnames
+     * @param {function} onFailure - Called on XHR failure.
+     */
+    realmInfo: function realmInfo(realm, onComplete, onFailure) {
```
shane-tomlinson, line 537: Is this tested via the tests for getSilentAssertion?

resources/static/common/js/user.js
shane-tomlinson, line 18: You know, the caching really only comes to play in the tests. I so wanna get rid of the global-ish state here, but that's for another PR as well.

resources/static/common/js/storage.js
```diff
@@ -775,6 +781,25 @@ BrowserID.Storage = (function() {
        * Clear any RP Redirection info
        */
       clear: clearRpRequestInfo
+    },
+
+    /*
+     * Info used regarding realms, such as if a user has signed into a
+     * realm.
+     */
+    realm: {
```
shane-tomlinson, line 804: Can we get a couple of explicit tests for these?

resources/static/dialog/js/misc/internal_api.js
shane-tomlinson, line 162: @jedp - HALP!

shane-tomlinson:
While testing this, I came across a problem: I can't log out!

STR

1) Set up dev environment to use a shimmed realm of 127.0.0.1
2) In tab 1, load up http://127.0.0.1:10001/?realm=127.0.0.1
3) In tab 2, load up http://localhost:10001/?realm=127.0.0.1
4) Sign in to the RP in tab 1
5) Sign in to the RP in tab 2
6) In tab 2, click "logout". Refresh the page. Still logged in... :(

lib/validate.js
seanmonstar: Now that I think about it, it doesn't really make sense, since https is required. I added this mostly to allow local testing, but now with shimmed realms, it's not needed. I'll just use the `hostname` instead.

lib/wsapi/realm_info.js
seanmonstar, line 24: correct. i'll add a comment say as much

seanmonstar:
@shane-tomlinson ah, I just found out why. The communication_iframe doesn't call `user.logout()`, instead it has it's own implementation. Can you of a reason why? I can't, so I'll likely just have it call `User.logout()`.

resources/static/common/js/user.js
seanmonstar, line 18: Yea... but the caching is useful, because there's a couple places along the way that we need to call realmInfo(), and that request makes an outbound request over the intertubes. It'd really slow things down to do that multiple times, so I need to cache the result.

But simply storing a cache of `realmInfo` that only stores one value felt naive, since you could technically call `watch()` with a different realm while it's in the middle of doing `getSilentAssertion()`, which would change the realm. Maybe. Stuff.

resources/static/common/js/network.js
seanmonstar, line 537: yea, getSilentAssertion calls User.isRealmValid, which calls network.realmInfo(realm).

resources/static/common/js/storage.js
seanmonstar, line 804: Ok. Though, it's using pretty much the exact same code as `siteInfo`, just under a different namespace...

seanmonstar:
@shane-tomlinson Newest commit tries to address your concerns. Please berate if I forgot any.

jaredhirsch:
@seanmonstar anything I can do to push this thing along?

seanmonstar:
@6a68 blockers are in a checklist at top of this PR.

currently, trying to deploy a couple of 123dones that use realms, so that security and @jedp can test against them

jedp:
There are some interesting issues with b2g.  (What else is new?)  I think we're going to need some platform code to make this work.

Basically, the removal of items from localStorage fires events that we can catch on normal web sites, but on b2g, data siloing and the discarding of the persona iframe when RPs are not using it (a we-will-kill-you-if-you-do-not-do-this requirement from platform) conspire to make it impossible for other RPs to hear about the logout unless we write some native code to fire `onlogout` for every RP with the same realm.  If that's the only solution, backwards compatibility with existing FirefoxOS devices is not possible.

I've filed this bug for FirefoxOS: https://bugzilla.mozilla.org/show_bug.cgi?id=919157

@shane-tomlinson maybe we can pow-wow quickly about this for a sanity check?  Make sure I've not got this completely wrong?

lloyd:
wild question.  If we didn't have to worry about logout events, how much simpler would this code be?  In that case, we would simply flip the logged-on bit at login time, and users would logout of each realm independently...  right?

callahad:
See my nay-saying re: logout events in https://github.com/mozilla/browserid/issues/3915#issuecomment-25108225 :)

jaredhirsch:
@seanmonstar I think, as far as automated tests go, having 2-3 long-lived subdomains of 123done is perfectly fine. They're distinct from the perspective of the same-origin policy, and that's the simplest solution I can think of.

I'd check that box and get on with the other 5. Do you need help with writing automated tests?

jaredhirsch:
Closing because upcoming API changes (see #3961) remove realms support. @seanmonstar please reopen if I'm off base.

PR #3857, opened 2013-09-04
RP branded emails.

shane-tomlinson:
- Create a new model on the front end, rp_info.js, to keep things sane.
- Add new backend validation types: color and image
- Add backend tests to check backgroundColor and siteLogo
- Update the copy in HTML and text emails as per @ryanfeeley's suggestions.
- Loads of email styling updates to work across webmail and mobile.
- Use a common _header.html.ejs and _footer.html.ejs shared across all emails so updates are simpler
- Simplify the email template configuration - only specify the template filename, not the full path.

### DIFF
--- a/lib/email.js
+++ b/lib/email.js
@@ -39,34 +39,45 @@ const templates = {
   "new": {
     landing: 'verify_email_address',
     subject: _("%(site)s: Complete sign-in"),
-    template: fs.readFileSync(path.join(TEMPLATE_PATH, 'new.ejs')),
-    templateHTML: fs.readFileSync(path.join(TEMPLATE_PATH, 'new.html.ejs'))
+    template: 'new.ejs',
+    templateHTML: 'new.html.ejs'
   },
   "reset": {
     landing: 'reset_password',
     subject: _("%(site)s: Complete password reset"),
-    template: fs.readFileSync(path.join(TEMPLATE_PATH, 'reset.ejs')),
-    templateHTML: fs.readFileSync(path.join(TEMPLATE_PATH, 'reset.html.ejs'))
+    template: 'reset.ejs',
+    templateHTML: 'reset.html.ejs'
   },
   "confirm": {
     landing: 'confirm',
     subject: _("%(site)s: Complete sign-in"),
-    template: fs.readFileSync(path.join(TEMPLATE_PATH, 'confirm.ejs')),
-    templateHTML: fs.readFileSync(path.join(TEMPLATE_PATH, 'confirm.html.ejs'))
+    template: 'confirm.ejs',
+    templateHTML: 'confirm.html.ejs'
   },
   "transition": {
     landing: 'complete_transition',
     subject: _("%(site)s: Complete sign-in"),
-    template: fs.readFileSync(path.join(TEMPLATE_PATH, 'transition.ejs')),
-    templateHTML: fs.readFileSync(path.join(TEMPLATE_PATH, 'transition.html.ejs'))
+    template: 'transition.ejs',
+    templateHTML: 'transition.html.ejs'
   },
 };
 
-// now turn file contents into compiled templates
+// turn template file contents into compiled templates
 Object.keys(templates).forEach(function(type) {
-  templates[type].template = ejs.compile(templates[type].template.toString());
+  var templatePath = path.join(TEMPLATE_PATH, templates[type].template);
+  var template = fs.readFileSync(templatePath);
+
+  templates[type].template = ejs.compile(template.toString(), {
+    filename: templatePath
+  });
+
   if (templates[type].templateHTML) {
-    templates[type].templateHTML = ejs.compile(templates[type].templateHTML.toString());
+    var templateHTMLPath = path.join(TEMPLATE_PATH, templates[type].templateHTML);
+    var templateHTML = fs.readFileSync(templateHTMLPath);
+
+    templates[type].templateHTML = ejs.compile(templateHTML.toString(), {
+      filename: templateHTMLPath
+    });
   }
 });
 
@@ -88,8 +99,17 @@ exports.setInterceptor = function(callback) {
   interceptor = callback;
 };
 
+function getBackgroundColor(backgroundColor) {
+  backgroundColor = backgroundColor || "45505b";
+  if (! /^#/.test(backgroundColor)) {
+    backgroundColor = "#" + backgroundColor;
+  }
+  return backgroundColor;
+}
+
+
 //TODO send in localeContext
-function doSend(email_type, email, site, secret, langContext) {
+function doSend(email_type, email, site, secret, langContext, backgroundColor, siteLogo) {
   if (!templates[email_type]) throw "unknown email type: " + email_type;
 
   // remove scheme from site to make it more human
@@ -107,12 +127,15 @@ function doSend(email_type, email, site, secret, langContext) {
     // log verification email to console separated by whitespace.
     console.log("\nVERIFICATION URL:\n" + public_url + "\n");
   } else {
+
     var templateArgs = {
       link: public_url,
       site: site,
       gettext: GETTEXT,
       format: format,
-      cachify: cachify
+      cachify: cachify,
+      backgroundColor: getBackgroundColor(backgroundColor),
+      siteLogo: siteLogo || ''
     };
 
     var mailOpts = {
@@ -133,7 +156,6 @@ function doSend(email_type, email, site, secret, langContext) {
     if (email_params.templateHTML) {
       mailOpts.html = email_params.templateHTML(templateArgs);
     }
-
     emailer.send_mail(mailOpts, function(err, success) {
       if (!success) {
         logger.error("error sending email to: " + email + " - " + err);
@@ -142,18 +164,18 @@ function doSend(email_type, email, site, secret, langContext) {
   }
 }
 
-exports.sendNewUserEmail = function(email, site, secret, langContext) {
-  doSend('new', email, site, secret, langContext);
+exports.sendNewUserEmail = function(email, site, secret, langContext, backgroundColor, siteLogo) {
+  doSend('new', email, site, secret, langContext, backgroundColor, siteLogo);
 };
 
-exports.sendConfirmationEmail = function(email, site, secret, langContext) {
-  doSend('confirm', email, site, secret, langContext);
+exports.sendConfirmationEmail = function(email, site, secret, langContext, backgroundColor, siteLogo) {
+  doSend('confirm', email, site, secret, langContext, backgroundColor, siteLogo);
 };
 
-exports.sendForgotPasswordEmail = function(email, site, secret, langContext) {
-  doSend('reset', email, site, secret, langContext);
+exports.sendForgotPasswordEmail = function(email, site, secret, langContext, backgroundColor, siteLogo) {
+  doSend('reset', email, site, secret, langContext, backgroundColor, siteLogo);
 };
 
-exports.sendTransitionEmail = function(email, site, secret, langContext) {
-  doSend('transition', email, site, secret, langContext);
+exports.sendTransitionEmail = function(email, site, secret, langContext, backgroundColor, siteLogo) {
+  doSend('transition', email, site, secret, langContext, backgroundColor, siteLogo);
 };
--- a/lib/email.js
+++ b/lib/email.js
@@ -12,6 +12,8 @@ logger = require('./logging.js').logger,
 url = require('url'),
 cachify = require('connect-cachify').cachify;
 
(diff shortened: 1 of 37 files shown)


shane-tomlinson:
@6a68 and @seanmonstar - @ryanfeeley and I worked on this at last week's work week. I have a follow on to this that makes use of the new model, rp_info, everywhere. Will issue a PR for that once this lands.

An ephemeral instance, branded-email.personatest.org, is set up to test on.

lib/wsapi/stage_user.js
```diff
@@ -78,7 +86,8 @@ exports.process = function(req, res) {
 
             res.json({ success: true, unverified: true });
 
-            email.sendConfirmationEmail(req.params.email, req.params.site, secret, langContext);
+            email.sendConfirmationEmail(req.params.email, req.params.site, secret, langContext,
```
shane-tomlinson, line 89: should we pass an rpInfo object like we do on the front end to avoid input parameter explosion?

shane-tomlinson:
data URIs can be large and the 10kb limit on requests causes failures. @lloyd suggested upping to 64 or 128kb.

lib/email.js
```diff
@@ -39,34 +39,45 @@ const templates = {
   "new": {
     landing: 'verify_email_address',
     subject: _("%(site)s: Complete sign-in"),
-    template: fs.readFileSync(path.join(TEMPLATE_PATH, 'new.ejs')),
-    templateHTML: fs.readFileSync(path.join(TEMPLATE_PATH, 'new.html.ejs'))
+    template: 'new.ejs',
```
seanmonstar, line 44: yay! clean!

lib/wsapi/stage_user.js
seanmonstar, line 89: That might be nice, but this currently works fine as well. WTB ES6 destructuring.

seanmonstar:
Looks good, it would seem that the request body test is failing in node v0.10.x

lib/wsapi/stage_user.js
shane-tomlinson, line 89: destructuring... < me drools />

shane-tomlinson:
Thanks @seanmonstar - I'll have a look at node 0.10.x

shane-tomlinson:
Data URIs do not work in images in GMail, unless I am escaping the emails incorrectly (will go check)

screen shot 2013-09-06 at 17 21 13

shane-tomlinson:
data URIs work in Yahoo!

screen shot 2013-09-06 at 17 46 12

jaredhirsch:
@shane-tomlinson I'll take a look at this, but here's some data to ponder.

I went ahead and signed up for litmus, the wonderful email screenshotting app. (Ping me if you want access--they give you a disposable email account to use with each individual test, it's great.)

[Here's](https://litmus.com/pub/196eeac/screenshots) a test run with screenshots of the new verification email sent by branded-email.personatest.org this morning.

For comparison, [here's](https://litmus.com/pub/df4eab7/screenshots) a test run with the current production emails.

Maybe we should clarify which email clients we want to support, and to what level? We could include this on MDN, similar to our browser support matrix. Campaignmonitor has a CSS [page](www.campaignmonitor.com/css/) that tracks feature support across their 10 most popular email clients--maybe we could use that list as a starting point?

jaredhirsch:
Also, what if we just don't send data-uris to gmail addresses? I know, apps-for-your-domain makes this hairy. We could check MX records though?

shane-tomlinson:
@6a68 - Thanks for the email photos, this is fantastic. It looks like I need to do some work! I stared at [campaign monitor](http://www.campaignmonitor.com/css/) list through most of this, it is very useful, looks like I missed out on some things though.

To avoid going too far into the rabbit hole of MX records and such, I just disabled dataURIs and thought we could do a subsequent PR. Since support for data URIs hasn't been officially released yet, I'm not too concerned yet.

lloyd:
what's timing of this feature landing?  (sorry to push, I'm just excited, it's awesome work)

shane-tomlinson:
@lloyd - next train if we can figure out the CSS issues. It's too large to go in this late in this train.

shane-tomlinson:
@6a68 - could you do another litmus test? I just pushed a change to the way the backgroundColors are written and manual tests in IE8 with GMail and Yahoo mail look good.

lib/email.js
```diff
@@ -88,8 +99,17 @@ exports.setInterceptor = function(callback) {
...
+function getBackgroundColor(backgroundColor) {
+  backgroundColor = backgroundColor || "45505b";
+  if (! /^#/.test(backgroundColor)) {
+    backgroundColor = "#" + backgroundColor;
+  }
+  return backgroundColor;
+}
+
+
 //TODO send in localeContext
```
jaredhirsch: is this a stale todo? I wonder if langContext is the thing this is asking for

ozten: Ya, you can nuke that comment. Sorry.

resources/email_templates/_header.html.ejs
```diff
@@ -0,0 +1,86 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
jaredhirsch, line 1: i didn't realize we use xhtml in our email templates. curious if there's any particular reason?

resources/email_templates/_header.html.ejs
```diff
@@ -0,0 +1,86 @@
...
+        </style>
+  </head>
+  <body style="-webkit-text-size-adjust: none;padding:0;margin:0;">
+    <center>
+    <table border="0" cellpadding="20" cellspacing="0" height="100%" width="100%"><!-- body area -->
+      <tr style="margin: 0;
+        padding: 0;
+        height: 100% !important;
+        width: 100% !important;
+        background: <%= backgroundColor %>
```
jaredhirsch: missing semicolon?

lib/validate.js
```diff
@@ -68,6 +69,21 @@ var types = {
     if (typeof x !== 'string' || !x.match(regex)) {
       throw new Error("not a valid origin");
     }
+  },
+  color: function(value) {
+    if (value.substr(0, 1) === '#') {
```
jaredhirsch: since we already source validator in this file, how about just using `validator.isHexColor` instead?

resources/email_templates/reset.html.ejs
```diff
@@ -1,112 +1,34 @@
...
-                </table><!-- #footer-->
-
-        </td>
-      </tr>
-    </table><!-- #container table -->
-    </center>
-
-  </body>
-</html>
+<% include _header.html.ejs %>
```
jaredhirsch, line 1: nice, now i get it. we're splitting this thing out to be easier to read. nice.

resources/static/dialog/js/modules/validate_rp_params.js
```diff
@@ -249,6 +249,11 @@ BrowserID.Modules.ValidateRpParams = (function() {
 
     dataMatches = inputLogoUri.match(dataUriRegex);
     if (dataMatches) {
+      if (inputLogoUri.length > (bid.MAX_SITE_LOGO_SIZE)) {
```
jaredhirsch, line 252: maybe we can add a check for a too-short/empty image as well, to make RP integration a tad easier? idk.

I guess this would be two separate things, and the case where it fails to match (because it's really an empty variable) is the more likely of the two. /me shrugs

resources/static/test/cases/common/js/user.js
```diff
@@ -149,7 +149,7 @@
 
     asyncTest(stageFuncName + " with XHR failure", function() {
       storage.addEmail(TEST_EMAIL);
-      var args = [lib[stageFuncName], TEST_EMAIL];
+      var args = [lib[stageFuncName].bind(lib), TEST_EMAIL];
```
jaredhirsch: i don't quite follow why we need to bind here

lib/email.js
```diff
@@ -88,8 +99,17 @@ exports.setInterceptor = function(callback) {
   interceptor = callback;
 };
 
+function getBackgroundColor(backgroundColor) {
+  backgroundColor = backgroundColor || "45505b";
```
jaredhirsch: I dunno about hard-coding the email backgroundColor default right here. Maybe it should be a constant?

lib/validate.js
```diff
@@ -68,6 +69,21 @@ var types = {
...
+      value = value.substr(1);
+    }
+
+    // Check if this is valid hex color
+    if (!value.match(/^[0-9a-fA-F]{3}$|^[0-9a-fA-F]{6}$/)) {
+      throw new Error('not a valid color: ' + value);
+    }
+  },
+  image: function(inputLogoUri) {
+    if (url.parse(inputLogoUri).protocol !== 'https:') {
```
jaredhirsch, line 81: maybe check inputLogoUri is a string before passing to url.parse, which'll barf if it isn't

resources/static/common/js/models/rp_info.js
```diff
@@ -0,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+BrowserID.Models.RpInfo = (function() {
```
jaredhirsch, line 4: i absolutely love this

resources/static/common/js/network.js
```diff
@@ -222,18 +222,21 @@ BrowserID.Network = (function() {
      * @param {string} email
      * @param {string} password
      * @param {string} origin - site user is trying to sign in to.
+     * @param {string} backgroundColor - backgroundColor to set in the email
      * @param {boolean} allowUnverified
```
jaredhirsch: the string is a lie (lines 224-226)

jaredhirsch: actually a lot of lies in this file's docstrings

jaredhirsch:
@shane-tomlinson just ran another litmus test, results are [here](https://litmus.com/pub/3d94397/)

Most of these are minor nits. Overall, resounding with win, great stuff! :surfer:

Feel free to merge after addressing (whether fixing, or just deciding to ignore) concerns above

resources/static/common/js/network.js
jaredhirsch: sorry, to be more explicit: please update the docstrings to match the function signatures

shane-tomlinson:
@6a68 - thanks for the thorough review, very helpful comments and great finds. I'll fix the remaining issues up on Monday when I'm back! Thanks again.

shane-tomlinson:
The footer text in outlook based emails looks whack, fix that before merging.

lib/validate.js
shane-tomlinson: Awesome, didn't realize this existed in validator. Requires an upgrade of the validator version.

resources/email_templates/_header.html.ejs
shane-tomlinson, line 1: I have no idea, I took what was there and extracted it.

resources/static/common/js/models/rp_info.js
shane-tomlinson, line 4: More of this is coming. I have another PR which replaces all the locally held RP type state with references to RpInfo, but it makes the PR blow up so I'll issue it once we get this PR merged.

resources/static/common/js/network.js
shane-tomlinson: good catch.

resources/static/test/cases/common/js/user.js
shane-tomlinson: I had to add the bind because I started using `this` in the called functions. I was not comfortable with this change (thanks for calling me out): `User` is a singleton, which allowed me to change the references of `this.rpInfo` to `User.rpInfo` and get rid of the `bind`.

resources/static/dialog/js/modules/validate_rp_params.js
shane-tomlinson, line 252: @6a68 - mmm.. We could check for too short of strings, though I'm not sure what the minimum length is. What would a 1x1 pixel be? If the string is empty/not filled in, validateSiteLogo is never called, so that case is avoided.

lib/email.js
```diff
@@ -12,6 +12,8 @@ logger = require('./logging.js').logger,
 url = require('url'),
 cachify = require('connect-cachify').cachify;
 
+const DEFAULT_BACKGROUND_COLOR = "45505b";
```
seanmonstar, line 15: config? meh

PR #3861, opened 2013-09-05
Ensure the window.close event is always sent along with the KPI event stream it belongs to.

### LINKED ISSUE #3794: KPI data conains many one event blobs (window.unload)
Looking at all of the blobs, an unexpectedly large number have short, hard to classify event_streams. The majority of these have only one event, and in most cases that event is "window.unload". (~17% of the kpi blobs since 2013-07-11 are blobs with only one event).

I haven't really been able to reproduce this, but I have been able to generate blobs without "window.unload" at the end, and seen only "window.unload" in my local storage, fwiw. I've seen this using Chrome, when closing the window instead of the dialog. I don't remember seeing an extra blob showing up in the kpi data, however, but I'll try again.

The % of these blobs has been growing, in particular right around the gmail announcement (where we had a spike in traffic overall).

It would be good to understand if this is a measurement error (sending an extra blob sometimes that doesn't represent a distinct session) or aborted dialog sessions, or something else entirely. (I suspect measurement error, but perhaps that is wishful thinking).

shane-tomlinson:
@kparlante, @6a68 and @seanmonstar - I could use some review here! Scary update to WinChan, I have yet to put a PR against that repo. Do you think this can be done another way using the existing callbacks, or does a new callback make more sense?

We incorrectly made the assumption that if a message was received by WinChan from the dialog, that all dialog processing was complete. This had the effect that if an assertion was generated in the dialog, a message with the assertion was sent to WinChan, _include.js took this as meaning all processing was complete in the dialog, and told the communication_iframe to submit the KPIs. The KPI event stream was cleared, and then the dialog would close, causing window.unload to be added to the recently cleared event stream.

To fix this, I added a new optional callback, close_cb, as input to WinChan.open. This callback is triggered due to a new 'close' message that is postMessage'd from the dialog's unload event handler. close_cb is then hooked up to send a message to the communication_iframe that the dialog is closed and that is _now_ safe to send the KPIs.

In addition, IE8 was unable to write the window.unload event to the event stream from the window.unload event handler. Instead, writing this final event must be done in window.beforeunload.

fixes #3794

### DIFF
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -1,6 +1,7 @@
 ;WinChan = (function() {
   var RELAY_FRAME_NAME = "__winchan_relay_frame";
   var CLOSE_CMD = "die";
+  var undefined;
 
   // a portable addListener implementation
   function addListener(w, event, cb) {
@@ -32,8 +33,8 @@
       // We must check for both XUL and Java versions of Fennec.  Both have
       // distinct UA strings.
       var userAgent = navigator.userAgent;
-      return (userAgent.indexOf('Fennec/') != -1) ||  // XUL
-             (userAgent.indexOf('Firefox/') != -1 && userAgent.indexOf('Android') != -1);   // Java
+      return (userAgent.indexOf('Fennec/') !== -1) ||  // XUL
+             (userAgent.indexOf('Firefox/') !== -1 && userAgent.indexOf('Android') !== -1);   // Java
     } catch(e) {}
     return false;
   }
@@ -85,7 +86,7 @@
      *                  6. caller upon reciept of 'ready', sends args
      */
     return {
-      open: function(opts, cb) {
+      open: function(opts, cb, close_cb) {
         if (!cb) throw "missing required callback argument";
 
         // test required options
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
A test ephemeral instance is available: https://window-unload.personatest.org

shane-tomlinson:
Do not merge yet, IE8 is submitting KPIs twice, once when the dialog closes, once when it re-opens. I am unsure if this is due to this PR or if it is existing behavior.

shane-tomlinson:
The IE8 problem is caused by this branch.

shane-tomlinson:
I take the comment back about IE8 problems are being caused by this branch. I am seeing this manifest in the current dev as well. I opened a new bug showing the behavior, #3863.

jaredhirsch:
@shane-tomlinson i'm probably retreading old ground here, but have we considered just assigning each dialog session a random, unique ID (maybe UUID) and streaming kpi events as they occur, or maybe batching on a short (5 sec?) interval?

It seems like firing events as they occur would relieve us of all the annoyances of trying to store and manage client-side KPI state.

If the user is interacting with the dialog, presumably we're doing xhrs left and right anyway, so mobile phones would still have the antenna running, and adding these xhrs wouldn't impact the battery much.

shane-tomlinson:
Closing in favor of #3864.

@6a68 - let's talk about your idea sometime, I'd be interested in hearing more. We implemented a UUID a while back, but dropped it because we figured out how to ship the KPIs all at once, maybe our current approach is too heavy handed.

PR #3862, opened 2013-09-06
[load_gen] improve details on failures to show original request and response code and body

jrgm:
When running load_gen, there are occasionally errors reported, which I'm pretty sure are database replication races. I don't believe that these actually happen in production (for the most part). The load_gen tool inherently does sequences of wsapi calls that a real user would never perform at the same speed. 

However, the errors are reported with so little detail that it's not possible to trace them. So this PR increases the errors to always show the response code and body, and the request method, host and URI.

Before:

```
error: (signup) responseNotOk: non 200 response code: 400
error: (signup) unknown: null response body
```

After:

```
error: (signup) responseNotOk: non 200 response code >>> {"code":400,"body":"Bad Request: that email does not belong to you","req":{"method":"POST","host":"login.anosrep.org","path":"/wsapi/cert_key"}}
error: (signup) responseNotOk: non 200 response code >>> {"code":400,"body":"","req":{"method":"GET","host":"login.anosrep.org","path":"/wsapi/fake_verification?email=foo%40loadtest.domain"}}
```

### DIFF
--- a/bin/load_gen
+++ b/bin/load_gen
@@ -218,8 +218,8 @@ function poll() {
             completed[act][2]++;
           } else {
             completed[act][1]++;
-            winston.error('('+act+') ' + err.toString());
           }
+          winston.error('('+act+') ' + err.toString());
         } else {
           completed[act][0]++;
         }
--- a/lib/load_gen/activities/interaction_data.js
+++ b/lib/load_gen/activities/interaction_data.js
@@ -6,7 +6,9 @@
  * this file is the "interaction_data" activity, which simulates 
  * the load of the client passing on interaction data for kpi metrics */
 
-var client = require('../../wsapi_client.js');
+const
+client = require('../../wsapi_client.js'),
+common = require('../common');
 
 var blob = { data: 
   { "event_stream":[["window.dom_loading",972],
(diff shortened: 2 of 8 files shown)


lib/wsapi_client.js
```diff
@@ -27,6 +27,23 @@ function startRequest() {
...
+    if (message._headers) {
+      req.host = message._headers.host;
+    }
+    req.path = message.path;
+
+    return JSON.parse(JSON.stringify(req));
+  }
+
+  return req;
+}
```
jrgm, line 45: And yeah, I am touching "private" API elements. I understand that this, theoretically, could change in the future.

lib/load_gen/activities/interaction_data.js
```diff
@@ -43,17 +45,12 @@ var blob = { data:
...
-    } else if (!r || r.code !== 200) {
-      cb("for interaction_data post response code is not 200: " + (r ? r.code : "no response"));
-    } else if (r.body !== '{"success":true}' ){
-      cb("interaction_data post not successful");
-    } else {
-      cb();
+    err = common.checkResponse(err, r);
+    if (err) return cb(err);
+    if (r.body.success !== true) {
+      return cb(common.error("interaction_data post not successful", null, r));
```
jrgm, line 52: common.checkResponse handles all the null response, non-200 and JSON.parse checking that I replace here.

bin/load_gen
```diff
@@ -218,8 +218,8 @@ function poll() {
             completed[act][2]++;
           } else {
             completed[act][1]++;
-            winston.error('('+act+') ' + err.toString());
           }
+          winston.error('('+act+') ' + err.toString());
         } else {
```
jrgm, line 223: This change means that 503's are always printed. (I think some might be misclassified as being errors from the toobusy module, so I want to log these for a while).

jaredhirsch:
:beers:

PR #3864, opened 2013-09-06
Remove the window.unload event in the KPI event stream.

shane-tomlinson:
@kparlante - this is a PR with the alternate proposal to remove the window.close event.

If we do not need the event for KPI processing, this allows us to avoid making changes to WinChan - see issue #3861

issue #3794


jaredhirsch:
@shane-tomlinson +1 to this approach, it seems more robust than twiddling winchan _and_ relying on unload/beforeunload, which can fail in so many ways.

Tests passing for me and looks good locally. Feel free to merge in :beers:

kparlante:
@shane-tomlinson  +1 KPI dashboard is not relying on (or currently using in any way) the window.close event.

PR #3865, opened 2013-09-06
Add the ability to clear certs from localStorage in the development menu

shane-tomlinson:
@jrgm - since we have been discussing this, what do you think? Worthy of inclusion into the main codebase?


jrgm:
r+ Looks good to me.

Issue #3778, opened 2013-08-13
Persona broken in packaged apps


mattbasta:
`navigator.id.watch` gets called successfully. There are no errors on the page.

After `navigator.id.request` gets called, though, this is the screen that loads:

http://cl.ly/image/2o2b0r2T1O1M

vladikoff:
@mattbasta Sorry, cannot reproduce, I've just tried several websites (i.e. http://123done.org/ ) and all of them worked fine. Is this inside of an app or something else? Are you using the latest version of the simulator (4.0)?

mattbasta:
This happened inside a packaged app loaded via "Add Directory". I'll check
what version of the simulator I'm using, but I believe it is the latest.

shane-tomlinson:
Is this a dup of #3614?

mattbasta:
Doesn't look like it. The dialog loads "correctly", but throws the error
shown.

Interestingly, it's also impossible to close the login dialog. You have to
shut down the simulator to get rid of it.

shane-tomlinson:
@mattbasta - That's not so awesome. Thanks for the extra info.

mattbasta:
Updated the title: I can reproduce the issue in on a Keon device as well (fully updated). You can download the package yourself here:

http://inferno.paas.allizom.org/

callahad:
Likely related to the hotfix we deployed to strictly check origins in winchan.

shane-tomlinson:
@callahad - packaged apps should be using native Persona, bypassing WinChan completely. @mattbasta - can you provide a set of STRs? I am unfamiliar with installing packaged apps on a device manually.

ozten:
Simulator STR
1) Install the FxOS simulator via Add-On https://addons.mozilla.org/en-US/firefox/addon/firefox-os-simulator/

2) In Fx Tools > Web Developer > Firefox OS Simulator

3) Start simulator

4) Download http://inferno.paas.allizom.org/package.zip

5) mkdir inferno ; cd inferno; unzip package.zip

5) Click Add Directory and locate `manifest.webapp`

6) Note: ![](http://dl.dropbox.com/u/10060532/Screenshots/T9VM.png) launches

7) Click gear icon

8) Click Sign In

Expected

Usual Persona flow

Actual

![](http://dl.dropbox.com/u/10060532/Screenshots/ynH0.png)

ozten:
[...]

PR #3876, opened 2013-09-11
3778 - verify app://{guid} urls

jedp:
This patch lets us use origins of the form `app://{guid}` with the verifier.

This should address some of the problems in Issue #3778

### DIFF
--- a/lib/verifier/certassertion.js
+++ b/lib/verifier/certassertion.js
@@ -55,6 +55,16 @@ function compareAudiences(want, got) {
     // case 1 & 1a
     // (app:// urls are seen on FirefoxOS desktop and possibly mobile)
     if (/^(?:https?|app):\/\//.test(got)) {
+      // If want and got are equal, then success.
+      if (want === got) {
+        return undefined;
+      }
+
+      // If they are not equal, parse them and see if they are equivalent.
+      // Note that urls like app://{guid} from FirefoxOS will not parse like
+      // normal urls.  They won't have a hostname. But that shouldn't matter in
+      // the case of guids: They should have been found equal in the test
+      // above.
       var gu = normalizeParsedURL(url.parse(got));
       got_scheme = gu.protocol;
       got_domain = gu.hostname;
@@ -73,12 +83,15 @@ function compareAudiences(want, got) {
     }
     if (!got_domain) throw "domain missing";
 
-    // now parse "want" url
+    // now parse "want" url - the url that was extracted from the assertion
     want = normalizeParsedURL(url.parse(want));
 
     // compare the parts explicitly provided by the client
     if (got_scheme && got_scheme !== want.protocol) throw "scheme mismatch";
     if (got_port && got_port !== want.port) throw "port mismatch";
+
+    // We unconditionally test domain equality; testing this last is crucial
+    // for ensuring that the provided origin isn't the empty string.
     if (got_domain !== want.hostname) throw "domain mismatch";
 
     return undefined;
--- a/tests/verifier-test.js
+++ b/tests/verifier-test.js
@@ -76,7 +76,11 @@ suite.addBatch({
     'http://fakesite.com:80 and http://fakesite.com:8000': matchesAudience(false),
     'https://fakesite.com:443 and https://fakesite.com:9000': matchesAudience(false),
 
-    'app://browser.gaiamobile.org and app://browser.gaiamobile.org:80': matchesAudience(true)
+    'app://browser.gaiamobile.org and app://browser.gaiamobile.org:80': matchesAudience(true),
+    'app://marketplace.firefox.com and app://marketplace.firefox.com': matchesAudience(true),
+    'app://{5041b155-d1c0-4957-9914-c3af6aa41cf1} and app://{5041b155-d1c0-4957-9914-c3af6aa41cf1}': matchesAudience(true), 
+    'app://{5041b155-d1c0-4957-9914-c3af6aa41cf1} and app://{5041b155-d1c0-4957-9914-c3af6aa41cf2}': matchesAudience(false), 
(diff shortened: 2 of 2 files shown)


jedp:
@warner did I do this ok?  Thanks!

lib/verifier/certassertion.js
```diff
@@ -55,6 +55,16 @@ function compareAudiences(want, got) {
...
+      // If want and got are equal, then success.
+      if (want === got) {
+        return undefined;
+      }
+
+      // If they are not equal, parse them and see if they are equivalent.
+      // Note that urls like app://{guid} from FirefoxOS will not parse like
+      // normal urls.  They won't have a hostname. But that shouldn't matter in
+      // the case of guids: They should have been found equal in the test
+      // above.
```
seanmonstar, line 67: but zomg the splosions!

really, though, `app://{guid}` isn't secure.

shane-tomlinson, line 67: Have you guys thought about `file://{guid}` for PhoneGap apps?

shane-tomlinson, line 67: @fmarier - I thought this was going to be in a separate verifier?

fmarier:
@shane-tomlinson That's what I thought as well.

@jedp Do we need this on the production verifier or is it just for testing? If we do need it in the production verifier, we should run it past Yvan.

jedp:
@seanmonstar I agree that there appear to be serious problems with `app://{guid}` urls in general (not unlike problems with `http://` urls, too).  Moreover, I have no idea how an app on device is supposed to use one of these assertions.  I do not understand how an app posting down to a server can provide any assurance that it is actually authorized to do so.  How can an app's server know how to trust the app that is talking to it?  I don't see any mechanism for that.  (Although, persona could be used to provide that mechanism through, say, a key fingerprint embedded in the assertion posted down.) 

@warner and I talked this over last night, but we could well have missed something.  If this patch opens up or enables any attacks, then it should obviously be rejected.

@shane-tomlinson this is tantamount to `file://{path}` urls, isn't it.  That did occur to me.  I didn't want to take that on as well here, though.

@shane-tomlinson , @fmarier I did initially put this in an ephemeral instance, but I'm worried about providing a testing platform that doesn't match production in its behavior.

Help!  Do we want to get together and talk about this?  I feel like it would be productive to get together with you guys and @warner and @basta and someone from gaia who understands the architecture of this whole `app://{guid}` business.

jaredhirsch:
@jedp I think having a meeting with marketplace folk would be good here. I'm strongly opposed to putting this in the regular production verifier, it's a horrific hack that subverts so much of what persona is intended to do. I'd lean more towards building a separate verifier for app testing, though I worry that apps will just ship with the dev verifier wired in (hence causing all the security problems in prod anyway).

kumar303:
>  How can an app's server know how to trust the app that is talking to it?

It can't, not when you allow `app://`. On Firefox OS any packaged app can arbitrarily pick a domain. For example, any app could choose `app://facebook.com` if they wanted to.

I have not been part of any deep discussion on this but has anyone looked at the oauth solutions we did for Firefox OS? https://developer.mozilla.org/en-US/docs/Web/Apps/Manifest?redirectlocale=en-US&redirectslug=Apps%2FManifest#redirects This is a very similar problem: a packaged app's domain can't be used to trust oauth redirect URLs so you have to define them up front in your manifest. I don't know if anything there applies to verifying assertions but maybe.

jaredhirsch:
@kumar303 that's exactly the sort of thing we need--some way to extend Persona's web-centric security/trust model into the wild world of anonymous apps with random identifiers. who is the best point of contact on the apps side to get into this further?

kumar303:
Maybe start by asking @fabricedesre about how Persona might be able to get a manifest hook? I'm not sure if that's the right way to go or not.

One correction to my last comment: setting a custom origin is only available to _privileged_ apps so it makes Jed's short term approach less scary. Privileged apps must be signed by Firefox Marketplace so we can put some checks in place to make sure an app doesn't steal someone's domain. We'll probably have to add a domain verification process for that though. Also, a long-term approach where multiple marketplaces sign apps is less clear.

jedp:
@kumar303, @6a68, I learned a little more about how apps talk to the servers behind them today.  Based on the `origin` in the manifest, packaged apps are granted xhr privileges for that domain only.  Apps can't just xhr any remote service; only the one that matches their origin.  So if a server receives an xhr from an app, the server is expected to trust that the app is legit, but it has no independent way of verifying.  The security of the system depends on Mozilla's vetting process being 100% air-tight, otherwise there are clear attacks.  For example, Recall that origins are assigned on device on a first-come first-served basis.  So say you are `kumarsfreshbeats.com` and you make an app with that origin.  If somebody creates an app and manages to get the same origin, `kumarsfreshbeats.com`, into the manifest, and that app is installed before your own app is, then the app can xhr to your server and you have no way of knowing the app is not your own.

Good news: ozten, warner, and I talked all this through with Travis Choma today.  He's going to be in MV tomorrow and has offered to meet with Marketplace to talk about how to move the ball forward (as the athletically inclined say) toward the correct fixes that will be best for Marketplace dev and testing.

Near term options include keeping an ephemeral verifier running (I have https://happyapp.personatest.org/verify), or just asking QA and dev to use `adb push` to get the dev marketplace app on device.  The advantage of the latter is that you _can_ have an `origin` parameter in the manifest of your choosing when you do it this way.  So even though this is an extra step for dev and QA, I think it's much better because it maintains parity between dev and prod.

fabricedesre:
Not sure if this is relevant there, but keep in mind that anyone can sideload a privileged app on its phone with a custom origin of its choice.

seanmonstar:
Privileged apps can ask for the permission `systemXHR`, which provides unrestricted XHR to any domain you want.

jaredhirsch:
closing this out, i think we're going to find a different way to support dev apps. @jedp [agrees](http://irclog.gr/#show/irc.mozilla.org/identity/321482)

PR #3877, opened 2013-09-11
groups simultaneous requests to the same resource

### LINKED ISSUE #3872: CSRF validation failure, token mismatch (was [ie10] regular error in dialog when running automation for reset-password-test)
I was reviewing some automation failures with IE10 and could see in the saucelabs video that the test failed due to an error screen shown at this step

https://github.com/mozilla/browserid/blob/dev/automation-tests/tests/reset-password-test.js#L121-L133

at line 129

I tried to get an example of the error, but could not. I guess I could extend the test to check that an error is not shown at that point and if so then grab it and stash it somewhere. But for now just noting that this failure is sometimes happening.

seanmonstar:
If we ever make more than 1 GET request to the same resource (including GET params), this will group them together into a single request. While the request is mid-flight, additional requests to the same resource will not start a new one, but add their listeners to the existing request.

This may not happen often, but when it does, we most likely never want the requests to be separate.

This likely fixes #3872, since right now, the dialog can sometimes call `/wsapi/session_context` twice at the start, one for `interaction_data`, and one for `cookie_check`. This is the likely cause of a race condition where both pass their empty session cookie, both requests generate a CSRF token, and then after one returns, we start a new request using the first token, but by then the second request finished and changed our cookie. Gross!

### DIFF
--- a/resources/static/common/js/modules/xhr.js
+++ b/resources/static/common/js/modules/xhr.js
@@ -275,7 +275,7 @@ BrowserID.Modules.XHR = (function() {
 
   function triggerHandlers(handlers, resp, xhrObj, textResponse) {
     for (var i = 0; i < handlers.length; i++) {
-      complete(handlers[i], resp, xhrObj, textResponse);
+      complete(handlers[i], _.extend({}, resp), xhrObj, textResponse);
     }
   }
 
--- a/resources/static/common/js/modules/xhr.js
+++ b/resources/static/common/js/modules/xhr.js
@@ -273,9 +273,11 @@ BrowserID.Modules.XHR = (function() {
     return errorInfo;
   }
 
-  function triggerHandlers(handlers, resp, xhrObj, textResponse) {
+  function triggerHandlers(handlers, response, xhrObj, textResponse) {
     for (var i = 0; i < handlers.length; i++) {
-      complete(handlers[i], _.extend({}, resp), xhrObj, textResponse);
+      var resp = response;
+      if (typeof resp === 'object') resp = _.extend({}, resp);
+      complete(handlers[i], resp, xhrObj, textResponse);
     }
   }
 
--- a/resources/static/common/js/modules/xhr.js
+++ b/resources/static/common/js/modules/xhr.js
@@ -75,7 +75,27 @@ BrowserID.Modules.XHR = (function() {
      * Abort all outstanding XHR requests
      * @method abortAll
      */
-    abortAll: abortAll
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@seanmonstar - I have _so_ wanted this for so long. Reviewing now.

resources/static/common/js/modules/xhr.js
```diff
@@ -240,6 +292,12 @@ BrowserID.Modules.XHR = (function() {
     return errorInfo;
   }
 
+  function triggerHandlers(handlers, resp, xhrObj, textResponse) {
+    for (var i = 0; i < handlers.length; i++) {
+      complete(handlers[i], resp, xhrObj, textResponse);
```
shane-tomlinson, line 280: There is a danger here of handlers modifying the response object, inadvertently interfering with each other. Perhaps create a copy of the response and pass it down?

resources/static/common/js/modules/xhr.js
```diff
@@ -90,6 +110,7 @@ BrowserID.Modules.XHR = (function() {
 
     self.outstandingRequests = {};
     self.outstandingTimers = [];
+    self.existingRequests = {};
```
shane-tomlinson: Can outstandingRequests be used instead, since it already tracks all outstanding requests?

shane-tomlinson: outstandingRequests is keyed on request.eventTime, but I think the request URI is equally valid. Will this interfere with POSTs? Do we ever have multiple outstanding POSTs to the same URI?

shane-tomlinson: If these can't be combined, can you rename existingRequests to existingGetRequests so the intent is clearer?

shane-tomlinson: So, after thinking about this more, I think we should be able to use outstandingRequests universally, even if it remains keyed by eventTime. There are at most only a few outstandingRequests at any given time, searching through the list in `getExistingRequest` should be very speedy.

resources/static/common/js/modules/xhr.js
seanmonstar, line 280: which argument? `resp`? or `xhrObj`?

shane-tomlinson, line 280: sorry, I wasn't precise in the least - `resp` in particular.

seanmonstar:
@shane-tomlinson hey! did your suggestions.

PR #3880, opened 2013-09-12
Clean up the cert invalidation code.

shane-tomlinson:
@seanmonstar or @6a68 - mind reviewing?

While doing the work for the "one time cert", I tried to decipher the cert invalidation code and was confused. To better understand it, I tore it apart and put it back together in a fashion that I could understand better.
- Rename cleanupIdentities to removeInvalidIdentities.
- Extract Method on the code that checked record validities
- Extract Method on code that checks if record is expired.


jaredhirsch:
R+, merging in

Issue #3579, opened 2013-06-25
[b2g] Feura Sans not served - cross origin errors when downloading fonts when origin is firefoxos.perona.org


shane-tomlinson:
@jedp and I are simulating prod in a dev environment with one server with two URLs: primaries.personatest.org is the A record and fx-primaries.personatest.org is a CNAME to primaries.personatest.org. The FirefoxOS device is pointed at fx-primaries.personatest.org. You can see the CORS errors below:

```
E/GeckoConsole( 1296): [JavaScript Error: "downloadable font: download failed (font-family: "Feura Sans" style:normal weight:bold stretch:normal src index:4): bad URI or cross-site access not allowed
....

E/GeckoConsole( 1296): [JavaScript Warning: "downloadable font: no supported format found (font-family: "Feura Sans" style:normal weight:bold stretch:normal src index:6)
E/GeckoConsole( 1296): source: (end of source list)" {file: "https://primaries.personatest.org/v/bed6e74109/production/en/dialog.css" line: 0 column: 0 source: "@font-face {
E/GeckoConsole( 1296):   font-family: "Feura Sans";
E/GeckoConsole( 1296):   font-style: normal;
E/GeckoConsole( 1296):   font-weight: 700;
E/GeckoConsole( 1296):   src: local("Feura Sans Bold"), local("FeuraSans-Bold"), url("https://primaries.personatest.org/v/9ee1ab1f19/fonts/en/feurasans-bold.eot#") format("embedded-opentype"), url("https://primaries.personatest.org/v/ab710ed2da/fonts/en/feurasans-bold.woff") format("woff"), url("https://primaries.personatest.org/v/22fb5eea7e/fonts/en/feurasans-bold.ttf") format("truetype"), url("https://primaries.personatest.org/v/9e85ff5af7/fonts/en/feurasans-bold.svg#Feura Sans") format("svg");
E/GeckoConsole( 1296): }"}]
```

The CORS headers on fonts are by default set to the A record name.

jedp:
Stupid question ... why is this even happening when FeuraSans is on device?  Why doesn't it pick it up locally and just be happy?

callahad:
Are the font family names maybe different?

shane-tomlinson:
@callahad - that is my guess, but Feura Sans is the one I was told to use by the Marketplace dudes. I will re-check.

callahad:
[...]

PR #3885, opened 2013-09-14
3579 - Feura Sans is now Fira Sans on FirefoxOS

jedp:
Fixes issue #3579.

Tested this by pushing it to https://train0619.personatest.org/ and pointing my unagi phone to it.  No more Feura Sans warnings.

@shane-tomlinson Did I miss any other changes that should be made?

(Phone is running latest mozilla-central trunk with gaia master)

### DIFF
--- a/resources/static/dialog/css/m.css
+++ b/resources/static/dialog/css/m.css
@@ -21,12 +21,13 @@
 
   /* Remove background image gradients set in style.css which cause
    * horizontal lines to appear while the dialog is loading.
-   * Set the default font to Feura Sans
+   * Set the default font to Fira Sans, which used to be called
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@jedp - we have to roll an entire new connect-fonts font pack for the new font. Only changing the name of the font will allow the new font name to be displayed on updated Unagi devices, but it will leave all other mobile users without the new font.

callahad:
This actually _should_ work, shouldn't it? FxOS will use the built-in Fira Sans, other devices will get the connect-fonts provided Feura Sans?

shane-tomlinson:
FWIW, we should _still_ roll a fira-sans font pack.

PR #3890, opened 2013-09-18
Awsbox 0.5.x upgrade

lloyd:
required to fix our dev environment, changes are minimal.


lloyd:
r+ing my own change because DNS is now moved over and without this, there's no way to deploy ephemeral instances.  go big or go home!  /me ducks

PR #3893, opened 2013-09-18
Update winston to the newest version to work with node 0.10.x

### LINKED ISSUE #3892: logging stops shortly after startup since upgrade to node v0.10.18
STR (in local env)

1) Set node version to v0.10.18
2) load up the dev server
3) in another window, visit <project_root>/var/log and `tail -f router.log`
4) load up the dev rp repeatedly
5) where are the logs?

shane-tomlinson:
@6a68 or @jrgm - would you guys mind reviewing?

fixes #3892


lloyd:
r+

Issue #3899, opened 2013-09-19
new-skool ephemerals http-proxy is getting saturated to MAX_FILES when running real browsers.


jrgm:
On the new-skool ephemeral's I'm seeing sockets leaked on the http-proxy until the proxy hits MAX_FILES, and stops accepting connections (causing run-all.js to start failing). 

This is not the same as https://github.com/mozilla/browserid/issues/3851. I see that the upgrade to http-proxy@0.10.3 is in place. 

While upgrading 123done.org to http-proxy@0.10.3 (using nodejs@0.10.7) solved the problem in that case, it's not solving things here. 

(Note the previous problem (GH-3851) was about hitting maxSockets of 100 in the socket pool (which would then take 2 minutes to expire); this is different.

On the new-skool ephemerals, there are sockets that get to ESTABLISHED state on the http-proxy process, and then they never goes away (which means we are not getting a FIN, RST, etc.? (or dropping the event socket events in node land))). 

I've tried switching up nodejs versions, http-proxy versions, and Connect versions, on an ephemeral box, but I'm not finding the root problem. But this is something about 304 responses. **curl/uncached requests for include.js are 200 OK; IMS requests hang**. 

But I'm done for the day; later.

(relate to https://github.com/mozilla/browserid/issues/3895 and https://github.com/mozilla/browserid/issues/3851)

lloyd:
hmmm.  at one point I did some serious tuning to up all limits on awsbox's image.  Don't know if that was lost and this is not a regression in application code, or if socket leaking is a new behavior...  

what are STR here?  just hammer an ephemeral deployment till it falls over?  Do I even need persona to repro this problem or can I repro with an awsbox running hello world - IOW - is this a persona specific issue, or an awsbox issue?

jrgm:
So, I tried to reduce this to just simple curl, but it's about real browsers running with caches. 

But, I put this script in place on current login.dev.anosrep.org at `~ec2-user/proxy_svr_est_watch.sh`:

```
#!/bin/sh                                                                                                           

PROXY_PID=$(ps -efww | grep /home/proxy/proxy_server.js | grep -v grep | awk '{ print $2 }')

while [ 1 ]; do 
    echo -n "`date`    "
    sudo lsof -P -n -p $PROXY_PID | grep ESTABLISHED | wc -l
    sleep 10
done
```

[...]

PR #3901, opened 2013-09-19
update to awsbox 0.5.4 - mitigates #3899

lloyd:
as stated, unblocks automated testing against ephemeral/dev instances.  Also, lots of regression fixes that will improve usability: https://github.com/mozilla/awsbox/blob/v0.5.4/ChangeLog#L1-L8

/cc @jrgm @shane-tomlinson @chilts


shane-tomlinson:
@lloyd - it says "we can't automatically merge" :(

lloyd:
@shane-tomlinson rebased, green now.

PR #3906, opened 2013-09-20
always decode returnTo before redirect

### LINKED ISSUE #3903: returnTo URLs are being re-encoded redirecting from email confirmation link.
If I pass `http://example.com/?q=don%27t` as the `returnTo` parameter to `navigator.id.request()`, after confirming my new persona account in the email (fallback auth provider), I am redirected to `http://example.com/?q=don%2527t`. It basically looks like the URL was sent through another pass of encoding which converts the `%` into `%25`.

On SUMO, we pass a lot of URLs that look like this because we support other languages with non-ascii characters. This is messing up the questions they are about to post which is bad. Please fix? :)

For reference, see https://bugzilla.mozilla.org/show_bug.cgi?id=918393

seanmonstar:
It turns out there are a couple places we do show the `returnTo` address, and to be sure we aren't carrying around potentially data, we'll continue to encode it as soon as we get it. So to fix this issue though, before redirecting, we'll always _de_code it, so it's back to the original value an RP passed.

fixes #3903

### DIFF
--- a/resources/static/common/js/helpers.js
+++ b/resources/static/common/js/helpers.js
@@ -101,6 +101,10 @@
     return String(email).replace(emailRe, "");
   }
 
+  function redirect(win, returnTo) {
+    win.location = decodeURI(returnTo);
+  }
+
   _.extend(helpers, {
     isObject: function(arg) {
       return Object.prototype.toString.apply(arg) === "[object Object]";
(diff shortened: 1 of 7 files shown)


jaredhirsch:
Silly Travis. Tests pass for me, looks reasonable, /me shrugs and hits the big button for great justice

PR #3908, opened 2013-09-22
update awsbox to 0.6.0, fixes #3899, fixes #3851

### LINKED ISSUE #3851: refresh our use of node-http-proxy on awsbox AMI to use the latest, http-proxy@0.10.3 when using nodejs 0.10.x
*\* TL;DR http-proxy@0.8.0 does not work correctly when run with nodejs >0.10.x **

I started writing this up as an issue for https://github.com/mozilla/123done, but since it can affect other browserid-related repos using awsbox deploys, that I'd put the issue in a central place.

I was seeing problems when running automation on 123done in parallel after its update this week to run on nodejs 0.10.7. HTTP requests would mysteriously hang, causing test failures. Anyways, in hindsight, pretty simple reason. 

(/cc @stephendonner in case he's noticing more flakiness recently with bidpom tests for webqa).

STR

1) load http://123done.org in Firefox or Chrome (probably any browser)

2) ssh ec2-user@123done.org

3) In the shell run this command:

```
while [ 1 ]; do date; sudo netstat -tanp | egrep -v 'LISTEN|redis-server|Active|Recv-Q|/sshd|:6379' |\
    awk '{ print $6 }' | sort | uniq -c | sort -k2; sleep 10; done
```

4) Simply hit reload in the browser and let the loads complete. Repeat...

5) After maybe 8 or 10 reloads the browser will spin waiting on network requests (check firebug, etc.)

At the point, you will something like this in the terminal window:

```
Fri Aug 30 22:08:28 UTC 2013
    206 ESTABLISHED
     47 TIME_WAIT
```

206 === 6 sockets from the browser to the proxy, 100 sockets from the proxy to 123done, and 100 sockets on 123done. node-http-proxy by default sets maxSockets to 100.

I duplicated this on an awsbox push of 123done code, and then downgraded nodejs to 0.8.24. With that version of nodejs, I can't reproduce the same issue. So it's an issue with the upgrade to nodejs 0.10.x.

Anyways, the problem is that connections in the proxy are not being shut down in some cases, so the connection pool gets saturated with two-minute long connections that have no real work to do anymore. 

I git-bisected this back to this commit in node-http-proxy@v0.9.0 - https://github.com/nodejitsu/node-http-proxy/commit/28a3b0a6f04197665d95f7574efb4289c259a647, in particular the return statement that is apparently preventing some necessary cleanup when run in nodejs v0.10.x.

Backporting that commit to node-http-proxy@v0.8.0 also works, but perhaps we should just refresh our use of node-http-proxy on awsbox AMI to use the latest, http-proxy@0.10.3, which appears to work for me (when using nodejs 0.10.x).

### LINKED ISSUE #3899: new-skool ephemerals http-proxy is getting saturated to MAX_FILES when running real browsers.

lloyd:
this will mean ephemeral instances now run nginx as a proxy, so it adds gzip support, performance improvements, and addresses the two issues mentioned in the subject.

### DIFF
--- a/lockdown.json
+++ b/lockdown.json
@@ -22,7 +22,7 @@
     "0.2.0": "c55013856c8194ec854a0cbec90aab5a04ce3ac5"
   },
   "awsbox": {
-    "0.5.4": "ac88be60ddb0bdfe923cd8b02697c92d81eb692f"
+    "0.6.0": "c57a8eb66f240ce4c6c9b62f1e9dc3a7c979a22a"
   },
   "awssum": {
     "1.1.0": "e010144aa516af78eca01352aab9203b816611bb"
@@ -31,7 +31,7 @@
     "1.1.0": "ea9ed3f7116044563bb0b14d47b2f38a684bac47"
   },
   "awssum-amazon-ec2": {
-    "1.2.0": "8bb53bd273f2a034bb9bc0a6ef55f5333f1efc00"
+    "1.3.2": "fb29a8dffaf6647440b93c64a1868317b4cc856c"
   },
   "awssum-amazon-route53": {
     "1.0.3": "85d1e6aeb22de14ed03e110cc887e482d0362aac"
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   },
   "devDependencies": {
     "vows": "git://github.com/cloudhead/vows#253ca34c",
-    "awsbox": "0.5.4",
+    "awsbox": "0.6.0",
     "irc": "0.3.3",
     "jshint": "0.9.1",
     "which": "1.0.5",


jaredhirsch:
works for me, merging in. @jrgm owe you a beer if I've missed something

Issue #3529, opened 2013-06-13
[kpi] Make metrics data available to kpi dashboard


kparlante:
It would be great if we could send router-metrics and verifier-metrics data to a kpiggybank collector. It would need to be processed to strip out information for user privacy, just as we do before making the data available to metrics.

cc @lloyd, @shane-tomlinson

shane-tomlinson:
@kparlante - @6a68 says that there may already be a JSONP API for that.

jaredhirsch:
@kparlante @shane-tomlinson Indeed, https://bugzilla.mozilla.org/show_bug.cgi?id=822126 has some backstory.

jedp:
This has been tagged with four stars: How can we close this bug with a PR to browserid?  Does this bug belong in pentaho or kpiggybank?

kparlante:
@jedp: The code to process those files should belong to the identity team, not the pentaho/metrics team, going forward. As the kettle scripts are replaced (python? node?), they could live in browserid, identity-ops, or a new repo.

kparlante:
@6a68: I don't think jsonp from the pentaho sites is the right way to do this -- we really don't need it to route through pentaho at all.

kparlante:
Moving out of browserid repo

shane-tomlinson:
Re-opening as a tracker until we get this merged.

jaredhirsch:
closed by merging #3939 :beers:

PR #3909, opened 2013-09-23
Issue 3529 metrics data kpi dashboard

shane-tomlinson:
@6a68 - here ya go man! 

issue #3529


shane-tomlinson:
@6a68 - I just finally saw the code you wrote in https://github.com/mozilla/browserid/compare/issue_3529_metrics_to_piggybank. It looks like you are collecting additional KPIs there that look useful, should I add those commits to this PR?

A lingering question for me is how to test the metrics=>KPIs (see https://github.com/mozilla/browserid/commit/5210ad5d3a065a6a80814fad914b48fc99968132#L2R49). KPIs had no tests before this PR and only a shell of a test here to make sure that /wsapi/interaction data does not blow up. Any ideas?

Also, if you want me to break this PR into components, I am happy to. @6a68 mentioned that he actually prefers one large PR because it takes less time to review in bulk.

@kparlante - this is for you as well!

jaredhirsch:
@shane-tomlinson I think you can safely ignore that old branch of mine, because that was a lazier version of what you're doing here: I was just adding kpi calls everywhere we had metrics calls.

I'm happy to review this as-is. If you're feeling funny about it, it's fine to split into pieces. Your call.

As far as unit testing the metrics->kpis part, how about this:
- split the problem space: ignore the code that normally generates the metrics, just test the transport
- input: pass the right sort of JSON into `kpi_data.store()`
- output: verify that a corresponding blob lands in kpiggybank

We could do this cheaply by mocking out just enough of the kpiggybank API (actually, maybe we just need a mock XHR?) to verify the stuff in the transmitted blob. I'll add more comments as I go through the PR, but that's my initial thought.

jaredhirsch:
OK, did a first pass. (It's funny, the whole second half of commits looks like those one-liners you find as you're polishing, I noticed the same thing in my long-lived branch.)

## High-level feedback:

I'd suggest doing a major pass to decouple event names (a description of application state), filter names (a description of the types of events filtered), and transport names (a description of the transport itself).

For instance, right now you cannot send an event to multiple endpoints, because we have to test that specific strings, like 'kpi' or 'statsd', are present in the event name. But the events themselves really just announce what's happened/happening at a specific point in the app. It doesn't matter what processes the event, really, or which transport will shuttle the event downstream.

The idea is we'd have filters with names like PerformanceDataFilter, UserInteractionFilter, SecurityEventFilter, ApplicationAvailabilityFilter. And we'd probably keep the transport names the same, though I'd suggest replacing MetricsTransport with KPiggybankTransport, describing exactly what the endpoint is.

This would let us not have to change event names each time we change transports/filters. Additionally, the event names could be straightforward, almost like documentation.

The filters would become more complex, but if we are able to keep the event names reasonably namespaced, we can avoid getting into the hell of 1000-line switch statements. I owe you good examples of event naming and namespacing, but I'm coming up empty at the moment.

This would also let us filter one event into multiple streams. (For instance, login failure matters from a security standpoint, should also land in the log files, and probably also matters from a user interaction counting perspective.)

This would let us change where we send a filter's output, or send the filter's output to multiple places. (This lets us audition heka in specific roles without depending on it, for free.)

Ah, also, this would let us remove the extra cef/statsd loggers, and really, really, just have one pipe (from the app's perspective).

Going to take a break for a bit, then I"ll throw in some lower-level feedback. I'll be sure to get online early tomorrow so we can chat in person, too.

shane-tomlinson:
Thanks for the feedback @6a68, super useful. I originally went down the route you suggested originally, with the decoupled event names and the elimination of the cef/statsd libraries.

I became afraid as the PR became even larger than it is now, and the comment in https://github.com/mozilla/browserid/blob/issue-3529-metrics-data-kpi-dashboard/lib/logging/cef_logger.js#L8-L9 made me question whether removing the libraries reduced the 'grepability' of finding cef/statsd events.

I think the goal is great, I'm trying to figure out how to get there without breaking expectations. Is that an incremental improvement on top of this PR or one to include in this PR?

shane-tomlinson:
Closing in favor of https://github.com/mozilla/browserid/pull/3939

Issue #3911, opened 2013-09-23
Test fails on post-limiting-test.js (on some combinations of distro/kernel/nodejs version)


mohit-agarwal:
![screenshot from 2013-09-24 04 14 19](https://f.cloud.github.com/assets/2356369/1195993/e430dbdc-24a1-11e3-9111-4e908d875e2d.png)

node version v0.10.18
npm version 1.3.8

Linux mohit-Inspiron-5420 3.2.0-51-generic #77-Ubuntu SMP Wed Jul 24 20:18:19 UTC 2013 x86_64 x86_64 x86_64 GNU/Linux

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=12.04
DISTRIB_CODENAME=precise
DISTRIB_DESCRIPTION="Ubuntu 12.04.2 LTS"

jrgm:
Yeah, I install 12.04.3 to get a newer 3.x kernel than what travis is installing (2.x) (don't ask me why that version on travis), and I can see that sometimes (but not always) ECONNRESET is returned instead of 'socket hang up' . For purposes of the test, I think just changing the test to accept either failure message is fine. So, if you want create a pull request, that would be good; if not, we'll get this fixed. Thanks!

e.g., `assert.ok(/Error: (socket hang up|read ECONNRESET)/.test(err.toString()));`

Try just running `WHAT_TESTS=back npm test`; you'll need phantomjs installed and mysqld installed and configured to run the `front` and `back_mysql` targets.

Issue #77, opened 2011-07-17
myfavoritebeer should point out that secure verification is server-based


benadida:

lloyd:
https://github.com/lloyd/myfavoritebeer.org/blob/master/static/js/main.js#L30

Now we should actually fix it.

PR #3912, opened 2013-09-24
Test fails on tests/post-limiting-test.js (on some combinations of distro/kernel/nodejs version)

mohit-agarwal:
In ubuntu version 12.04.2/12.04.3 , I see that sometimes (but not always) ECONNRESET is returned instead of 'socket hang up' .

So 'tests/post-limiting-test.js' needs to be edited to expect either 'socket hang up' or ECONNRESET. Then only will it be able to pass all the tests. For purposes of the test, changing the test to accept either failure message is fine.

So I edited the line 70 of the 'tests/post-limiting-test.js' as:
     assert.ok(/Error: (socket hang up|read ECONNRESET)/.test(err.toString()));

This will ensure that the post-limiting-test.js is passed.

### DIFF
--- a/tests/post-limiting-test.js
+++ b/tests/post-limiting-test.js
@@ -67,7 +67,7 @@ function addTests(port, path) {
         });
       },
       "fails": function (err) {
-        assert.ok(/socket hang up/.test(err.toString()));
+	assert.ok(/Error: (socket hang up|read ECONNRESET)/.test(err.toString()));
       }
     }
   });


jaredhirsch:
you're seeing this on which version of node and which kernel?

mohit-agarwal:
Node version v0.10.18

Linux 3.2.0-51-generic #77-Ubuntu SMP Wed Jul 24 20:18:19 UTC 2013 x86_64 x86_64 x86_64 GNU/Linux

jaredhirsch:
@mohit-agarwal thanks for the feedback. I'm hesitant to merge this until I better understand the source of this bug, just patching the test seems like it might be ignoring the diagnostic value of the tests.

If I understand this correctly, node on ubuntu sometimes throws an ECONNRESET instead of hanging up cleanly--is this an artifact of how the tests are structured, or are we not handling some exception higher in the code? Have you seen funky ECONNRESETs thrown when running locally?

Downloading an ubuntu ISO now to try to repro, I'll update this PR in a bit.

jrgm:
@6a68 - it wasn't mentioned that this PR fixes GH-3911. 

I have reproduced this on an Ubuntu 12.04 LTS newly installed. But, yeah, despite my suggestion to fix the test, it had been in the back of my mind to go back and look at the reason for the difference. It feels to me like a race, and either message is equivalent to the post being blocked. But I could well be wrong, and there's a deeper bug. 

But, get out yer strace, tcpdump and ngrep! ;-)

jaredhirsch:
@jrgm thanks for the pointer--I hadn't seen that bug. installing now.

jaredhirsch:
@jrgm I ran out of time on this, your call on just merging or digging deeper. Sorry @mohit-agarwal, thanks a lot for contributing, don't want to block you.

jrgm:
@6a68 - so I take back my ill-advised crack about "real operating system" ;-)

I reproduced the 0.10.18 on an awsbox. I then installed 0.10.10, and the bug wasn't there. Git-bisecting nodejs pointed me to https://github.com/joyent/node/commit/255650f4d9738fd55800d9f16942f74dcb19f8a9, which was introduced in 0.10.16. But after looking closely at tcpdump, I could see that there was a subtle change is packet timing that I think explains this. Before the commit above, for the test, which is on localhost (with a MTU of 16436), the entire request would arrive in one packet. So the client would just wait for the response, which would be a FIN packet from the server, and the client would report 'socket hang up'. But with commit 255650f, now almost all the request (>10KB) would be in one packet, with a few trailing bytes to follow. So the server would initiate a destroy of the socket on the first packet, and the second packet would rejected with RST.

Er, I think that explains it. At any rate, either way, the connect limit is doing the right thing. So I'm going to merge this, since either 'socket hang up' or 'ECONNRESET' on the client can be a valid response.

PR #3913, opened 2013-09-24
Ie11

shane-tomlinson:
@callahad - this is something I started about a month ago, but have not completed.

### DIFF
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -5,7 +5,7 @@
   // a portable addListener implementation
   function addListener(w, event, cb) {
     if(w.attachEvent) w.attachEvent('on' + event, cb);
-    else if (w.addEventListener) w.addEventListener(event, cb, false);
+    else if (w.addEventListener) w.addEventListener(event, cb);//, false);
   }
 
   // a portable removeListener implementation
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -17,12 +17,19 @@
   // checking for IE8 or above
   function isInternetExplorer() {
     var rv = -1; // Return value assumes failure.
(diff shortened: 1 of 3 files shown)


resources/static/common/js/lib/winchan.js
```diff
@@ -5,7 +5,7 @@
   // a portable addListener implementation
   function addListener(w, event, cb) {
     if(w.attachEvent) w.attachEvent('on' + event, cb);
-    else if (w.addEventListener) w.addEventListener(event, cb, false);
+    else if (w.addEventListener) w.addEventListener(event, cb);//, false);
```
shane-tomlinson, line 8: Figure out a better way for this.

shane-tomlinson, line 8: IE11 has addEventListener and doesn't do the right thing if you pass `false`

seanmonstar, line 8: wtf

jaredhirsch, line 8: :-\ @callahad ping your IE contacts about this, seems like a browser bug.

[msdn](http://msdn.microsoft.com/en-us/library/ie/ff975245%28v=vs.85%29.aspx) claims this is supported since ie9

jaredhirsch:
the dialog works for me in ie11.personatest.org in modern.ie's preview VM (11.0.9431.0)

jaredhirsch:
i can also login/logout of ie11.123done on win 7/ie9 and xp/ie8.

jaredhirsch:
@callahad, shane and I will try to get the IE11 hack fixed up and mergeable tomorrow.

currently I'm thinking maybe we wrap addEventListener in a try/catch to handle FF blowing up when 'false' is omitted

shane-tomlinson:
Closing in favor of #3933

Issue #3905, opened 2013-09-20
watch() API logs out on page load with iOS 7 Safari


jamesshore:
I'm getting early reports from users that they can't login when using Safari on iOS 7. Chrome on iOS 7 works fine. I'm using the watch() API. I'll update this issue as I have more information.

**Update: Issue confirmed.** To reproduce the issue, use Safari on iOS 7 to log into a site, such as http://123done.org, that uses the navigator.id.watch() API. After logging in, reload the page. (Some sites will reload the page automatically when you login.) Persona will call the `onlogout` function and log you out.

**Original report:**

Here's what I know so far:
1. I saw several users (but not all) show up in my logs with a login immediately followed by a logout, repeated several times. It looks like users are logging in, automatically being logged out, and then manually trying again.
2. I emailed every user that experienced this problem about an hour ago. I've gotten the following replies so far:

"That happened in Safari. I'll check it when I get a chance. Thanks."

"it seems the new Safari 7 on iOS 7 has some trouble with the new Persona API. Sorry, I don't know what goes wrong there, but I upgraded the iPad to iOS 7 just yesterday. Anyway, Chrome works fine, even on iOS 7."
1. (Possibly unrelated.) I recently upgraded to the watch() API, so it's possible there's a bug in my implementation of it. My code is very simple, though. Here it is:

``` javascript
// Copyright (c) 2012-2013 Titanium I.T. LLC. All rights reserved.
/*global window, document, navigator, XMLHttpRequest, alert, setupPersona, $*/

(function() {
    "use strict";

    window.setupPersona = function(email) {
        $(".tdjs-signin").click(onSignin);
        $(".tdjs-signout").click(onSignout);
        navigator.id.watch({
            loggedInUser: email,
            onlogin: postLoginRequest,
            onlogout: postLogoutRequest
        });
    };

    function onSignin(event) {
        event.preventDefault();
        navigator.id.request({
            siteName: "Let’s Code JavaScript"
        });
    }

    function onSignout(event) {
        event.preventDefault();
        navigator.id.logout();
    }

    function postLoginRequest(assertion) {
        var post = $.post("/v3/auth/login", "assertion=" + assertion);
        post.done(function(data) {
            if (data === "ok") {
                window.location.reload();
            }
            else {
                alert("Login failed. Please try again.");
                // Must logout on failure or Persona will keep trying to validate assertion. But that will trigger
                // onlogout (postLogoutRequest()) synchronously, which will cause a window reload, so the logout must be
[...]

PR #3914, opened 2013-09-24
mitigate change in iOS7 privacy behavior - no longer logout users upon reload

lloyd:
Mitigation for issue #3905

if desired, we should hotfix this into stage or prod. /cc @callahad @6a68 @shane-tomlinson @jrgm 

deployed at http://ios7.123done.org

### DIFF
--- a/resources/static/common/js/browser-support.js
+++ b/resources/static/common/js/browser-support.js
@@ -95,6 +95,5 @@ BrowserID.BrowserSupport = (function() {
      */
     isIOS: isIOS
   };
-
 }());
 
--- a/resources/static/common/js/storage.js
+++ b/resources/static/common/js/storage.js
@@ -158,6 +158,14 @@ BrowserID.Storage = (function() {
     }
   }
 
+  function storageCheckSet() {
+    storage.storageCheck = "true";
+  }
+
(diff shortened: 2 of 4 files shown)


callahad:
> if desired, we should hotfix this into stage or prod.

**We must hotfix this.** User's can't log into Persona-based sites on iOS. ([Video](http://www.youtube.com/watch?v=Yq4MPKJNpyo))

@jrgm What do you need to make hotfixing a possibility?

jaredhirsch:
works on iOS 7 MobileSafari and Chrome: able to log into 123done, not logged out on reload.

i was concerned about how we are setting the storage check key (per IE8 worries mentioned elsewhere in the storage.js file), but can't repro any problems seen in #1637, so I think this is good to hotfix.

jaredhirsch:
merging :beers:

lloyd:
so @callahad - I'll leave it to you to work with @jrgm and delegate hotfixing / cherry picking this into an appropriate train, I'm out for now.

jrgm:
We can take this in the delayed 09.11 branch, which I still have to get on stage.

callahad:
This fix is now live in production

PR #3924, opened 2013-09-27
evaluate nothing in include.js when native support is available

lloyd:
There's no reason to evaluate a bunch of code when we have native support.  let's abort earlier.

Also, there's no reason to test for .request when .mozId is defined - this affords greater flexibility in native implementations.

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -116,13 +116,6 @@
 
   if (!navigator.id) {
     navigator.id = {};
-    // Is there a native implementation on this platform?
-    // If so, hook navigator.id onto it.
-    if (navigator.mozId) {
-      navigator.id = navigator.mozId;
-    } else {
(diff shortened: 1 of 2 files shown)


callahad:
we probably want to version the APIs so that the shim can still apply when the browser supports an old version of the protocol, right?

lloyd:
@callahad you mean version the object exposed by the browser?  unsure.

But that's somewhat orthogonal to this pull request, right?

callahad:
yeah, orthogonal. sorry, I'm just totally dropping stream of consciousness stuff all over GitHub :)

lloyd:
No sweat.  That's usually my job (random). :)

Dan Callahan notifications@github.com wrote:

yeah, orthogonal. sorry, I'm just totally dropping stream of consciousness stuff all over GitHub :)

---

PR #3933, opened 2013-09-30
fix IE11

### LINKED ISSUE #3593: Persona dialog blank on MSIE 11 on Windows 8.1
Robert Kaiser (KaiRo on moz IRC) reports that "Persona doesn't work on MSIE 11 on Win 8.1"

He explained on the `#planning` channel on Jun 27:

```
[16:11:06] <KaiRo>   jedp: the sign in popup comes up completely empty, the arrow design is there, but no textbox for filling in an email address - in the end, I guess someone will need to get an image from http://windows.microsoft.com/en-us/windows-8/preview-download and test directly with that browser
[16:12:01] <KaiRo>   jedp: of course, we can't be sure if they'll ship with that UA or if what breaks Persona is a bug on their side
[16:12:49] <KaiRo>   jedp: oh, and I turned off any kind of exception lists or anything they might automatically download - I installed all updates but manually
[16:13:10] <jedp>    KaiRo thanks for the details.  I'll file an issue.
[16:13:29] <KaiRo>   jedp: the latter is just in case they might send different UAs to different sites based on such a downloaded list
```

shane-tomlinson:
@6a68, @callahad, @seanmonstar 
- update WinChan's ua sniff to handle IE11
- remove the extra  after the  class.

A test instance is set up on ie11.personatest.org

fixes #3593

### DIFF
--- a/resources/views/dialog_layout.ejs
+++ b/resources/views/dialog_layout.ejs
@@ -36,7 +36,7 @@
      */ %>
   <title><%= format(gettext("Mozilla Persona: %s"), [gettext(title)]) %></title>
 </head>
-  <body <% if (showLoading) { %>class="loading"i<% } %>>
+  <body <% if (showLoading) { %>class="loading"<% } %>>
       <% if (useJavascript !== false && enable_development_menu) { %>
         <a href="#" id="showDevelopment">&nbsp;</a>
       <% } %>
--- a/resources/static/common/js/lib/winchan.js
+++ b/resources/static/common/js/lib/winchan.js
@@ -17,12 +17,19 @@
   // checking for IE8 or above
   function isInternetExplorer() {
     var rv = -1; // Return value assumes failure.
+    var ua = navigator.userAgent;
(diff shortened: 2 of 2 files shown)


shane-tomlinson:
Corresponding PR against WinChan is at mozilla/winchan#34

jaredhirsch:
@shane-tomlinson what about the bit where you needed to remove `false` from `addEventListener`?

shane-tomlinson:
@6a68 - I retested with it in and had no problem. I think I misdiagnosed.

resources/views/dialog_layout.ejs
```diff
@@ -36,7 +36,7 @@
      */ %>
   <title><%= format(gettext("Mozilla Persona: %s"), [gettext(title)]) %></title>
 </head>
-  <body <% if (showLoading) { %>class="loading"i<% } %>>
+  <body <% if (showLoading) { %>class="loading"<% } %>>
```
seanmonstar, line 39: lol nice

Issue #3856, opened 2013-09-04
internal.setPersistent sets `remember` field instead of `logged_in` field to indicate user wants a silent assertion.


shane-tomlinson:

shane-tomlinson:
This means users who sign in using the internal API are not getting the full goodness of silent assertions.

shane-tomlinson:
Is internal.setPersistent actually used anywhere?

callahad:
Hi! To help us better focus, I'm "closing" all issues that have been open for more than six months. These have been tagged "cleanup-2014" so that we can go back and review them in the future.

For more information, check out this thread: http://thread.gmane.org/gmane.comp.mozilla.identity.devel/7394

If you believe this bug is still a major issue for you, please comment, submit a pull request, or discuss it on our mailing list: https://lists.mozilla.org/listinfo/dev-identity

Sorry for GitHub notification spam!

PR #3936, opened 2013-09-30
internal.setPersistent sets `remember` field instead of `logged_in` field to indicate user wants a silent assertion.

mohit-agarwal:
In  `resources/static/test/cases/dialog/js/misc/internal_api.js` and `resources/static/dialog/js/misc/internal_api.js` , internal.setPersistent sets `remember` field instead of `logged_in` field to indicate user wants a silent assertion.

This means users signing in using internal API are not getting the full goodness of silent assertions.

So we should replace 'remember' field with 'logged_in' field.

internal.setPersistent function is defined in `resources/static/dialog/js/misc/internal_api.js`.

Fixes Issue #3856.

### DIFF
--- a/resources/static/dialog/js/misc/internal_api.js
+++ b/resources/static/dialog/js/misc/internal_api.js
@@ -61,7 +61,7 @@
 
     user.checkAuthentication(function(authenticated) {
       if (authenticated) {
-        storage.site.set(origin, "remember", true);
+        storage.site.set(origin, "logged_in", true);
       }
 
       complete(!!authenticated || null);
--- a/resources/static/test/cases/dialog/js/misc/internal_api.js
+++ b/resources/static/test/cases/dialog/js/misc/internal_api.js
@@ -55,7 +55,7 @@
     internal.setPersistent(ORIGIN, function(status) {
       strictEqual(status, null, "user is not authenticated should not succeed in setting persistent");
 
-      testUndefined(storage.site.get(ORIGIN, "remember"), "remember status not set");
+      testUndefined(storage.site.get(ORIGIN, "logged_in"), "logged_in status not set");
       testUndefined(storage.site.get(ORIGIN, "email"), "email not set");
       start();
     });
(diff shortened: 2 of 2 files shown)


seanmonstar:
I don't understand perfectly the difference between `logged_in` and `remember`. @shane-tomlinson halp?

shane-tomlinson:
Travis-ci is lying https://travis-ci.org/mohit-agarwal/browserid/builds/12066533 looks all green to me.

shane-tomlinson:
@seanmonstar - This is super fuzzy in my head and I am digging in the logs to find out more. (f1e45ae29e2899df61cc6ae031d594b6994963b8, 02f89b43f8a62eaf44159e1d9b56f33cde175fe2)

In a period long ago (2011), we introduced the notion of persistent sign ins, where users could select whether or not to remember the email association at a site. The user had to select a checkbox for an email address to be remembered. In localStorage, we saved whether the checkbox was selected and what the email address was.

After a few months, we ditched the "remember" and assumed users always wanted the association remembered. We began storing a bit that serves a similar purpose - "logged_in". "logged_in" says that if an email address is stored, and logged_in is set, a silent assertion can be generated when calling navigator.id.watch. 

If logged_in is false, but an email address is stored, no silent assertion is generated, but the email address is selected in the email picker. The latter can happen if the user signed in to an RP from the dialog, but then the RP calls navigator.id.logout().

So, to get to the purpose of this PR. I believe the PR is good as stands, but see my message on dev-identity about the possibility of removing this code entirely.

shane-tomlinson:
@mohit-agarwal, @seanmonstar - after trying to find out more history of what exactly "remember" was used for, I realized my suggestion to change "remember" to "logged_in" is incorrect. "logged_in" is actually an email address and not a boolean.

The name "logged_in" is confusing. This leads me to realize there are two fields with similar purposes - "email" and "logged_in", which seems silly. Can "email" and "logged_in" email ever be different?

The purpose of "logged_in" is for silent assertions out of the .watch API, "email" is the email that is displayed to the user in the email picker. Should these be consolidated and "logged_in" converted to a boolean?

@mohit-agarwal - Thanks for the PR, I want to hold off merging until we straighten out the confusion.

shane-tomlinson:
@mohit-agarwal - I am going to close this until we figure out whether we need BrowserID.internal.setPersistent. 

Sorry to have you do work that may not be merged, but your effort here is not wasted - it is forcing us to think about how to properly handle this. If we decide to continue to expose BrowserID.internal.setPersistent, we will use this pull request as a starting point, update logged_in to take an email address, and move on.

Thanks again!

PR #3940, opened 2013-10-01
Extract rp info

shane-tomlinson:
@6a68 & @seanmonstar - this finishes the work started in #3857 where I introduced the rp_info object. There were a lot of localized bits of RP related state that are all removed and pushed into this object.

An ephemeral instance, rp-info.personatest.org is available to test.

### DIFF
--- a/resources/static/dialog/js/misc/state.js
+++ b/resources/static/dialog/js/misc/state.js
@@ -79,7 +79,7 @@ BrowserID.State = (function() {
       // screen.
       var actionInfo = {
         email: info.email,
-        siteName: self.siteName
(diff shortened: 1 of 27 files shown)


resources/static/dialog/js/misc/state.js
```diff
@@ -627,7 +606,10 @@ BrowserID.State = (function() {
 
   }
 
+  var und;
```
seanmonstar, line 609: just a nit: why use such a short variable name? What's wrong with `var undefined;`? It makes the code more clear everywhere else where this variable is used, instead of trying to recall what `und` is.

Won't block review.

Issue #2647, opened 2012-10-25
.watch() should throw an exception if loggedInUser is not a valid type


callahad:
Valid types are: `undefined`, `null`, `false`, String

Anything else should result in an exception

See mozilla/id-specs@4558cb25a1f5d4d542b7b46cd18fe16cf890b926

mohit-agarwal:
We need to throw an exception if loggedInUser is not a valid type. So If I throw an error with a message, would that work?

 `throw new Error("loggedInUser is not a valid type");`

This would be thrown when `loggedInUser` is not a valid type.

jaredhirsch:
@mohit-agarwal that sounds good to me! feel free to submit a PR with the changes. would also be good to add a unit test to ensure the exception is thrown if the type is wrong.

seanmonstar:
@6a68 trololol. would be great to have _any_ unit tests for include.js

shane-tomlinson:
@seanmonstar, @6a68 - ooooohihihahaha..  testing of include.js.. I can't find the open issue about this.

A shell: https://github.com/mozilla/browserid/blob/dev/resources/static/test/cases/include.js

seanmonstar:
this is in train-2013.10.23

PR #3941, opened 2013-10-01
.watch() should throw an exception if loggedInUser is not a valid type

mohit-agarwal:
Valid types for `loggedInUser` are: `undefined`, `null`, `false`, String

`typeof undefined === 'undefined';`
`typeof null === 'object';`
`typeof false === 'boolean';`

.watch() will throw an Error if loggedInUser is not a valid type.

Fixes Issue #2647.

### DIFF
--- a/resources/static/common/js/helpers.js
+++ b/resources/static/common/js/helpers.js
@@ -122,6 +122,14 @@
       return Object.prototype.toString.apply(arg) === "[object Array]";
     },
 
+    is_Null: function(arg) {
+      return arg === null;
+    },
+
+    is_Undefined: function(arg) {
+      return (typeof arg === 'undefined');
+    },
+
     /**
      * Get an email from a DOM element and validate it.
      * @method getAndValidateEmail
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -317,6 +317,11 @@
       // back compat support for loggedInEmail
       checkRenamed(options, "loggedInEmail", "loggedInUser");
       loggedInUser = options.loggedInUser;
+      if(typeof loggedInUser !== 'undefined' &&
+	 typeof loggedInUser !== 'string' &&
+	 typeof loggedInUser !== 'object')
+	      throw new Error("loggedInUser is not a valid type");
+
 
       _open_hidden_iframe();
     }
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -317,11 +317,12 @@
       // back compat support for loggedInEmail
       checkRenamed(options, "loggedInEmail", "loggedInUser");
       loggedInUser = options.loggedInUser;
-      if(typeof loggedInUser !== 'undefined' &&
-	 typeof loggedInUser !== 'object' &&
-	 typeof loggedInUser !== 'boolean' &&
-	 typeof loggedInUser !== 'string')
-	      throw new Error("loggedInUser is not a valid type");
+      if (!is_Null(loggedInUser)  &&
+          !is_Undefined(loggedInUser) &&
+          !isString(loggedInUser))
+      {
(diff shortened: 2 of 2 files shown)


resources/static/include_js/_include.js
```diff
@@ -317,6 +317,12 @@
       // back compat support for loggedInEmail
       checkRenamed(options, "loggedInEmail", "loggedInUser");
       loggedInUser = options.loggedInUser;
+      if(typeof loggedInUser !== 'undefined' &&
+	 typeof loggedInUser !== 'object' &&
```
shane-tomlinson: @mohit-agarwal - Can you make two changes:
- create a temp variable called `loggedInUserType` and set it to `typeof loggedInUser`. Compare against that instead of repeatedly calling `typeof loggedInUser`
- Reformat to use spaces instead of tabs.

shane-tomlinson:
Testing now.

resources/static/include_js/_include.js
shane-tomlinson: Now that I am digging in further, scratch the original feedback.

New feedback:
- Reformat to use spaces instead of tabs.
- Add space between the `if` and `(`.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/navigator.id.watch?redirectlocale=en-US&redirectslug=DOM%2Fnavigator.id.watch), loggedInUser must be one of:
- `undefined`
- `null`
- A string

Neither booleans nor random objects should be accepted. 

`null` is of type `object`, but so are `{ my: "random object" }` and `new String("this is a string")`. 

To positively check for null, and only null:

```
itemToCheck === null
```

The corresponding "not null" test is:
`loggedInUser !== null`.

Strings are a bit tricky, because `typeof "this is a string"` === `string`. `typeof new String("this is a string")` === `object`. :/

A positive test for a string is:

```
Object.prototype.toString.apply(itemToCheck) === "[object String]"
```

It's corresponding negative test:

```
Object.prototype.toString.apply(loggedInUser) !== "[object String]"
```

shane-tomlinson:
Thanks @mohit-agarwal! If you can make these changes, I can re-review and test!

resources/static/include_js/_include.js
jaredhirsch: @shane-tomlinson @mohit-agarwal At risk of complicating this PR slightly more, I think the right approach is to modify the [include.js generation script](https://github.com/mozilla/browserid/blob/dev/scripts/create_include.js#L38-L40) to insert the type-checking functions in [common/helpers.js](https://github.com/mozilla/browserid/blob/dev/resources/static/common/js/helpers.js#L109-L123), that seems cleaner and removes duplication.

It really should be a one-liner to add that to the include-create script, shouldn't be too hairy to add.

mohit-agarwal:
@shane-tomlinson I have made the necessary changes, please review the PR. I have inserted type-checking functions in common/helpers.js.

resources/static/common/js/helpers.js
```diff
@@ -122,6 +122,14 @@
       return Object.prototype.toString.apply(arg) === "[object Array]";
     },
 
+    is_Null: function(arg) {
```
shane-tomlinson, line 125: @mohit-agarwal - Our standards have been to use camelCase without `_` for function names. Can you do one last edit and I'll do a final test?

mohit-agarwal, line 125: @shane-tomlinson : I did that initially, but it gave me errors as isNull and isUndefined are already defined and are being used in different files.

shane-tomlinson, line 125: @mohit-agarwal - ok, I'll have a bit closer look at the code, perhaps we already have these functions in which case we don't need to add a second copy?

shane-tomlinson:
Closing in favor of #3979

Issue #2910, opened 2013-01-07
BrowserID.internal.get() should allow application-defined claims


lhoward:
We have a profile for using BrowserID in the GSS-API/SASL, so that non-web applications may authenticate using BrowserID. It is described at http://www.padl.com/~lukeh/gss-browserid.html.

The mechanism supports channel binding (so the authentication can, for example, be bound to a TLS channel) and can perform a Diffie-Hellman key exchange in order to establish a shared session key, used for message protection services and fast re-authentication. Both of these parameters need to be included in the assertion, so they are signed with the user's private key.

In an earlier implementation of the mechanism, these were encoded in the audience URL. Our current implementation overrides jwcrypto.assertion.sign() so that it can insert the addition claims immediately prior to signing. As this is obviously implementation dependent and unsupported, I would like to ask that when an assertion is requested from a native app - that is, the usage is #NATIVE and BrowserID.internal.get() is called - that it be possible to submit application-specific claims for inclusion in the assertion. This could be an additional parameter or a new key on the options object.

ozten:
@lhoward - Had a chance to discuss this and Issue #2913 in the bug triage meeting.

The core team is excited about your work on a BrowserID GSS mechanism. Working through these ideas and sharing updates on the mozilla.dev.identity mailing list is awesome.

We're going to close these two bugs for the following reasons:
- Issues against mozilla/browserid should be very specific reports of a bug against the core system. They should include steps to reproduce, actual results, and expected results
- GSS-API/SASL design should take place on the mailing list, not this issue tracking system
- GSS-API/SASL implementation issues would be best tracked in that project

We are really excited about the potential for GSS-API and another SASL mechanism! Keep up the great work in your project and in our mailing list.

shane-tomlinson:
Re-opening since @lhoward has issued #3952 to address this.

PR #3952, opened 2013-10-07
 Allow UA to submit arbitrary claims to assertion

lhoward:


shane-tomlinson:
@jedp - would this help you out as well?

shane-tomlinson:
Thanks for the PR @lhoward!

You are right that the additional claims being tucked into rp-info.js seems a bit strange, but it does seem like the most logical place.

Additional things I'd like to see:
- A test or two to ensure only objects can be passed in to `experimental_userAssertedClaims`
- A test to ensure only the internal API can pass in `experimental_userAssertedClaims`, unless we decide to open the feature to everyone (see list discussion https://groups.google.com/d/msg/mozilla.dev.identity/VWgVVjCdEdM/jxBPASO3198J)

I can help with these in my fork, and add you as a contributor, or you can do them in your own branch.

resources/static/dialog/js/modules/validate_rp_params.js
```diff
@@ -313,6 +332,14 @@ BrowserID.Modules.ValidateRpParams = (function() {
     return email;
   }
 
+  function validateUserAssertedClaims(claims, name) {
+    if (typeof claims !== "object") {
```
jaredhirsch: this check will allow objects, arrays, and `null`. what kind of format do we want to allow?

jaredhirsch:
@shane-tomlinson if I understand correctly, webrtc wanted to add some additional fields inside the assertion. @ryanseys would this change be useful in that case?

ryanseys:
This would be very useful. An arbitrary payload field that you could sign
would be amazing.

resources/static/dialog/js/modules/validate_rp_params.js
shane-tomlinson: @lhoward - I wrote a bit of info on correctly checking objects yesterday in https://github.com/mozilla/browserid/pull/3941#discussion_r6823625.

Instead of `typeof claims !== 'object'`, you may want to use `Object.prototype.toString.apply(claims) === "[object Object]"`

As @6a68 said, strings, arrays, and null are all Objects, but only objects return "[object Object]" when converted using Object.prototype.toString.apply. You can experiment with this in browser's developer console.

```
> Object.prototype.toString.apply({a:"b"})
"[object Object]"
> Object.prototype.toString.apply(new Object({a:"b"}))
"[object Object]"
> Object.prototype.toString.apply("this is a string")
"[object String]"
> Object.prototype.toString.apply(null)
"[object Null]"
> Object.prototype.toString.apply([])
"[object Array]"
```

lhoward: Ah, thanks Shane. Well, this should definitely be fixed in the patch then.

lhoward:
@shane-tomlinson regarding your second comment, it is already restricted to the internal API - is there something I'm missing? see isInternalRPAPI()

shane-tomlinson:
@lhoward - You can ignore that request. I am wondering whether this can be safely opened up to the navigator.id.request/watch API as well. @callahad, @jedp and @lloyd?

lloyd:
why not land internal-only first, then later consider external exposure?

resources/static/dialog/js/modules/validate_rp_params.js
```diff
@@ -313,6 +332,14 @@ BrowserID.Modules.ValidateRpParams = (function() {
     return email;
   }
 
+  function validateUserAssertedClaims(claims, name) {
+    if (Object.prototype.toString.apply(claims) === "[object Object]") {
```
shane-tomlinson, line 336: @lhoward - I was just writing up unit tests for this and came across this. Is this supposed to be `===` or `!==`. Do you want to force this to be an object or a non-object?

lhoward, line 336: Should be an object. If the test is inverted, my bad, hadn't tested this yet. Will check later.

shane-tomlinson, line 336: @lhoward - I can take care of it! Thanks for the clarification.

shane-tomlinson:
fixes #2910

shane-tomlinson:
Manually merged after updating validateUserAssertedClaims and adding unit tests. Thanks @lhoward!

Merge commit: 3a0e7ffa9ca35c4ac2b0f9d919a769c7c08ce7c9

lhoward:
Testing gss_browserid against http://login.dev.anosrep.org/sign_in#NATIVE (rather than the production URL), I now see this error:

2013-10-16 01:07:30.570 gss-client[33750:f0b] {
    MessageLevel = ErrorMessageLevel;
    MessageSource = JSMessageSource;
    MessageType = LogMessageType;
    lineNumber = 0;
    message = "Unable to post message to https://mockmyid.com. Recipient has origin https://login.dev.anosrep.org.\n";
    sourceURL = "";
}

on both our normal and userAssertedClaims branches.

Further, if userAssertedClaims is set, I also get a transient error dialog with “See more info / relay frame could not be found”; this then disappears, the login window is shown and the authentication completes.

Probably this is related to some other change (as my original patch worked fine). It's late here, so will have to look into tomorrow. Perhaps this is better discussed on the list.

-- Luke

PR #3953, opened 2013-10-07
Remove `lockdown`, use `npm shrinkwrap`

mmalecki:
For documentation on `npm shrinkwrap` see
https://npmjs.org/doc/shrinkwrap.html.

The reason wanting to make that change is that I `npm shrinkwrap` is more the de facto standard for locking down `npm` dependencies (I wanted to start working on upgrading `express` but ran into `lockdown` error).

I read in #2022 that you use different sets of dependencies in production and development, and that's still easy with shrinkwrap: `vim package.json; rm npm-shrinkwrap.json; npm i; npm shrinkwrap`.


lloyd:
I would _love_ to move to shrinkwrap, but we can't because shrinkwrap doesn't check for bitwise equivalence.  No checksums.  That's the primary motivation behind lockdown, so we can guarantee that the same versions tested (and security reviewed) are used in production.

blocked on isaacs/npm#3141

PR #3938, opened 2013-09-30
String consistency for desktop and mobile, and some HTML email tweaks

ryanfeeley:
Email line breaks, removed px from td height in HTML email and Sign in with Persona » Sign to RP with… on mobile.

PR #3957, opened 2013-10-08
Feeley email updates

shane-tomlinson:
@ryanfeeley - I manually merged what I could from PR #3938 and then manually edited the rest. How is this looking for you?


shane-tomlinson:
@ryanfeeley - the split line footer looks a bit off. Is this how you wanted it to look?

screen shot 2013-10-08 at 17 17 33

ryanfeeley:
@shane-tomlinson Thanks for the preview Shane! You are right. I don't want a line-wrap in the context.

shane-tomlinson:
The footer is a single line again:

screen shot 2013-10-09 at 10 53 31

shane-tomlinson:
If the RP does not specify a siteName, the dialog shows "Sign in to with..."
screen shot 2013-10-09 at 11 13 04

ryanfeeley:
Domain? Or drop the "to"? (will need i18n consideration)

shane-tomlinson:
@ryanfeeley - I added the domain and merged.

PR #3960, opened 2013-10-09
a stateless "goldilocks" api

seanmonstar:
keeps current watch() api, but makes it "stateless" if RP doesn't want
session management. to indicate this, simply don't pass an `onlogout`
handler to watch().

navigator.id.watch({ onlogin: function(assertion) {} });
navigator.id.request(optionsAsNormal);

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -4,9 +4,9 @@ Ben Adida <ben@adida.net>
 Brian J Brennan <brianloveswords@gmail.com>
 Brian Warner <warner-github@lothar.com>
 Bryan Clark <clarkbw@gmail.com>
-Dan Callahan <dan.callahan@gmail.com>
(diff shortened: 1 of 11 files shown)


seanmonstar:
Two notes:
1. Since there's really only 1 option to pass to watch() for this, I'd think it would be nice to allow passing a function instead of an object with an onlogin parameter. 
2. There's some confusion over whether we're keeping the same function names, or introducing configure and prompt. Changing it in this PR is really easy, the only changes would be in include.js.

seanmonstar:
oh dangit. i didnt realize there was a mozilla/goldilocks branch, and this PR automatically chose it as the target. closing and making new versus dev.

callahad:
Yeah, the other branch was Lloyd's slash-and-burn, goldilocks-only experiment.

PR #3968, opened 2013-10-10
email verification metrics

### LINKED ISSUE #3528: [kpi] Add email-verification-metrics.json
It would be great if we could count email verifications, aggregated by date. We're interested in knowing if new users are successful, and knowing the counts can help us validate the kpi data and count cases where users verify in a different browser, etc.

Similar to router-metrics.json and verifier-metrics.json.

cc: @shane-tomlinson, @lloyd

### LINKED ISSUE #3920: send /wsapi/stage_* events to kpiggybank from backend
Because there are still some kinks to work out w.r.t. the KPIs coming from the client side, @kparlante has asked that we use data from the backend to count the number of emails that are staged as well as the number that are verified.

Link to #3528 and #3873, #3863

shane-tomlinson:
@6a68  and @kparlante - another metrics one coming your way!
- Send stage events to kpiggybank.
- Send email verification metrics to kpiggybank.

available to test on https://logging-shane.personatest.org

fixes #3528 
fixes #3920

### DIFF
--- a/lib/logging/transports/filters/metrics.js
+++ b/lib/logging/transports/filters/metrics.js
@@ -10,8 +10,12 @@ const _ = require('underscore');
 const coarse = require('../../../coarse_user_agent_parser');
 
 var MessageMatches = {
-  signin: true,
-  verify: true
+  'signin': true,
+  'verify': true
+  'complete_email_confirmation.success': true,
+  'complete_reset.success': true,
+  'complete_transition.success': true,
+  'complete_user_creation.success': true
 };
 
 var RegExpMatches = [
--- a/lib/logging/transports/filters/metrics.js
+++ b/lib/logging/transports/filters/metrics.js
@@ -12,13 +12,11 @@ const coarse = require('../../../coarse_user_agent_parser');
 var MessageMatches = {
   'signin': true,
   'verify': true
-  'complete_email_confirmation.success': true,
-  'complete_reset.success': true,
-  'complete_transition.success': true,
-  'complete_user_creation.success': true
 };
 
 var RegExpMatches = [
+  /complete_(?:[^\.]+)\.success/,
+  /stage_(?:[^\.]+)\.success/,
   /^metrics\.report\./
 ];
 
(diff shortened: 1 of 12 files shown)


jaredhirsch:
looking now.

lib/logging/transports/filters/metrics.js
```diff
@@ -10,11 +10,13 @@ const _ = require('underscore');
 const coarse = require('../../../coarse_user_agent_parser');
 
 var MessageMatches = {
-  signin: true,
-  verify: true
+  'signin': true,
+  'verify': true
```
jaredhirsch, line 14: hmm, this is kinda tricky. do we need to make this change? if so, maybe it's better to use an array of strings for matches and do an indexOf check. I could easily see one of us forgetting to make the key a string.

jaredhirsch:
Not something to address in this PR, but I'm thinking how nice it'd be to have signals like `wsapi.complete.<name of wsapi>`.

Going a step further, this also makes me think that we could also insert `wsapi.begin.<name of wsapi>`.

I tend to overgeneralize, though. The signal names are fine as they stand :-)

lib/logging/transports/filters/metrics.js
jaredhirsch, line 14: In interest of keeping things moving, if you think this change is worth making, it's probably better to make that change in a separate PR.

jaredhirsch:
OK, looking good and tests passing. Merging in.

lib/logging/transports/filters/metrics.js
shane-tomlinson, line 14: @6a68 - I wasn't sure how to best handle this for performance. As a side note, we do use arrays for RegExps. I am worried about what happens at scale whether logging will become the bottleneck. :/

jaredhirsch, line 14: You know, that's an interesting question. First off, I'm not sure how much of an impact this logging code will have on the app as a whole. Maybe we could run two ephemeral instances, one with logging disabled, point load_gen at both of them, and see if the logging change alone produces a meaningful change in resource use.

Still, though, this does look like it might be one of those rare cases where micro-optimizations might actually have a significant impact. I'm reminded of the now a bit old, but still wonderfully informative, [talk](http://www.slideshare.net/newmovie/know-yourengines-velocity2011) Dave Mandelin gave at Velocity a couple of years ago.

PR #3969, opened 2013-10-10
Add statsd metric for looking up /.well-known/browserid

### LINKED ISSUE #1704: add statsd metrics to /wsapi/address_info for /.well-known/browserid lookup times via proxy
add statsd metrics to /wsapi/address_info for /.well-known/browserid lookup elapsed times via http proxy

shane-tomlinson:
Who wants to take it?

fixes #1704

### DIFF
--- a/lib/logging/transports/filters/statsd-timing.js
+++ b/lib/logging/transports/filters/statsd-timing.js
@@ -13,19 +13,28 @@ var TimingMessageMatches = {
 };
 
 var TimingRegExpMatches = [
+  /^elapsed_time\.(.*)/
 ];
 
-exports.test = function(msg) {
+function getMatch(msg) {
   if (msg in TimingMessageMatches) return true;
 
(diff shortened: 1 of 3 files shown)


jaredhirsch:
High-level question: is it better to record the elapsed time and do the math inside the app, or is it better for the app to log timestamps so that arbitrary intervals can be measured downstream?

I'm not sure of the answer.

jrgm:
Re: @6a68's question: don't we get both a timestamped logline **and** a statsd metric from the new logging arrangement?

shane-tomlinson:
@6a68 - updated to handle success and failure cases separately.

shane-tomlinson:
This is what an entry looks like in the logs:

```
{"elapsed":246,"level":"info","message":"elapsed_time.deep_fetch_well_known.error","timestamp":"2013-10-16T13:24:43.255Z"}
```

jaredhirsch:
OK, looking now.

jaredhirsch:
looks good, tests passing. R+, merging in. :beers:

PR #3971, opened 2013-10-10
Avoid double-counting the /sign_in page when users return from authenticating with their primary.

### LINKED ISSUE #3970: /sign_in metric counted twice if user redirects to primary
With #3939, we have the ability to send kpiggybank the number of users who see the /sign_in dialog directly from the backend. There is a problem, users who redirect to a primary IdP are counted twice, the first time when they first open the dialog, the second time when they return to from the IdP.

shane-tomlinson:
I am slightly uncomfortable with this and am opening to get feedback from @6a68, @kparlante and @seanmonstar.
- Instead of using the #AUTH_RETURN hash, use ?AUTH_RETURN, which is sent to the backend and can be used to avoid double count.
- Update tests.
- Get rid of the testmob.org include, it has been reaped.

fixes #3970

### DIFF
--- a/lib/logging/logging.js
+++ b/lib/logging/logging.js
@@ -37,7 +37,7 @@ const MetricsKpiggybankTransport = require('./transports/metrics-kpi');
 exports.logger.add(LogFileTransport, {});
 exports.logger.add(StatsdTransport, {});
 exports.logger.add(MetricsFileTransport, {});
-exports.logger.add(MetricsKpiggybankTransport, {});
+exports.logger.add(MetricsKpiggybankTransport.getInstance(), {}, true);
 
 exports.enableConsoleLogging = function() {
   exports.logger.add(winston.transports.Console, {
--- a/lib/logging/middleware/metrics.js
+++ b/lib/logging/middleware/metrics.js
@@ -6,20 +6,51 @@
  * Metrics middleware that reports when the signin dialog is opened.
  */
 
-const logger = require("../logging").logger;
 const urlparse = require('urlparse');
+const _ = require('underscore');
+const logger = require("../logging").logger;
 
 
 // utility function to log a bunch of stuff at user entry point
 module.exports = function(req, res, next) {
-  if (req.url === '/sign_in') userEntry(req);
+  // NOTE - this only counts the first time the user sees the dialog, it does
+  // not count if the user returns from a primary. req.url will be
+  // /sign_in?AUTH_RETURN or /sign_in?AUTH_RETURN_CANCEL for users who are
+  // returning from their IdP.
+  if (req.url === '/sign_in')
+    signInEntry(req);
+  else if (req.url === '/sign_in?AUTH_RETURN')
+    idpAuthReturnEntry('idp.auth_return', req);
+  else if (req.url === '/sign_in?AUTH_RETURN_CANCEL')
+    idpAuthReturnEntry('idp.auth_cancel', req);
+
   next();
 };
 
-function userEntry(req) {
+function signInEntry(req) {
+  entry('signin', req, { rp: getReferer(req) });
+}
+
+function idpAuthReturnEntry(type, req) {
+  entry(type, req, { idp: getReferer(req) });
+}
+
+function entry(type, req, data) {
+  logger.info(type, _.extend({
+    browser: req.headers['user-agent'],
+    // IP address (this probably needs to be replaced with the
+    // X-forwarded-for value
+    ip: getIpAddress(req)
+  }, data));
+}
+
+function getIpAddress(req) {
   var ipAddress = req.connection.remoteAddress;
   if (req.headers['x-real-ip']) ipAddress = req.headers['x-real-ip'];
+  return ipAddress;
+}
 
+function getReferer(req) {
   var referer = null;
   try {
     // don't log more than we need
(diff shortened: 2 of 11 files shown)


lib/logging/middleware/metrics.js
```diff
@@ -12,6 +12,8 @@ const urlparse = require('urlparse');
 
 // utility function to log a bunch of stuff at user entry point
 module.exports = function(req, res, next) {
+  // NOTE - this only counts the first time the user sees the dialog, it does
+  // not count if the user returns from a primary.
```
seanmonstar: oh right, because "/sign_in" !== "/sign_in?AUTH_RETURN". Took me a moment.

shane-tomlinson: I should probably say that in the comment, eh?

kparlante:
@shane-tomlinson : its an expansion of scope, but it would be useful to count the ?AUTH_RETURN case separately. In that case, the referrer is the IdP, which would give us a window on # of unique IdPs (and a heads up if new IdPs show up).

shane-tomlinson:
@kparlante - once #3968 is merged, I'll complete this. I have most of it done, but want to use the testing from that PR to enable clean testing of parts of this PR.

jaredhirsch:
#3968 is merged, sorry I've been lagging recently :beers:

jaredhirsch:
@shane-tomlinson OK, so my concern coming in was that the querystring would be visible to other servers, and maybe get cached. Looking at [our caching behavior for wsapis](https://github.com/mozilla/browserid/blob/dev/lib/wsapi.js#L341) and the HTTP 1.1 spec, I think we should be good.

Does anything show up in the querystring that could be spoofed and cause problems? Trying to think through potential issues caused by that being visible to servers.

lib/logging/middleware/metrics.js
```diff
@@ -6,20 +6,51 @@
...
+}
+
+function idpAuthReturnEntry(type, req) {
+  entry(type, req, { idp: getReferer(req) });
+}
+
+function entry(type, req, data) {
+  logger.info(type, _.extend({
+    browser: req.headers['user-agent'],
+    // IP address (this probably needs to be replaced with the
```
jaredhirsch, line 41: maybe now is the time to finally remove this slightly misleading comment?

I think we are specifically not ever logging the real IP address, it's always 127.0.0.1, is it worthwhile to bother including it?

shane-tomlinson:
@6a68 - this should be cached in the same way as `/sign_in`, which is to 'allow caching, but require revalidation via ETag' - see https://github.com/shane-tomlinson/browserid/blob/99b30ac00e638398902a395296a7b0bb74cb0ad7/lib/static/views.js#L42. I must admit, that max-age=0 seems like it would disable caching.

I have removed the IP address from the logs since it was always 127.0.0.1

lib/logging/middleware/metrics.js
shane-tomlinson, line 41: @6a68 - on ephemeral instances, the real address is actually logged. For instance, this is what I see in router.log when running Selenium tests from Sauce:

```
{"browser":"Mozilla/5.0 (Windows NT 6.1; WOW64; rv:22.0) Gecko/20100101 Firefox/22.0","ip":"199.27.108.155","rp":"http://logging-shane.123done.org","level":"info","message":"signin","timestamp":"2013-10-16T15:22:02.903Z"}
```

As you can see, I put the ip address stuff back in after seeing this. Note, we do NOT send the IP address off to kpiggybank.

lib/logging/logging.js
```diff
@@ -37,7 +37,7 @@ const MetricsKpiggybankTransport = require('./transports/metrics-kpi');
 exports.logger.add(LogFileTransport, {});
 exports.logger.add(StatsdTransport, {});
 exports.logger.add(MetricsFileTransport, {});
-exports.logger.add(MetricsKpiggybankTransport, {});
+exports.logger.add(MetricsKpiggybankTransport.getInstance(), {}, true);
```
shane-tomlinson, line 40: I use a singleton here so that I can get the same transport instance for testing.

shane-tomlinson:
This code is live on https://logging-shane.personatest.org

shane-tomlinson:
I added a new commit to log new primary users, which addresses #3985

lib/logging/middleware/metrics.js
jaredhirsch, line 41: Ah, that makes sense.

lib/logging/transports/filters/metrics.js
```diff
@@ -2,9 +2,25 @@
...
+ * stage_reverify.success
+ * stage_transition.success
+ * stage_user.success
+ * complete_email_confirmation.success
+ * complete_reset.success
+ * complete_transition.success
+ * complete_user_creation.success
+ * idp.auth_return
+ * idp.auth_cancel
+ * idp.create_new_user
```
jaredhirsch, line 22: this list doesn't seem to include metrics.report.*

jaredhirsch, line 22: eh, if I'm correct, just add this in a tiny follow-on PR.

jaredhirsch:
Looks good, tests pass. R+, merging.

Issue #1752, opened 2012-06-14
node-statsd leaks Client global variable


seanmonstar:
the `Client` function in node-statsd isn't a named function, or prefixed with `var`, so it leaks.

```
$ node
> var statsd = require('./lib/statsd');
undefined
> Client
[Function]
```

PR #3972, opened 2013-10-10
upgrade to node-statsd 0.0.7

seanmonstar:
- newer version works on windows
- removes global StatsD
- can be moved out of optionalDependencies

### DIFF
--- a/lib/logging/transports/statsd.js
+++ b/lib/logging/transports/statsd.js
@@ -4,17 +4,11 @@
 
 const util = require('util');
 const winston = require('winston');
+const StatsD = require("node-statsd").StatsD;
 const logger = require('../logging').logger;
 const config = require('../../configuration');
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
fixes #1752

shane-tomlinson:
Running tests now.

lib/logging/transports/statsd.js
```diff
@@ -4,18 +4,12 @@
 
 const util = require('util');
 const winston = require('winston');
+const StatsD = require("node-statsd").StatsD;
 const logger = require('../logging').logger;
 const config = require('../../configuration');
 const timing_filter = require('./filters/statsd-timing');
 const increment_filter = require('./filters/statsd-increment');
 
-var StatsD = false;
```
shane-tomlinson, line 12: I always like when the net amount of code is reduced. This felt gross.

PR #3977, opened 2013-10-15
Train 2013.10.09 hotfix 3976 undefined sitename

### LINKED ISSUE #3976: RP displayed as "undefined" in email verification screen when siteName not specified
Both myfavoritebeer.org and 123done.org use siteName so the names appear.

screen shot 2013-10-15 at 14 22 32

shane-tomlinson:
@seanmonstar - whatcha think, I've been overloading @6a68 with PRs lately.

Before merge, I want to make sure this isn't going to bite us anywhere else.

fixes #3976

### DIFF
--- a/resources/views/test.ejs
+++ b/resources/views/test.ejs
@@ -116,7 +116,6 @@
     <script src="/common/js/javascript-extensions.js"></script>
     <script src="/common/js/lib/bidbundle.js"></script>
     <script src="/build/code_version.js"></script>
-    <script src="https://testmob.org/include.js"></script>
     <script src="/common/js/gettext.js"></script>
     <script src="/common/js/lib/dom-jquery.js"></script>
     <script src="/common/js/lib/hub.js"></script>
--- a/resources/static/common/js/models/rp_info.js
+++ b/resources/static/common/js/models/rp_info.js
@@ -68,7 +68,7 @@ BrowserID.Models.RpInfo = (function() {
     },
 
     getSiteName: function() {
-      return this.siteName;
+      return this.siteName || this.getHostname();
     },
 
     getSiteLogo: function() {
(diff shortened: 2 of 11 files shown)


resources/static/dialog/js/modules/check_registration.js
```diff
@@ -18,9 +18,11 @@ BrowserID.Modules.CheckRegistration = (function() {
       options = options || {};
 
       self.checkRequired(options, "email", "rpInfo");
+      var rpInfo = options.rpInfo;
```
shane-tomlinson: can get rid of this line since rpInfo is only used once.

resources/static/dialog/js/modules/verify_primary_user.js
```diff
@@ -59,6 +59,8 @@ BrowserID.Modules.VerifyPrimaryUser = (function() {
     options = options || {};
     var self = this;
 
+    var rpInfo = self.rpInfo;
```
shane-tomlinson: get rid of this line since rpInfo is only used once.

shane-tomlinson:
@seanmonstar - an ephemeral instance, undefined-sitename.personatest.org is set up to test with.

resources/views/test.ejs
```diff
@@ -116,7 +116,6 @@
     <script src="/common/js/javascript-extensions.js"></script>
     <script src="/common/js/lib/bidbundle.js"></script>
     <script src="/build/code_version.js"></script>
-    <script src="https://testmob.org/include.js"></script>
```
jaredhirsch, line 119: I think this is a stray change from another PR?

shane-tomlinson, line 119: testmob.org got reaped. :(

jaredhirsch, line 119: shane pointed out in IRC that this fix was merged separately into dev, but not stage. that's why it is here

jaredhirsch:
R+, merging.

Some yak-shaving discussion happened between me and shane in IRC, mentioning for history but it's not important enough to paste in here and reformat: http://irclog.gr/#show/irc.mozilla.org/identity/330185

PR #3979, opened 2013-10-15
Issue2647

shane-tomlinson:
@mohit-agarwal - I am opening a new PR to address #2647, this is your PR with some tests included. How does this look to you?


mohit-agarwal:
@shane-tomlinson : Ya this seems fine now..It is passing all the tests.

Issue #2497, opened 2012-09-17
login.persona.org should be a true IdP


benadida:
with authentication and provision endpoints, so a native implementation can fallback to it normally.

lloyd:
interesting...  Is this fully designed?  For the web case, would we use the authentication and provisioning endpoints?  Or would these endpoints be _only_ a convenience for native implementations?

I think working through the implementation for 30 minutes will shake out most of the issues.  I am glad to do this, and all I need is full, yet terse, context on all of your ideas about how this will work and why.

The greatest issue I see right now is in explicitly defining the overlapping roles of the current persona servers and the native implementation.  Understanding precisely how much of the implementation will actually get moved to the native implementation is critical in understanding what exactly the authentication endpoint will display to a user and whether this is a new orthogonal feature for native implementations, or a small restructure of our existing resources.

benadida:
@lloyd this is the thing we spent a few hours discussing a while back, and I believe you were included. This has to do with appending the GET parameter to the .well-known, the same solution we're using for BigTent.

So yes, I believe this is fully designed, and written up in the spec, and I'm pretty sure you were part of that discussion, but of course we can discuss again if there are any doubts!

For the shimmed case, I don't think we would need to use the auth and provisioning endpoints explicitly, since we're trying to have a dynamic single-page web app. But it's not just a convenience for native implementations, it is _required_. Otherwise, native implementations can't use the fallback IdP.

lloyd:
Woah there, Ben:  I'm not pushing back here.

I remember this idea, I understand the importance, I just don't remember the feasibility assessment and understanding precisely how we would pull it off.

I just don't feel like this project is sufficiently clear to open an issue about.

I'm curious about things like:
1. if the user already has a persona account and then moves to their phone and enters an email from their existing account, are the asked for their password?
2. if the user enters an email that is not known by persona, do they have to choose a persona password to record the verification?  At what point is that password picked?  Is it inside the Auth endpoint?

I'm just trying to figure out the nature of the relationship between the phone implementation and our servers.  I've presented the persona servers in the past as a combined "Implementation Provider" (shim) - and "Secondary identity authority" - now we're trying to split these two things apart.  

I just want to make sure we don't hit an artery.

/cc @ozten @jedp

benadida:
@lloyd sorry, my tone was off because I was writing too quickly before running out on an errand. I didn't mean to sound defensive.

I do think we have thought it through quite fully, but let's check.

[...]

PR #3982, opened 2013-10-16
Issue 2497 refactor to idp

ozten:
Background for pull request: Issue https://github.com/mozilla/browserid/issues/2497#issuecomment-26439900

Creating a pull request to try to get some frontend help.

Thanks to @shane-tomlinson for previously reviewing and getting it this far.

Known Issues:
- siteName and other options aren't passed in from Desktop code, I'm investigating in gecko

Needs Frontend <3
- rpinfo changes in dev broke this patch, minimal hardcoded changes were made to get it working again
  - Need help doing this correctly


seanmonstar:
fabulous! _gets reviewing goggles_

lib/static/views.js
```diff
@@ -456,17 +475,17 @@ exports.setup = function(app) {
...
+    var proxyIdps = config.get('proxy_idps') || {};
     // Can we delegate authority to BigTent?
     if (proxyIdps[domain]) {
       res.json({ 'authority': proxyIdps[domain] });
       // Be a Secondary
     } else {
-      res.json({ 'public-key': publicKey.toSimpleObject() });
+      res.json({
+        'public-key': publicKey.toSimpleObject(),
+          'provisioning': '/provision',
```
seanmonstar: indent is off

resources/static/common/js/lib/hub.js
```diff
@@ -12,6 +12,7 @@ Hub = (function() {
       currID = 0;
 
   function on(message, callback, context) {
+    console.log('HUB ON ' + message);
```
seanmonstar: reminder to kill these `console.logs` before merge

lib/static_resources.js
```diff
@@ -123,6 +123,60 @@ var dialog_js = und.flatten([
...
+  '/common/js/renderer.js',
+  '/common/js/lib/module.js', 
+  '/common/js/class.js', 
+  '/common/js/tooltip.js',
+  '/common/js/mediator.js',
+  '/common/js/validation.js',
+  '/common/js/helpers.js',
+  '/common/js/gettext.js',
+  '/common/js/browser-support.js',
+  'common/js/enable_cookies_url.js',
```
shane-tomlinson: is this supposed to have a `/` prepended?

lib/wsapi/discovery.js
```diff
@@ -0,0 +1,56 @@
...
+  'domain': 'hostname'
+};
+exports.i18n = false;
+
+
+// determine public hostname for use as the issuer for
+// secondary addresses
+const HOSTNAME = urlparse(config.get('public_url')).host;
+
+try {
```
shane-tomlinson, line 27: It seems like this try/catch could be generalized and placed into `primary.js`

shane-tomlinson, line 27: my bad, `secrets.js`

lib/wsapi/discovery.js
```diff
@@ -0,0 +1,56 @@
...
+  logger.error("can't read public key, exiting: " + e);
+  process.nextTick(function() { process.exit(1); });
+}
+
+exports.process = function(req, res) {
+  var domain = req.params.domain;
+  var wellKnown;
+  primary.checkSupport(domain, function(err, r) {
+    if (!err && r && r.urls) {
+      wellKnown = {
```
shane-tomlinson, line 39: Can both of these well known settings be extracted into functions?

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,96 @@
...
+      ANIMATION_TIME = 500,
+      EMAIL_PRESENT_NEXT = 'button:visible',
+      sc;
+
+  var Module = bid.Modules.PageModule.extend({
+    start: function(options) {
+      options = options || {};
+      var self = this;
+
+      window.resizeTo(700, 440); // Gross
```
shane-tomlinson: resizeBy might give you better mileage, resizeTo includes browser chrome which can differ depending on toolbars.

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,96 @@
...
+          var email = info.email;
+          // TODO: Are we sure that we authed as email?
+          moduleManager.start("set_password", info);
+          mediator.subscribe("password_set", function(msg, info) {
+            // TODO merge broke us - user.setRPInfo must now be called like in
+            // resources/static/dialog/js/modules/dialog.js line 130, 286
+            //      params.origin = user.getOrigin();
+            var rpInfo = bid.Models.RpInfo.create({origin: 'http://192.168.186.138:10001'});
+            user.setRpInfo(rpInfo);
+            dialogHelpers.createUser.call(self, email, info.password, function(info) {
```
shane-tomlinson: Can these be extracted into functions? The pyramid o' doom is getting awfully deep.

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,96 @@
...
+
+        mediator.subscribe("new_user", function(msg, info) {
+          var email = info.email;
+          // TODO: Are we sure that we authed as email?
+          moduleManager.start("set_password", info);
+          mediator.subscribe("password_set", function(msg, info) {
+            // TODO merge broke us - user.setRPInfo must now be called like in
+            // resources/static/dialog/js/modules/dialog.js line 130, 286
+            //      params.origin = user.getOrigin();
+            var rpInfo = bid.Models.RpInfo.create({origin: 'http://192.168.186.138:10001'});
```
shane-tomlinson: ya, we need this info for the origin somehow. Are all the RP parameters passed here? If so, we'll need tos/pp, siteLogo and backgroundColor as well.

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,96 @@
...
+                  } else {
+                    self.renderError("error", {
+                      action: {
+                        title: "error in waitForUserValidation 2",
+                        message: "Expected 'complete', but got '" + msg + "'"
+                      }
+                    });
+                  }
+                },
+                function(err) {
```
shane-tomlinson: I wanna kill the error rendering with a spoon.

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,96 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+BrowserID.Modules.IdPAuthentication= (function() {
```
shane-tomlinson, line 4: so, FWIU, this is a super mini state machine. With as yuck as the state machine is, we may want to use the generalized portion of it here if there is ever a point where the user can go back a screen within the IdP. I can help you with that if necessary.

resources/views/authenticate_layout.ejs
```diff
@@ -0,0 +1,78 @@
...
+                <div class="contents">
+                    <div class="loadingSpinner"></div>
+                    <h2 class="center"><%= gettext('Connecting to Persona...') %></h2>
+                </div>
+            </div>
+        </div>
+      </section>
+
+      <% } %>
+
```
shane-tomlinson, line 96: I believe the error/waiting/delay sections will need to be added here to handle when those events occur.

resources/views/provision.ejs
```diff
@@ -0,0 +1,65 @@
...
+function fail() {
+  var msg = "user is not authenticated as target user";
+  navigator.id.raiseProvisioningFailure(msg);
+}
+
+navigator.id.beginProvisioning(function(email, duration) {
+  getJSON('/wsapi/session_context', function(session, text, xhr) {
+    if (session.authenticated) {
+      getJSON('/wsapi/list_emails', function(data, text, xhr) {
+        if (data.emails.indexOf(email) === -1) {
```
shane-tomlinson, line 25: Is there ever a chance in hell we'll use this universally? If so, [].indexOf will break in IE8.

resources/views/provision.ejs
```diff
@@ -0,0 +1,65 @@
...
+              type: 'POST',
+              contentType: 'application/json',
+              data: JSON.stringify(params),
+              dataType: 'text',
+              success: function(certificate, status, xhr) {
+                navigator.id.registerCertificate(certificate);
+              },
+              error: function(xhr, status, err) {
+                // TODO: Handle errors
+                if (console && 'function' === typeof console.log) {
```
shane-tomlinson: There is navigator.id.raiseProvisioningFailure I believe for this case.

resources/static/dialog/js/modules/idp_authentication.js
shane-tomlinson: So... What we need is a way to get those parameters into this dialog....  We could drag over the channel stuff that allows you to call navigator.id.registerChannel, and then pass them in that way. We could use query strings. We could possibly stuff that info into local or sessionStorage? Indexeddb? Dunno what the best route forward is, would like to find out what capabilities you have.

shane-tomlinson:
So... The devil hiding in the details, password resets and transition states. Are they going to be handled?

resources/static/common/js/lib/hub.js
ozten: Ya, the version of Desktop codebase that this runs against has horrible dev tools, so I need these crutches.

lib/wsapi/discovery.js
ozten, line 27: This call is made from 7 sites in the code. Not all of them wrap in a try/catch.
I'd rather not do this change which is outside of the scope of this PR.

lib/wsapi/discovery.js
ozten, line 39: Sure, what do you mean exactly? Something like this?

```
function prepareWellKnown(pubKey, provUrl, authUrl) {
  provUrl = provUrl || config.get('public_url') + '/provision';
  ...
  return {
    "public-key": pubKey,
    provisioning: provUrl,
    ...
  }
}
```

resources/views/provision.ejs
ozten, line 25: If IE8 implements persona natively and re-uses these flows, I'll eat my hat. :)

resources/static/dialog/js/modules/idp_authentication.js
ozten: Do you have a SHA for before and after the rpinfo stuff?

What are the inputs and were do they come from? We should probably solve this in content space, but natively is also an option.

ozten:
> The devil hiding in the details, password resets and transition states. Are they going to be handled?

Can we land this and start getting UX feedback, before implementing those features?

resources/static/dialog/js/modules/idp_authentication.js
ozten, line 4: I'm trying to do this, but not really sure what to do. Can you make a pull request off of issue-2497-refactor-to-idp with a sketch (or complete implementation) of what you mean?

ozten:
Trying to get awsbox working, I've merged from dev, but got bitrot for the second time.

I was able to see that logging.js moved and fix that.

The other issue is that the dialog just comes up blank.

To unblock myself, I've copied over scripts/deploy.js and changed the dependency to awsbox@0.6.2... but we'll have to figure out what changed about the dialog code that has broken the initial screen display for the new /auth screen. Any ideas?

jaredhirsch:
@ozten I'll do some git twiddling on Monday, see if I can cleanly apply your changes on top of current dev.

jaredhirsch:
ok, looking now.

jaredhirsch:
@ozten nice work so far!

I was able to cleanly cherry-pick all the changes onto current dev (except for the commits merging dev into this branch). My branch is at https://github.com/6a68/browserid/tree/issue-2497-idp-rebased-not-merged.

One tiny logging path change was needed to get tests running, that's on the HEAD of that branch, [a7fd566](https://github.com/6a68/browserid/commit/a7fd566e59f4110089617414820d86ad70504c54).

The local RP seems fine to me on my branch, I can log in and out.

ozten:
@6a68 is rocking the merge from dev to update this branch, you rule!

Please try to reproduce the blank dialog I was seeing.

You can download a Persona enabled Desktop build:
http://ftp.mozilla.org/pub/mozilla.org/firefox/try-builds/aking@mozilla.com-62351505bca9/

Set a couple preferences (about:config):

```
dom.identity.enabled = true
dom.identity.persona_fallback = http://127.0.0.1:10002
toolkit.identity.debug = true
```

Restart Firefox.

dom.identity.persona_fallback defaults to our new ephemeral server https://desktoppersona.personatest.org/ but you can change this to any environment.

Steps to reproduce:

1) Launch Firefox with a clean profile
2) Go to  http://127.0.0.1:10001
3) Click Get an assertion
4) Enter foo@bar.com

Actual: empty dialog

Expected: prompt to set a password

jaredhirsch:
@ozten happy to help :-)

I've got the special FF build, I toggled the settings you mentioned (actually tried using the desktoppersona fallback as well as localhost), but I'm not sure what to do next.

When I try to log into RPs, I click the persona button but nothing happens. Tried 123done, mozillians, nothing. Also not seeing errors in the JS console--maybe there's a desktop-specific console for the native code that I'm unaware of?

I'll poke a bit more at my branch and yours, see if I can at least get an empty dialog using your branch.

ozten:
So in nightly if you go to about:config and search for identity does it look like this 

Clicking get Assertion should look like this

jaredhirsch:
It looks like the blank dialog error was caused by using hashes, but the recently-landed #3971 I think means we should be using querystring instead.

@ozten said this got him past the blank dialog problem in a local branch, I've updated my [issue-2497-idp-rebased-not-merged](https://github.com/6a68/browserid/tree/issue-2497-idp-rebased-not-merged) branch with the fix (commit (9e8ed0f)[https://github.com/6a68/browserid/commit/9e8ed0f]).

jaredhirsch:
@ozten i don't have the `toolkit.identity.enabled` pref at all, but I have the other three matching your screenshot.

The second image is broken for me.

ozten:
Thanks Jared!

Using your patch, I see `Error: Relay frame could not be found`.

Thanks for continuing to dig for breaking changes from merging.

jaredhirsch:
I have my current branch running on https://rebasedidp.personatest.org. Will pick this up again in the morning. :beers:

jaredhirsch:
@ozten looks like [this line](https://github.com/6a68/browserid/blob/issue-2497-idp-rebased-not-merged/resources/static/dialog/js/modules/idp_authentication.js#L43) probably shouldn't have that IP hard-coded. I'm not totally clear on where we can get the proper domain info in this context, though. any ideas?

jaredhirsch:
heh, I caused the Relay Error: we look for the hash, not the querystring, when we [setup the channel](https://github.com/6a68/browserid/blob/issue-2497-idp-rebased-not-merged/resources/static/dialog/js/modules/dialog.js#L98). Removing that, or changing that code to look for ?NATIVE as window.location.search, returns us to the mysterious empty dialog situation we had before. Continuing to dig.

ozten:
> @ozten looks like this line probably shouldn't have that IP hard-coded. I'm not totally clear on where we can get the proper domain info in this context, though. any ideas?

Yes, I've noted (here in this pull request) that I need help. I've hardcoded rpInfo to unblock myself locally. I've asked for either:

1) A patch to do this correctly or
2) An explaination of what changed and what I need to do or
3) the SHA that introduced rpinfo changes and I'll figure it out

rpInfo was the first bitrot that broke this pull request (for context).

Perhaps related is that the Desktop isn't sending in all the options that come into the navigator.id.request call. I'm going to look into that part, as it's desktop code.

Thanks again.

jaredhirsch:
@ozten ah, ok. now it all makes sense. i'll keep digging in the dialog code.

lib/static_resources.js
```diff
@@ -123,6 +123,60 @@ var dialog_js = und.flatten([
     '/dialog/js/start.js'
   ]]);
 
+var idp_authenticate_min_js = '/production/:locale/idp_authenticate.js';
```
fmarier, line 128: Suggested comment:

```
// Bundle for user agents wanting to make use of the fallback IdP
```

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,108 @@
...
+  function passwordSetComplete(self, email) {
+    return function(msg, info) {
+      // TODO merge broke us - user.setRPInfo must now be called like in
+      // resources/static/dialog/js/modules/dialog.js line 130, 286
+      //      params.origin = user.getOrigin();
+      var rpInfo = bid.Models.RpInfo.create({origin: 'http://192.168.186.138:10001'});
+      user.setRpInfo(rpInfo);
+      dialogHelpers.createUser.call(self, email, info.password, function(info) {
+        if (info.success) {
+          // TODO Desktop code isn't passing through this information
```
ozten: I think this problem is in Desktop code. Also, we could land this pull request in production without breaking or bothering anyone, since they are new, un-used paths. I'm fine with either landing hard coded settings here or removing them.

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,108 @@
...
+          // TODO: Are we sure that we authed as email?
+          navigator.id.completeAuthentication();
+        });
+
+        mediator.subscribe("new_user", newUserComplete(self));
+
+        moduleManager.start("authenticate", {
+          email: email
+        });
+        $(EMAIL_PRESENT_NEXT).click();
```
fmarier, line 93: This click of the Next button is necessary to simulate `emailHint`. Once we have `emailHint` we should make use of it here and remove this small hack.

I'm not suggesting we should block on this, but maybe there should be a `TODO: use emailhint instead of this hack` comment.

resources/static/pages/js/idp_authenticate.js
```diff
@@ -0,0 +1,58 @@
+// In the style of pages/js/start.js
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+$(function() {
+  "use strict";
+
+  window.resizeTo(700, 440); // Gross
```
ozten: I don't think we need to do this, since it happens as a side-effect of calling start on our new module. Right?

lib/wsapi/discovery.js
```diff
@@ -0,0 +1,56 @@
...
+        "public-key": JSON.parse(r.publicKey.serialize()), // gross
+        authentication: r.urls.auth,
+        provisioning: r.urls.prov,
+      };
+    } else {
+      if (err) {
+        logger.info('"' + domain + '" primary support is misconfigured, falling back to secondary: ' + err);
+      }
+      wellKnown = {
+        "public-key": publicKey.toSimpleObject(),
```
fmarier: Suggestion: if you were to rename `publickey` to `fallbackpubkey` (or something similar) then it wouldl be more obvious that this is a constant and not anything that's coming from the callback.

Obviously it's not a big deal, but I thought I'd mention it because I had to look at that twice while I was reading this bit.

resources/static/common/js/error-messages.js
```diff
@@ -158,6 +158,10 @@ BrowserID.Errors = (function(){
 
     xhrError: {
       title: gettext("Communication Error")
+    },
+
+    waitForUserValidation: {
+      title: "Waiting for user validation"
```
fmarier, line 164: No `gettext()`? Is this error message not meant to be translated?

ozten, line 164: No `gettext` is intentional, happy to change this.

Most error messages aren't translated. They end up hidden behind "more info". I think the idea is that it is useful for developers and would be a burden on the L10n community for not much benefit.

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,108 @@
...
+      mediator.subscribe("password_set", passwordSetComplete(self, info.email));
+    };
+  }
+
+  function passwordSetComplete(self, email) {
+    return function(msg, info) {
+      // TODO merge broke us - user.setRPInfo must now be called like in
+      // resources/static/dialog/js/modules/dialog.js line 130, 286
+      //      params.origin = user.getOrigin();
+      var rpInfo = bid.Models.RpInfo.create({origin: 'http://192.168.186.138:10001'});
```
fmarier: Where is that hard-coded IP address coming from?

resources/static/common/js/error-messages.js
fmarier, line 164: Fair enough. I don't think it would be all that useful to end-users.

fmarier:
We should get checkmyidp.org to use the new discovery API as a final check at the end. Filed as https://github.com/shane-tomlinson/checkmyidp.org/issues/8.

tests/discovery-test.js
```diff
@@ -0,0 +1,85 @@
...
+
+var suite = vows.describe('discovery');
+
+const TEST_DOMAIN = "example.domain",
+      TEST_EMAIL = "test@" + TEST_DOMAIN,
+      TEST_ORIGIN = 'http://127.0.0.1:10002';
+
+const SECONDARY_TEST_DOMAIN = "example.com";
+
+// an explicitly disabled domain
```
fmarier, line 30: We should make use of this in a test. Let's ensure that `disabled: true` domains always "delegate" to the fallback.

resources/static/dialog/js/modules/idp_authentication.js
shane-tomlinson, line 4: OTOH, we should hold off on adding the state machine functionality until we have more direction. No need to add complexity where it's not yet needed.

ozten:
@fmarier or @shane-tomlinson I've removed temporary hacks, console.log, etc. We should be good to merge as discussed.

General note: The Goldilocks API breaks the Native implementation, so use http://192.168.186.138:10001/watch.html instead of http://192.168.186.138:10001/

ozten:
G R E E N

resources/static/dialog/js/modules/idp_authentication.js
```diff
@@ -0,0 +1,104 @@
...
+      navigator.id.beginAuthentication(function(email) {
+        mediator.subscribe("authentication_success", function(msg, info) {
+          // TODO: Are we sure that we authed as email?
+          navigator.id.completeAuthentication();
+        });
+
+        mediator.subscribe("new_user", newUserComplete(self));
+
+        var rpInfo = bid.Models.RpInfo.create({
+          origin: "http://127.0.0.1:10001",
```
fmarier: Did you forget to take these out?

fmarier:
r+ from me. makes Travis happy. tested by Shane and Austin.

ozten:
All the man needed was a Guiness at the airport!

Thanks!

PR #3984, opened 2013-10-17
remove obsolete reference to env.ZERIGO_DNS_KEY and env.PERSONA_DEPLOY_DNS_KEY 

jrgm:
Also, add a way to specify default domain with env.PERSONA_DEPLOYMENT_DOMAIN (instead of assuming .personatest.org)

### DIFF
--- a/scripts/deploy.js
+++ b/scripts/deploy.js
@@ -9,7 +9,7 @@ child_process = require('child_process');
 
 /*
  * A thin wrapper around awsbox that expects certain env
- * vars and invokes awsbox for ya to deploy a VM. 
+ * vars and invokes awsbox for ya to deploy a VM.
  */
 
 if (!process.env['AWS_ID'] || ! process.env['AWS_SECRET']) {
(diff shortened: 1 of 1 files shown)


scripts/deploy.js
```diff
@@ -38,21 +31,17 @@ if (process.argv.length > 1 &&
...
-    } else if (options.n) {
-      options.u = scheme + options.n + ".personatest.org";
-    }
+  var scheme = (options.p ? 'https' : 'http') + '://';
 
-  } else {
-    console.log('WARNING: No DNS key defined in the environment!  ' +
-                'I cannot set up DNS for you.  We\'ll do this by IP.');
+  if (process.env['PERSONA_DEPLOYMENT_HOSTNAME']) {
+    options.u = scheme + process.env['PERSONA_DEPLOYMENT_HOSTNAME'];
```
shane-tomlinson, line 41: huh, I never knew about this.

Issue #3987, opened 2013-10-17
Explore using `mozbrowser` on ToS/PP in native persona to allow JS to run


mattbasta:
Our ToS/PP are still broken on native persona because we use JS to load them (and the way we do L10n means it'll be really terrible to change that). Can you guys experiment with `mozbrowser` or something like that to allow those documents to load with scripts? Surely there's a way to pull it off, otherwise hosted apps wouldn't be able to run JS ;)

shane-tomlinson:
@mattbasta - is there some documentation on mozbrowser? I can give it a try once I have some info.

shane-tomlinson:
Some links around the webs:

https://github.com/mozilla-b2g/gaia/pull/7178
https://groups.google.com/forum/#!msg/mozilla.dev.webapi/LFuVuUvqT6Q/pcOcvXIo6sIJ
https://bugzilla.mozilla.org/show_bug.cgi?id=714861

mattbasta:
It's pretty straightforward. You just add the mozbrowser attribute to an
iframe. The only stipulation is that you need to request the permission for
it, but that shouldn't be an issue for persona.

lloyd:
@mattbasta because in firefox os persona is remote content, we can't actually do that.

mattbasta:
Can you explain the risk of allowing JS in the ToS/PP iframe? If Peraona is
remote content then I don't understand how it has any privileges which
could be compromised.

lloyd:
because your private key lives in local storage in the domain, the attack would be identity compromise.

mattbasta:
How exactly could the iframe extract information from the parent's localStorage? That would violate the same origin policy.

shane-tomlinson:
@mattbasta - IIRC in the flurry of last minute FFOS activity, we chose to sandbox the iframe to avoid bad acting RPs from frame busting and taking over the trusted UI. I am not sure if we ever experimented to see our worries were well founded, but we took the cautious approach to avoid shooting ourselves in the foot.

mattbasta:
I've filed a pull request (as you can see above).

A few notes:
- If you don't specify `allow-top-navigation`, it shouldn't be possible to frame-bust the page.
- Even if the page did, it's effectively the same as if they called `window.open`, yeah?
- Everything should be running in its own origin anyway

Please let me know if I'm making faulty assumptions.

shane-tomlinson:
[...]

PR #3988, opened 2013-10-17
Allow scripts in inline ToS/PP pages (issue #3987)

mattbasta:
This allows scripts to run in inlined ToS/PP pages, but not to access local assets. Even if an attacker specified a page on the persona origin, it would run in a unique origin that would not have access to the local storage.


jaredhirsch:
closing, parent bug has been closed as well

PR #3993, opened 2013-10-22
update db for convict's falsiness

### LINKED ISSUE #3989: upgrade to convict 0.4.1 broke local testing with back_mysql
The commit was from https://github.com/mozilla/browserid/pull/3973, but it points out that there may be some issues elsewhere with about defaults with new convict that are baked into the code that uses the old convict. 

STR:
1. git clone git://github.com/mozilla/browserid.git
2. cd browserid
3. npm install
4. WHAT_TESTS=back_mysql npm test

Expected: tests pass

Actual: 

```
> browserid@1.0.0-b2 test /home/ec2-user/tmp/browserid
> ./scripts/test
>>> Now Running: API level unit tests using MySQL
Can't run tests: can't connect to the database
```

Notes:
1. I can reproduce this on linux or osx.
2. Travis tests pass because .travis.yml sets `MYSQL_USER=root` in the environment.

This can be "fixed" with this change:

```
diff --git a/lib/db/mysql.js b/lib/db/mysql.js
index 84e97dc..65a1b83 100644
--- a/lib/db/mysql.js
+++ b/lib/db/mysql.js
@@ -121,7 +121,7 @@ exports.open = function(cfg, cb) {
   };

   Object.keys(options).forEach(function(param) {
-    options[param] = (cfg[param] !== undefined ? cfg[param] : options[param]);
+    options[param] = (cfg[param] ? cfg[param] : options[param]);
     if (options[param] === undefined) delete options[param];
   });
```

but seems better to me to file this as an issue to track down any other places where we might be bit by the same pattern of `falsiness` changes in meaning with the new convict.

Marking as 5-star.

seanmonstar:
fixes #3989

### DIFF
--- a/lib/db/mysql.js
+++ b/lib/db/mysql.js
@@ -121,7 +121,7 @@ exports.open = function(cfg, cb) {
   };
 
   Object.keys(options).forEach(function(param) {
-    options[param] = (cfg[param] !== undefined ? cfg[param] : options[param]);
+    options[param] = cfg[param] || options[param];
     if (options[param] === undefined) delete options[param];
   });
 


jaredhirsch:
looking now

jaredhirsch:
yup, this fixes the unit tests on mysql. :beers:, merging in

PR #3997, opened 2013-10-25
Extract session context objects into their own objects.

shane-tomlinson:
Extract session context from user.js & network.js into two models, user-context and network-context.

### DIFF
--- a/resources/static/common/js/user.js
+++ b/resources/static/common/js/user.js
@@ -22,9 +22,8 @@ BrowserID.User = (function() {
       pollDuration = POLL_DURATION,
       stagedEmail,
       stagedPassword,
-      userid,
-      auth_status,
-      PERSONA_ORG_AUDIENCE = "https://login.persona.org";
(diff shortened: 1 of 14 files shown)


resources/static/common/js/user.js
```diff
@@ -916,10 +923,13 @@ BrowserID.User = (function() {
...
         network.listEmails(function(server_emails) {
-          // update our local storage map of email addresses to user ids
-          if (userid) {
-            storage.updateEmailToUserIDMapping(userid, server_emails);
-          }
+          withContext(function(context) {
+            var userid = context.getUserId();
+            // update our local storage map of email addresses to user ids
+            if (userid) {
+              storage.updateEmailToUserIDMapping(userid, server_emails);
```
shane-tomlinson, line 930: This feels backwards, perhaps storage.js should have access to user-context and passing the user-id could be avoided.

seanmonstar:
looks good, clean.

PR #4012, opened 2013-11-05
CSP headers

### LINKED ISSUE #3271: Add CSP headers to login.persona.org
While we've writing about [CSP](https://developer.mozilla.org/en-US/docs/Security/CSP) in a number of other bugs, it doesn't look like we have a bug open for actually adding CSP headers to `login.persona.org`.

Even if we still have eval's in our code, we could serve a `Content-Security-Policy` header with `script-src: unsafe-eval` and a `X-Content-Security-Policy` header with `options eval-script`.

The big win for us would be that we could disable all inline scripts.

seanmonstar:
- use hood to add security headers
- adds CSP, X-Content-Type-Options: nosniff
- adjusts HSTS and X-FRAME-OPTIONS

fixes #3271

### DIFF
--- a/lib/configuration.js
+++ b/lib/configuration.js
@@ -409,10 +409,41 @@ var conf = module.exports = convict({
     format: String,
     default: "DENY"
   },
+  csp: {
+    headers: {
+      doc: 'The headers used for CSP',
+      default: ['Content-Security-Policy']
+    },
+    policy: {
+      doc: "Content-Security-Policy",
+      format: Object,
+      default: {
+        // defaulting to nothing! if you want it, turn it on
+        'default-src': ['none'],
+        'script-src': ['self'],
+        'style-src': ['self', 'unsafe-inline'],
+        'connect-src': ['self'],
+        'font-src': ['self'],
+        'img-src': '*', // to load siteLogos
+        'frame-src': '*' // must load provisioning iframes :(
+      }
+    }
+  },
+  hsts: {
+    maxAge: {
+      format: 'duration',
+      default: 10886400
+    },
+    includeSubdomains: {
+      format: Boolean,
+      default: true
+    }
+  },
   measure_dom_loading: {
     doc: "Measure the time it takes the DOM to start loading by embedding an inline script in the document HEAD",
     format: Boolean,
-    default: false
+    default: false,
+    env: 'MEASURE_DOM_LOADING'
   },
   forcible_issuers: {
     doc: "Hostnames which this Persona instance will issue identies for. Used in b2g only.",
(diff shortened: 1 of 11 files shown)


fmarier:
(This isn't really a problem with the pull request, but rather a comment for hood...)

I think that we should only send the official `Content-Security-Header`. I've had of severe bugs in previous implementations of CSP (especially on Safari) and I don't think the benefit of enabling CSP on older browsers outweights the risk of breaking these browsers.

lib/configuration.js
```diff
@@ -409,10 +409,35 @@ var conf = module.exports = convict({
     format: String,
     default: "DENY"
   },
+  csp: {
+    policy: {
+      doc: "Content-Security-Policy",
+      format: Object,
+      default: {
+        // cant use default-src, since that includes iframes, and we
```
fmarier: Ideally, we should have `default-src 'none'` and then enable the things we need (like `frame-src *` as per your comment).

seanmonstar: oh snap. i didn't realize you could use `*` to reset one of the options. I didn't use `default-src` because I thought that meant `frame-src` was doomed.

seanmonstar:
interesting about the older CSP headers. I didn't know of any bugs, and saw that others use them also. I can get rid of them in hood, but I also allowed the option to specify your own headers, which I can easily do here.

fmarier:
I deployed this code to awsbox and I'm not sure why, but I'm not getting the HSTS header:

```
$ curl --head https://francois3271.personatest.org/include.js
HTTP/1.1 200 OK
Server: nginx/1.3.16
Date: Thu, 07 Nov 2013 14:19:44 GMT
Content-Type: application/javascript
Content-Length: 17572
Connection: keep-alive
X-Powered-By: Express
Content-Security-Policy: default-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; connect-src 'self'; font-src 'self'; img-src *; frame-src *;
X-Content-Type-Options: nosniff
Vary: Accept-Encoding,Accept-Language
Cache-Control: public, max-age=0
ETag: "17572-1383824474000"
X-Frame-Options: DENY
Access-Control-Allow-Origin: https://francois3271.personatest.org
```

fmarier:
Should we have different CSP policies for the different services? The one that's in there at the moment looks good for the shim/fallback idp, but I'm guessing we could have stricter ones for `static.login.persona.org` and `verifier.login.persona.org`.

seanmonstar:
Ah yes, I had been foiled testing against an awsbox while npm was down. I'll test and setup an awsbox also.

seanmonstar:
As for having different policies: it seems CSP is only useful for pages of HTML. So, /verify only returns JSON from a POST, and static only includes styles, images, and scripts. I don't think the CSP would do anything for either. Still worth it?

seanmonstar:
There was some code in wsapi that was adding `req.connection.proxySecure`, but it happened after `hsts` did it's check. I moved that code to a tiny middleware, and included it just before hood everywhere.

Confirmed that HSTS was sent on this awsbox: https://csp.personatest.org

seanmonstar:
@fmarier whatcha think?

fmarier:
You're right, CSP only makes sense (and is only executed) for HTML / XHTML pages.

This looks good to me. I've tested it on an awsbox and ran the unit tests. r+

PR #4017, opened 2013-11-07
Issue 3867: Change footer icon which is blurry and cropped in dialog

Somsubhra:
Changed the base64 image at the footer of the dialog and kept the image size according to the aspect ratio.


fmarier:
Did you also change the source PNG?

Somsubhra:
No. I just changed the base64 code. That would be persona_watermark_footer_sm.png?

fmarier:
Yes, the base64 string is a straight `base64 dialog/i/persona_watermark_footer_sm.png`, we should make sure the source and the CSS file are kept in sync.

Issue #1946, opened 2012-07-06
Please support a more traditional login flow


st3fan:
BrowserID currently requires a 'full' browser because it depends on javascript, frames and popup windows.

My request is to support a more classic login flow where an app can simply redirect to BrowserID.org/login?callback=http://myapp.com/foo and browserid will redirect back to the application on the specified callback, passing it the assertion. The app can then verify the assertion as usual and proceed.

This has the advantage that BrowserID will work with minimal effort on more limited browsers like for example Android and iOS applications that simply open a WebView or browser on devices that simply do not have the UI to support popup windows. Or third party iOS browsers that generally do not support popup windows.

This will also greatly improve things for native applications that run just BrowserID in a webview. They can then pass a callback with a custom protocol that can be captured by the native app more easily. This is a common technique on both Android and iOS.

benadida:
@st3fan importantly, this has the big disadvantage that it ties down everything to browserid.org, which is a centralized architecture we want to move away from.

I think we should explore better SDKs for mobile without forcing a centralization of the protocol.

st3fan:
I understand. Maybe browserid.org can just be the springboard and do the discovery and redirect to another service if needed?

gerv:
st3fan: can sites even detect that they are in a UIWebView, as opposed to Mobile Safari? Would you have to detect that the popup open failed?

benadida: only, surely, for limited capability browsers? Which are always going to need the browserid.org JS shim anyway, so rely on browserid.org.

st3fan:
@gerv Yeah we can probably detect that.

benadida:
@gerv @st3fan so one idea I've had is that I think we can detect when window.open fails and have our JS shim do a redirect flow based on the _same_ API in that case. We need to prototype this, but what do you think?

(note that this would also help things work in Chrome for iOS, which has a funky window.open due to iOS restrictions.)

shane-tomlinson:
Closing as same fix as #2034

eevee:
I don't think #2034 has quite addressed the same problem.  I love the idea of browserid but am absolutely unwilling to make an otherwise static-ish website require client scripting just to log in.

Would it be possible to allow a site to implement its own discovery-redirect-springboarding, doing the usual JS work on the server side without needing to rely on browserid.org?

shane-tomlinson:
@eevee - could you explain your idea and the reasons why you'd like to do the scripting server side in more detail?

kayaelle:
[...]

PR #4021, opened 2013-11-13
declaritive api

seanmonstar:
**WIP**. Currently for discussion and code review purposes.

Still missing:
- [ ] POST from our /verify_email endpoint back to RP returnTO

### DIFF
--- a/example/rp/declaritive.ejs
+++ b/example/rp/declaritive.ejs
@@ -1,0 +1,213 @@
+<!DOCTYPE html>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<html>
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="initial-scale=1.0; width=device-width;">
+<title>
+Persona Relying Party
+</title>
+<style type="text/css">
+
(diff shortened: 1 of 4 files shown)


seanmonstar:
linking to #1946

eevee:
whoa _awesome_.  thanks so much for doing this!  glad to see it's being well-received so far too  :)

fmarier:
There's a lot of good stuff in here that has nothing to do with the declarative API (cleanups, jshints, etc.). Maybe we could split these things into a separate pull request and merge that one first?

seanmonstar:
yea perhaps. i was just working in include.js, and getting annoyed how jshint would warn me on every save, so I fixed all the warnings :D

eevee:
i'm pretty close to needing an auth system for a project; any indication whether this is likely to be adopted?  there was no opposition in the thread, at least  :)

callahad:
Persona is in maintenance mode for the moment, as the Cloud Services group at Mozilla is heads-down on Firefox Accounts and a revision to Firefox Sync. I'd suggest going ahead with one of our existing APIs instead.

jaredhirsch:
@seanmonstar it seems like the include.js portions of this have been pulled into a separate PR. Would you mind splitting that stuff out from this one?

Also, I hate to be that guy, but the spelling is **declarAtive**, would you mind fixing that throughout?

:beers:

callahad:
I'm going to close this as wontfix for now. Even though I like the design, I don't think merging in a completely new API is in our best interest right now. :(

PR #4024, opened 2013-11-14
use hostname from auth_url in 3rd party cookie message

### LINKED ISSUE #4022: 3rd party cookies disabled message doesn't show delegated domain
mozilla.com delegates to login.mozilla.org, but if a user doesn't accept 3rd party cookies, they are told to whitelist mozilla.com.

seanmonstar:
fixes #4022

### DIFF
--- a/resources/static/dialog/js/misc/state.js
+++ b/resources/static/dialog/js/misc/state.js
@@ -312,8 +312,13 @@ BrowserID.State = (function() {
         add: !!self.addPrimaryUser,
         email: self.email,
         rpInfo: self.rpInfo,
-        idpName: info.idpName || URLParse(info.auth_url).host
+        idpName: info.idpName,
+        idpHost: URLParse(info.auth_url).host
       });
+
+      if (!info.idpName) {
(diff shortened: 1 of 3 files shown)


resources/static/dialog/js/misc/state.js
```diff
@@ -312,9 +312,14 @@ BrowserID.State = (function() {
         add: !!self.addPrimaryUser,
         email: self.email,
         rpInfo: self.rpInfo,
-        idpName: info.idpName || URLParse(info.auth_url).host
+        idpName: info.idpName,
+        idpHost: URLParse(info.auth_url).host
       });
 
+      if (info.idpName) {
```
fmarier: Are you sure you don't mean the opposite here? `if (!info.idpName) {`

seanmonstar: yea, you're right. woops.

seanmonstar:
ok, reversed that if clause.

fmarier:
Looks good!

PR #4045, opened 2013-12-09
winston out, intel in

seanmonstar:
So, this replaces our usage of winston with [intel](http://seanmonstar.github.io/intel). Why?
- We get names for our loggers, so that each message has more context as to where it's coming from.
- Much of the logging moves to configuration, making it easier to test different environments.
- My [super-basic-dont-rely-on](https://github.com/seanmonstar/intel/blob/master/benchmark/logging.js) benchmarks indicate 2x faster than winston.
- Uses `WritableStream`'s internal buffering, instead of drain events: https://github.com/seanmonstar/intel/blob/master/lib/handlers/stream.js#L22
- Because :heart:


shane-tomlinson:
@seanmonstar - travis keeps failing. :(

seanmonstar:
I restarted the failed jobs. Looks green now.

shane-tomlinson:
Can you update the `/bin` directory?

```
[app@ip-10-244-99-124 bin]$ ./load_gen
UNCLE: Error: Cannot find module 'winston'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/home/app/code/bin/load_gen:14:17)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
[app@ip-10-244-99-124 bin]$
```

lib/logging/cef_logger.js
```diff
@@ -38,6 +38,14 @@ var Logger = function(options) {
  *              transactions.
  */
 
+const SEVERITY_MAP = {
+  // cef => intel
+  info: 'info',
```
shane-tomlinson, line 43: I did this with winston at one point too, the reason I didn't change CEF logging over is because of a comment somewhere that said "we want to see CEF logs distinct from normal logs". I like this way more.

lib/keysigner/ca.js
```diff
@@ -7,7 +7,7 @@
 var jwcrypto = require('jwcrypto'),
     cert = jwcrypto.cert,
     secrets = require('../secrets.js'),
-    logger = require('../logging/logging.js').logger;
+    logger = require('../logging/logging.js').getLogger('bid.keysigner.certify');
```
shane-tomlinson: Does it make sense to keep the namespaces the same as the filename so they can easily be found without resorting to grep?

lib/logging/filters/metrics.js
```diff
@@ -22,15 +22,14 @@
  * idp.create_new_user
 */
 
+const util = require('util');
```
shane-tomlinson: Is `util` used?

lib/logging/filters/metrics.js
```diff
@@ -22,15 +22,14 @@
...
-const coarse = require('../../../coarse_user_agent_parser');
-
-var MessageMatches = {
-  'signin': true,
-  'verify': true
-};
+const coarse = require('../../coarse_user_agent_parser');
 
 var RegExpMatches = [
+  /^signin\b/,
```
shane-tomlinson, line 30: I dig it, any perf regressions?

lib/logging/filters/statsd-increment.js
```diff
@@ -2,23 +2,17 @@
...
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const config = require('../../../configuration');
+const config = require('../../configuration');
 const PREFIX = "browserid." + config.get('process_type') + ".";
 
 
-var IncrementMessageMatches = {
-  "assertion_failure": true,
-  "uncaught_exception": true
```
shane-tomlinson, line 11: should `uncaught_exceptions` still increment statsd?

lib/logging/filters/statsd-increment.js
```diff
@@ -27,7 +21,7 @@ exports.test = function(msg) {
 };
 
 exports.toType = function(msg) {
-  return PREFIX + msg;
+  return PREFIX + String(msg).split(' ')[0];
```
shane-tomlinson: The split is duplicated elsewhere, is this something that could be put into intel?

lib/logging/handlers/kpi.js
```diff
@@ -0,0 +1,89 @@
...
+  intel.Handler.call(this, options);
+}
+util.inherits(MetricsKpiggybankHandler, intel.Handler);
+
+MetricsKpiggybankHandler.BATCH_SIZE = config.get('kpi.metrics_batch_size');
+
+var PiggyProto = MetricsKpiggybankHandler.prototype;
+
+PiggyProto.emit = function(record, callback) {
+  // TODO config?
```
shane-tomlinson: TODO!

lib/logging/handlers/statsd.js
```diff
@@ -0,0 +1,51 @@
...
+}
+util.inherits(StatsdHandler, intel.Handler);
+
+StatsdHandler.prototype.emit = function(record, callback) {
+  if ( ! this.statsd) return callback();
+  var meta = record.args[ record.args.length - 1 ];
+
+  if (increment_filter.test(record.message)) {
+    this.statsd.increment(increment_filter.toType(record.message), meta);
+  }
```
shane-tomlinson: nit, formatting.

lib/logging/logging.js
```diff
@@ -21,31 +21,46 @@
...
+const config = require('../configuration');
 
-const LogFileTransport = require('./transports/log-file');
-const StatsdTransport = require('./transports/statsd');
-const MetricsFileTransport = require('./transports/metrics-file');
-const MetricsKpiggybankTransport = require('./transports/metrics-kpi');
+module.exports = intel;
+// go through the configuration and determine log location
+// for now we only log to one place
+// FIXME: separate logs depending on purpose?
```
shane-tomlinson: FIXME!

lib/logging/handlers/kpi.js
```diff
@@ -0,0 +1,89 @@
...
+
+PiggyProto.isQueueFull = function isQueueFull() {
+  return this.queue.length >= MetricsKpiggybankHandler.BATCH_SIZE;
+};
+
+PiggyProto.getQueue = function getQueue() {
+  return this.queue;
+};
+
+PiggyProto.flush = function flush() {
```
shane-tomlinson: I like the name change.

lib/logging/handlers/kpi.js
```diff
@@ -0,0 +1,89 @@
...
+function MetricsKpiggybankHandler(options) {
+  options = options || {};
+  this.queue = [];
+  intel.Handler.call(this, options);
+}
+util.inherits(MetricsKpiggybankHandler, intel.Handler);
+
+MetricsKpiggybankHandler.BATCH_SIZE = config.get('kpi.metrics_batch_size');
+
+var PiggyProto = MetricsKpiggybankHandler.prototype;
```
shane-tomlinson: Something about PiggyProto makes this harder for me to follow. Maybe an `_.extend(MetricsPiggybankHandler.prototype, {...`?

lib/wsapi/interaction_data.js
```diff
@@ -38,7 +37,7 @@ exports.process = function(req, res) {
       });
     }
 
-    kpiTransport.push(kpi_json);
+    logger.info(kpi_json);
```
shane-tomlinson, line 40: Will this show all of the KPI data on the console and in the file logs? I made the direct call to kpiTransport.push to avoid that.

tests/kpi-test.js
```diff
@@ -11,8 +11,10 @@ const wsapi = require('./lib/wsapi');
 const config = require('../lib/configuration');
 const kpi_data = require('../lib/kpi_data');
 const HttpMock = require('./lib/http-mock');
-const logger = require('../lib/logging/logging').logger;
-const KpiTransport = require('../lib/logging/transports/metrics-kpi');
+const KpiHandler = require('../lib/logging/handlers/kpi');
+const intel = require('../lib/logging/logging');
```
shane-tomlinson, line 15: Do you need both intel and logger or do you just need the logger?

tests/kpi-test.js
```diff
@@ -155,13 +155,24 @@ suite.addBatch({
       ];
       config.set('kpi.send_metrics', true);
 
-      var kpiTransport = new KpiTransport();
+      var bid = intel.getLogger('bid');
+      var handler;
+      for (var i = 0; i < bid._handlers.length; i++) {
```
shane-tomlinson: Can you use Array.filter here? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

shane-tomlinson: Or Array.find? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

tests/metrics-header-test.js
```diff
@@ -92,12 +94,23 @@ suite.addBatch({
 // Listen for actual messages that are sent to the KPI transport.
 // reset the transport queue between each test run to ensure we only get the
 // messages we care about.
-var kpiTransport = KpiTransport.getInstance();
+var kpiHandler;
+var bid = intel.getLogger('bid');
+for (var i = 0; i < bid._handlers.length; i++) {
```
shane-tomlinson: Same .find

tests/metrics-header-test.js
```diff
@@ -119,7 +132,7 @@ suite.addBatch({
       config.set('kpi.send_metrics', true);
 
       sendRequestToMetricsMiddleware('/sign_in', 'https://123done.org');
-      return kpiTransport.getItem('signin');
+      return kpiHandler.getQueue()[0][0];
```
shane-tomlinson, line 132: so confused!

tests/statsd-test.js
```diff
@@ -26,7 +28,6 @@ var IncrementMessagesToTest = {
   "wsapi_code_mismatch.wsapi_url":
       "browserid.vows.wsapi_code_mismatch.wsapi_url",
   "wsapi.wsapi_url": "browserid.vows.wsapi.wsapi_url",
-  "uncaught_exception": "browserid.vows.uncaught_exception"
```
shane-tomlinson, line 29: :( Does anything happen with the uncaught exceptions?

tests/statsd-test.js
```diff
@@ -35,16 +36,15 @@ for (var logMessage in IncrementMessagesToTest) {
...
+        this.statsdTransport.handle(logger.makeRecord(logger._name, logger.INFO, logMessage, [logMessage, {
           field: "value"
-        }, this.callback);
+        }])).done(this.callback);
       },
-      "succeeds": function(err, success) {
-        assert.isNull(err);
-        assert.equal(true, success);
+      "succeeds": function(err) {
+        assert.ifError(err);
```
shane-tomlinson, line 51: I did not know this was an assertion!

shane-tomlinson:
@seanmonstar - intel is a nice lib! All of my concerns are inline in comments already, I have not yet load tested this to see how it performs. I am seeing a substantial number of Selenium failures and am unsure whether they are caused by this PR or by something else.

shane-tomlinson:
When running the selenium tests locally, this is the error I see:

```
Caused by: org.openqa.selenium.remote.ErrorHandler$UnknownServerException: Unable to locate element: {"method":"css selector","selector":"#login"}
Build info: version: '2.26.0', revision: '18041', time: '2012-11-01 19:33:38'
System info: os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.9', java.version: '1.7.0_40'
Driver info: driver.version: unknown
    at <anonymous class>.<anonymous method>(file:///var/folders/x6/zd3bzsyd08v932ymg7t283rc0000gn/T/anonymous6389230538218104218webdriver-profile/extensions/fxdriver@googlecode.com/components/driver_component.js:8319)
    at <anonymous class>.<anonymous method>(file:///var/folders/x6/zd3bzsyd08v932ymg7t283rc0000gn/T/anonymous6389230538218104218webdriver-profile/extensions/fxdriver@googlecode.com/components/driver_component.js:8328)
    at <anonymous class>.<anonymous method>(file:///var/folders/x6/zd3bzsyd08v932ymg7t283rc0000gn/T/anonymous6389230538218104218webdriver-profile/extensions/fxdriver@googlecode.com/components/command_processor.js:10337)
    at <anonymous class>.<anonymous method>(file:///var/folders/x6/zd3bzsyd08v932ymg7t283rc0000gn/T/anonymous6389230538218104218webdriver-profile/extensions/fxdriver@googlecode.com/components/command_processor.js:10342)
    at <anonymous class>.<anonymous method>(file:///var/folders/x6/zd3bzsyd08v932ymg7t283rc0000gn/T/anonymous6389230538218104218webdriver-profile/extensions/fxdriver@googlecode.com/components/command_processor.js:10282)
15:22:59.550 WARN - Exception: Unable to locate element: {"method":"css selector","selector":"#login"}
Command duration or timeout: 32 milliseconds
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html
Build info: version: '2.26.0', revision: '18041', time: '2012-11-01 19:33:38'
System info: os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.9', java.version: '1.7.0_40'
Driver info: driver.version: EventFiringWebDriver
Session ID: f93008e5-18ca-3249-91a8-d14810b88532
15:23:00.113 INFO - Executing: [find element: By.selector: #login] at URL: /session/1386948151112/element)
15:23:00.644 WARN - Exception thrown
```

jrgm:
@shane-tomlinson actually, the ssl cert for testidp.org has expired, so a lot of selenium tests _are_ failing at the moment.

seanmonstar:
@shane-tomlinson you rock!

lib/logging/cef_logger.js
seanmonstar, line 43: yea, this still has cef being used explicitly, but everything sent to cef is also passed back to us as a normal log.

lib/logging/filters/metrics.js
seanmonstar, line 30: maybe, but with everything else happening it's hard to tell. I'd imagine RegExp is slower than String Equality, but..

lib/logging/filters/statsd-increment.js
seanmonstar: I had a brain fart. intel concats all arguments to `logger.info(...args)` into a single `record.message`, but they're left individually on `record.args`. So instead of splitting on `record.message`, I'll pass `record.args[0]` from the statsd handler.

lib/logging/handlers/kpi.js
seanmonstar: Oh right. I had a question, since logging is now setup from the `logging` config option, if we needed a separate option to send metrics. It seemed like if we didn't want to send metrics, we shouldn't include this handler.

We could either get rid of `kpi.send_metrics`, or I could check `kpi.send_metrics` in `lib/configuration` and if true/false, make sure we are/arent using this handler.

lib/logging/logging.js
seanmonstar: reading the code below, it seems this is fixed. It just felt a little janky. But it works as it did before, so I'll kill the comment.

lib/wsapi/interaction_data.js
seanmonstar, line 40: Oh, I figured we'd want to see it in the logs also. Any reason not to?

lib/logging/handlers/kpi.js
seanmonstar: hokay

tests/kpi-test.js
seanmonstar: `Array#find` is in ES6, so we don't have it in v8/node yet. I could use filter, but I didn't want to keep looping after I found the handler, hence the `break`.

tests/metrics-header-test.js
seanmonstar: Same, can't use `find`.

tests/metrics-header-test.js
seanmonstar, line 132: It's because the queue is an array of (entry, callback) tuples. So, `queue[0] = [entry, callback]`, so to get the entry, its `queue[0][0]`. I'll added comments to the `getQueue` method in the KpiHandler class saying as much.

seanmonstar:
@shane-tomlinson I rebased in fixes for many of the comments. Some I left questions back for you.

I have some work in [intel@master](https://github.com/seanmonstar/intel/commit/2204bc78115ee0cb2a771df7475e0d696900dbdf) that will allow me to remove several of the filters from `lib/logging/filters` and have them be in the config, removing more files and long live config!

lib/wsapi/interaction_data.js
shane-tomlinson, line 40: @jrgm, @lloyd or @kparlante - any increased user data exposure if KPIs are displayed in the console & added to the log files? I am specifically worried about correlating data sets with IP addresses.

Running locally, I do not see KPI data in either the logs or console, doing some remote testing now.

tests/kpi-test.js
shane-tomlinson: Good point about ES6 - what about http://documentcloud.github.io/underscore/#find

shane-tomlinson:
@seanmonstar - can you add this diff?

``` diff
diff --git a/lib/logging/filters/statsd-increment.js b/lib/logging/filters/statsd-increment.js
index 6f6fd10..cedf569 100644
--- a/lib/logging/filters/statsd-increment.js
+++ b/lib/logging/filters/statsd-increment.js
@@ -12,8 +12,6 @@ var IncrementRegExpMatches = [
   /^wsapi\./
 ];

-uncaught_exceptions
-
 exports.test = function(msg) {
   for (var i = 0, regExp; regExp = IncrementRegExpMatches[i]; ++i) {
     if (regExp.test(msg)) return true;
```

shane-tomlinson:
@seanmonstar - looking manually through the logs on a test instance, the entire configuration is printed when the processes start, meaning SMTP, database & other `secret` info is placed in the logs. Can this be filtered for production environments?

shane-tomlinson:
https://github.com/mozilla/persona/pull/4045#issuecomment-30557484 remove more files!

shane-tomlinson:
@seanmonstar - When attempting to run the load tests, I am seeing 503s when creating the initial users. This happened before, but I was able to complete the user creation ~50% of the time before, now I cannot at all. Quite possibly user error on my part.

``` bash
[app@ip-10-105-2-239 bin]$ BROWSERID_FAKE_VERIFICATION=true ./load_gen -s https://intel-shane.personatest.org
```

lib/wsapi/interaction_data.js
jrgm, line 40: > @jrgm, @lloyd or @kparlante - any increased user data exposure if KPIs are displayed in the console & added to the log files? I am specifically worried about correlating data sets with IP addresses.

I don't understand your question here. Can you rephrase?

jrgm:
> @seanmonstar - looking manually through the logs on a test instance, the entire configuration is printed when the processes start, meaning SMTP, database & other secret info is placed in the logs. Can this be filtered for production environments?

Not sure which you are asking to have turned off. Previously, it would put the config in the log on production environments, but not bother annoying the user in dev where it would have also been put in the console. I want the production dump to stay as-is.

tests/kpi-test.js
seanmonstar: Never knew about that, I'll use `_.find()`.

seanmonstar:
woops, I left in that `uncaught_exception` syntax error so I'd notice before commiting. Didn't do much good.

seanmonstar:
@shane-tomlinson I had goofed on the `uncaught_exception` stuff. I had pulled it out, and was going to log ALL logged exceptions, but realized that wouldn't tell us which ones were uncaught. So, I've split them into separate statsd counters, and all should be right in the world.

lib/logging/cef_logger.js
shane-tomlinson, line 43: Cool, I like it.

shane-tomlinson:
Thanks for the feedback about the production dump @jrgm.

shane-tomlinson:
Thanks for the changes @seanmonstar!

From a dev standpoint, I am good with this PR as it is. Doing local load testing, I am seeing a slight performance regression from the current dev. Method:

3 windows & a stopwatch needed.

Window 1:
`sudo htop`, then sort by `TIME+`

Window 2:

```
BROWSERID_FAKE_VERIFICATION=1 NODE_ENV=test_mysql npm start
```

Stopwatch:
Set to count down 1:30.

Window 3:

```
bin/load_gen -s http://127.0.0.1:10002 -m 250000
```

Start the stopwatch and load_gen at the same time.

In the window with htop, keep track of the TIME+ of the router.

Stop load_gen when the stopwatch completes at 1:30.

With the current dev, the router uses ~0:00.17
With intel, the router uses ~0:00.20

@jrgm - you have far more fu at load testing than I, could you give this a whirl and see what you think?

shane-tomlinson:
@seanmonstar - I am seeing failures when running the backend tests locally:

```
tests/statsd-test.js
····✗·✗·✗·✗··················


    assertion_failure - increments
      ✗ and reports to statsd
        »
        actual expected

        browserid.vows.assertion_failure { field: 'value' }
         // statsd-test.js:52

    wsapi_code_mismatch.wsapi_url - increments
      ✗ and reports to statsd
        »
        actual expected

        browserid.vows.wsapi_code_mismatch.wsapi_url { field: 'value' }
         // statsd-test.js:52

    wsapi.wsapi_url - increments
      ✗ and reports to statsd
        »
        actual expected

        browserid.vows.wsapi.wsapi_url { field: 'value' }
         // statsd-test.js:52

    uncaught_exception - increments
      ✗ and reports to statsd
        »
        actual expected

        browserid.vows.uncaught_exception { field: 'value' }
         // statsd-test.js:52
  ✗ Broken » 25 honored ∙ 4 broken (3.650s)
  %                                                                                 [~PERSONA_ROOT]
```

seanmonstar:
@callahad fixed @shane-tomlinson latest comment.

only other thing i'd point out, is that the json is likely a little different than what winston output. is that terrible?

jaredhirsch:
@seanmonstar this was the thing that killed the 10.23 release, right? Perf regressions with `toobusy` and logging? Looking at the code, doubt I'll get the review done today (been a while...)

Still seeing unit test errors locally for me, might be my setup?

```

  tests/email-throttling-test.js
·····dbwriter (28355): ERROR bid.cef: Throttling email staging request { signature: 'EMAIL_LOAD',
········  

  tests/stalled-mysql-test.js
·········browserid (27970): ERROR bid.cef: emailKnown failed { signature: 'DB_FAILURE',
·······························  
```

jaredhirsch:
@seanmonstar ah yup, looks like travis is fine. mind rebasing?

seanmonstar:
rebased!

lib/primary.js
```diff
@@ -320,7 +320,7 @@ exports.getPublicKey = function(domain, cb) {
 // Is issuingDomain allowed to issue certifications for emails from
 // emailDomain.
 exports.delegatesAuthority = function (emailDomain, issuingDomain, cb) {
-  exports.checkSupport(emailDomain, function(err, r) {
+  exports.checkSupport(emailDomain, function onSupportChecked(err, r) {
```
jaredhirsch, line 323: yay function names :beers:

lib/logging/filters/metrics.js
jaredhirsch, line 30: This is an interesting one. I know that modern browsers have separate JITs for regexps, so regexps should be blazing fast in general. However, there are loads of string tricks in modern browsers too--I don't know of any related to string _comparison_, but that doesn't mean anything.

I also don't know if the logger will run through the regexp often enough to meaningfully warm the regexp JIT.

I guess we'll just see if the whole thing is performant enough, and get into details later, if we need to.

package.json
```diff
@@ -39,6 +39,7 @@
     "gobbledygook": "0.0.3",
     "hood": "0.2.1",
     "i18n-abide": "0.0.14",
+    "intel": "0.4.0",
```
jaredhirsch, line 42: still the version we'd want to use?

jaredhirsch:
@seanmonstar Do you know anything about toobusy load testing for this patch? Been a while, I could use some help getting going. Maybe we can fill out the infamous [empty section](https://github.com/mozilla/persona/blob/dev/docs/LOAD_GENERATION.md#usage) of the load_gen documentation that explains how to use the tool. paging @jrgm and/or @jbonacci :-)

Issue #2181, opened 2012-07-30
add a /robots.txt file 


jrgm:
From a chat in IRC with @floatingatoll, but it might be useful to add a /robots.txt file and enumerate just those things that are useful to users in finding us. 

For example, search google with site:login.persona.org and site:login.anosrep.org
1. both sites are indexed about the same, but the general user should never care that staging login.anosrep.org exists so ask to not be indexed on stage and dev
2. the search returns /unsupported_dialog, /verify_email_address, /sign_in, /add_email_address. Not particulary useful information to the general user especially since the latter three all show an error if you follow the link from the google search.

lloyd:
good idea!  Let's do some SEO!

jaredhirsch:
I guess, arguably, this could get done by Beta 2. Leaving it open & handing to @lloyd. Reassign as you will.

jrgm:
This is still a valid issue, and not so hard to do (a view that provides different robot advice on prod vs. stage/dev/ephemeral). Good first bug.

es:
Hey @lloyd, mind if I work on the issue?

lloyd:
@EmilS All yours!

shane-tomlinson:
Done! Merged with c0b2bba5b334f73b7eb340107dad4864067e2582. Thanks @Somsubhra

PR #4049, opened 2013-12-11
Issue #2181. Added a robots.txt

Somsubhra:
Added a robots.txt based on whether it is production or development version.

### DIFF
--- a/lib/static/views.js
+++ b/lib/static/views.js
@@ -295,6 +295,16 @@ exports.setup = function(app) {
     renderCachableView(req, res, 'about.ejs', {title: _('About'), fullpage: false});
   });
 
+  app.get("/robots.txt", function(req, res) {
+    if (isProductionEnvironment()) {
+      var robots = fs.readFileSync(path.join(__dirname, "..", "..", "resources", "robots.txt")).toString();
+    } else {
+      var robots = 'User-agent: *\nDisallow: /\n';
+    }
+    res.setHeader('Content-Type', 'text/plain; charset=utf8');
+    res.send(robots);
+  });
+
   function pickLocale(supported, lang) {
     var templateLocale = 'en',
         locale = i18n.localeFrom(lang);
--- a/lib/static/views.js
+++ b/lib/static/views.js
@@ -296,13 +296,19 @@ exports.setup = function(app) {
   });
 
   app.get("/robots.txt", function(req, res) {
-    if (isProductionEnvironment()) {
-      var robots = fs.readFileSync(path.join(__dirname, "..", "..", "resources", "robots.txt")).toString();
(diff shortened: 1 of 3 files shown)


lib/static/views.js
```diff
@@ -295,6 +295,16 @@ exports.setup = function(app) {
     renderCachableView(req, res, 'about.ejs', {title: _('About'), fullpage: false});
   });
 
+  app.get("/robots.txt", function(req, res) {
+    if (isProductionEnvironment()) {
+      var robots = fs.readFileSync(path.join(__dirname, "..", "..", "resources", "robots.txt")).toString();
```
callahad: Can we do this asynchronously?

lib/static/views.js
```diff
@@ -295,6 +295,22 @@ exports.setup = function(app) {
     renderCachableView(req, res, 'about.ejs', {title: _('About'), fullpage: false});
   });
 
+  app.get("/robots.txt", function(req, res) {
+    cacheAllTheThings(res);
+    res.setHeader('Content-Type', 'text/plain; charset=utf8');
+    if (isProductionEnvironment()) {
+      fs.readFile(path.join(__dirname, "..", "..", "resources", "robots.txt"), function(err, data) {
```
shane-tomlinson, line 302: We should be able to cache the robots.txt file, or does cacheAllTheThings take care of this?

resources/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+Disallow: /sign_in
+Disallow: /unsupported_dialog
+Disallow: /verify_email_address
+Disallow: /add_email_address
```
shane-tomlinson, line 5: Several endpoints should be added here:
- `/cookies_disabled`
- `/reset_password`
- `/confirm`
- `/complete_transition`
- `/auth`
- `/unsupported_dialog_without_watch`
- `/relay`
- `/provision`
- `/build`
- `/production`

lib/static/views.js
shane-tomlinson, line 302: granted, requesting robots.txt is probably not requested very frequently.

shane-tomlinson:
Thanks for this PR @Somsubhra! Can you also add a test in `/tests/page-requests-tests.js` around line 63 where we test whether valid requests are responded to?

lib/static/views.js
Somsubhra, line 302: cacheAllTheThings should probably take care of caching.

Somsubhra:
Sure, I'll add the tests. I'll follow up with a patch adding more entries to robots.txt and a test

shane-tomlinson:
Manually merging, thanks @Somsubhra!

PR #4057, opened 2013-12-21
Add schema to email for one-click sign-in action in the inbox

ryanseys:
This is very much an MVP. I would like feedback on labels strings ("Complete sign-in").

Is this also the correct place in code to put such a change?

### DIFF
--- a/resources/email_templates/new.html.ejs
+++ b/resources/email_templates/new.html.ejs
@@ -1,5 +1,12 @@
 <% include _header.html.ejs %>
 
+  <div itemscope itemtype="http://schema.org/EmailMessage">
+    <meta itemprop="description" content="Complete your sign-in"/>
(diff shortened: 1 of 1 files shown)


resources/email_templates/new.html.ejs
```diff
@@ -1,5 +1,12 @@
 <% include _header.html.ejs %>
 
+  <div itemscope itemtype="http://schema.org/EmailMessage">
+    <meta itemprop="description" content="Complete your sign-in"/>
+    <div itemprop="action" itemscope itemtype="http://schema.org/ViewAction">
+      <link itemprop="url" href="<%= link %>"/>
+      <meta itemprop="name" content="Complete sign-in"/>
```
seanmonstar, line 7: We'd probably want to localize this string.

PR #4060, opened 2014-01-02
Prevent email clients from linkifying site domains.

daisylb:
As reported in [this mailing list thread](https://groups.google.com/d/msg/mozilla.dev.identity/bbQVtYLXhk8/WFgu14TG1YsJ) (plus my reply, which ~~has yet to be approved by mods~~ I accidentally didn't actually send to the list), mail clients like trying to be clever and linkifying things like URLs.

This _should_ prevent that, although I couldn't test it because I couldn't figure out how to get my local Persona clone to send an email and/or display it to console.

### DIFF
--- a/resources/email_templates/confirm.ejs
+++ b/resources/email_templates/confirm.ejs
@@ -1,4 +1,4 @@
-<%= format(gettext('Click this link to confirm access to %(site)s:'), { site: site })%>
+<%= format(gettext('Click this link to confirm access to %(site)s:'), { site: site.replace('.', '\uFEFF.') })%>
 <%= link %>
 
 <%= gettext('If you were NOT attempting this sign-in, please ignore this email.') %>
--- a/resources/email_templates/confirm.html.ejs
+++ b/resources/email_templates/confirm.html.ejs
@@ -2,7 +2,7 @@
 
   <tr>
     <td align="center">
-      <%- format(gettext('Click this link to confirm access to %(site)s:'), { site: '<strong>' + site + '</strong>' })%>
+      <%- format(gettext('Click this link to confirm access to %(site)s:'), { site: '<strong>' + site.replace('.', '&#xFEFF;.') + '</strong>' })%>
     </td>
   </tr>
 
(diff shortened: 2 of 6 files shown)


fmarier:
We should probably not put HTML entities in the plain text version of the emails since I suspect a number of mail clients will not parse them properly.

daisylb:
Of course, silly me. I've swapped them for raw Unicode characters instead.

jaredhirsch:
Hi @adambrenecki - thanks for submitting this PR! Sorry it got a bit stale :beers:

Do you have an example of a mail client that actually exhibits the problem that this PR aims to fix?

As far as testing the patch, we use [nodemailer](https://github.com/andris9/Nodemailer) to send mail, so you can test by adding your personal email account's smtp info to `config/local.json`. You can have a look at [lib/config.js](https://github.com/mozilla/persona/blob/dev/lib/configuration.js#L160-L176) to see what the format of the parameters should be; I think something of the form `smtp: {host: 'host', user: 'foo@bar.com', pass: 's3cret', port: 25}` should work.

callahad:
FWIW, this fix makes me nervous for some reason. Seems like it might be a bit too fragile / hacky. :(

daisylb:
@6a68 AirMail and GMail do. I've heard Apple Mail and Outlook do - I'll test both later. 

@callahad Yeah, I couldn't find a less hacky way to do it. There's no way to tell mail clients 'no, don't linkify this' (or if there is, I couldn't find it). Apple Mail has a `<meta>` tag that disables it, but the others don't.

Aside: on further inspection, it seems \u2060 is a more appropriate character to use than \uFEFF - I'll change that shortly.

jaredhirsch:
Just a quick update - we're currently focused on getting a release out the door, but I'll take a look once that's done.

callahad:
I'm going to be a jerk and close this as wontfix.

At the end of the day I'm just super uncomfortable with the risk / benefit tradeoff here. :(

I really appreciate the pull request, @adambrenecki :) Thank you.

PR #4062, opened 2014-01-09
Pull vows via https, not git protocol

callahad:
hardik in IRC had a ton of trouble installing Persona because of this. For some
reason he wasn't able to clone over git://, but https works just fine.

This should also be marginally faster since we don't have to clone an entire
repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "connect-fonts-feurasans": "0.0.3"
   },
   "devDependencies": {
-    "vows": "git://github.com/cloudhead/vows#253ca34c",
+    "vows": "https://github.com/cloudhead/vows/tarball/253ca34c",
     "awsbox": "0.6.2",
     "irc": "0.3.6",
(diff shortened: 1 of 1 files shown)


seanmonstar:
is there any reason we can't reference a version in npm?

callahad:
Trying to go with a minimal change, since we were explicitly referencing that SHA before.

I'll take a quick peek at the vows history and see where 253ca34c ended up...

seanmonstar:
since this references the same SHA, I can merge this, and we look into using an npm version after?

callahad:
NPM is currently very outdated compared to this SHA:

cloudhead/vows@253ca34c was committed 2013-08-29. It's still the current head of master.
0.7.0 is the latest in NPM, and is much older, cloudhead/vows@c683d8877da94354c4dc428e7f64995afb377441 was committed on 2012-11-16.

PR #4063, opened 2014-01-09
bumped client-sessions and hood

seanmonstar:
- client-sessions removed node-proxy dependency, added more crypto options
- hood added Report-Only ability to csp()


seanmonstar:
@callahad lets merge this, so Windows peeps don't have to compile node-proxy, yea?

PR #4068, opened 2014-01-15
check observers.logout, not onlogout

### LINKED ISSUE #4067: logging into RP on stage logs out immediately after
Goldilocks came home, breaking everything. You big jerk.

Problem:
`if (!observers.onlogout) rp_api = 'stateless';`

It's supposed to be `logout`, without the `on`, and so all requests converted to goldilocks, and so everyone's onlogout() was being called. Blocks train-2013.10.23 deployment: https://bugzilla.mozilla.org/show_bug.cgi?id=958719

seanmonstar:
fixes #4067

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -339,7 +339,7 @@
       var rp_api = api_called;
       if (rp_api === "request") {
         if (observers.ready) rp_api = "watch_with_onready";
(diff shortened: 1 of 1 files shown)


fmarier:
Code looks good. I assume you've tested it and made sure it fixes the problem we had on stage.

r+

PR #4070, opened 2014-01-15
allow loggedInUser: false

### LINKED ISSUE #4069: Incorrect check of type for loggedInUser (from GH-2647)
This code:

https://github.com/shane-tomlinson/browserid/blob/dev/resources/static/include_js/_include.js#L314-L320

is just wrong according to the original statement at https://github.com/mozilla/persona/issues/2647

```
'Valid types are: undefined, null, false, String. Anything else should result in an exception'
```

`false` is not an allowed value by that check.

You can reproduce this with the STR at https://bugzilla.mozilla.org/show_bug.cgi?id=958719#c4, and loading https://popcorn.webmaker.org/. In the error console, you will see 'Error: loggedInUser is not a valid type'.

seanmonstar:
fixes #4069

target is train-2013.10.23

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -311,6 +311,9 @@
       // back compat support for loggedInEmail
       checkRenamed(options, "loggedInEmail", "loggedInUser");
       loggedInUser = options.loggedInUser;
+      if (loggedInUser === false) {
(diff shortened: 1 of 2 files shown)


callahad:
R+

PR #4071, opened 2014-01-15
allow loggedInUser: false

### LINKED ISSUE #4069: Incorrect check of type for loggedInUser (from GH-2647)

seanmonstar:
fixes #4069

target is dev

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -311,6 +311,9 @@
       // back compat support for loggedInEmail
       checkRenamed(options, "loggedInEmail", "loggedInUser");
       loggedInUser = options.loggedInUser;
+      if (loggedInUser === false) {
+        loggedInUser = null;
+      }
       if (!isNull(loggedInUser)  &&
           !isUndefined(loggedInUser) &&
           !isString(loggedInUser))
--- a/resources/static/test/cases/include.js
+++ b/resources/static/test/cases/include.js
@@ -45,14 +45,15 @@
     });
   });
 
-  test("watch only accepts null, undefined, or a string for loggedInUser", function() {
-    // a string, null, and undefined are valid
+  test("watch only accepts null, undefined, false, or a string for loggedInUser", function() {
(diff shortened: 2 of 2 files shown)


resources/static/include_js/_include.js
```diff
@@ -311,6 +311,9 @@
       // back compat support for loggedInEmail
       checkRenamed(options, "loggedInEmail", "loggedInUser");
       loggedInUser = options.loggedInUser;
+      if (loggedInUser === false) {
+        loggedInUser = null;
```
jaredhirsch, line 315: I dunno about this fix. I think the better thing is to add an isFalse check to the conditional on the next line:

`if (!isNull && !isUndefined && !isString && !isFalse)`

callahad:
I get @6a68's concern. At the same time, tests pass, and this effectively coalesces the `false` case into the known-good `null` case. :shipit:

callahad:
(Premise: we test the living daylights out of this before going live. Also, we remove the loggedInUser stuff in a month or two once Goldilocks is known-good.)

resources/static/include_js/_include.js
seanmonstar, line 315: I did this because later, in `_open_hidden_iframe()`, this value is passed to the comm_iframe. And there, the value is compared to `null`, `undefined`, and strings. Instead of adding `false` checks everywhere, it seemed less error-prone to just convert to `false` to `null`.

PR #4081, opened 2014-01-25
include.js: all properties should be passed to navigator.id.watch() now

seanmonstar:


jaredhirsch:
@seanmonstar I am mildly terrified of all the changes I see in here, especially since we have no include tests.

Would you mind adding a description of what this patch is doing (and not doing...) :-)

resources/static/include_js/_include.js
```diff
@@ -15,9 +17,9 @@
 
     function getInternetExplorerVersion() {
       var rv = -1; // Return value assumes failure.
-      if (nav.appName == 'Microsoft Internet Explorer') {
+      if (nav.appName === 'Microsoft Internet Explorer') {
         var ua = nav.userAgent;
-        var re = new RegExp("MSIE ([0-9]{1,}[\.0-9]{0,})");
+        var re = new RegExp("MSIE ([0-9]{1,}[\\.0-9]{0,})");
```
callahad, line 22: I don't think that backslash should be escaped.

```
MSIE ([0-9]{1,}[\\.0-9]{0,})
```

Regular expression visualization

[Debuggex Demo](https://www.debuggex.com/r/V6175NjAp8-cerZu)

I'd suggest this regexp literal instead:

```
/MSIE ([0-9]+\.[0-9]+)/
```

It matches [all known UA Strings](http://useragentstring.com/pages/Internet%20Explorer/) for IE 2.0-10.6, though it omits the beta tag from versions like `5.0b1`.

resources/static/include_js/_include.js
```diff
@@ -308,6 +334,7 @@
       // NOTE: Do not modify without reading GH-2017
       observers.ready = options.onready || null;
 
+      checkDeprecated(options, 'loggedInUser');
```
callahad, line 337: Should we be deprecating this, here? `loggedInUser` shouldn't really be deprecated independent of the _entire_ Observer API.

callahad:
Looks good to me, modulo the regex and deprecation warnings for Observer. Seems like we should keep Observer un-deprecated until we can get Goldilocks into production and make sure it works.

resources/static/include_js/_include.js
seanmonstar, line 22: In this case, since the backslash is in a string, it's escaping the period, which does nothing. So it becomes in regex `/[.0-9]/` which matches anything. In order to keep the period escaped in the regex, you have to escape the backslash in a string. JSHint noticed this error.

However, there's no reason this regex needs to be built with the constructor. It's not building the string from variables. I didn't pay attention when fixing it, I was just cleaning up jshint warnings.

resources/static/include_js/_include.js
seanmonstar, line 337: that's true. We should also checkDeprecated `onlogout`, `onmatch`, `onready`.

resources/static/include_js/_include.js
callahad, line 22: wat? js, why you so crazy?

shane-tomlinson, line 22: @callahad - only because it's a string used to create a regexp and not a regexp itself.

jaredhirsch:
@seanmonstar mind rebasing?

callahad:
I believe we can close this in favor of my patches on the meta Goldilocks
bug. Also, in light if recent re-examining of how Goldilocks needs to work,
I don't think this patch does what we want.

The jshint cleanups are valuable, and should be forward ported once we
close the meta bug.

jaredhirsch:
Sounds good, closing.

Issue #3270, opened 2013-04-18
The xhr_delay error message doesn't specify which WSAPI call is timing out


fmarier:
When an XHR to a WSAPI call is taking longer than 10 sec to return, we show the "We are sorry, this is taking longer than it should." error message.

It would be great to have a "show more details" link like on other error pages and to include the name of the WSAPI call it was trying to make.

In the `xhrDelay()` function of `resources/static/common/js/xhr.js`, it looks like we are passing the request information (`reqInfo`) along with that event. However, `delayStart()` (in `resources/static/common/js/modules/xhr_delay.js`) isn't passing it on when rendering the error page.

khuevu:
Hi fmarier, can I work on this issue?

seanmonstar:
@khuevu it's yours!

PR #4094, opened 2014-02-16
Fix for issue #3270: The xhr_delay error message doesn't specify which W...

khuevu:
I have added the `Get more info` link to the load.ejs template. The additional info here only contains network info while the additional info in error pages have an action info object. `renderDelay` is invoked at lower level than `renderError`, the action info is not available.


seanmonstar:
This looks awesome! Thanks for including a test as well!

PR #4100, opened 2014-03-03
set secureProxy=true when running behind an ssl proxy

jrgm:
I'm not really thrilled with this change (1. should I add a config to say 'over ssl but using ssl termination out of process, and 2. client-sessions should be easier to use/better documented for this very common case).

### DIFF
--- a/lib/wsapi.js
+++ b/lib/wsapi.js
@@ -339,17 +339,21 @@ exports.setup = function(options, app) {
 
   app.use(WSAPI_PREFIX, express.cookieParser());
   app.use(WSAPI_PREFIX, express.bodyParser());
+
+  var cookieOpts = {
+    path: '/wsapi',
+    httpOnly: true,
+    maxAge: config.get('authentication_duration_ms')
+  };
+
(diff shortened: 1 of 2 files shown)


jrgm:
I guess I didn't explain the "why" of this change: basically the upgrades in client-session and its dependent cookies leads to what was pointed out in https://github.com/mozilla/node-client-sessions/issues/30#issuecomment-36080390. I'd also noticed a couple of weeks ago that awsbox deployments were not setting cookies for /wsapi (and probably a production deploy would have been broken too).

jaredhirsch:
Thanks @jrgm! Any thoughts on how we could test this bug fix? (I'm happy to write the tests if you're not up for it, I could use a guide though--been a while.)

jrgm:
@6a68 - added a test for this. Please have a look when you get spare time.

jaredhirsch:
:beers:

jaredhirsch:
@jrgm Works for me. About your comments in the pull request, if you think a config would be good, go ahead and file a followup bug.

PR #4108, opened 2014-03-23
Fix #3866 - Use RP branding on login popup in main site.

### LINKED ISSUE #3866: Dialog does not use RP branding when signing in to the main site.
We did all that work, we should use the features!

screen shot 2013-09-06 at 17 16 35

Osmose:
Only includes the `siteLogo` if the page is being served over SSL; local development sign-in fails if you don't do this since it's not HTTPS. I considered using a data URI, but couldn't figure out a way to include it that would stay updated with the image file and wouldn't require something silly that's currently beyond my skills, like a (whatever the equivalent to context processors from Django are in express-land) that adds the data URI to the template context for embedding in `layout.ejs`.

Here's a preview of what it looks like for me (using a data URI temporarily to make the image display):
screen shot 2014-03-23 at 2 43 16 pm


seanmonstar:
\o/

I've kicked Travis, once it's green I'll merge.

PR #4120, opened 2014-04-10
update rpm spec for nodejs 0.10.x

jrgm:
- require 0.10
- get rid of obsolete deps for npm, openssl-devel, etc.,
- subversion for L10N work
- make gmp mandatory for _this_ rpm

@gene1wood, r?

### DIFF
--- a/scripts/browserid.spec
+++ b/scripts/browserid.spec
@@ -11,8 +11,8 @@ URL:           https://github.com/mozilla/browserid
 Source0:       %{name}.tar.gz
 BuildRoot:     %{_tmppath}/%{name}-%{version}-%{release}-root
 AutoReqProv:   no
-Requires:      openssl, nodejs >= 0.8.17
-BuildRequires: gcc-c++, git, jre, make, npm, openssl-devel, expat-devel, nodejs >= 0.8.17
+Requires:      nodejs-svcops >= 0.10.26, gmp
+BuildRequires: gcc-c++, git, subversion, make, gmp, gmp-devel, nodejs-svcops >= 0.10.26
 
 %description
 persona server & web home for persona.org


gene1wood:
We'll want to update these as well
- https://github.com/mozilla/identity-ops/blob/master/chef/cookbooks/persona-webhead/attributes/default.rb#L2
- https://github.com/mozilla/identity-ops/blob/master/chef/cookbooks/persona-keysign/attributes/default.rb#L2
- https://github.com/mozilla/identity-ops/blob/master/chef/cookbooks/persona-dbwrite/attributes/default.rb#L2

And we'll need to build a version of nodejs at the version you're looking at and upload it to 

https://s3.amazonaws.com/mozilla-identity-us-standard/rpms/

jrgm:
@gene1wood thanks for pointing out the chef changes needed. I'll send you a PR.

re: nodejs rpm: Fxa is already using nodejs-svcops at  0.10.26-1.el6 in the mozilla-services-aws yum repo

(Although, there will be a 0.10.27 real soon now which will update openssl to 1.0.1g, and we should consider whether we need/want to build that version as our standard nodejs 0.10.x version).

jrgm:
@gene1wood or @6a68 - look good? We'll need to also merge this back to train-2014.04.16

gene1wood:
@jrgm we'll probably also want to update the version of `express` and `connect` in `package.json` since building with nodejs 0.8.17 gives these warnings

```
npm WARN engine express@2.5.0: wanted: {"node":">= 0.4.1 < 0.7.0"} (current: {"node":"0.8.17","npm":"1.2.0"})
npm WARN engine connect@1.7.2: wanted: {"node":">= 0.4.1 < 0.7.0"} (current: {"node":"0.8.17","npm":"1.2.0"})
```

And maybe others (that we can figure out when we try to build on 0.10.26 for the first time.

Issue #4114, opened 2014-03-31
tests failing on IE{8,9,10,11}


jrgm:
After the fix in https://github.com/mozilla/persona/pull/4113 lands, there are still some broken tests, but only for IE versions.

There failures are the below (but they are a bit flaky and will sometimes pass).

`git bisect` points at https://github.com/mozilla/persona/commit/d181cef71e29669dfc450599fc0ed5581a80c0db as the "cause". Now that's seemingly unrelated to the failures below, but I think there is a bit of cleanup after each test needed. If I remove the calls to watch in that test suite on current `dev` branch, all the tests pass on IE versions.

```
1.common/js/user: tokenInfo with a bad token without returnTo info, no returnTo in results (1, 1, 2)Rerun62 ms

1.returnTo is undefinedExpected:  "undefined"

Result:  "string"

Diff:  "undefined" "string" 

Source:     
   at Anonymous function (https://jrgmb.personatest.org/test/cases/common/js/user.js:641:7)
   at complete (https://jrgmb.personatest.org/common/js/helpers.js:63:7)
   at Anonymous function (https://jrgmb.personatest.org/common/js/user.js:675:9)
   at complete (https://jrgmb.personatest.org/common/js/helpers.js:63:7)
   at success (https://jrgmb.personatest.org/common/js/network.js:257:11)
   at complete (https://jrgmb.personatest.org/common/js/helpers.js:63:7)
   at triggerHandlers (https://jrgmb.personatest.org/common/js/modules/xhr.js:280:7)


2. common/js/user: authenticateWithAssertion with valid assertion (1, 2, 3)Rerun97 ms

1.failedExpected:  false

Result:  true

Diff:  false true 

Source:     at Anonymous function (https://jrgmb.personatest.org/test/cases/common/js/user.js:718:9)
   at complete (https://jrgmb.personatest.org/common/js/helpers.js:63:7)
   at Anonymous function (https://jrgmb.personatest.org/common/js/user.js:470:9)
   at Anonymous function (https://jrgmb.personatest.org/common/js/user.js:341:7)
   at complete (https://jrgmb.personatest.org/common/js/helpers.js:63:7)
   at withContext (https://jrgmb.personatest.org/common/js/network.js:36:18)
   at withContext (https://jrgmb.personatest.org/common/js/user.js:339:5)


3. common/js/user: getAssertion with unconfirmed computer status - cert is removed after getAssertion (1, 0, 1)Rerun111 ms

[...]

PR #4121, opened 2014-04-12
Backport 4114 fix to 2013.10.23 train.

jaredhirsch:
Note: this is just backporting the fix that was landed in dev already. See #4114 for more.


jaredhirsch:
Merging my own. Travis is green and the diff looks sane.

PR #4129, opened 2014-05-15
Add a Do Not Track policy

fmarier:
This will prevent Persona from being blocked by privacy-protecting
add-ons such as the EFF's Privacy Badger.

https://www.eff.org/privacybadger#how_to_comply_with_dnt

The temporary policy is copied verbatim from https://www.eff.org/files/dnt-policy-preliminary.txt

https://github.com/EFForg/privacybadgerfirefox/issues/98

### DIFF
--- a/lib/static/views.js
+++ b/lib/static/views.js
@@ -302,6 +302,13 @@ exports.setup = function(app) {
     res.send(robots);
   });
 
+  var dntPolicy = fs.readFileSync(path.join(__dirname, "..", "..", "resources", "dnt-policy.txt")).toString();
+  app.get("/.well-known/dnt-policy.txt", function(req, res) {
+    cacheAllTheThings(res);
+    res.setHeader('Content-Type', 'text/plain; charset=utf8');
+    res.send(dntPolicy);
+  });
+
   function pickLocale(supported, lang) {
     var templateLocale = 'en',
         locale = i18n.localeFrom(lang);
--- a/resources/dnt-policy.txt
+++ b/resources/dnt-policy.txt
@@ -1,0 +1,15 @@
+PRELIMINARY DNT POLICY
+
+This domain interprets DNT as a request for an opt out of collection and
+retention of visitors' reading habits, which we will respect, subject to
+reasonable exceptions that respect user privacy. 
+
+This is a temporary document.  A full DNT Policy is being drafted at
+https://eff.org/dnt-policy .  When that document is finished, this domain may
+decide to adopt it.
+
+You can agree to this same policy by posting it at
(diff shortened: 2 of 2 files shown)


lib/static/views.js
```diff
@@ -302,6 +302,13 @@ exports.setup = function(app) {
     res.send(robots);
   });
 
+  var dntPolicy = fs.readFileSync(path.join(__dirname, "..", "..", "resources", "dnt-policy.txt")).toString();
+  app.get("/.well-known/dnt-policy.txt", function(req, res) {
+    cacheAllTheThings(res);
+    res.setHeader('Content-Type', 'text/plain; charset=utf8');
+    res.send(dntPolicy);
+  });
```
ozten, line 310: Looks great! Restarted Travis.

ozten:
Sounds like an npm error or we're testing for Node 0.8.x and that support has been lost?
https://gist.github.com/ozten/7df4b8e2afcdb5e29110

fmarier:
BTW, I wasn't able to test this change using awsbox. It keeps timing out on me while running `scripts/deploy.js deploy`.

This file needs to be available from both https://login.persona.org/.well-know/dnt-policy.txt and https://static.login.persona.org/.well-know/dnt-policy.txt for Persona to be allowed by Privacy Badger.

shane-tomlinson:
This is really interesting, I didn't know this was a thing!

fmarier:
@shane-tomlinson yeah, it's a fairly new extension, but I really like the approach :)

chrismore:
I think this is what is causing problems for me too. I couldn't log into an etherpad team website with persona using my ldap account. It worked for all other emails except ldap. I disabled the EFF plugin and I was then able to log in fine.

fmarier:
@chrismore from the team etherpad site, you can click on the privacy badger icon (in the URL bar) and allow `login.persona.org` and `static.login.persona.org`:

privacy_badger_options

You may also need to do that on `login.mozilla.org` too.

jaredhirsch:
@ozten yeah, 0.8 seems to be failing on travis. I've opened a bug (#4133) to remove 0.8 from the travis configs. This should be fine, since we already merged a commit that moves the server from 0.8 to 0.10.

@fmarier I have been having problems with deploying awsboxes, as well as just getting the local service to log me in (bidbundle errors in the dialog, grr). I guess we have some bitrot to clean up.

At any rate, I can verify that this patch works locally, since the local URL for webheads serves up the policy (`127.0.0.1:10002/.well-known/dnt-policy.txt`), as does the local static URL (`127.0.0.1:10010/.well-known/dnt-policy.txt`). So, merging. :beers:

PR #4132, opened 2014-05-23
Fix typo in getRPAPI; forced all RPs to Goldilocks

### LINKED ISSUE #4131: Mistakenly forcing RPs to stateless ("Goldilocks") API instead of normal Observer
We tried to deploy 90e48d0, but had to roll back due to issues with Goldilocks.

Sites like https://voo.st, https://mozillians.org, and https://air.mozilla.org would accept a login, then immediately log the user back out.

http://123done.org does _not_ exhibit this issue.

There's evidence that we were treating them as using the stateless ("Goldilocks") API before logging in: I still had a few tabs open that had initially loaded with 90e48d0 as login.persona.org. When I called .request() in one of those, after rolling back, the popup reported: "improper usage of API: Error: invalid value for rp_api: stateless"

Hypothesis: We're incorrectly identifying sites as using the stateless API.

**Mechanism of action:**

Page load / login works normally, but because we think the RP wants to be stateless, we clear the email -> RP association (from which `loggedInUser` is derived) right before login finishes and we send the assertion.

When the RP page reloads after login, we (correctly) treat it as classic Observer and check what the site says `loggedInUser` should be (`foo@example.com`) against our local email -> RP association (`undefined`).

Persona then triggers `onlogout` to rectify the mismatch between the RP and its own belief.

callahad:
fixes #4131

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -342,7 +342,7 @@
       var rp_api = api_called;
       if (rp_api === "request") {
         if (observers.ready) rp_api = "watch_with_onready";
-        else if (observers.onlogout) rp_api = "watch_without_onready";
+        else if (observers.logout) rp_api = "watch_without_onready";
         else rp_api = "stateless";
       }
 


callahad:
(This fixes the issue in dev; a sister to commit 583d13595ed037493b1179459c3c43110ed153c2 for `train-2014.05.23`)

jaredhirsch:
Giving travis a kick, not sure why half the jobs errored out.

callahad:
They were all the Node 0.8 jobs. We could also merge the stuff on
`train-2014.04.16` that switches us over to Node 0.10 (or rebase
`train-2014.05.23` onto that)... The branches are a tiny bit messy at the
moment :)

jaredhirsch:
@callahad Hunh, the patch that switches us to node 0.10 was merged into dev last month (commit 79e6339, pull request #4120). It looks like the 5-23 train branched off of the 4-16 train, which is kinda funny. Was there some reason we didn't branch off of dev? Any thoughts on tossing that train and making a new one from dev?

I'm having some problems getting Persona running locally (will open a bug momentarily), and I'm also unable to create an ephemeral instance (emailed the list about this recently), so I can't verify the change is running properly. However, half the tests _do_ pass, it's a tiny change, seems logical, and we'll verify it on stage. Merging.

PR #4136, opened 2014-06-02
use awsbox at latest revision 

jrgm:


jaredhirsch:
What if we instead rev the awsbox version number and just update package.json to use that?

jrgm:
> What if we instead rev the awsbox version number and just update package.json to use that?

Not sure if we currently have to keys to publish a new module to npm, so I just want to cut out the middleman (npm). (And actually, I meant that to be at master; updated PR).

jaredhirsch:
Well, it doesn't affect production, so I guess not pinning versions (using master vs a particular sha) is fine. Merging.

jaredhirsch:
whoops. kicked travis, will check again later

jaredhirsch:
but r+ from me, feel free to merge when it's green @jrgm

seanmonstar:
The only maintainer of [awsbox on npm](https://www.npmjs.org/package/awsbox) is @lloyd !

jrgm:
It's green.

Issue #4134, opened 2014-05-30
Meta: Goldilocks Polish


callahad:
The new "Goldilocks" API is now available in production, and all basic functionality works.

I'd like to make a few ergonomic and functional tweaks before announcing it as "final," namely:
- [x] The `experimental_emailHint` request option should be renamed / aliased to just `email`.
- [x] The `onready` callback should still be available with Goldilocks.
- [x] Visual customization options `siteName`, `siteLogo`, and `backgroundColor` should move to `.watch()` instead of `.request()`.
- [x] ToS / PP functionality should be removed from Goldilocks.

**EDIT:** ToS / PP was not removed from Goldilocks; instead, it continues to work and a deprecation warning is logged to the console.

ozten:
> ToS / PP functionality should be removed from Goldilocks.

Any documentation around this change? Would removing it break any Mozilla websites from a legal perspective? I wonder if we should measure usage in the wild before removing?

callahad:
There was some discussion on list [back in September](http://thread.gmane.org/gmane.comp.mozilla.identity.devel/6176), and Lloyd pretty much convinced me: it's not being used, and it feels out of place with the idea of focusing Persona on email verification.

There's also the broader goal of reducing complexity so that we can actually keep Persona maintained in the face of our reduced manpower.

callahad:
All PRs merged. Closing metabug.

tilgovi:
it looks like you meant to close this but didn't

callahad:
Hah! You're right. Thanks :)

PR #4138, opened 2014-06-02
Allow the use of "onready" in Goldilocks

callahad:
Related to #4134


callahad:
Gah, shouldn't try to commit while watching the WWDC keynote. Rebasing and squashing this branch...

PR #4141, opened 2014-06-02
Prefer passing display opts to watch, not request

callahad:
This patch is backwards compatible.

Console warnings are issued when display options (siteName, siteLogo,
backgroundColor) are passed to .request().

Rationale:

In the future, we hope that browsers will have native support for the BrowserID
protocol. It would be nice if those browsers could display properly themed "sign
into this site" buttons in the native browser UI.

If display options are passed to .request(), they're not known until the user
clicks "sign in" on the site itself. On the other hand, if they're passed to
.watch(), they're known by the browser as soon as the site loads, and can be
used without requiring direct user interaction with the page.

Related to #4134

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -149,6 +149,9 @@
       ready: null
     };
 
+    // dialog appearance options (siteLogo, siteName, backgroundColor)
+    var displayOpts = {};
+
     var loggedInUser;
 
     var compatMode = undefined;
@@ -321,6 +324,10 @@
         throw new Error("loggedInUser is not a valid type");
       }
 
+      // Save dialog appearance options
+      if (options.siteName) displayOpts.siteName = options.siteName;
+      if (options.siteLogo) displayOpts.siteLogo = options.siteLogo;
+      if (options.backgroundColor) displayOpts.backgroundColor = options.backgroundColor;
 
       _open_hidden_iframe();
     }
@@ -363,6 +370,15 @@
       if (options.privacyPolicy && !options.termsOfService) {
         warn("privacyPolicy ignored unless termsOfService also defined");
       }
+
+      if (options.siteLogo) warn("Please pass siteLogo to .watch() instead of .request()");
+      if (options.siteName) warn("Please pass siteName to .watch() instead of .request()");
+      if (options.backgroundColor) warn("Please pass backgroundColor to .watch() instead of .request()");
+
+      // Allow request to override display opts passed to watch, for UI testing.
+      options.siteLogo = options.siteLogo || displayOpts.siteLogo;
+      options.siteName = options.siteName || displayOpts.siteName;
+      options.backgroundColor = options.backgroundColor || displayOpts.backgroundColor;
 
       options.rp_api = getRPAPI();
       var couldDoRedirectIfNeeded = (!needsPopupFix || api_called === 'request' || api_called === 'auth');
@@ -534,9 +550,6 @@
         opts.termsOfService = passedOptions.termsOfService || undefined;
         opts.privacyURL = passedOptions.privacyURL || undefined;
         opts.tosURL = passedOptions.tosURL || undefined;
-        opts.siteName = passedOptions.siteName || undefined;
-        opts.siteLogo = passedOptions.siteLogo || undefined;
-        opts.backgroundColor = passedOptions.backgroundColor || undefined;
         opts.experimental_emailHint = passedOptions.experimental_emailHint || undefined;
         // api_called could have been set to getVerifiedEmail already
         api_called = api_called || "get";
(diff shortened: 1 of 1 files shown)


resources/static/include_js/_include.js
```diff
@@ -321,6 +324,10 @@
         throw new Error("loggedInUser is not a valid type");
       }
 
+      // Save dialog appearance options
+      if (options.siteName) displayOpts.siteName = options.siteName;
+      if (options.siteLogo) displayOpts.siteLogo = options.siteLogo;
+      if (options.backgroundColor) displayOpts.backgroundColor = options.backgroundColor;
```
callahad, line 330: Reviewer: This is inside `internalWatch`, which both `.watch()` and legacy `.get()` invoke. We simply copy the display options out into a context variable, the same way we handle copying the state callbacks into the `observers` variable. This ensures they're available to `internalRequest` later on.

resources/static/include_js/_include.js
```diff
@@ -364,6 +371,15 @@
...
 
+      if (options.siteLogo) warn("Please pass siteLogo to .watch() instead of .request()");
+      if (options.siteName) warn("Please pass siteName to .watch() instead of .request()");
+      if (options.backgroundColor) warn("Please pass backgroundColor to .watch() instead of .request()");
+
+      // Allow request to override display opts passed to watch, for UI testing.
+      options.siteLogo = options.siteLogo || displayOpts.siteLogo;
+      options.siteName = options.siteName || displayOpts.siteName;
+      options.backgroundColor = options.backgroundColor || displayOpts.backgroundColor;
+
```
callahad, line 383: Reviewer: This is inside `internalRequest`, which both `.request()` and legacy `.get()` invoke. We pull display options from the context variable that we populated above, but we allow options passed directly to `.request()` to take precedence. I'm not sure how I feel about that, but it seems potentially important for designers, since they'll want to experiment with `backgroundColor`, and `.watch()` can only be called once per page load.

By letting `.request()` override those values, you can still sit in a JS console and run `navigator.id.request({backgroundColor: '#F00'})` as you're testing stuff out.

resources/static/include_js/_include.js
```diff
@@ -556,7 +569,10 @@
               callback = null;
             }
           },
-          onlogout: function() {}
+          onlogout: function() {},
+          siteName: passedOptions.siteName,
+          siteLogo: passedOptions.siteLogo,
+          backgroundColor: passedOptions.backgroundColor
```
callahad, line 576: Reviewer: Lastly, this is inside the legacy `.get()` implementation. Instead of adding the display options to `opts` and passing it to `internalRequest`, we pass them to `internalWatch`. This avoids console warnings for users of the oh-so-deprecated yet tragically popular `.get()`.

resources/static/include_js/_include.js
jaredhirsch, line 383: This seems like an easy place to introduce bugs (say, if you pass different params to request and watch). Designers are already used to refreshing the page to pick up CSS changes. I don't think it's valuable to allow the display options to be overridden in this way.

resources/static/include_js/_include.js
jaredhirsch, line 330: You don't need the guard clauses at all, just do `displayOpts.foo = options.foo` for each of these.

(The type-related behavior in this file is ridiculous, it needs serious cleanup...)

resources/static/include_js/_include.js
jaredhirsch, line 576: I'm not totally sure I follow. Why do we not want to show console warnings for people using deprecated APIs?

resources/static/include_js/_include.js
jaredhirsch, line 383: Hmm, I think we should also update the test RP to use watch instead of request.

Best way to do this is probably to tuck the displayOpts into web storage (cookies or localstorage, whatever), then refresh the page. I'm happy to help with this if you have questions or get stuck :-)

callahad, line 383: Sounds good. Will update patch to 1) not allow overriding in `.request`, and 2) update the test RP.

Instead of web storage, I'll just have a manual button to trigger `.watch();`, which I've found useful when testing on jsfiddle.

resources/static/include_js/_include.js
callahad, line 576: We don't show warnings for folks using visual customization with `.get`, since there would be no way for them to solve it. "Pass those to .watch? But I don't have a .watch!"

Another way to think about it is that .get is deprecated, but passing those options are not separately deprecated.

Once Goldilocks is stable, we'll add full-on deprecation warnings to Get and Observer, and publish migration guides to Goldilocks

callahad:
@6a68 Sounds like you're r+, aside from: 1) updating test RPs, and 2) disallowing overriding from .request?

resources/static/include_js/_include.js
jaredhirsch, line 383: Yeah, that sounds great.

jaredhirsch:
Yup, sounds like a plan.

callahad:
Travis is timing out on connecting to MySQL, but the tests pass for me on Linux.

I've got me an r+ from @Osmose. Updating the Test RPs, then merging.

callahad:
Actually, deferring test rp updates to a second pull request. Merged this PR manually: e1e7f3062619aa870227b74148615366de4e3399

PR #4142, opened 2014-06-02
Stateless (Goldilocks) API ignores ToS / PP

callahad:
Easy to revert if we need to, but let's start by keeping the Goldilocks surface area as absolutely minimal as possible. More discussion in the issue below:

Related to #4134

### DIFF
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -360,9 +360,7 @@
 
       if (options.rp_api === "stateless") {
         if (options.termsOfService || options.privacyPolicy) {
-          warn("stateless API does not support termsOfService or privacyPolicy, ignoring");
-          delete options.termsOfService;
-          delete options.privacyPolicy;
+          warn("Deprecation notice: termsOfService and privacyPolicy options are deprecated in the stateless (\"Goldilocks\") Persona API, and will eventually be removed. More info: https://github.com/mozilla/persona/issues/4134#issuecomment-44859342");
         }
       }
 
--- a/resources/static/include_js/_include.js
+++ b/resources/static/include_js/_include.js
@@ -356,6 +356,16 @@
       checkRenamed(options, "tosURL", "termsOfService");
       checkRenamed(options, "privacyURL", "privacyPolicy");
 
+      options.rp_api = getRPAPI();
+
+      if (options.rp_api === "stateless") {
+        if (options.termsOfService || options.privacyPolicy) {
+          warn("stateless API does not support termsOfService or privacyPolicy, ignoring");
+          delete options.termsOfService;
+          delete options.privacyPolicy;
+        }
+      }
+
       if (options.termsOfService && !options.privacyPolicy) {
         warn("termsOfService ignored unless privacyPolicy also defined");
       }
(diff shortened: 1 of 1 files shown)


resources/static/include_js/_include.js
```diff
@@ -356,6 +356,16 @@
       checkRenamed(options, "tosURL", "termsOfService");
       checkRenamed(options, "privacyURL", "privacyPolicy");
 
+      options.rp_api = getRPAPI();
+
+      if (options.rp_api === "stateless") {
```
jaredhirsch, line 361: Would you mind updating the existing tests in `test/cases/dialog/js/modules/validate_rp_params.js`, and also adding some tests for this new functionality?

jaredhirsch:
I gotta be honest, I am not crazy about this change, because it doesn't simplify our code, and it makes our API situation even more complex for RPs (" we'll show your ToS/PP, _sometimes_ ").

I'd prefer either ripping ToS/PP out completely, or leaving it alone. This project has been dormant for months and had accreted a lot of product complexity, and I think we're fine to move fast and start really streamlining.

resources/static/include_js/_include.js
callahad, line 361: Happy to, but this isn't new functionality, per se. `getRPAPI` is live in production. :)

callahad:
We can't rip out ToS/PP quite yet, out of respect for existing RPs, but this establishes helps us get there in the future. This patch creates a line in the sand that ToS/PP doesn't exist in Goldilocks, but it does exist in the soon-to-be-deprecated Get and Observer APIs.

We take a negligible bump in complexity for the moment so that we can _actually_ drop ToS/PP once Goldilocks is the only supported API.

callahad:
@6a68 With updated tests, are you willing to r+ this?

From what I can tell, our options are:
1. Nuke ToS/PP entirely, surprising RPs and breaking our deprecation contract.
2. Allow ToS/PP, sticking us with this ill-fated feature for the foreseeable future.
3. Accept this patch, limiting ToS/PP to deprecated APIs.

This change is literally the addition of 5 functional lines of code; I'm not sure it adds meaningful complexity, and it buys us quite a bit of win in the future.

resources/static/include_js/_include.js
jaredhirsch, line 361: Hmm, I was thinking we could use the existing test file to ensure that ToS/PP would fail if passed when rp_api == 'stateless'. That might still be useful?

But I think what we really want is to probably split this file into the BrowserSupport piece and the "really load Persona" pieces (internalWatch, internalRequest, the public API), and test both separately, the former with UA strings from Browserscope via npm useragent package, the latter with a fake WinChan. I'll just do this in a separate bug, though :-)

jaredhirsch:
Alright, you've convinced me, let's take the gradual approach :-)

Osmose:
Talked with @callahad, we decided that removing the functionality is a bit too risky to do outright (especially because it'd block me from doing a django-browserid release since it's already on Goldilocks but still supports some Mozilla sites that rely on the arguments), so we're going even more gradual and just warning for the time being.

And given that there's no failure to test, only warnings, this looks r+ to me. :D

callahad:
Travis is timing out on connecting to MySQL, but the tests pass for me on Linux.

With @Osmose's r+, merging.

PR #4145, opened 2014-06-04
updates for awsbox changes

jrgm:
1. add missing requirements for subversion, gmp-devel (and emacs) to .awsbox.json to work with new awsbox
2. until we get new awsbox ami images built, this needs to be >= 0.10.25


callahad:
Way to sneak in emacs, there ;)

PR #4166, opened 2014-07-21
fix persona rpm spec file for train-2014.07.19 release

relud:
r? @6a68 or @callahad


jaredhirsch:
Doh! Forgot to bump the specfile, thanks!

Issue #4179, opened 2014-08-22
 NODE_ENV=test_mysql node tests/stalled-mysql-test.js failing with Travis(0.10.31) and anything(0.10.30)


jrgm:
See: https://travis-ci.org/mozilla/persona/jobs/32913897

and

[12:56:37] travis-ci: [travis-ci] mozilla/persona#4291 (dev - e1e7f30 : Dan Callahan): The build failed.
[12:56:37] travis-ci: [travis-ci] Change view : https://github.com/mozilla/persona/compare/08e221291fbe...e1e7f3062619
[12:56:37] travis-ci: [travis-ci] Build details : http://travis-ci.org/mozilla/persona/builds/32913895
[13:40:33] callahad:  jrgm: By the way, node 0.10.31 should fixes the stalled-mysql-test.js failures. https://github.com/joyent/node/commit/6f043940bdcbfb5272be8ae959cd74b9fb5cf4f8
[13:41:11] callahad:  Not sure why that test run still fails
[13:41:55] callahad:  re-runs locally
[13:42:59] jrgm:      callahad: yeah, I had tracked down that same commit. But 0.10.31 didn't seem to fix it for me on linux.
[13:44:27] jrgm:      callahad: I was just about to file an issue to track, but then it passed with 0.10.31 on osx for me
[13:47:39] callahad:  jrgm: Hrm. That whole test run passes on OS X for me, too
[13:48:03] jrgm:      callahad: ... and now it passes for me on linux. So why fail on Travis with 0.10.31? (blah, blah, clocks, blah, blah, multi-tenant, blah, blah, ubuntu, ..)
[13:48:54] callahad:  goes to build 10.31 on Debian

hemanth:
Same issue. @jbonacci

callahad:
Whoops, the mass-closing script misfired. Reopening.

abendigo:
I have found the problem.  In stalled-mysql-tests.js,
STALL_MYSQL_WHEN_PRESENT is set to point to a non-existent directory,
and so addStallDriverBatch fails to create the .stall file.  I will try to find a fix, and submit a pull request.

abendigo:
Woohoo! Bumping the version of the temp module seemed to fix it! As soon as my branches travis build passes, I will submit a Pull Request.

callahad:
Hah! You're kidding? That's awesome :)

abendigo:
Pull Request submitted: https://github.com/mozilla/persona/pull/4204

PR #4181, opened 2014-08-23
Allow SVG siteLogos to be specified as data: URIs

callahad:
The only legitimate MIME type for SVGs is "image/svg+xml", but our whitelist
contained "image/svg" instead.

As a result, data-URI encoded SVGs were not allowed as the siteLogo.

### DIFF
--- a/resources/static/test/cases/dialog/js/modules/validate_rp_params.js
+++ b/resources/static/test/cases/dialog/js/modules/validate_rp_params.js
@@ -245,6 +245,17 @@
     testExpectValidationFailure({siteLogo: URL});
   });
 
+  asyncTest("get with data:image/svg+xml;... siteLogo - allowed", function() {
+    testExpectValidationSuccess({
+        siteLogo: "data:image/svg+xml;base64,FAKEDATA"
+      },
+      {
+        siteLogo: "data:image/svg+xml;base64,FAKEDATA"
+      }
+    );
+  });
+
+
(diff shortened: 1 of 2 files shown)


callahad:
Tests pass locally. Travis is just acting up (#4179).

Manual verification also looks good.

@Osmose or @6a68, could I get a review?

Osmose:
Code looks good, r+. Logic for the change makes sense too.

If you expected me to run this code YOU'RE OUTTA YOUR MIND and should let me know to run the code.

resources/static/test/cases/dialog/js/modules/validate_rp_params.js
```diff
@@ -245,6 +245,17 @@
...
+    testExpectValidationSuccess({
+        siteLogo: "data:image/svg+xml;base64,FAKEDATA"
+      },
+      {
+        siteLogo: "data:image/svg+xml;base64,FAKEDATA"
+      }
+    );
+  });
+
+
```
Osmose, line 258: Extra space I assume?

Osmose, line 258: Also wtf is with this new github comment style it's freaking me out

callahad:
self-merging into future

callahad:
Fixed in https://github.com/mozilla/persona/commit/c80fe2fabe569321e0fac77cf6055edf81e0d615

PR #4182, opened 2014-08-23
Completely overhaul test RP. One page, all APIs.

callahad:
Still needs CSS and returnTo love, but should be much more functional than
the existing test RP.

Basically, I wanted a single page where I could test all of the APIs, unexpected interactions, etc.

This does that.


callahad:
@jrgm @Osmose @6a68 

New hotness awaits at http://127.0.0.1:10001/ -- could I get a review, please?

(No serious review / testing needed, since this is only the test RP, but I'd at least like a :thumbsup: or a :thumbsdown: before I hitting the green button)

example/rp/index.html
```diff
@@ -5,199 +5,205 @@
...
+    <a href="/?prod">Prod</a>
+    <a href="/?fxos">FxOS</a>
+  </p>
 
+  <div class=config>
+    <fieldset>
+      <legend>navigator.id.watch()</legend>
+      <ul>
+        <li>
+          <label for=watchOnlogin>onlogin</label>
```
Osmose, line 38: Ah god where are the quotes what madness have you wrought

example/rp/index.html
```diff
@@ -5,199 +5,205 @@
...
+          <input type=checkbox checked id=watchOnlogin />
```
Osmose, line 39: No quotes but self-closing tags what year is it

(In sane-er words, no need to self-close these, it's HTML5!)

Osmose:
Just idly glancing through, I'm not comfortable enough to :thumbsup: this all by my lonesome. :P

example/rp/index.html
callahad, line 39: There's a reason I'm a backend dev :) This is pretty hacky at the moment;
cleaning up and styling separately.

And yes, siteLogo is indeed intended to require HTTPS.

seanmonstar, line 39: I view that as a mistake in the html5 spec. Self-closing is preferred!!!11!

seanmonstar:
Eh, I pressed it. Looked good to me...

example/rp/index.html
```diff
@@ -5,199 +5,205 @@
...
+  </div>
+
+  <div class=logging>
+    <h2>Event Log</h2>
+
+    <ol id="eventLog"></ol>
+
+    <h2>Assertions</h2>
+
+    <ol id="assertionLog"></li>
```
pdehaan, line 204: Should this be `</ol>` instead of `</li>`?

http://validator.w3.org/check seems to be barfing on this line and the closing `</div>` on the following line (which is probably just due to the mismatching tags on L204).

example/rp/index.html
pdehaan, line 38: +1 to using quotes around attributes in the future cleanup PR.

example/rp/index.html
pdehaan, line 204: Not sure if it's worth it, but it'd be nice if we could maybe add Grunt to this repo and throw a few random linting tasks like JSHint (which I think may be handled in a mocha test or something, IIRC), and JSONLint, and HTMLLint, etc.

Catching things like `<ol></li>` is very hard to do manually in a PR, but Grunt+Travis will catch them pretty easily.

callahad:
Today I learned:
1. I should not be trusted to hand-code HTML in a new editor without something like jshint and htmltidy (or jade / haml!) to protect me from myself.
2. You guys are still watching this repo.
3. Seanmonstar has the right attitude.

;)

Things will likely get worse before they get better. I have a local branch that's all gussied up with Yahoo's purecss.io stuff, but it makes the markup completely ridiculous.

Issue #4171, opened 2014-08-04
Tweak logging for heka consumption


kparlante:
Sample logs from @shane-tomlinson back in October: https://gist.github.com/shane-tomlinson/e30b4f266b7f1e14435d

@dannycoates is working on a logging standard that we may want to adopt (or move towards): https://gist.github.com/dannycoates/0c7d04f9aa8d3908007a

Will update w/decisions after "data team" meeting tomorrow.

kparlante:
We should go with: https://mana.mozilla.org/wiki/display/CLOUDSERVICES/Logging+Standard
Example:

```
{
   "rp":"http://logging-shane.123done.org",
   "type":"signin",
   "user_agent":{
      "os":"Windows NT 6.1",
      "browser":"Firefox",
      "version":22
   },
   "timestamp":1382014800000
},
```

Becomes

```
{
   "Timestamp":1382014800000,
   "Type":"persona.server",
   "Hostname":<hostname>,
   "EnvVersion":1,
   "Pid":<processid>,
   "rp":"http://logging-shane.123done.org",
   "event":"signin",
   "user_agent":<entire user agent string>
},
```

Another example, just the "after" version:

```
{
   "Timestamp":1382014800000,
   "Type":"persona.client",
   "Hostname":<hostname>,
   "EnvVersion":1,
   "Pid":<processid>,
   "event_stream":["window.dom_loading", "xhr_complete.GET/wsapi/session_context"],
   "sample_rate":1,
   "lang":"en",
   "new_account":false,
   "screen_width":1024,
   "screen_height":768,
   "orphaned":true,
   "rp_api":"watch_with_onready",
   "number_emails":0,
   "number_sites_signed_in":0,
   "number_sites_remembered":0,
   "user_agent":<entire user agent string>
}
```

pdehaan:
@kparlante The OCD in me wants to know why we have some keys w/ Uppercase like `Timestamp` or `Type`, but others are lowercase like `event` or `user_agent`. Seems inconsistent and i cannot unsee.

kparlante:
@pdehaan: yeah, uppercase are the Heka envelope fields.

seanmonstar:
@pdehaan CANNOT UNSEE

![ducks wear dog masks](https://cloud.githubusercontent.com/assets/51479/4096086/c9961554-2fbc-11e4-9b46-9ab5bf1917d2.jpeg)

pdehaan:
@seanmonstar 

[...]

PR #4191, opened 2014-08-30
Add support for logging to heka

shane-tomlinson:
This is for @kparlante, but I could also use some review from @6a68 or @callahad!

issue #4171


callahad:
Whoops, reopened.

Looks like this would log to stdout by default, where previously we would remain silent if a logPath wasn't specified. Is that what we want?

What needs to happen to get this into production and talking to a Heka instance?

cc: @stomlinson @kparlante

PR #4207, opened 2014-11-20
Make send-from address for verification configurable.

mythmon:
@relud was complaining about this to me, so I fixed it.

r?


jrgm:
r=jrgm looks good to me. Thanks!

PR #4208, opened 2014-11-21
Train 0719 ses patch

jrgm:
/cc @callahad, @relud, @mythmon, @stephendonner 

This is just the incremental update to the train that is in stage and prod to use @mythmon's fix to make email sender configurable for AWS SES.


callahad:
@jrgm LGTM. r+

Issue #4224, opened 2015-09-24
Continue button isn't enabled when login form is autofilled


rnewman:
See Bug 1203385 (Firefox for Android). Also affects Firefox for iOS (WebKit), and has been reported on desktop, too.

STR from that bug:

Steps to reproduce:
1. go to bugzilla.mozilla.org
2. try sign in with persona
3. a box appears  which tells to write email
4. don't write anything.  just press on the writing area and see a email suggestion comes.
5. press the suggested email.

Actual results:

see down. the continue button is black

Expected results:

continue button should bee blue and work fine

![screenshot_2015-09-10-09-46-35](https://cloud.githubusercontent.com/assets/91722/10081176/71a9e008-62a8-11e5-8094-5b1fded5eaa6.png)

krydos:
Hi, 
I do not see any other (and simple) way than use setInterval() function. I searched for some libraries which can help us to avoid the issue but I noticed that all of them are using setInterval() function internally. 

I decided to not include any additional libraries because fix is pretty simple... Hope it's ok.

PR #4226, opened 2015-11-02
fixed issue #4224

krydos:
continue button should be enabled even in case
of autofill the email field.

I do not see any other (and simple) way than use **setInterval()** function. I searched for some libraries which can help us to avoid the issue but I noticed that all of them are using **setInterval()** function internally.


krydos:
tests are failed because there is issue with 
GET / tests

I'm not sure my code is related to this...  I will try to fix it too... hope after this the build fill be succesed.

PR #4230, opened 2016-01-15
Add a notice about decomissioning the service.

rfk:
I can't actually build this thing locally due to some dependency problem.  I tested out the changes by making them live in the web console:

screen shot 2016-01-14 at 18 20 05

screen shot 2016-01-14 at 18 19 23

Frankly I'm not sure it's worth adding this here, see more discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1239786

### DIFF
--- a/resources/static/pages/css/style.css
+++ b/resources/static/pages/css/style.css
@@ -782,3 +782,22 @@ article.flexible .info {
   margin-bottom: 30px;
 }
 
+#content .notice {
+    text-align: center;
+    background: #8A241B;
+    border: 1px solid #710B02;
+    padding: 5px;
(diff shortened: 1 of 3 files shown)


callahad:
r+, but waiting on merging per discussion in Bugzilla.

relud:
because i build persona on every deploy, and I was deploying an update to www.persona.org, I found that I now need util-linux (for /bin/flock) to build persona. not sure if that helps.

callahad:
Tiny nit: Instead of "This **website** will be unavailable" can/should we say "This **service** will be unavailable"?

PR #4236, opened 2016-10-06
feat: Add a shutdown notice to the dialog and all main site pages.

shane-tomlinson:
@rfk and @relud - this is to add a Persona shutdown notice everywhere. 😢

### DIFF
--- a/resources/static/common/css/style.css
+++ b/resources/static/common/css/style.css
@@ -594,3 +594,23 @@ footer .help {
 #development input[type=text] {
   padding: 3px 5px;
 }
+
+.notice {
+  text-align: center;
+  background: #8A241B;
+  border: 1px solid #710B02;
+  padding: 5px;
+  padding-left: 15px;
+  padding-right: 15px;
+}
+
+.notice p {
+  padding: 0.4em;
+}
+
+.notice a {
+  color: #fff;
+  border-bottom: 1px dotted #fff;
+  font-weight: normal;
+}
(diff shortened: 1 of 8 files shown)


relud:
I don't actually understand how these changes work. It looks like the notice is being moved from 2 places to 2 different places. I'm assuming that the 2 new places somehow get included everywhere, including the approximate location of the previous places?

rfk:
> I'm assuming that the 2 new places somehow get included everywhere, 
> including the approximate location of the previous places?

Yes.  Adding it to layout.ejs causes it to appear on all pages of the persona.org static site, including where it was before, like this:

screen shot 2016-10-07 at 14 06 07

Adding it to dialog_layout.ejs causes it to appear in the user-facing login dialog, like this:

screen shot 2016-10-07 at 14 06 49

@shane-tomlinson one think I did notice, is that clicking on the "More info" link in the dialog doesn't work for me.  Right-clicking and opening in a new tab does, but plain left-clicking does not.  I wonder if it's intercepting navigation attempts and rejecting ones it doesn't recognize, as it seems to want to open it in the originating window.

Otherwise, this LGTM, thanks!

shane-tomlinson:
@rfk - thanks! fixed the click issue. The odd thing was I tested that, the link was being partially obscured. If you clicked on the right side of the link everything worked as expected, from the left side, no.

relud (approved):

PR #4237, opened 2016-11-16
Persona end of life.

shane-tomlinson:
* All frontend routes show a "Persona has shutdown... See more" message.
* All wsapi requests return 410 (Gone).
* All navigator.id calls are gutted except those that open the Persona EOL dialog.

So long, and thanks for all the fish.


relud:
of the components dbwrite, keysign, browserid, route, and static, which ones are still needed with this?

rfk:
> +81 −2,367

Brutal.  Thanks @shane-tomlinson, I will try to take this for a spin soon.

> of the components dbwrite, keysign, browserid, route, and static, which ones
> are still needed with this?

My guess is that dbwrite and keysign are not needed, but will also do no harm if still active.  But @shane-tomlinson would better know the details.

shane-tomlinson:
> My guess is that dbwrite and keysign are not needed, but will also do no harm if still active. But @shane-tomlinson would better know the details.

I believe that route and static are the only ones that are still needed, [at least that's all I have running locally](https://github.com/mozilla/persona/pull/4237/files#diff-755317fea4d1d26fcc3fc9c656e1a092L42).

shane-tomlinson:
@rfk - I updated this PR to remove all processes but `route` and `static` as well as add tests to ensure all WSAPI routes return a 410 when requested.

rfk:
This works well in my local testing.

Trying to use the auth dialog:

screen shot 2016-11-23 at 06 56 18

Trying to access the persona management site:

screen shot 2016-11-23 at 06 55 43

Trying to talk directly to the service API:

screen shot 2016-11-23 at 06 58 47

rfk:
> wants to merge 1 commit into train-2014.07.19

I suggest that we don't merge this into the train branch, but rather cut a new train branch with this in it.  That will make it easier to roll back to the live version if we need to.

rfk:
@shane-tomlinson could you please rebase this when you get a chance?

PR #4238, opened 2016-11-28
Update decommission notice to make it more final

rfk:
This changes the shutdown notice to indicate that the service is definitely going away in December, rather than "after November". It also makes it redder so it stands out more:

screen shot 2016-11-29 at 09 47 58

@shane-tomlinson @relud r?  This will be better than having the site continue to say "after November 30th" once it is in fact after November 30th.

### DIFF
--- a/resources/static/common/css/style.css
+++ b/resources/static/common/css/style.css
@@ -597,8 +597,8 @@ footer .help {
 
 .notice {
   text-align: center;
-  background: #8A241B;
-  border: 1px solid #710B02;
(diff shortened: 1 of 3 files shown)


rfk:
Slight tweak to the wording, suggested by @ckarlof:

screen shot 2016-11-29 at 12 19 23

shane-tomlinson:
@rfk - shippit.

shane-tomlinson (approved):

PR #4239, opened 2017-02-02
2012.11.23 hotfix issue2822

husinbetawi:


rfk:
Can you please say more about this PR and why it's necessary, esp. given that the hosted service has been shut down?

